PR #2, opened 2016-11-01
Adding a draft for models.

gabelula:
To discuss.


models/asset.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+var mongoose = require('mongoose');
```
wyattjoh: Should import the mongoose that's in the current folder, ie:

``` js
const mongoose = require('../mongoose');
```

Also, should use const wherever possible, never use var.

wyattjoh (changes requested):
Should use `const/let` instead of `var` in ES6. Also don't import raw `mongoose`, instead grab the one from the mongoose file as in the comments listed below.

models/asset.js
gabelula: ok. Done!

wyattjoh (approved):

wyattjoh (changes requested):
Foreign keys are not referenced using models. Decided to hold off until resolved.

gabelula:
I wonder if we should also add "children" to the comment's model (we have parent right now).

models/comments.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
```
okbel: This file must be `comment.js` instead of `commentS.js`

gabelula:
Anything else to be able to merge this one? @wyattjoh @okbel @jde

models/comments.js
gabelula: Done

models/action.js
```diff
@@ -0,0 +1,35 @@
...
+    createdAt: 'created_at',
+    updatedAt: 'updated_at'
+  }
+});
+
+/**
+ * Finds an action by the id.
+ * @param {String} id  identifier of the action (uuid)
+*/
+ActionSchema.methods.findById = function(id) {
```
riley: this should be a static e.g. `ActionSchema.statics.findById = function (id) {`

riley (changes requested):
just one little thing I see

riley (approved):

wyattjoh (approved):

PR #4, opened 2016-11-01
Adding coral embed stream

davidgljay:
## What does this PR do?

Adds `coral-embed-stream` to the newly structured `talk` repo.

## How do I test this PR?

1. `npm install`
2. From the repo root, run `npm run embed-start`
3. Dev server should start.
4. Comment stream should load with mock data.
5. Posts won’t work, and a few 404s are expected until we get the endpoints up.

## Notes

- I’ve left the original web pack in place for reference, I’m not certain what the vision is for a central webpack so wanted to keep that separate. If run independently, this webpack file will build to the target dist directory.


riley:
I'm getting this error @davidgljay 

screen shot 2016-11-01 at 4 34 13 pm

davidgljay:
@riley This issue should be fixed. Also, I removed Dynamic Containers since there's no longer a user case for them and replaced with vanilla react.

Note: CommentStream is now has a gigantic render function, and under most circumstances it would be worth breaking up. I've avoided doing so
1. To limit the scope of the refactor to what's absolutely necessary.
2. Because it's likely that there will be benefits to having one giant render function down the road, and there's no downside other than unfashionableness.

riley (approved):
lgtm! 💯

PR #5, opened 2016-11-02
Code base - Scripts, linting, tests

okbel:
# Code base - Scripts, linting, tests

## Pre-Git
`pre-git` makes sure that some commands run before some git actions.

`npm run lint` runs in each `git commit`
`npm run test` runs in each `git push`

## Node versions
I’m using `nvm` for node versioning
The node version we are using is `v7.0.0`. This way we avoid using so many dependencies using the latest node features. [https://github.com/creationix/nvm](https://github.com/creationix/nvm)

This changes may look super simple but I thinks it's going to be helpful for the quick development of Talk.

## Things we *NEED* to discuss before we merge this PR

- `standard`: Are we going to use `standard`?
- Please give your feedback so we can add the codebase in this PR for Talk!

Thanks!


jde:
Love the pre-git list/test hooks!

wyattjoh:
Downside to using these hooks is that it's required to be installed on all clients machines, it doesn't prevent adding code that hasn't been linted from a machine without the hook. It would make sense to add linting to our CI process that ensured that the code is good before even allowing a merge. Everyone should be using an editor that performs linting on the codebase, that way no linting errors will bubble up.

wyattjoh:
The only thing different from standard that the eslint that I added to this project is the requirement of semicolons. I really believe that it improves readability of the code and reduces frustrations. What about standard is more preferable? The rules applied by `eslint:recommended` make it much easier to read. I think it would be worth it for us to adopt it as it applies stricter rules to the code style.

You can see the rules applied by this standard here: http://eslint.org/docs/rules/ anything with a checkmark is applied when `eslint:recommended` is used.

I had nothing but chaos dealing with elkhorn's style and I think it caused more errors than not. Linting for JS is to me the closest thing to compiling the code to check for syntax errors.

All in all I prefer the use of `eslint:recommended` across our FE/BE code.

okbel:
We can have both! We can add linting and testing to the CI as well. 

I think we all need to run tests in our machines. So every new PR that lands here, has correct linting and test coverage (without having to wait). Also, I think this approach encourages us to build more tests. _(I don't want to wait for the CI to run all the tests if I forgot to run them before pushing. That seems to me like a waste of time)_

About linting, it's the same for me. We can skip the `client` folder, or add `.eslintrc.json` file in the client folder if this takes a lot of our time. But the purpose of the PR is to raise awareness that we are handling two code styles.

wyattjoh:
Using git hooks, CI, and editor linting integrations sounds like a perfect stack to me @okbel :)

riley:
if we add the linting step to CI then we can accept PRs from other people and just lint them ourselves. Each dev on our team should have the proper linting set up in their editor, but I've found depending solely on this doesn't always work in practice :)

`eslint:recommend` is fine with me. I'd like to get some consensus, since it's the whole team working on the same repo at this point. @gabelula @impronunciable @davidgljay @jde

gabelula:
More strict code style and more tests seems good to me.

davidgljay:
Two questions:

1) If we apply this do we need to go through and refactor all of our existing code with the new styles in order to deploy to CI? If so we're signing up for a useful-but-not-user-facing chunk of work, and we should be mindful of it.

2) There's a floating `no-used-vars` linting error with React. Would this prevent us from deploying via CI?

wyattjoh:
@davidgljay we can add rule exceptions and pull them out as time goes on for the client folder as we already have to turn on JSX support to the linter. Specifcially around the `no-unused-vars` we just have to say:

```
...
"rules": {
  "no-unused-vars": [
    1
  ]
}
...
```

Which would move the `no-unused-vars` to a warning rather than an error.

Regarding the CI integration, this is only provided as the "last gate" rather than the only gate. As @riley mentioned, linting + tests should be ran locally before committing anyways.

riley:
I added some extra eslint rules if you guys are interested.

riley (approved):
great work! 👍

PR #7, opened 2016-11-02
start settings schema

riley:
# in progress

## What is this?

create a settings schema. please comment on what settings I'm missing.

Obviously we're not going to have all these settings exposed at first, and I'll create methods as we go along.


models/setting.js
```diff
@@ -0,0 +1,35 @@
+const mongoose = require('../mongoose');
+const uuid = require('uuid');
```
gabelula: the ID was missing or we need to remove this requirement.

models/setting.js
```diff
@@ -0,0 +1,35 @@
...
+const Schema = mongoose.Schema;
+
+const SettingSchema = new Schema({
+  anonymous_users: {type: Boolean, default: false},
```
gabelula: can we divide somewhow the settings and make comments on them? They are a lot and a big block of code.

jde:
@riley Great work thinking out all these settings. I'd like to be careful to only implement them one at a time as we have needs though.

Let's start with a super simple schema:

```
   moderation: enum {'pre', 'post'}
```

To satisfy the story and we can include more settings as are relevant.

Please comment out the potential settings that you put together in a block below, or move them into a SETTINGS_ROADMAP.md file. We will get to them in short enough time.

routes/api/settings/index.js
```diff
@@ -0,0 +1,14 @@
+const express = require('express');
+const router = express.Router();
+const path = require('path');
+const Setting = require(path.resolve(__dirname, 'models/setting'));
+
+router.get('/settings', (req, res, next) => {
+  Setting.getSettings().then(res.json).catch(next);
+});
+
+router.put('/settings', (req, res, next) => {
```
wyattjoh: Remove `/settings`, add to the parent router to mount on `/settings`

wyattjoh (changes requested):

models/setting.js
```diff
@@ -3,8 +3,7 @@ const Schema = mongoose.Schema;
 
 const SettingSchema = new Schema({
   id: {type: String, default: '1'},
```
gabelula, line 5: shouldn't be 

default: uuid.v4,
 unique: true

for the id ?

gabelula (changes requested):

models/setting.js
riley: I thought we'd discussed hardcoding it to `"1"` since there is only ever one record in this collection, and there's an init file I added to `/bin` that creates the record with an id of `1`

gabelula: ok. I missed that part of the conversation. But you are right! Only 1 there is.

gabelula (approved):

bin/init.js
```diff
@@ -0,0 +1,16 @@
+const MongoClient = require('mongodb').MongoClient;
+const url = process.env.TALK_MONGO_URL;
```
wyattjoh: I'd rather see a graceful fallback like:

``` js
const url = process.env.TALK_MONGO_URL || 'localhost';
```

And remove the error throwing below.

package.json
```diff
@@ -64,6 +64,7 @@
     "imports-loader": "^0.6.5",
     "json-loader": "^0.5.4",
     "mocha": "^3.1.2",
+    "mongodb": "^2.2.11",
```
wyattjoh: Shouldn't need the native driver directly here

bin/init.js
```diff
@@ -0,0 +1,16 @@
+const MongoClient = require('mongodb').MongoClient;
```
wyattjoh: Should use the mongoose client and not the native driver.

wyattjoh (changes requested):

wyattjoh (approved):

PR #8, opened 2016-11-02
Mocking stream endpoint

davidgljay:
## What does this PR do?

Creates a stream endpoint that returns a static json object and adjusts the frontend to expect data of that format.

## How do I test this PR?

1. `npm run build`
2. `npm start`
3. Visit `http://localhost:3000/client/coral-embed-stream/samplearticle.html`
4. The page should render appropriately (though you won’t be able to post comments.)
5. Your server logs should say ‘Stream endpoint has been hit with asset_id assetTest’. (This is temporarily there for verification.)

## Notes

- I went ahead and added an `npm build` script for efficiency, someone with more of a handle on the architecture (i.e. @wyattjoh  or @okbel) should feel free to refactor.
- I went ahead and set up a path for hosting static built files. Same as above, feel free to refactor.


client/coral-framework/store/actions/items.js
```diff
@@ -89,26 +78,48 @@ export const appendItemRelated = (item_id, property, value) => {
...
+          return response.ok ? response.json() : Promise.reject(response.status + ' ' + response.statusText)
+        }
+      )
+      .then((json) => {
+
+        /* Sort comments by date*/
+        let rootComments = []
+        let childComments = {}
+        const sorted = json.sort((a,b) => b.created_at - a.created_at)
+        sorted.reduce((prev, item) => {
```
jde: From @riley: since this block doesn't return things by keys, we should use a for each here.

client/coral-framework/store/actions/items.js
```diff
@@ -89,26 +78,48 @@ export const appendItemRelated = (item_id, property, value) => {
...
+            childComments[item.parent_id] = children.concat(item.id)
+          }
+        }, {})
+
+        dispatch(addItem({
+          id: assetId,
+          comments: rootComments
+        }))
+
+        Object.keys(childComments).reduce((prev, key) => {
```
jde: From @riley: since this block doesn't return things by keys, we should use a for each here.

routes/api/stream/index.js
```diff
@@ -0,0 +1,34 @@
+const express = require('express');
+
+const router = express.Router();
+
+router.get('/', (req, res) => {
+  console.log('Stream endpoint has been hit with asset_id ', req.query.asset_id)
+  res.setHeader('Content-Type', 'application/json');
```
jde: res.json() will take care of the header and stringify for you.

routes/api/stream/index.js
```diff
@@ -0,0 +1,34 @@
+const express = require('express');
+
+const router = express.Router();
+
+router.get('/', (req, res) => {
```
jde, line 5: Add a third param called `next` here, which can be used to pass errors down the first error handler it finds down the chain.

See: https://expressjs.com/en/guide/error-handling.html

client/coral-framework/store/actions/items.js
```diff
@@ -214,13 +211,19 @@ export function postItem (data, type, id) {
...
+    }
+    const options = {
+      method: 'POST',
+      body: JSON.stringify(action)
     }
-    dispatch(appendItemArray(item, action, user))
-    return fetch('/v1/action/' + action + '/user/' + user + '/on/item/' + item, options)
+
+    dispatch(appendItemArray(id, type, user_id))
+    return fetch('api/v1/comments/' + id + '/actions', options)
```
gabelula: we are not using api/v1 , right?

routes/api/stream/index.js
riley: we'll add it later. I think linting might fail if there are unused vars, and this is a mock for now.

riley (approved):
lgtm 💯

PR #9, opened 2016-11-03
Added coral admin

impronunciable:
Just added the coral-admin client side code


jde (approved):
Notes from review:

We look forward to refactoring the webpack build process to include all client work and enforce the coral-[embed]-bundle paradigm for predictable accessing client code.

PR #11, opened 2016-11-03
make the settings routes actually work

riley:
## What does this PR do?

adds ability to pull json directly off `req.body` by adding some middleware. Add some doc blocks.

## How do I test this PR?

- make a GET request to `/api/v1/settings/` and you should get the current settings (assuming you've run `/bin/init.js`
- make a PUT request to `/api/v1/settings/` with a json payload of something like `{"moderation": "post"}` and you should get back the updated settings object.

@coralproject/tech

### DIFF
--- a/app.js
+++ b/app.js
@@ -6,7 +6,6 @@ const bodyParser = require('body-parser');
 
 const app = express();
 
-app.use(bodyParser.urlencoded({extended: true}));
 app.use(bodyParser.json());
 
 app.use('/api/v1', require('./routes/api'));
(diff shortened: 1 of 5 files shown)


app.js
```diff
@@ -2,9 +2,13 @@
 
 // Initialize application framework.
 const express = require('express');
+const bodyParser = require('body-parser');
 
 const app = express();
 
+app.use(bodyParser.urlencoded({extended: true}));
```
wyattjoh: Shouldn't add this middleware unless we have urlencoded payloads.

riley: ah, makes sense.

wyattjoh (approved):

PR #14, opened 2016-11-04
Linting Error

okbel:
## What does this PR do?
This PR solves  ` 6:3  error  Don't use process.exit(); throw an error instead  no-process-exit`linting error
## How do I test this PR?

- Run `npm run lint`
- It shouldn't trigger any errors

@coralproject/tech

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -35,7 +35,7 @@
         "no-throw-literal": [2],
         "yoda": [1],
         "no-path-concat": [2],
-        "no-process-exit": [2],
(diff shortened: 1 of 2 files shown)


wyattjoh:
Why not we actually throw an error instead of disabling the linting rule?

okbel:
I'm in favor of doing that. Like the mongo connection. I thought this was a blocker! Working on it.

okbel:
@wyattjoh Updated and passing!

gabelula (approved):
It seems good.

PR #15, opened 2016-11-04
Comment api

gabelula:
I'm still working on the other routes for comments but want to start the discussion on this one going.

## What does this PR do?

Implements  routes on
- POST comment
- GET stream

## How do I test this PR?

npm start
curl the endpoints

@coralproject/tech


models/user.js
```diff
@@ -12,12 +11,12 @@ const UserProfileSchema = new Schema({
...
-  timestamps: {
-    createdAt: 'created_at',
-    updatedAt: 'updated_at'
-  }
+// },{
+//   _id: false,
+//   timestamps: {
+//     createdAt: 'created_at',
+//     updatedAt: 'updated_at'
+//   }
```
riley: Why did this get commented out? Are we not using timestamps for the users?

routes/api/comments/index.js
```diff
@@ -1,7 +1,23 @@
...
 
+//==============================================================================
+// Validations on parameters
+//==============================================================================
+
+router.param('comment_id', function(res, req, next, comment_id) {
+  console.log('validations on comment id ');
+  req.comment_id = comment_id;
+  next();
+});
```
riley: This is neat, but why do this instead of just doing a lookup on `req.params` in the other route handlers?

routes/api/comments/index.js
```diff
@@ -10,8 +26,23 @@ router.get('/:comment_id', (req, res) => {
...
+  });
+  comment.save(function(err, comment) {
+    if(err) {
+      res.status(500);
+      return next(err);
+    }
+    res.status(201);
+    res.send(comment.id);
+    next();
+  });
```
riley, line 35: This block could be 

```
comment.save().then(({id}) => {
  res.status(201).send(id);
}).catch(error => {
  next(error);
});
```

I'd like to avoid callbacks if possible, I feel like it makes the code a bit easier to reason about. I still have nightmares about my old js with a million `if (err) { // something }`

routes/api/stream/index.js
```diff
@@ -1,33 +1,17 @@
...
-      'id': 'def',
-      'type': 'comment',
-      'body': 'Another comment',
-      'created_at': new Date().getTime() - 400000,
-      'asset_id': 'assetTest'
-    }
-  ]).catch(next);
+  const comments = Comment.findByAssetId(req.query.asset_id) || [];
+  const users = User.findByIdArray(comments.map((comment) => comment.author_id));
+  const actions = Action.findByItemIdArray(comments.map((comment) => comment.id));
```
riley: Ok, this sums up the beauty/horribleness of javascript. `comments`, `users` and `actions` here are all `Promise`s when this code is run, and will _eventually_ resolve to arrays at some (unknown) point in the future. So below, when we call `res.json`, you'll get something crazy in the json. So to solve this, you need to pass all the Promises to `Promise.all` like 

```
Promise.all([comments, users, actions]).then(([comments, users, actions]) => {
  // comments, users, actions will be resolved to arrays here.
}).catch(error => {
  // handle error
});
```

[Promise.all MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all)

routes/api/stream/index.js
```diff
@@ -1,33 +1,17 @@
...
+
+  res.json(comments.concat(users).concat(actions)).catch(next);
```
riley: `res.json` is not a Promise so it won't have a `catch` method here.

riley (changes requested):

models/user.js
gabelula: Not sure how it ended commented. I will add it again.

routes/api/comments/index.js
gabelula: Yes. Removing them until we really need to use this.

routes/api/comments/index.js
gabelula: Done.

wyattjoh:
Once my docker-compose branch gets merged into this branch @gabelula it will select the right version of node that will resolve the issue that you're having now on CI.

gabelula:
Ready to review.

riley (approved):

riley:
awesome! works and tests pass 💯

PR #19, opened 2016-11-07
Asset

jde:
## What does this PR do?

This PR adds the concept of the asset to Talk. It includes:

* the asset model with basic crud
* asset endpoints
* tests 
*docs

## How do I test this PR?

npm run test

@coralproject/tech


app.js
```diff
@@ -4,6 +4,10 @@
 const express = require('express');
 
 const app = express();
+const bodyParser = require('body-parser');
+
+app.use(bodyParser.json()); // for parsing application/json
+app.use(bodyParser.urlencoded({extended: true})); // for parsing application/x-www-form-urlencoded
```
wyattjoh: When do we have `application/x-www-form-urlencoded` encoded bodies? It should be false here otherwise.

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+
+  return Asset.findOne({id});
+
+};
+
+/**
+ * Finds a asset by its url.
+ * @param {String} url  identifier of the asset (uuid).
+*/
+AssetSchema.statics.findByUrl = function(url) {
```
davidgljay, line 59: This looks good, but we should be clear on what we want to support with findBy functions and what we should handle with queries.

bin/init.js
```diff
@@ -3,5 +3,5 @@ const defaults = {id: '1', moderation: 'pre'};
 
 Setting.update({id: '1'}, {$setOnInsert: defaults}, {upsert: true}).then(() => {
   console.log('created settings object.');
-  process.exit();
+//  process.exit(); // Removed to satisfy hooks
```
wyattjoh: This should be resolved upstream, remove comment and merge with master to resolve.

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+    default: 'article'
+  },
+  headline: String,
+  summary: String,
+  section: String,
+  subsection: String,
+  authors: [String],
+  publication_date: Date
+},{
+  _id: false,
```
wyattjoh: Should not use `_id: false` here, Mongoose needs is for root documents in a collection :(

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+});
+
+
+/**
+ * Search for assets. Currently only returns all.
+*/
+AssetSchema.statics.search = function(query) {
+
+  return Asset.find(query);
+
```
wyattjoh, line 42: Remove extra space around this return. Should add `.exec()` here else this will never execute.

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+  }
+});
+
+
+/**
+ * Search for assets. Currently only returns all.
+*/
+AssetSchema.statics.search = function(query) {
+
+  return Asset.find(query);
```
davidgljay: Why is `find` being wrapped here? If we expect a use case for this to be more complex in the future should we just wrap it then?

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+
+};
+
+/**
+ * Finds a asset by its url.
+ * @param {String} url  identifier of the asset (uuid).
+*/
+AssetSchema.statics.findByUrl = function(url) {
+ 
+  return Asset.findOne({'url': url});
```
wyattjoh: Should add `.exec()` here else this will never execute.

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+
+};
+
+/**
+ * Finds an asset by its id.
+ * @param {String} id  identifier of the asset (uuid).
+*/
+AssetSchema.statics.findById = function(id) {
+
+  return Asset.findOne({id});
```
wyattjoh: Should add .exec() here else this will never execute.

models/asset.js
```diff
@@ -0,0 +1,110 @@
...
+*/
+AssetSchema.statics.upsert = function(data) {
+
+  // If an id is not sent, create one.
+  if (typeof data.id === 'undefined') {
+    data.id = uuid.v4();
+  }
+
+  // Perform the upsert against the id field.
+  let updatePromise = Asset.update({id: data.id}, data, {upsert: true})
```
wyattjoh: Should add .exec() here else this will never execute.

routes/api/asset/index.js
```diff
@@ -0,0 +1,48 @@
...
+router.get('/', (req, res) => {
+
+  let query = {};
+
+  if (typeof req.query.url !== 'undefined') {
+    query.url = req.query.url;
+  }
+
+  Asset.search(query)
+    .then((asset) => {
```
davidgljay, line 15: It seems like `query` and `findByUrl` can both be used here, is there a reason why we to need to keep both functions?

davidgljay (commented):
I have a question about the why there's both a `query` and `findByUrl` function, otherwise this looks good.

routes/api/asset/index.js
```diff
@@ -0,0 +1,48 @@
...
+      res.json(asset);
+    });
```
wyattjoh: Needs a `.catch(next);` here, all promises must be handled.

routes/api/asset/index.js
```diff
@@ -0,0 +1,48 @@
...
+
+});
+
+// Get an asset by id
+router.get('/:id', (req, res) => {
+
+  Asset.findById(req.params.id)
+    .then((asset) => {
+      res.json(asset);
+    });
```
wyattjoh: Needs a `.catch(next);` here, all promises must be handled.

routes/api/asset/index.js
```diff
@@ -0,0 +1,48 @@
...
+// Upsert an asset and return the affected document.
+router.put('/', (req, res) => {
+
+  Asset.upsert(req.body)
+    .then((asset) => {
+
+      res.json(asset);
+
+    })
+    .catch((err) => {
```
wyattjoh: Needs a `.catch(next);` here, all promises must be handled. do not send json back here using `res.json` and instead rely on the error handler.

swagger.yaml
```diff
@@ -3,7 +3,7 @@ info:
   title: Talk API
   description: A commenting platform from The Coral Project. https://coralproject.net
   version: "0.0.1"
-host: api.talk-coralproject.net
+host: talk-stg.coralproject.net/api/v1
```
wyattjoh: Host should not include the `/api/v1`, that is part of the `basePath` below.

swagger.yaml
```diff
@@ -158,6 +158,38 @@ paths:
...
           description: OK
+  /asset:
+    get:
+      tags:
+        - Asset
+      description: Get an asset by id.
+      responses:
+        200:
+          description: OK
+        404:
```
wyattjoh: It seems that the express app does not return a 404 when the returned asset is not there (aka, empty/null), remove this error state to reflect the current state of the API.

swagger.yaml
```diff
@@ -158,6 +158,38 @@ paths:
...
+          description: Not Found
+    put:
+      tags:
+        - Asset
+      description: Upsert an asset.
+      responses:
+        204:
+          description: OK
+  /asset/url/{url}:
```
wyattjoh: I would prefer to see this in a more url safe manner as:

```
/asset?url={{ urlencoded_url }}
```

Because we can urlencode a query param and not a route param.

swagger.yaml
```diff
@@ -158,6 +158,38 @@ paths:
       responses:
         204:
           description: OK
+  /asset:
```
wyattjoh, line 165: Should really specify the ID that you are getting it by as a query param as:

```
/asset?id={{ urlencoded_id }}
```

swagger.yaml
```diff
@@ -212,3 +244,43 @@ definitions:
...
+        description: The subsection that the asset is in.
+      authors:
+        type: string
+        description: An array of the authors for this asset.
+      publication_date:
+        type: date
+        desctipion: When this asset was published.
+
+
+
```
wyattjoh: Remove extra spaces.

tests/asset.js
```diff
@@ -0,0 +1,110 @@
+/* eslint-env node, mocha */
+const Asset = require('../models/asset');
+
+const expect = require('chai').expect;
+let chai = require('chai');
```
wyattjoh: All these `let`'s should be `const`'s

wyattjoh (changes requested):

models/asset.js
jde: My thought was that it's good practice to keep all database access in the model. This will lead to the fewest variations on database access and let us make intelligent decisions about what we want/don't want each item to do in one place.

routes/api/asset/index.js
jde: findByUrl is designed to return a single asset by url, which is a key use case. Search is meant to return an array of results. You are correct that both are not yet strictly needed as more variations of code elsewhere can account for them.  The use case for search will come up as soon as we want to list urls (which I do in testing.) I'm up for discussing whether or not we want to be very strict in our functionality buildout or if this makes sense.

app.js
jde: Removed. I was using that in postman, but do not need to.

bin/init.js
jde: Fixed by merge.

models/asset.js
jde: Updated. I was mistaken for it's purpose.

models/asset.js
jde: Added.

models/asset.js
jde: Added.

routes/api/asset/index.js
jde: Added the catch block.

swagger.yaml
jde: Yes. I had already made that change in the code. Updating the docs to reflect.

mongoose.js
```diff
@@ -1,6 +1,6 @@
 const mongoose = require('mongoose');
 const enabled = require('debug').enabled;
-const url = process.env.TALK_MONGO_URL || 'mongodb://localhost';
+const url = process.env.TALK_MONGO_URL || 'mongodb://localhost/coral-talk';
```
wyattjoh: Should not select the database.

tests/asset.js
jde: Updated

mongoose.js
jde: My thought was to have a default db, but I can get behind enforcing people to explicitly set or end up writing to /test/

swagger.yaml
jde: Removed.

swagger.yaml
jde: Updated host and base path.

wyattjoh (approved):

PR #20, opened 2016-11-07
fix(coral-admin): Adapt to the new schema.

impronunciable:
Most things are not working yet but at least it bundles


client/coral-admin/package.json
```diff
@@ -4,14 +4,13 @@
   "description": "",
   "main": "index.js",
   "scripts": {
-    "build": "./node_modules/.bin/webpack --config webpack.config.js",
+    "build": "./node_modules/.bin/webpack --config webpack.config.js --watch",
```
wyattjoh: Should create a `build-watch` as `build` will be used by CI eventually.

jde (approved):
Bundles are building and files are serving.

PR #21, opened 2016-11-07
add tests to settings models/endpoints

riley:
## What does this PR do?

Adds test coverage to `Setting` model and corresponding endpoints.

## How do I test this PR?

- `npm test`
- tests should pass

@coralproject/tech


tests/routes/api/settings/index.js
```diff
@@ -0,0 +1,62 @@
...
+  before(() => {
+    return Setting.update({id: '1'}, {$setOnInsert: defaults}, {upsert: true});
+  });
+
+  it('should respond to a PUT with new settings', () => {
+    chai.request(app)
+    .put('/api/v1/settings')
+    .send({moderation: 'post'}, (err, res) => {
+      expect(err).to.be.null;
+      expect(res).to.have.status(204);
```
jde, line 46: Would it be safer to return 200 with the settings object in the body? This way a fe could update it's state and ensure that what the user sees exactly reflects what's in the db.

This is a larger style question for our apis.  Approving PR.

jde (approved):

tests/routes/api/settings/index.js
riley: that's what I had before, but @wyattjoh pointed out that it's more inline with REST to just return something along the lines of "it worked" instead of returning stuff from a PUT.

PR #22, opened 2016-11-07
Comment api implementation

gabelula:
## What does this PR do?

Adds endpoints for comments.

## How do I test this PR?

npm test

@coralproject/tech


models/comment.js
```diff
@@ -44,6 +46,38 @@ CommentSchema.statics.findByAssetId = function(asset_id) {
   return Comment.find({asset_id});
 };
 
+/**
+ * Change the status of a comment.
+ * @param {String} id  identifier of the comment  (uuid)
+ * @param {String} status the new status of the comment
+*/
+CommentSchema.statics.changeStatus = function(id, status) {
+  return Comment.update({'id': id}, {$set: {'status': status}}, {upsert: false}).then(() => {
```
riley: So you could change this to `Comment.findOneAndUpdate` [Model#findOneAndUpdate](http://mongoosejs.com/docs/api.html#query_Query-findOneAndUpdate) and pass `{new: true}` into the options hash and get back the updated comment for free, so you don't have to make a second db query.

So it would be `Comment.findOneAndUpdate({id}, {$set: {status}}, {upsert: true, new: true}).then(comment => {`

models/comment.js
```diff
@@ -44,6 +46,38 @@ CommentSchema.statics.findByAssetId = function(asset_id) {
...
+    Comment.findById(id).then((comment) => {
+      return comment;
+    }).catch((err) => {
```
riley: I usually don't have the `catch` statements here, I put them in the route or wherever the Promise returned from this function is used. Once the error is "caught", you can't do anything else with it. Wherever you implement `Comment.changeStatus` you probably want to send an error back over the network or something in a `catch` statement there.

routes/api/comments/index.js
```diff
@@ -22,34 +26,59 @@ router.get('/:comment_id', (req, res, next) => {
...
-    body: req.query.body,
-    author_id: req.query.author_id,
-    asset_id: req.query.asset_id,
-    parent_id: req.query.parent_id,
-    status: req.query.status
+    body: req.body.body,
+    author_id: req.body.author_id,
+    asset_id: req.body.asset_id,
+    parent_id: req.body.parent_id,
+    status: req.body.status
```
riley: if you want to, you can use [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to pull the properties off `req.body` to make for less repetition.

`const {body, author_id, asset_id, parent_id, status} = req.body;`

and then you would just use the variables as normal.

`let comment = new Comment({body, author_id, asset_id, parent_id, status});`

routes/api/comments/index.js
```diff
@@ -22,34 +26,59 @@ router.get('/:comment_id', (req, res, next) => {
...
   }).catch(error => {
     next(error);
   });
-
 });
 
-router.put('/:comment_id', (req, res) => {
-  res.send('Update a comment');
+router.post('/:comment_id', (req, res, next) => {
+  Comment.findById(req.params.comment_id).then((comment) => {
```
riley, line 37: same thing here, change this to `findOneAndUpdate` with `{new: true}`

routes/api/comments/index.js
```diff
@@ -22,34 +26,59 @@ router.get('/:comment_id', (req, res, next) => {
...
+    comment.body = req.body.body;
+    comment.author_id = req.body.author_id;
+    comment.asset_id = req.body.asset_id;
+    comment.parent_id = req.body.parent_id;
+    comment.status = req.body.status;
+
+    comment.save().then((comment) => {
```
riley: so i'd probably just return `comment.save()` here, and have another `then()` on line 53 so we're not getting too nested. Also, if the save fails for some reason, your `catch` will still work.

routes/api/comments/index.js
```diff
@@ -22,34 +26,59 @@ router.get('/:comment_id', (req, res, next) => {
...
+  }).catch(error => {
+    next(error);
+  });
 });
 
-router.delete('/:comment_id', (req, res) => {
-  res.send('Delete a comment');
+router.delete('/:comment_id', (req, res, next) => {
+  Comment.findById(req.params.comment_id).then((comment) => {
+    comment.remove().then(() => {
```
riley: add a `return` statement here so that if the remove fails, your `catch` will still work.

riley (changes requested):
looking good, just a few changes

riley (approved):

PR #30, opened 2016-11-08
fix a race condition in the tests

riley:
## What does this PR do?

I thought that `it` statements ran in series in chai tests. This is incorrect and caused a race condition since I was hitting the network and causing tests to randomly fail.

## How do I test this PR?

- `npm run tests`
- tests should pass

@coralproject/tech


kgardnr (approved):
Nice! Test pass on multiple runs locally.

PR #34, opened 2016-11-08
Update readme

kgardnr:


README.md
```diff
@@ -1,23 +1,35 @@
...
+Make sure you have Docker running first and then run `docker-compose up -d`
+
 ### Testing
 `npm test`
 
 ### Lint
 `npm run lint`
 
+### Helpful URLs
+Bare comment stream: http://localhost:5000/client/coral-embed-stream
```
jde: The default port comes out to 3000 for me.  Is it 5000 on yours?

kgardnr: Oh yeah if I use docker, it comes out as 5000, but we can just note PORT so they can run on whatever port they're using.

davidgljay (approved):

PR #42, opened 2016-11-08
Update settings in ui

riley:
## What does this PR do?

You can update moderation pre/post in coral-admin!

## How do I test this PR?

- `npm run build`
- go to `/configure`
- click the `Enable pre moderation checkbox`
- click Save
- wow

@coralproject/tech


.eslintrc.json
```diff
@@ -1,4 +1,8 @@
 {
+    "globals": {
+        "fetch": true,
+        "Headers": true
+    },
```
wyattjoh: These global's aren't necessary. Right now, the `client` folder isn't linted. When it is, we can specify `browser: true` in the `env` object which will enable these globals automatically.

Recommended action, remove the `globals` object.

wyattjoh (changes requested):

jde:
The settings aren't saving. When I save the endpoint replies with a null.  Refreshing the page, the setting is not loaded.

wyattjoh (approved):

Issue #48, opened 2016-11-09
Incorporate bin/init.js into application startup


jde:
Currently we need to run bin/init.js to start up settings. Failure to do so leads to an opaque error and lots of debugging. I'd like to avoid extra steps that we need to know to do whenever possible.

Move the init.js step into the bootstrapping part of the server start. In this case, we can check to see if settings have been initialized and initialize them if they have not yet been initialized.

jde:
Fixed by #55

PR #55, opened 2016-11-10
initialize settings in an init file before app accepts requests

riley:
## What does this PR do?

It was extremely frustrating to try and debug a problem and it ended up being that the tests nuked the db and settings were missing. This happened a few times a day, wasting time. I've moved the init script into the app start-up lifecycle, so the app will not accept requests until it's all ready to go.

## How do I test this PR?

- `npm start`
- curl `http://localhost:3000/api/v1/settings`
- response should not be `null`

@coralproject/tech

#48

### DIFF
--- a/models/setting.js
+++ b/models/setting.js
@@ -12,6 +12,14 @@ const SettingSchema = new Schema({
 });
 
 /**
+ * this is run once when the app starts to ensure settings are populated
+ * @return {Promise} null initialize the global settings object
+ */
+SettingSchema.statics.init = function (defaults) {
+  return this.update({id: '1'}, {$setOnInsert: defaults}, {upsert: true});
+};
+
+/**
  * gets the entire settings record and sends it back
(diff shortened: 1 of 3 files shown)


models/setting.js
```diff
@@ -12,6 +12,14 @@ const SettingSchema = new Schema({
 });
 
 /**
+ * this is run once when the app starts to ensure settings are populated
+ * @return {Promise} null initialize the global settings object
+ */
+SettingSchema.statics.init = function (defaults) {
+  return this.update({id: '1'}, {$setOnInsert: defaults}, {upsert: true});
```
jde: We should only initialize settings when the settings object is not already initialized.

Running this every time the server starts will overwrite already existing settings with defaults. 

We need to check to see if settings are already initialized, and if they are, do not overwrite with defaults.

jde (approved):

PR #56, opened 2016-11-10
Add settings to stream return

jde:
## What does this PR do?

Adds a settings object to the response of the GET stream endpoint. The settings object currently contains the pre/post moderation setting, which the FE needs to know how to message the user upon posting. This object will contain future settings and may be used to serve dynamic settings per asset or user in the future.

## How do I test this PR?

npm run test

or

look for the settings object in your friendly neighborhood `/api/v1/stream?asset_id=xxx` response.

@coralproject/tech

### DIFF
--- a/routes/api/stream/index.js
+++ b/routes/api/stream/index.js
@@ -1,44 +1,46 @@
+
 const express = require('express');
+const router = express.Router();
 
 const Comment = require('../../../models/comment');
 const User = require('../../../models/user');
 const Action = require('../../../models/action');
-
 const Setting = require('../../../models/setting');
 
-const router = express.Router();
 
-// Find all the comments by a specific asset_id.
+// Find all the comments, users, actions by a specific asset_id + settings.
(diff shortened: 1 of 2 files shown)


davidgljay (changes requested):
This duplicates with PR https://github.com/coralproject/talk/pull/57, which I did in parallel. Let's touch base about which approach is best.

jde:
Agreed. We can discuss. This method returns settings along with data in a single request, which I believe will lead to faster load times than making a settings request, waiting, then making a data request.

We can also consider the idea @wyattjoh brought up that would package settings into the steam code when it initializes. This would be fastest of all but would create a hard requirement to only load settings through the /embed/stream route.

jde:
Thinking more deeply about this, I like the solution in #57 as the path forward. Wrapping settings into various calls like /stream and writing code into those reducers to pull it out has a bad smell. Having an explicit settings endpoint is clear, simple and doesn't have pre-optimization.

PR #58, opened 2016-11-10
Add embedding routes and homepage

jde:
## What does this PR do?

Creates new templated embedding scheme for embeds.
Adds a home page (article with stream embed).
Removed abandoned duplicated copy of Embed screen.

## How do I test this PR?

Test endpoints:
/
/embed/stream
And the embed code from the admin panel.

@coralproject/tech


routes/index.js
```diff
@@ -0,0 +1,22 @@
+const express = require('express');
+
+const router = express.Router();
+
+router.use('/api/v1', require('./api'));
+router.use('/admin', require('./admin'));
+
+router.use('/embed/:embed', (req, res, next) => {
```
wyattjoh: Move this to a `/routes/embed` folder and require it as above.

wyattjoh (changes requested):

routes/index.js
jde: Done. Right. We talked about this already. :)

wyattjoh (approved):

PR #59, opened 2016-11-10
Moved the moderation queue to its own folder and change the route to …

gabelula:
## What does this PR do?

the new route is /api/v1/queue/comments/pending

## How do I test this PR?

npm test

@coralproject/tech

…/queue/comments/pending


routes/api/queue/index.js
```diff
@@ -0,0 +1,30 @@
...
+//==============================================================================
+
+// Returns back all the comments that are in the moderation queue. The moderation queue is pre or post moderated,
+// depending on the settings. The :moderation overwrites this settings.
+// Pre-moderation:  New comments are shown in the moderator queues immediately.
+// Post-moderation: New comments do not appear in moderation queues unless they are flagged by other users.
+router.get('/comments/pending', (req, res, next) => {
+  Setting.getModerationSetting().then(function({moderation}){
+    let moderationValue = req.query.moderation;
+    if (typeof moderationValue === 'undefined' || moderationValue === undefined) {
```
jde: `typeof moderationValue === 'undefined'` is a good check. `moderationValue === undefined`is not valid. This simply checks against a variable called _undefined_, which we are just assuming is undefined.

It would be better to say:

`if (typeof moderationValue === 'undefined' || !moderationValue) {`

This covers all the falsey cases ('', null, 0, false), which is good practice in a case like this where we are not using _flow_ or another type checking framework.

routes/api/queue/index.js
```diff
@@ -0,0 +1,30 @@
...
+      moderationValue = moderation;
+    }
+    Comment.moderationQueue(moderationValue).then((comments) => {
```
jde: Having a single queue method in the comment is ok for simple queues. 

If/when queues take on properties beyond simple querying I would recommend creating a queue model that contains all the logic and uses the Comments model to do the basic operations.

tests/routes/api/queue/index.js
```diff
@@ -0,0 +1,105 @@
...
+        expect(res).to.have.status(200);
+        expect(res.body[0]).to.have.property('id', 'def');
+        done();
+      });
+  });
+
+  it('should return all the pending comments as pre moderated', function(done){
+    chai.request(app)
+      .get('/api/v1/queue/comments/pending')
+      .query({'moderation': 'pre'})
```
jde: The pre/post moderation settings are controlled via the Settings object. Mixing responsibility for how to set this setting will lead to confusion. 

When we test different settings, we should call the settings endpoint to set them. The api should react by checking the settings object alone.

Please refactor here and anywhere in the code base that doesn't use Settings to determine the moderation method.

transfer.py
```diff
@@ -0,0 +1,49 @@
+from pymongo import MongoClient
```
jde: I'd like to keep scripts like this that are specific to internal/private data out of the repo. Devs seeing this who do not have/know of Wapo data will be very confused. I'll create a private repo for internal scripts such as this one.

jde (changes requested):

routes/api/queue/index.js
```diff
@@ -0,0 +1,30 @@
...
+// Get Routes
+//==============================================================================
+
+// Returns back all the comments that are in the moderation queue. The moderation queue is pre or post moderated,
+// depending on the settings. The :moderation overwrites this settings.
+// Pre-moderation:  New comments are shown in the moderator queues immediately.
+// Post-moderation: New comments do not appear in moderation queues unless they are flagged by other users.
+router.get('/comments/pending', (req, res, next) => {
+  Setting.getModerationSetting().then(function({moderation}){
+    let moderationValue = req.query.moderation;
```
jde: Moderation should come from the Settings model only.

routes/api/queue/index.js
gabelula: Yes. It seems good. We  can move it that way.

tests/routes/api/queue/index.js
gabelula: I was thinking that this endpoint will be called when we need to show a preview of the queue that would have those two sets of moderation.

transfer.py
gabelula: Yes. It went in by mistake :/ I removed it.

routes/api/queue/index.js
gabelula: Done

jde (approved):

PR #61, opened 2016-11-11
I18n

davidgljay:
## What does this PR do?

Updates internationalization across talk.

## How do I test this PR?

1. Load the embed stream.
2. Enter `localStorage.setItem('locale','es-US')` in the browser console.
3. Text should switch to spanish (some translations missing.)

## Notes

- After discussions with @dan and @gaba I have decided to go with a consistent translation file for each package. All languages are included in the same file to avoid the need to modify imports when adding languages.
- Some spanish translations have been marked as todo with ""¡traduceme!"" rather than trusting google translate.

@coralproject/tech


davidgljay:
Thanks a ton @gaba!

gabelula (approved):
I would love to have each language in its own file but that is something we can look again in the future when everything is set. This is a great move for the first draf of translations.

PR #62, opened 2016-11-11
first pass at permalink box

riley:
## What does this PR do?

First pass at a permalink box as [outlined here](https://www.pivotaltracker.com/story/show/130310029). I'm going to put the code for reaching out of the iframe and scrolling to the comment in another PR.

Thoughts: I tried to do as little styling as possible, which is why it looks kinda janky. Should I be more/less opinionated? How should I allow for configuring?

## How do I test this PR?

- `npm install`
- `http://localhost:3000/admin/embed/stream/preview`
- there should be a preview button with a link to an `/:asset_id#:comment_id`

@coralproject/tech


client/coral-plugin-permalinks/PermalinkButton.js
```diff
@@ -0,0 +1,90 @@
...
+      this.setState({copyFailure: true});
+    }
+
+    setTimeout(() => {
+      this.setState({copyFailure: null, copySuccessful: null});
+    }, 4500);
+  }
+
+  render () {
+    const publisherUrl = 'http://nytimes.com/';
```
gabelula: Can we add it to the settings now and grab it from there?

client/coral-plugin-permalinks/PermalinkButton.js
```diff
@@ -0,0 +1,90 @@
...
+
+    return (
+      <div className={`${name}-container`} style={styles}>
+        <button onClick={this.toggle} className={`${name}-button`}>
+          <i className={`${name}-icon material-icons`} aria-hidden={true}>link</i>
+          Permalink
```
gabelula: i18n is already in master. Can you merge master and use i18n?

gabelula (changes requested):

client/coral-plugin-permalinks/PermalinkButton.js
jde: When embedding these comments, we can't rely on any concept of a base url. These comments may be on properties across sub domains or even domains.

Our best bet is to look at the url that the comment stream is on, use that and append something that says where the view goes to.

client/coral-plugin-permalinks/translations.js
```diff
@@ -0,0 +1,12 @@
+export default {
```
gabelula: the strings should be a translation.json file

gabelula (approved):

PR #63, opened 2016-11-12
Like button

davidgljay:
## What does this PR do?

Adds a like button and the ability to remove a flag to talk.

## How do I test this PR?

1. Like a comment in the stream.
2. The button should change color and the like count should increase.
3. Reload the page. The button should be black but the count should remain.
4. Like and then unlike the comment a second time. The count should decrease to its original number.
5. Reload the page, the appropriate number should display.
6. Flag and unflag a comment. The button should return to its "unflaged" state and you should receive an appropriate notification. 

@coralproject/tech


riley:
When I like a comment and then reload, it's black and has a count of 1 (correct). But when I reload again, like, reload, unlike, and reload the count no longer there. is this correct? I guess we don't have the concept of a user yet. I wasn't sure what "original number" was in #4 above. @davidgljay

client/coral-framework/store/actions/items.js
```diff
@@ -241,9 +241,45 @@ export function postAction (item_id, action_type, user_id, item_type) {
...
+*
+*/
+
+export function deleteAction (item_id, action_type, user_id, item_type) {
+  return () => {
+    const action = {
+      action_type,
+      user_id
+    };
+    const options = {
```
riley: I would recommend pulling these options hashes out to an init function like [here](https://github.com/coralproject/talk/blob/master/client/coral-admin/src/actions/settings.js#L13) if only because we're going to be adding auth to all this stuff, and it make everything easier.

client/coral-framework/store/actions/items.js
```diff
@@ -241,9 +241,45 @@ export function postAction (item_id, action_type, user_id, item_type) {
...
+      method: 'DELETE',
+      headers: {
+        'Content-Type':'application/json'
+      },
+      body: JSON.stringify(action)
+    };
+
+    return fetch(`/api/v1/${item_type}/${item_id}/actions`, options)
+      .then(
+        response => {
```
riley: it's also really nice to have the responseHandler be in one place too.

riley (changes requested):

davidgljay:
@riley I think you have an extra reload in there. With the way that users are being mocked now once you reload you shouldn't be able to 'unlike' until you've 'liked' in that browser session. So it should be:
1. Like
2. Reload
3. Like and unlike.
4. Reload.
5. Like count should be the same as it was at step 2.

riley:
nice! everything is working now 👍

riley (approved):

PR #67, opened 2016-11-15
Added fix for webpack building

wyattjoh:
## What does this PR do?

- Ensures that the building of assets is done with the correct environment flags.

## How do I test this PR?

Try building assets with `npm run build` and `npm run build-watch` and note the differencing in NODE_ENV from the resulting JS.

@coralproject/tech


gabelula (approved):
It looks good.

PR #69, opened 2016-11-16
Multi assets

jde:
## What does this PR do?

Extends the current single asset/single stream model to mock out an entire news site. 
Adds links to the admin section from the mock article pages.

## How do I test this PR?

Navigate to any url that looks like this:

/assets/[title-of-article]

You should get a faux article titled "title of article".

A unique comment stream should be on that article. Comments and actions on a faux article should only appear on that faux article.

@coralproject/tech


davidgljay (approved):
Looks good!

PR #71, opened 2016-11-16
reset password endpoints

riley:
## What is this?

We're using JWTs to verify the user and enable resetting user passwords from the API. This PR adds the functionality to do this.

## How to test

Since the UI is not in master for this, you'll just have to use `curl` or Postman.

- POST to `http://localhost:3000/api/v1/user/request-password-reset` with the body of the request as json with `{"email": "your_email"}`
- if your email is associated with a local profile in the db, you will get an email from coralcore@mozillafoundation.org at that address with a link to reset your password. If no user was found, a 500 is thrown
- the big token in the email only shows up if `NODE_ENV != 'production'`
- copy that token
- make a POST request to `http://localhost:3000/api/v1/user/update-password` with a json payload of `{"token": "the_token", "password": "new_password"}`
- you should get a 204 response that your password was updated.

I have no idea how to test the email component of this PR.


routes/api/user/index.js
```diff
@@ -70,4 +71,59 @@ router.post('/:user_id/role', (req, res, next) => {
...
+  }
+
+  User
+    .createJWT(email)
+    .then(token => {
+      const options = {
+        subject: 'Password Reset Requested - Talk',
+        from: 'coralcore@mozillafoundation.org',
+        to: email,
+        html: `<p>We received a request to reset your password. If you did not request this change, you can ignore this email.
```
jde: Please move the email text into a view so that we can easily keep track of them and edit them without needing to touch js code.

services/mailer.js
```diff
@@ -0,0 +1,43 @@
+const nodemailer = require('nodemailer');
+const sgTransport = require('nodemailer-sendgrid-transport');
+const debug = require('debug')('talk:mail');
+const options = {
+  auth: {
+    api_key: process.env.TALK_SENDGRID_APIKEY
```
jde: Please use an SMTP_CONNECTION_STRING var instead of custom sendgrid code. This will allow us to connect to just about any email service.

jde (changes requested):

services/mailer.js
```diff
@@ -14,7 +8,7 @@ if (!process.env.TALK_SENDGRID_APIKEY) {
         '///////////////////////////////////////////////////////////////');
 }
 
-const transporter = nodemailer.createTransport(sgTransport(options));
+const transporter = nodemailer.createTransport(`smtps://apikey:${process.env.TALK_SENDGRID_APIKEY}@smtp.sendgrid.net`);
```
jde: Instead of building the transport string like this, we should use an env var called TALK_SMTP_CONNECTION_URL and pass that directly into createTransport().  This will allow us to use sendgrid or any other provider.

routes/api/user/index.js
jde: Also, associating a language with the template filename and using the i18n language name will give us the ability to send multilingual emails!

For example, this email could be:

/views/emails/reset_password_EN.ejs

services/mailer.js
```diff
@@ -0,0 +1,37 @@
+const nodemailer = require('nodemailer');
+const debug = require('debug')('talk:mail');
+
+if (!process.env.TALK_SMTP_STRING) {
+  debug('\n///////////////////////////////////////////////////////////////\n' +
+        '///   TALK_SENDGRID_APIKEY should be defined if you would   ///\n' +
```
jde: Comments still say TALK_SENDGRID_APIKEY.  Please update to new var.

routes/api/user/index.js
```diff
@@ -52,4 +58,61 @@ router.post('/:user_id/role', (req, res, next) => {
...
+  if (!email) {
+    return next();
+  }
+
+  User
+    .createJWT(email)
+    .then(token => {
+      const options = {
+        subject: 'Password Reset Requested - Talk',
+        from: 'coralcore@mozillafoundation.org',
```
jde: Please update to "no-reply" to make sure our core email doesn't make it out into the world of spammers and password-forgetters!

models/user.js
```diff
@@ -12,6 +14,13 @@ const USER_ROLES = [
   'moderator'
 ];
 
+if (!process.env.TALK_SESSION_SECRET) {
+  debug('\n////////////////////////////////////////////////////////////\n' +
+        '///   TALK_SESSION_SECRET must be defined to encode      ///\n' +
+        '///   JSON Web Tokens and other auth functionality       ///\n' +
+        '////////////////////////////////////////////////////////////');
+}
+
```
wyattjoh, line 22: Should just throw an error here instead to stay consistent with the other error states of the application. Debug is also never printed unless the DEBUG env var is set.

models/user.js
```diff
@@ -394,6 +410,55 @@ UserService.findByIdArray = (ids) => {
 };
 
 /**
+ * Creates a JWT from a user email. Only works for local accounts.
+ * @param {String} email of the local user
+ */
+UserService.createJWT = function (email) {
```
wyattjoh: Should be named something more specific to the intention, like `UserService.createPasswordResetToken` which describes its use case.

models/user.js
```diff
@@ -394,6 +410,55 @@ UserService.findByIdArray = (ids) => {
...
+
+      return token;
+    });
+};
+
+/**
+ * verifies a jwt and returns the associated user
+ * @param {String} token the JSON Web Token to verify
+ */
+UserService.verifyToken = function (token) {
```
wyattjoh: Same as above, try changing it to `UserService.verifyPasswordResetToken`.

models/user.js
```diff
@@ -394,6 +410,55 @@ UserService.findByIdArray = (ids) => {
...
+      resolve(decoded);
+    });
+  })
+  .then(decoded => {
+    /**
+     * TODO: check the jti from this decoded token in redis
+     * and make an entry if it does not exist.
+     * reject if entry already exists.
+     */
+    return this.findOne({id: decoded.userId});
```
wyattjoh: Prefer the use of `UserService.findById` over using `this.find` to stay consistent.

routes/admin/index.js
```diff
@@ -1,11 +1,15 @@
...
 
 router.get('/embed/stream/preview', (req, res) => {
   res.render('embed-stream', {basePath: '/client/embed/stream'});
 });
 
+router.get('/password-reset/:token', (req, res, next) => {
+  // render a page or something?
+  res.send('ok');
+});
+
```
wyattjoh, line 12: Is this route not implemented?

services/mailer.js
```diff
@@ -0,0 +1,37 @@
+const nodemailer = require('nodemailer');
+const debug = require('debug')('talk:mail');
+
+if (!process.env.TALK_SMTP_CONNECTION_URL) {
+  debug('\n///////////////////////////////////////////////////////////////\n' +
+        '///   TALK_SMTP_CONNECTION_URL should be defined if you would   ///\n' +
+        '///   like to send password reset emails from Talk          ///\n' +
+        '///////////////////////////////////////////////////////////////');
+}
```
wyattjoh, line 8: Same as above, prefer `throw` versus logging out.

wyattjoh (changes requested):

routes/admin/index.js
riley: i think this is going to be part of @okbel's changes in the `passport` branch (or after that)

models/user.js
```diff
@@ -454,7 +455,7 @@ UserService.verifyToken = function (token) {
      * and make an entry if it does not exist.
      * reject if entry already exists.
      */
-    return this.findOne({id: decoded.userId});
+    return this.findById(decoded.userId);
```
wyattjoh: Should be UserService.findById, and these service functions should not use the function keyword, prefer use of arrow functions.

wyattjoh (approved):

PR #72, opened 2016-11-16
Added documentation to the user model + abstracted it

wyattjoh:
## What does this PR do?

Adds documentation to the User model, abstracts the Model as a service.

## How do I test this PR?

Run tests.

@coralproject/tech


models/user.js
```diff
@@ -2,32 +2,72 @@ const mongoose = require('../mongoose');
...
-  roles: {
-    type: [{type: String, enum: ['admin', 'moderator']}]
-  }
+  }, {
+    _id: false
+  })],
+
+  // Roles provides an array of roles (as strings) that is associated with a
+  // user.
+  roles: [String]
```
gabelula, line 67: You are removing the enum in roles, why?

gabelula: ok. I see that you are enforcing it when adding the role.

gabelula (approved):

PR #74, opened 2016-11-18
add(coral-admin): Syntax highlighting on links

impronunciable:
If there are links on a comment those are being highlighted on the moderation queue like this:

highlighted links


gabelula (approved):
It looks good.

PR #77, opened 2016-11-18
Action bugfixes

davidgljay:
## What does this PR do?

Fixes the following bugs:

1. Reply button on replies does not open reply box.
2. Like button fails to work when replies are rendered on the screen.
3. Clicking "flag" and reloading the page increments the like count.
4. Notifications block the last comment in the stream.

## How do I test this PR?

1. Click the "Reply" button on a reply. The reply box should open on the parent comment.
2. Create a stream with several comments and replies. Like buttons should still work.
3. Click the "flag" button and reload the page. The like count on that comment should be unchanged.
4. Post a comment. The notification should not cover the last comment in the stream.

## Note
PR also allows the stream to be direclty accessed outside of an embed at `/embed/stream` for development purposes.

@coralproject/tech


wyattjoh:
I have a improved version of the getActionSummaries that I'll push up here in a sec.

gabelula:
1. The reply box appear on the parent comment but not the Post button. It also is not working.

gabelula (changes requested):
The part of posting to a reply comment seems not to be working.

davidgljay:
This PR has been merged with https://github.com/coralproject/talk/pull/79

PR #79, opened 2016-11-19
User integration

davidgljay:
## What does this PR do?

Integrates user information into comments and actions.

## How do I test this PR?

1. Open the comment stream.
2. You should not be able to like or flag.
3. Log in.
4. You should be able to post a comment. This comment should have the right name.
5. Like and flag a comment.
6. Check admin panel, the post requests to the server should include your user id.
7. Click the "Reply" button on a reply. The reply box should open on the parent comment.
8. Create a stream with several comments and replies. Like buttons should still work.
9. Click the "flag" button and reload the page. The like count on that comment should be unchanged.
10. Post a comment. The notification should not cover the last comment in the stream.

@coralproject/tech


gabelula:
The moderation queue still has 'Anonymous' for all the user names.

riley:
I can't flag a second comment after I test flagged the first one.

davidgljay:
@riley: I tried a few different combinations but couldn't reproduce this flagging error. Can you give more details?

gabelula:
Is it ok to be able to flag a comment that you already liked?

losowsky:
@gabelula yes

gabelula:
Is ok to be able to like and/or flag my own comment?

gabelula:
@davidgljay Can you change it so I can't flag or like my own comment?

davidgljay:
@gaba We haven't had a user story for not being able to like or flag one's own comment. I can put it in as a separate PR, but let's not block these bugfixes for it.

gabelula (approved):
It looks good.

PR #80, opened 2016-11-21
Wordlist

wyattjoh:
## What does this PR do?

This PR adds support for the wordlist as well as cleaning up the comments API + documenting it.

## How do I test this PR?

Run `npm run test`.

@coralproject/tech


routes/api/comments/index.js
```diff
@@ -1,147 +1,111 @@
...
-  Comment.findByStatus('rejected').then(comments => {
-    res.status(200).json(comments);
+  let query;
+
+  if (req.query.status) {
+    query = Comment.findByStatus(req.query.status);
+  } else if (req.query.action_type) {
+    query = Comment.findByActionType(req.query.action_type);
+  } else {
+    query = Comment.all();
```
davidgljay, line 15: I'm curious why we're exposing this end point. I can think of situations where returning all comments in the system would be sensitive, when is it useful?

routes/api/comments/index.js
```diff
@@ -1,147 +1,111 @@
...
-  console.log(req.params);
+router.post('/:comment_id/actions', (req, res, next) => {
+
+  const {
+    user_id,
+    action_type
+  } = req.body;
+
   Comment
-    .removeAction(req.params.comment_id, req.body.user_id, req.body.action_type)
```
davidgljay, line 139: So we're now deleting actions by referencing the specific action id rather than the user, item_id and action type? Sounds fair, though we'll need some updates not included in this PR to make action deletion work.

davidgljay (changes requested):
Looks good, but these proposed changes will break action updates from the frontend. Let's make the necessary updates before merging.

routes/api/comments/index.js
wyattjoh: This endpoint is used by the admin app to retrieve flagged/rejected comments. When we are ready with the FE, middleware will be here to prevent anyone except admin's from hitting the endpoint.

routes/api/comments/index.js
wyattjoh: Because the action summaries contain the action id for the current user, it would be the preferable way to remove actions. It's much cleaner REST'wise to have that considering the data for displaying the "like" box is right there too.

davidgljay (approved):

PR #85, opened 2016-11-22
Scroll to comment

riley:
## What does this PR do?

Fixes how permalinks work. I pulled the styles out into a separate file instead of inlining them in the component. The permalink itself is generated by just looking at the parent document's url instead of using the asset id.

## How do I test this PR?

- open `http://localhost:3000/admin/embed/preview`
- make a bunch of comments
- get the permalink for one of the early-ish comments
- open that link in a new tab
- the page should scroll down

@coralproject/tech


riley:
closing because of some weirdness with github

PR #92, opened 2016-11-23
Scroll to comment

riley:
## What does this PR do?

Fixes how permalinks work. I pulled the styles out into a separate file instead of inlining them in the component. The permalink itself is generated by just looking at the parent document's url instead of using the asset id. I'm also hosting pym locally in the build instead of relying on the flaky npr one.

I had to resort to polling to get some of this stuff to work. Feel free to point and laugh only if you can think of a better way to implement this.

## How do I test this PR?

- open `http://localhost:3000`
- make a bunch of comments
- get the permalink for one of the early-ish comments
- open that link in a new tab
- the page should scroll down

@coralproject/tech


jde:
When testing the PR, I'm seeing a few things not working:

1) When testing on the preview link as described above, the url generated is incomplete:
screenshot 2016-11-25 10 25 54
This test is irrelevant, as the stream is an embed and will only be used on articles. 

2) When testing on a page with an embedded stream, the link generates correctly but the page does not scroll down to the comment in question. There is no visual way to know which comment has been linked to.

impronunciable:
@riley Please solve conflicts

impronunciable:
booooom

impronunciable (approved):

PR #94, opened 2016-11-23
The field is action_type and not action.

gabelula:
## What does this PR do?

It fix the fetch of flagged comments to query action_type and not action. 

## How do I test this PR?

Look at flags comments in the moderation queue.


client/coral-admin/src/AppRouter.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import {Router, Route, IndexRoute, browserHistory} from 'react-router';
 
-import ModerationQueue from 'containers/ModerationQueue';
-import CommentStream from 'containers/CommentStream';
-import Configure from 'containers/Configure';
+import ModerationQueue from 'containers/ModerationQueue/ModerationQueue';
```
jde: Please update filenames to follow the convention that all containers end with "...Container.js".

jde (approved):

PR #95, opened 2016-11-25
Deprecate stream preview

jde:
## What does this PR do?

Removes the non-standard /embed/stream/preview route. 

At this point we are only supporting one method of embedding: the iframe code published by the admin screen. This preview was embedding the stream directly into the page via script tag. This was causing confusion among devs.

## How do I test this PR?

Functionally, not much has changed. I've just removed the route that shouldn't be hit directly.


routes/admin/index.js
```diff
@@ -1,12 +1,8 @@
 const express = require('express');
 const router = express.Router();
 
-router.get('/embed/stream/preview', (req, res) => {
-  res.render('embed-stream', {basePath: '/client/embed/stream'});
-});
-
-// this route is expecting there to be a token in the hash
-// see /views/password-reset-email.ejs
+// GET /password-reset expects an OpenID token in the hash.
```
wyattjoh: This isn't an OpenID token, replace with:

```js
// Get /password-reset expects a signed token (JWT) in the hash.
```

wyattjoh (changes requested):

wyattjoh (approved):

PR #99, opened 2016-11-28
Corrected action upserting functionality

wyattjoh:
## What does this PR do?

Actions created now can not be duplicated for a given item_id, item_type, user_id, action_type.

## How do I test this PR?

- Try and double like or double flag a comment! It shouldn't create two likes/flags!


models/comment.js
```diff
@@ -157,20 +157,17 @@ CommentSchema.statics.changeStatus = function(id, status) {
...
-  // check that the comment exist
-  let action  = new Action({
-    action_type: action_type,
-    item_type: 'comment',
-    item_id: id,
-    user_id: user_id
-  });
-  return action.save();
-};
+CommentSchema.statics.addAction = (item_id, user_id, action_type) => Action.insertUserAction({
```
jde, line 165: I'm loving how thin our routes end up when we rely on middleware for things like auth and db features such as upsert for storage behavior.

jde (approved):

PR #101, opened 2016-11-28
Settings Structure and Coral UI Components

okbel:
# Settings Structure and Coral UI Components

## What does this PR do?

This PR adds:
- Structure for the Settings Tab within the EmbedStream
- Container Structure
- TabBar
- Tab
- Button (now takes `full` as a prop to use full width)
- Base Styles and Material Styles
- Bio Tab (not functional yet)
- CommentHistory Tab (not functional yet)

## Usage 
TabBar takes `onChange` to handle the click  of the Tabs, `activeTab` and `cStyle` like most of the Coral components (Stands for CoralStyle)

``` javascript
        <TabBar onChange={this.handleTabChange} activeTab={activeTab} cStyle='material'>
          <Tab>All Comments (120)</Tab>
          <Tab>Profile Settings</Tab>
        </TabBar>
```

## Arch Notes: 
All the functionality must be within the Container Component. (`SettingsContainer.js`) and everything that is not related to UI, like fetching or so, must be in the main State, not as a class method. *Components* within the `Components` folder should be stateless.

``` javascript
const mapDispatchToProps = dispatch => ({
  getBio: () => dispatch(),
  getHistory: () => dispatch(),
  handleSaveChanges: () => dispatch(),
  handleCancel: () => dispatch()
});
```

It's important to keep the consistency across containers and stateless components.


riley:
could you make it so that you can't select the text on the tab title?

client/coral-embed-stream/src/CommentStream.js
```diff
@@ -46,7 +51,8 @@ const mapDispatchToProps = (dispatch) => ({
   },
   appendItemArray: (item, property, value, addToFront, itemType) =>
     dispatch(appendItemArray(item, property, value, addToFront, itemType)),
-  logout: () => dispatch(logout())
+  logout: () => dispatch(logout()),
+  changeTab: activeTab => dispatch(changeTab(activeTab))
```
riley: can we just delete this entire `mapDispatchToProps`? These are all just passthroughs and (unless I'm misunderstanding something) are adding needless complexity and room for typos.

client/coral-embed-stream/src/CommentStream.js
```diff
@@ -90,137 +96,148 @@ class CommentStream extends Component {
...
-                          </div>
-                          <div className="replyActionsRight">
-                            <FlagButton
+
+          <TabBar onChange={this.props.changeTab} activeTab={activeTab}>
+            <Tab><Count id={rootItemId} items={this.props.items}/></Tab>
+            <Tab>Settings</Tab>
+          </TabBar>
+
+          <ContentBox show={activeTab === 0}>
```
riley: could we call this something like `<TabContent>` so we know they belong together? or is this being used somewhere else?

client/coral-framework/actions/embedStream.js
```diff
@@ -0,0 +1,22 @@
+import {base, handleResp, getInit} from '../helpers/response';
```
riley: could we get a comment block at the top here explaining generally what this file does? `embedStream` seems a bit vague (and I'm a bit dense)

client/coral-settings/containers/SettingsContainer.js
```diff
@@ -0,0 +1,63 @@
+import React, {Component} from 'react';
+import {connect} from 'react-redux';
+
+import {} from '../../coral-framework/actions/embedStream';
```
riley: what does this line accomplish?

client/coral-settings/containers/SettingsContainer.js
```diff
@@ -0,0 +1,63 @@
...
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = state => ({
+  embedStream: state.embedStream.toJS()
+});
+
+const mapDispatchToProps = dispatch => ({
```
riley: again, do we need this `mapDispatchToProps` at all?

riley (changes requested):
looking good! just a few changes/clarifications

client/coral-framework/actions/embedStream.js
okbel: Sure 👍  Every view should have it's reducer and its actions. So this should be every action that we can do within the embedStream view

client/coral-settings/containers/SettingsContainer.js
okbel: This will import the actions that embedStream will have in the future. As I explained, this is a PR without the functionality for bio. So there is where we should grab the actions for CommentHistory and Bio

client/coral-framework/actions/embedStream.js
riley: I feel like that architecture got us into trouble with Ask. We had a "Forms" action/reducer that went crazy as we added features. I advocate breaking things up by how they're stored in the db like `Users` `Comments` or whatever instead of matching up to a view. Also if we want to move some actions to another file, it makes it weird.

client/coral-embed-stream/src/CommentStream.js
okbel: This was discussed with Riley via Zoom

client/coral-embed-stream/src/CommentStream.js
okbel: The name of that Component it's not strictly related to the tab. But we can change the name

okbel:
## Changes Added 

- Tab title cannot be selected
- `ContentBox` renamed to `TabContent`
- `embedStream` reducers, actions were deleted. and actions were added to the `CommentStream.js` class. (_I don't entirely agree with this choice_)

## Meeting Notes and Changelog notes
The usage of the `mapDispatchToProps` was explained in a separate meeting due to its complexity and importance.
I wouldn't encourage having multiple states. It blurs the meaning of the mainState. Also, I strongly believe that the correct usage of this structure will help us enormously in the long run. It's proven to be scalable. It's important to emphasise that this structure wasn't used in Ask. 

## Output

Tabs functionality: (more styling coming)
out

client/coral-embed-stream/src/CommentStream.js
riley: yeah, I was just confused. I think this is a better way to do it than how we did it in cay.

riley (approved):

PR #102, opened 2016-11-28
Config auth fix

riley:
## What does this PR do?

The auth cookies are not sent by default with our `fetch` library. @okbel had created helper classes, but we didn't implement them everywhere. Fixed now!

Also, I snuck in a tiny refactor of the item actions to remove loops in favor of functional programming for clarity. Ssh

## How do I test this PR?

- log into the admin
- you should be able to see comments to be moderated and should be able to update config settings


okbel:
Looks great!

But we also need to add it here `client/coral-admin/src/services/talk-adapter.js`

```
Promise.all([
  fetch('/api/v1/queue/comments/pending'),
  fetch('/api/v1/comments?status=rejected'),
  fetch('/api/v1/comments?action_type=flag')
])
.....
```

okbel (approved):

PR #103, opened 2016-11-28
Admin Logout

okbel:
## What does this PR do?
- Adds logout
- Solves that glitch with the Communicorn

## How do I test this PR?

- Log in as an Admin
- You should see the Settings Icon. Click it
- Log out


gabelula (changes requested):
Move https://code.getmdl.io/1.2.1/material.indigo-pink.min.css into our own CDN

gabelula (approved):
We move content into own CDN in other PR after discussing in github issues.

PR #109, opened 2016-11-28
Settings in stream

davidgljay:
## What does this PR do?

Moves settings to the stream call

## How do I test this PR?

1. Load the comment stream
2. Check the network tab, there should not be a call to `api/v1/settings`
3. Check the redux store, the settings should be visible in the `config` section of the store
4. All tests should pass


routes/api/stream/index.js
```diff
@@ -30,29 +29,18 @@ router.get('/', (req, res, next) => {
...
+  .then(([asset, settings]) => {
+    // Get the sitewide moderation setting and return the appropriate comments
+    switch(settings.moderation){
+    case 'pre':
+      return Promise.all([Comment.findAcceptedByAssetId(asset.id), asset, settings]);
+    case 'post':
+      return Promise.all([Comment.findAcceptedAndNewByAssetId(asset.id), asset, settings]);
+    default:
+      return Promise.reject(new Error('Moderation setting not found.'));
     }
```
wyattjoh, line 39: The switch was replaced because the default described in the database is "pre". Because there can be only two states it made sense to not complain in the event of a missing setting. I'd prefer to retain this as the single if check as before and simply add the settings object to the returned `Promise.all` below.

wyattjoh (changes requested):

routes/api/stream/index.js
davidgljay: @wyattjoh, this change has been implemented.

wyattjoh (approved):

PR #112, opened 2016-11-28
keep it DRY. add getInit all over the place. update tests

riley:
## What does this PR do?

the `helpers/response` that provides the init params for our fetch requests was not implemented everywhere, causing all sorts of regressions. I think we've gotten most (if not all) at this point. the helpers class was also duplicated, which seemed like a recipe for disaster. 

@wyattjoh pointed out that a better way to do this would be to just return the fetch Promise instead of the params so we never forget. That will come in a later PR.

## How do I test this PR?

- update settings and stuff
- post comments
- `npm test`
- everything should work


davidgljay (approved):
Looks good.

PR #115, opened 2016-11-29
Bugfix: Like refresh

davidgljay:
## What does this PR do?

If a user likes a comment, that comment now stays liked by that user on page refresh.

## How do I test this PR?

1. Like a comment.
2. Reload the page.
3. The like button should stay blue.
4. Unlike the comment by clicking it.
5. Reload the page.
6. The like button should be black.


kgardnr:
Dupe.

PR #117, opened 2016-11-29
retreiving comments based on settings was backwards

riley:
## What does this PR do?

The stream endpoint was returning the wrong comments e.g. `pre` comments when `post` was set. This PR fixes it and updates the tests.

## How do I test this PR?

- remove all your comments
- have moderation to `post`
- make a comment. it should appear
- refresh the page. it should still be there.


davidgljay (commented):
Confirmed: the comment stream now behaves as expected.

davidgljay (approved):

PR #119, opened 2016-11-29
implement coralApi fetch

riley:
## What does this PR do?

We were wasting a lot of time implementing `fetch` with my previous `getInit` function. If you hand-rolled your fetch request (tsk tsk) it would cause regressions even if `getInit` was updated. Or if we added auth and your response handler wasn't updated things would break.

so from now on, you're not going to implement `fetch` on your own. Import `coralApi` from the `helpers/response.js` in coral-framework. The signature is `coralApi(<url>, <options>)` where `options` will override the defaults set in the helper. No more having to worry about auth and all that stuff!

## How do I test this PR?

- `npm test`
- tests should all still pass


davidgljay (changes requested):
I get strange behavior when logged out. The user display name is blank and the 'Sign Out' button still appears. Clicking this button generates a 'Not Authorized' error. 
screen shot 2016-11-29 at 5 12 01 pm

davidgljay (approved):

PR #120, opened 2016-11-30
User Bio and Settings Tab

okbel:
## What does this PR do?

This PR adds

- User Bio and Settings Tab
- Tooltip Component (for the user bio hovering the Author name)
- RestrictedContent
- Adds spinner to embedStream
- Banned/Suspended User UI (with RestrictedContent) 
- authUser !=== userData (settings, bio)

## RestrictedContent

You can wrap all the secured content within the `RestrictedContent` Container it will show a default message `This content is not available` if restricted is true. If no component is provided you can send a message.

``` javascript
// Params:
restricted: <Boolean>
restrictedComp: <ReactElement>
message: <String>
``` 

### Usage

``` javascript
<RestrictedContent restricted={!loggedIn} restrictedComp={NotLoggedIn}>
 // Restricted Stuff
</RestrictedContent>
```

## How do I test this PR?

LoggedIn
- Check the settings tab
- Add a user bio
- Hover names in the embedStream

Logged Out
 - Check the settings tab


client/coral-embed-stream/src/CommentStream.js
```diff
@@ -109,152 +85,181 @@ class CommentStream extends Component {
...
-                          <Content body={reply.body}/>
-                          <div className="replyActionsLeft">
-                              <ReplyButton
-                                updateItem={this.props.updateItem}
-                                id={replyId}
-                                showReply={reply.showReply}/>
-                              <LikeButton
+          {loggedIn && <UserBox user={user} logout={this.props.logout} />}
+          {/* Add to the restricted param a boolean if the user is suspended*/}
+          <RestrictedContent restricted={false} message={lang.t('suspendedAccountMsg')}>
```
davidgljay: If possible we should handle language in these components directly rather than in Comment Stream. This will make it easier to swap components in and out of commentStream in the future.

client/coral-plugin-commentbox/CommentBox.js
```diff
@@ -1,6 +1,7 @@
 import React, {Component, PropTypes} from 'react';
 import {I18n} from '../coral-framework';
 import translations from './translations.json';
+import {Button} from 'coral-ui';
```
davidgljay, line 4: Is `coral-ui` published to npm? I thought we were trying to minimize custom UI in talk. No objections, just want to make sure that this reflects an intentional change in our earlier decisions.

client/coral-settings/components/NotLoggedIn.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import styles from './NotLoggedIn.css';
+
+export default () => (
+  <div className={styles.message}>
+    <div>
+      <a>Sign In</a> to access Settings
```
davidgljay: Is there a reason this isn't a live link?

davidgljay (changes requested):
A few questions. The only change I'd request is a notification when the user's bio saves successfully. (Right now it gives the user no feedback.)

client/coral-plugin-commentbox/CommentBox.js
okbel: No, it's not published. It's in the 'coral-ui' root folder. These are the basic needed components that go with the Coral Styleguide, so in the future we can add something like, customStyles and modify them all by just importing that. (like we did with Cay)

okbel:
@davidgljay Ready to be tested! I run into a lot of merge issues, if someone can check that I didn't delete anything from CommentStream before merging, that would be great

davidgljay:
One more question: is the cancel button supposed to do anything?

screen shot 2016-12-01 at 10 51 18 am

davidgljay:
The "Cancel" button seems to generate a SAVE_BIO_REQUEST.

davidgljay (approved):

PR #122, opened 2016-12-02
Banned word list

davidgljay:
## What does this PR do?

Adds the ability to define banned words in the admin panel. Automatically rejects comments which include these words and notifies the user.

## How do I test this PR?

1. Go to `http://localhost:3000/admin/configure`
2. Click 'Banned words list`
3. Enter one or more comma separated words. "Save" button should turn green.
4. Click "Save Changes"
5. Navigate to `http://localhost:3000`
6. Enter a comment with a banned word.
7. You should receive a notification informing you that your comment includes a banned word.
8. Go to `http://localhost:3000/admin`
9. You should see you comment under the "rejected" tab.

## Note

A few spanish translations are missing in `/client/coral-admin/src/translations.json` and in `/client/coral-framework/translations.json`. Search for the phrase `¡traduceme!` if you'd like to help out.


jde:
Comments submitted in this branch (with Post mod set) do not immediately appear. Refreshing afterward shows them.

davidgljay:
@jde I'm not able to reproduce this error, perhaps it was addressed by the latest merge? Please confirm that it still appears on your system.

riley:
I'm getting that error too when I try to post a comment from a new user with "post" as the moderation setting. here is the error I get in my console (the first error is on page load with existing comments, the second is when I try to create a new comment):

screen shot 2016-12-06 at 11 44 31 am

riley:
ok, I pushed a fix that's going into a different PR to get rid of the first error above, but I'm still getting the same `author_id is undefined`.

davidgljay:
@riley I've addressed the `auhor_id` error and updated the branch to reflect changes in the way that statuses are handled. Can you take a look again?

riley:
I put in `ducks` as a banned word and hit save, but was still able to leave comments with no notifications. I had post-moderation on if that matters. The comment posted immediately, and the comment was still there with ducks in it after reload.

davidgljay:
@riley Just pushed a merge, though I couldn't reproduce your error. Can you ping me when you get online and we'll investigate?

gabelula:
Yes. I had the same issue as Riley. I was able to publish a comment (post-moderation) with the banned word.

gabelula (changes requested):
It is not banning words for me (using Firefox  & post-moderation settings).

davidgljay:
Ahh, I see. Looks like there was a second update to the way that status is passed back to the frontend. (It went from `status` being an array to the array being in `status_history`). The issue was introduced by a merge.

This api has changed multiple times while this PR has been getting reviewed for merge, we should find a way to manage these sorts of changes.

jde:
Agreed that we need a more solid process for communicating api changes. In this case, it was changed then changed back. There are a few of these issues causing dev woes. @kgardnr and I are rolling out some process for the new year.

gabelula (approved):

PR #125, opened 2016-12-02
Adjusting height of login button to match current comment

davidgljay:
## What does this PR do?

Adjusts height of the login box when clicking "like" or "flag" button further down the stream.

## How do I test this PR?

1. Create several comments.
2. Click "like" or "flag" on one of the comments lower in the stream.
3. Login box should appear in front of that comment.
4. The iframe should expand to account for the login box.


okbel:
Is this about the height or the position? If is about the position shouldn't this get fixed with `position:fixed` Within the dialog? @davidgljay

davidgljay:
It's about the height. Say that I click "Like" at a comment at the bottom of the stream. The login box appears, but the bottom of the box is cut off because the iFrame isn't big enough. This PR adjusts the height of the iFrame to prevent this problem.

okbel (approved):

PR #126, opened 2016-12-03
Adds a variable to the possible translations.

gabelula:
## What does this PR do?

Adds a variable to the strings to be included in the translation.

## How do I test this PR?

Adds a translation with a variable and check that it works.

## Riley Edit

added ability to accept `n` arguments to translation function. so the string:

`Hello {0} {1}!`

can be called with `lang.t('header.greeting'), 'Riley', 'Davis')` and it'll be translated to

`Hello Riley Davis!`

### Second edit

now you can translate like so > `Hello {0} {1}. Your first name is {0}` 
With `lang.t('header.greeting', 'Riley', 'Davis')` => `Hello Riley Davis. Your first name is Riley`

### DIFF
--- a/client/coral-framework/modules/i18n/i18n.js
+++ b/client/coral-framework/modules/i18n/i18n.js
@@ -1,5 +1,7 @@
 import timeago from 'timeago.js';
 import esTA from '../../../../node_modules/timeago.js/locales/es';
+import has from 'lodash/has';
+import get from 'lodash/get';
 
 /**
  * Default locales, this should be overriden by config file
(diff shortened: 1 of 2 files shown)


client/coral-framework/modules/i18n/i18n.js
```diff
@@ -58,8 +60,9 @@ class i18n {
         return key;
       }
 
-      const val = String(translation);
+      let val = String(translation);
       if (val) {
+        val = val.replace('/\{0\/g', att);
```
davidgljay: I tested this line and it didn't seem to work as desired. Should it be:

`val = val.replace(/\{0\}/g, att);`?

davidgljay (changes requested):

client/coral-framework/modules/i18n/i18n.js
gabelula: YES!!

davidgljay (approved):

PR #128, opened 2016-12-05
Status history

wyattjoh:
Added support for comment status history with user moderation actions being tracked as well as asset specific queue endpoints.


models/comment.js
```diff
@@ -1,9 +1,37 @@
...
+const Schema = mongoose.Schema;
 const uuid = require('uuid');
 const Action = require('./action');
 
-const Schema = mongoose.Schema;
+/**
+ * The Mongo schema for a Comment Status.
+ * @type {Schema}
+ */
+const StatusSchema = new Schema({
```
jde, line 10: Could we add a field in here to store the kind of action that was taken in order to set this status:

* "moderator" - an action explicitly taken by a moderator
* "blacklist" - a status set because of a match against the blacklist
* "" - default setting

We could possibly used the assigned_by field to capture this information, however I wouldn't want to use a single field to capture both a moderator's id and something like "blacklist"

routes/api/settings/index.js
```diff
@@ -4,19 +4,21 @@ const Setting = require('../../../models/setting');
...
-  Setting
-    .getSettings()
-    .then(settings => {
-      res.json(settings);
-    })
-    .catch(next);
+  Setting.retrieve().then((settings) => {
+    res.json(settings);
+  })
+  .catch((err) => {
```
davidgljay, line 10: This seems more verbose than the previous implementation of catch, why the change?

routes/api/stream/index.js
```diff
@@ -26,8 +26,8 @@ router.get('/', (req, res, next) => {
         return asset;
       }),
 
-    // Get the public settings.
-    Setting.getPublicSettings()
+    // Get the moderation setting from the settings.
+    Setting.retrieve()
```
davidgljay, line 30: This will return all settings including banned words to the client, is that the intention?

models/comment.js
wyattjoh: We should create an additional story for that, currently the story doesn't have this field added.

routes/api/settings/index.js
wyattjoh: This way we don't bind the next call, this was on @okbel's suggestion I believe.

routes/api/stream/index.js
wyattjoh: Not really, I'll take a peek at a solution.

davidgljay (approved):

PR #129, opened 2016-12-06
Asset comment settings

okbel:
## What does this PR do?

This PR adds

- Coral UI 
  - Checkbox
- Configure Stream Tab Functionality
- UX/UI

Note: 'Pre-Moderate Comments Containing Links' is ready on the frontend, once we have that in the backend, we'll tie that up.

## How to test this?

- Go to the comment stream
- Go to the third tab, Configure Stream
- Configure the stream 
- Click some checkbox, they will turn green and as soon as you hit apply the will turn back grey. 💃 (approved by Emma )


okbel:
Would you mind taking a look at this one, @davidgljay and @wyattjoh? //cc @kgardnr

routes/api/asset/index.js
```diff
@@ -82,18 +82,11 @@ router.post('/:asset_id/scrape', (req, res, next) => {
...
     .overrideSettings(req.params.asset_id, req.body)
-    .then(() => {
-
-      res.status(204).end();
-    })
-    .catch((err) => {
-      next(err);
-    });
-
+    .then(({settings}) => res.status(200).send({settings}))
```
wyattjoh: Updates can not return a payload, this should be reverted to return 204 without a body.

wyattjoh (changes requested):

wyattjoh:
This should not be using the `Asset.status` field, and instead should defer to what we have in the close-comments (https://github.com/coralproject/talk/tree/close-comments) branch.

okbel:
@wyattjoh I'll update the PR as soon as `close-comments` branch gets merged

client/coral-configure/components/ConfigureCommentStream.js
```diff
@@ -0,0 +1,51 @@
...
+      <h3>Configure Comment Stream</h3>
+      <p>
+        As an admin you may customize the settings for the comment stream for this article
+      </p>
+      <Button
+        className={styles.apply}
+        cStyle={changed ? 'green' : 'darkGrey'}
+        onClick={handleApply}
+      >
+        Apply
```
davidgljay: This and other text should be handled via i18n.

client/coral-configure/components/ConfigureCommentStream.css
```diff
@@ -0,0 +1,37 @@
+.container {
```
davidgljay, line 1: Earlier we made a decision to handle all comment stream css in default.css for ease of reconfigurability. I'm open to revisiting that decision (it probably generally makes sense to reexamine our strategy for customizing style), but we should be intentional about it before changing norms.

client/coral-framework/reducers/config.js
```diff
@@ -1,30 +1,28 @@
-/* @flow */
-
 import {Map} from 'immutable';
-import * as actions from '../actions/config';
+import * as actions from '../constants/config';
```
davidgljay, line 2: This makes sense as a standard, but it's not reflected for items and notifications. Should we make that update here or put it in a separate PR? If we're going to refactor we should make it consistent throughout.

davidgljay (changes requested):
One change requested around i18n and a few standards questions.

client/coral-configure/components/ConfigureCommentStream.css
okbel: I agree. The idea of having the `default.css` was for customizing and creating themes. This idea still remains. But components need to have core styling such as positions, and vertical-aligns. The app must still work regardless of the `default.css`. That being said, shouldn't the work of `default.css` only be for theming and overriding purposes?

okbel:
@wyattjoh @davidgljay Everything updated. Thanks!

gabelula (approved):
Ready to merge.

gabelula (approved):

gabelula (approved):

PR #133, opened 2016-12-08
Flag username

davidgljay:
## What does this PR do?

Allows a user to flag a username or comment and provide details.

## How do I test this PR?

1. Click "Report" underneath a comment.
2. Select whether to report a comment or username. Click continue.
3. Select a reason or enter one via "other." Click continue.
4. If you flagged the comment, the flag should be red.
5. If you flagged the username, it should stay black.
6. Check the network panel, action should be posted to the server at `/v1/api/comments/:comment_id/actions` or `/v1/api/users/:user_id/actions`


impronunciable:
- I really like how it works :)

- The popup is not hiding even clicking outside

screen shot 2016-12-08 at 12 55 02 pm

client/coral-embed-stream/src/CommentStream.js
```diff
@@ -272,8 +274,8 @@ const mapDispatchToProps = (dispatch) => ({
   getStream: (rootId) => dispatch(getStream(rootId)),
   addNotification: (type, text) => dispatch(addNotification(type, text)),
   clearNotification: () => dispatch(clearNotification()),
+  postAction: (item, action, itemType, field, detail) => dispatch(postAction(item, action, itemType, field, detail)),
```
impronunciable: At this point we should use an object instead of positional arguments

client/coral-framework/actions/items.js
```diff
@@ -217,11 +217,12 @@ export function postItem (item, type, id) {
 *
 */
 
-export function postAction (item_id, action_type, user_id, item_type) {
+export function postAction (item_id, action_type, item_type, field, detail) {
```
impronunciable: same as above, too many positional params

client/coral-plugin-flags/FlagButton.js
```diff
@@ -1,48 +1,179 @@
...
+  }
+
+  getPopupMenu = (step) => {
+    switch(step) {
+    case 1: {
+      return {
+        header: lang.t('step-1-header'),
+        options: [
+          {val: 'user', text: lang.t('flag-username')},
+          {val: 'comments', text: lang.t('flag-comment')},
```
impronunciable: remove extra comma

impronunciable: I  think there are extra commas in every array, if this is intended that's ok

models/action.js
```diff
@@ -35,12 +37,14 @@ ActionSchema.statics.findById = function(id) {
  * @param {String} action the new action to the comment
  * @return {Promise}
  */
-ActionSchema.statics.insertUserAction = ({item_id, item_type, user_id, action_type}) => {
+ActionSchema.statics.insertUserAction = ({item_id, item_type, user_id, action_type, field, detail}) => {
```
impronunciable: Too many params

client/coral-embed-stream/src/CommentStream.js
riley: I'm using onClickOutside or something like that in Permalink if you want to check it out @davidgljay

routes/api/comments/index.js
```diff
@@ -148,13 +148,8 @@ router.put('/:comment_id/status', authorization.needed('admin'), (req, res, next
...
 
 router.post('/:comment_id/actions', (req, res, next) => {
-
-  const {
-    action_type
-  } = req.body;
-
   Comment
-    .addAction(req.params.comment_id, req.user.id, action_type)
+    .addAction(req.params.comment_id, req.user.id, req.body)
```
wyattjoh: We should extract only the parameters we need from the body, not the whole one, this would result in anyone being able to stuff whatever they want in our db.

models/comment.js
```diff
@@ -284,11 +284,11 @@ CommentSchema.statics.pushStatus = (id, status, assigned_by = null) => Comment.u
  * @param {String} action the new action to the comment
  * @return {Promise}
  */
-CommentSchema.statics.addAction = (item_id, user_id, action_type) => Action.insertUserAction({
+CommentSchema.statics.addAction = (item_id, user_id, action) => Action.insertUserAction({
   item_id,
   item_type: 'comment',
   user_id,
-  action_type
+  action
```
wyattjoh: This style (hiding fields inside this new `action` value) makes this difficult to read.

models/user.js
```diff
@@ -546,3 +547,17 @@ UserService.addBio = (id, bio) => (
...
+ * @param {String} item_id  identifier of the user  (uuid)
+ * @param {String} user_id  user id of the action (uuid)
+ * @param {String} action the new action to the user
+ * @return {Promise}
+ */
+UserService.addAction = (item_id, user_id, action) => Action.insertUserAction({
+  item_id,
+  item_type: 'comment',
+  user_id,
+  action
```
wyattjoh: Same as above.

routes/api/user/index.js
```diff
@@ -157,4 +157,16 @@ router.put('/:user_id/bio', (req, res, next) => {
     });
 });
 
+router.post('/:user_id/actions',  authorization.needed(), (req, res, next) => {
+
+  User
+    .addAction(req.params.comment_id, req.user.id, req.body)
```
wyattjoh: Same as above, we shouldn't just pull the entire body.

wyattjoh (changes requested):

models/comment.js
davidgljay: Agreed, @dan requested this change. I can revert.

models/action.js
davidgljay: The alternative is to pass in a whole object, which seems messy (and frought, see @wyatt's comments below.) I'm leaving this as is.

impronunciable:
I still can't get rid of the popup

impronunciable:
Functionality is there :)

routes/api/user/index.js
```diff
@@ -158,9 +158,15 @@ router.put('/:user_id/bio', (req, res, next) => {
 });
 
 router.post('/:user_id/actions',  authorization.needed(), (req, res, next) => {
+  console.log('Hit action endpoint');
```
wyattjoh: Seems this one snuck through

tests/routes/api/comments/index.js
```diff
@@ -350,7 +350,7 @@ describe('/api/v1/comments/:comment_id/actions', () => {
       return chai.request(app)
         .post('/api/v1/comments/abc/actions')
         .set(passport.inject({id: '456', roles: ['admin']}))
-        .send({'action_type': 'flag'})
+        .send({'user_id': '456', 'action_type': 'flag'})
```
wyattjoh: This shouldn't actually do anything, should revert this line.

routes/api/user/index.js
davidgljay: Oops! My apologies.

jde:
This is looking amazing!  Two details:

1) Agree that we need a close button on the last step. Clicking off the modal "works" but I would argue is not a good UI solution.
2) When data is put in the action collection, the `item_id` field is `null`:

6pyfcs2uw6_ckdvthzf3n00toru1dhluhpqklrhbxio

jde:
@losowsky confirmed the need to close:

We should always have the ability to close the box (currently if you're in the flow, there's not way to get out.) Clicking off the box should always close it (it doesn't close now if you're on step 2.)

wyattjoh (approved):

PR #140, opened 2016-12-08
Simplifies redux's handling of actions in CommentStream.

davidgljay:
## What does this PR do?

Updates CommentSream to automatically make most actions available as props. This prevents us from needing to update actions in commentStream whenever they change.

## How do I test this PR?

1. Tests should pass.
2. Review code change and form opinions.
3. Opinions should be mostly positive.

### DIFF
--- a/client/coral-embed-stream/src/CommentStream.js
+++ b/client/coral-embed-stream/src/CommentStream.js
@@ -29,11 +29,6 @@ import RestrictedContent from '../../coral-framework/components/RestrictedConten
 import SuspendedAccount from '../../coral-framework/components/SuspendedAccount';
 import CloseCommentsInfo from '../../coral-framework/components/CloseCommentsInfo';
 
-const {addItem, updateItem, postItem, getStream, postAction, deleteAction, appendItemArray} = itemActions;
-const {addNotification, clearNotification} = notificationActions;
-const {logout, showSignInDialog} = authActions;
-const {updateOpenStatus} = configActions;
-
 class CommentStream extends Component {
 
   constructor (props) {
@@ -59,8 +54,8 @@ class CommentStream extends Component {
 
   static propTypes = {
     items: PropTypes.object.isRequired,
-    addItem: PropTypes.func.isRequired,
-    updateItem: PropTypes.func.isRequired
+    auth: PropTypes.object.isRequired,
+    config: PropTypes.object.isRequired
   }
 
   componentDidMount () {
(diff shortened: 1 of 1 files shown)


client/coral-embed-stream/src/CommentStream.js
```diff
@@ -284,20 +279,14 @@ const mapStateToProps = state => ({
...
-  appendItemArray: (item, property, value, addToFront, itemType) => dispatch(appendItemArray(item, property, value, addToFront, itemType)),
-  handleSignInDialog: () => dispatch(authActions.showSignInDialog()),
-  logout: () => dispatch(logout()),
-  updateStatus: status => dispatch(updateOpenStatus(status))
-});
+// Ties all imported actions to dispatch and makes them available in props.
+const mapDispatchToProps = (dispatch) => {
+  const actions = Object.assign({}, itemActions, notificationActions, authActions, configActions);
+
+  return Object.keys(actions).reduce((dispatchActions, key) => {
```
okbel, line 286: I don't think we should do this. Adding another iteration attempts to the performance. And the reason behind `mapDispatchToProps` is that we are 100% aware of what our component can do. 
Something that we can do to make this file smaller is abstracting the functionality into smaller components. This Component is making a lot of things and it shouldn't. 

 Breaking this into smaller components like we have with the `coral-sign-in` `coral-settings` or `coral-configure`, in my opinion, is the way to go.

davidgljay: That seems fine, I'm interested in:

1) Defining and sticking to some established API of actions that we make available to plugin developers eventually.
2) Not asking plugin developers to worry about a mapDispatchToProps or mapStateToProps function. 

It may be that there are multiple parts of our application (in-stream admin settings, the main comment stream, user settings, etc) that a plugin developer can connect into. If so we should maybe draw a distinction between a part of the application (`coral-sign-in`) that we don't really intend to be replaceable by a plugin and a part of the architecture (`coral-plugin-like-button`) that we do.

jde: +1 to keeping the mapping and imports explicit.

I agree we will eventually need to start drawing distinctions for plugin developers. I'm not sure we're ready to start making those distinctions yet though.

My recommendation is to:

1.  break up CommentStream into a series of smaller components (plugins for those that can be useful elsewhere, more components in the commentstream embed for ones that are specific.)
1. finish out the features through .5 alpha (more or less) keeping things as simple and straight forward as possible, also focusing on team practices toward keeping a stable build, smaller prs, etc... during that time
1. do a plugin review and add more definition to them once we have simple, stable code for all of our features.

gabelula (changes requested):
Should we have a Zoom meeting about this and agree on what to do next?

client/coral-embed-stream/src/CommentStream.js
davidgljay: It seems like there's enough resistance that I'll go ahead and remove this suggestion.

PR #134, opened 2016-12-08
Close comments

impronunciable:
Adds functionality for closing comments on a thread and a closed comments message in the admin

## How to test?

- Go to the admin settings
- Create a `closed comments` message
- Go to a thread
- There is a third tab for closing comments
- Close the comments
- Go back to the first tab, the comments tab
- It should display that message

PR #144, opened 2016-12-09
Adds status field to the asset and changeStatus

gabelula:
## What does this PR do?

- Adds status field to the Asset model
- Adds changeStatus method to the Asset model
- It changes the endpoint to change status to use that new method
- Adds tests

## How do I test this PR?

- Check that the stream gets closed and open and that it works.


impronunciable (commented):
looks good

gabelula:
This is still relevant as we do not have backend for the status.

wyattjoh:
The closed status of the asset is fully represented by the closedAt field, so i don't see the point of this as it seems like we are duplicating effort.

wyattjoh:
See this PR https://github.com/coralproject/talk/pull/134 for the closed at field which provides the Opened/Closed status.

PR #150, opened 2016-12-09
remove cruft by pulling out some immutable.js stuff

riley:
## What does this PR do?

I'm going to go through the dumb components and removing where we directly access immutable.js objects because it's just noise. I'm starting from the bottom up, so there will be several of these PRs. The goal is to convert the `mapStateToProps` to call `toJS()` at render time so that we don't have to call `comments.get('status')` or whatever every single time.

I've modified `ModerationQueue` and `CommentStream` in this PR.

## How do I test this PR?

- go to the admin
- you should still be able to see the comments in ModerationQueue and interact with them. No functionality change should have happened


client/coral-admin/src/containers/ModerationQueue/ModerationQueue.js
```diff
@@ -79,6 +79,11 @@ class ModerationQueue extends React.Component {
     const {comments, users} = this.props;
     const {activeTab, singleView, modalOpen} = this.state;
 
+    const c = comments.toJS();
+    const premodIds = c.ids.filter(id => c.byId[id].status === 'premod');
```
jde: I like the standard of moving logic like this out of the fray and using a meaningful variable name.

client/coral-admin/src/components/CommentList.js
```diff
@@ -119,16 +121,20 @@ export default class CommentList extends React.Component {
...
+    let {active} = this.state;
+
+    const users = this.props.users.toJS();
+    const comments = this.props.comments.toJS();
 
     return (
       <ul className={`${styles.list} ${singleView ? styles.singleView : ''}`} {...key}>
         {commentIds.map((commentId, index) => {
-          const comment = comments.get(commentId);
+          console.log('inside the map', typeof commentId, commentId, typeof active, active);
```
jde: Please remove

jde (changes requested):

client/coral-admin/src/components/CommentList.js
```diff
@@ -81,12 +83,12 @@ export default class CommentList extends React.Component {
...
     const {active} = this.state;
     // check boundaries
-    if (active === null || !commentIds.size) {
-      this.setState({active: commentIds.get(0)});
-    } else if (direction === 'up' && active !== commentIds.first()) {
-      this.setState({active: commentIds.get(commentIds.indexOf(active) - 1)});
-    } else if (direction === 'down' && active !== commentIds.last()) {
-      this.setState({active: commentIds.get(commentIds.indexOf(active) + 1)});
+    if (active === null || !commentIds.length) {
+      this.setState({active: head(commentIds)});
```
jde: I would recommend against using a library method with the native object already has a method that accomplishes the same thing.

client/coral-admin/src/components/CommentList.js
```diff
@@ -81,12 +83,12 @@ export default class CommentList extends React.Component {
...
+    } else if (direction === 'up' && active !== head(commentIds)) {
```
jde: I would recommend against using a library method with the native object already has a method that accomplishes the same thing. Does _.head([]) provide deeper functionality than [].first()?

riley: `[].first()` isn't a method unless you add it, to my knowledge.

riley: `.first()` is an immutable thing afaik

davidgljay (approved):
Looks good! Looking through this I've found some weirdness in the action (where we pull state into an action using getState rather than handling it in a reducer, which makes way more sense.) But that's tangential enough to this that it doesn't matter. Approved!

Issue #149, opened 2016-12-09
Configure Stream tab should not be visible for non admins


jde:
### Expected behavior

Users who don't have the admin role (including non-logged in users), should not see or have any way of gaining access to the Configure Stream tab in the comment stream.

### Actual behavior

The tab is visible for everyone, but if you don't have access we see this:

![configure](https://cloud.githubusercontent.com/assets/259957/21062932/32648d0c-bdf7-11e6-9a00-a45311177c8f.png)

### Steps to reproduce behavior

Go to a stream
log out
You shouldn't see the Configure Stream tab

Go to a stream
log in as a non-admin
You shouldn't see the Configure Stream tab

okbel:
Ready to be tested! https://github.com/coralproject/talk/pull/160 // cc @kgardnr

okbel:
Merged!

PR #160, opened 2016-12-12
Configure Tab

okbel:
## What does this PR do?
- Solves issue https://github.com/coralproject/talk/issues/149
- Adds prop `restricted` to the `<Tab>` Component (optional)
- Adds `isAdmin` to the auth.state in the CommentStream

## How do I test this PR?
As a visitor you should not be able to see the third tab.
As an Admin you should be able to see the Configure Tab


davidgljay (approved):
Looks good!

PR #161, opened 2016-12-12
Open/Close comments stream

gabelula:
## What does this PR do?

- Adds a status (for open, closed) to the Asset as we not only need to close the status but sometimes open it again. 
- Fix the endpoint and FE to open/close the stream.

## How do I test this PR?

- open the stream
- reload the page
- check that is still open
- close the stream
- reload the page
- check that is still close


wyattjoh:
I don't see why this change is necessary, what additional status's do we need to handle? This seems more fragile to me versus the `closedAt` timestamp.

client/coral-framework/actions/config.js
```diff
@@ -31,3 +23,31 @@ export const updateConfiguration = newConfig => (dispatch, getState) => {
...
+const openStream = () => ({type: actions.OPEN_COMMENTS});
+const closeStream = () => ({type: actions.CLOSE_COMMENTS});
+
+export const updateOpenStatus = status => dispatch => {
+  if (status === 'open') {
+    dispatch(openStream());
+    dispatch(updateOpenStream({closedAt: null}));
+  } else {
+    dispatch(closeStream());
+    dispatch(updateOpenStream({closedAt: new Date().getTime(), closedMessage: ''}));
```
okbel: Why are we sending an empty closed message? Shouldn't we use the one that's in there when we close the stream?

okbel (changes requested):

client/coral-framework/actions/config.js
davidgljay: @okbel I think a null closed message means that the stream is open, while a date indicates the time at which it will close (`Date().getTime()` means close it now.)

routes/api/asset/index.js
```diff
@@ -90,10 +90,19 @@ router.put('/:asset_id/settings', (req, res, next) => {
...
+  const {
+    closedAt = null,
+    closedMessage = null
+  } = req.body;
   Asset
-    .update({id: req.params.asset_id}, {status: req.query.status})
-    .then(() => res.status(204).end())
+    .update(id, {
+      closedAt: closedAt,
+      closedMessage: closedMessage
```
davidgljay: So closed messages are set by asset and then default to the global message if none is set?

davidgljay (commented):
Looks good to me, but I know @wyattjoh had a few changes to request.

client/coral-framework/actions/config.js
gabelula: Not really. We need to change the FE to add the possiblity of the moderators to add a custom Closed message per Comment stream. 

A stream is open when the ClosedAt is null or a future date.

routes/api/asset/index.js
gabelula: Yes. That is how it should work. I will add the default of global message.

client/coral-framework/actions/config.js
okbel: @impronunciable Dan was working on that Feature. Let's see what he thinks @gabelula @davidgljay

impronunciable: We need to have at least a default closed message, an empty string is not an option

gabelula: The stream is open if closedAt is null or "in the future". The closedMessage should be the ones from settings OR the one that we send it. 

We should have a way to set the closedMessage in the FE for a specific asset.

wyattjoh (approved):

davidgljay (approved):
Checks out for me, changes look good.

okbel (approved):

PR #163, opened 2016-12-12
My comments

riley:
## What does this PR do?

this adds the My Comment list on the user bio tab as a plugin. This is a first pass, and it's going to raise some product questions.
- should we mark banned comments as such?
- should we even show banned comments
- do we want to show metadata like how many likes a comment got?
- etc.

Things to check out:
- I updated the test suite so that the components could be mounted in a browser environment with jsdom
- Comment and CommentHistory
- I made an `asset` actions file since it made sense with what I was doing

## How do I test this PR?

- `npm install`
- `npm run test`
- log in and click the Settings tab to view your comments.


routes/api/assets/index.js
```diff
@@ -40,6 +40,20 @@ router.get('/', (req, res, next) => {
...
+  Asset.findMultipleById(assetIds)
+    .then(assets => {
+      res.json(assets);
+    })
+    .catch(error => {
+      error.status = 500;
+      next(error);
+    });
+});
+
```
wyattjoh: I thought it we had discussed that the hydration on the user comments endpoint would be better?

routes/api/comments/index.js
```diff
@@ -121,6 +121,18 @@ router.post('/', wordlist.filter('body'), (req, res, next) => {
   });
 });
 
+router.get('/user/:user_id', (req, res, next) => {
```
wyattjoh: This route conflicts with existing routes:

```
/api/v1/comments/:comment_id
/api/v1/comments/user/:user_id
```

It should instead be:

```
/api/v1/comments?user_id=...
```

As you are just listing comments.

wyattjoh (changes requested):

client/coral-framework/actions/assets.js
```diff
@@ -0,0 +1,19 @@
+import coralApi from '../helpers/response';
+import {addItem} from './items';
+
+export const FETCH_MULTIPLE_ASSETS = 'FETCH_MULTIPLE_ASSETS';
+export const RECEIVE_MULTIPLE_ASSETS = 'RECEIVE_MULTIPLE_ASSETS';
+export const FAILURE_MULTIPLE_ASSSETS = 'FAILURE_MULTIPLE_ASSSETS';
+
+export const fetchMulitpleAssets = ids => {
+  return dispatch => {
+    dispatch({type: FETCH_MULTIPLE_ASSETS});
```
okbel: We are using `_REQUEST` `_SUCCESS` `_FAILURE` as postfix

Can we change those?

okbel (changes requested):

routes/api/comments/index.js
wyattjoh: Note that this will also resolve the assets being on the comments for you.

wyattjoh: As well as hydrating other data, see https://github.com/coralproject/talk/blob/7ef75fa92c60cf45c6868fbb9a81d77b7f87e506/routes/api/comments/index.js#L12-L63

client/coral-framework/reducers/user.js
```diff
@@ -30,6 +33,10 @@ export default function user (state = initialState, action) {
   case actions.SAVE_BIO_SUCCESS:
     return state
       .set('settings', action.settings);
+  case actions.RECEIVE_COMMENTS_BY_USER:
+    return state.set('myComments', fromJS(action.comments));
+  case assetActions.RECEIVE_MULTIPLE_ASSETS:
+    return state.set('myAssets', fromJS(action.assets));
```
okbel, line 39: `return state.set('myAssets', action.assets);` 
 `action.comments` instead of using fromJS is not working?

routes/api/assets/index.js
riley: of course, I forgot. I'll do that in this PR

okbel (approved):

wyattjoh (approved):

PR #165, opened 2016-12-13
Set char count as admin

davidgljay:
## What does this PR do?

Allows an admin to set the character count. 
Pivotal story: https://www.pivotaltracker.com/n/projects/1863625

## How do I test this PR?

1. Visit `http://localhost:3000/admin/configure`
2. You should be able to check and uncheck the character count.
3. Enter a letter instead of a number in the character count field. You should see an error and be unable to save.
4. Enter an appropriate number and save.
5. `charCount` and `charCountEnable` should be sent to the server as an integer and boolean respectively.

### DIFF
--- a/client/coral-admin/src/translations.json
+++ b/client/coral-admin/src/translations.json
@@ -57,7 +57,7 @@
       "community": "Community",
       "closed-comments-desc": "Write a message for closed threads",
       "closed-comments-label": "Write a message...",
-      "comment-count-header": "Limit Content Length",
+      "comment-count-header": "Limit Comment Length",
       "comment-count-text-pre":  "Comments will be limited to ",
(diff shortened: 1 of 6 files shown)


client/coral-admin/src/translations.json
```diff
@@ -96,6 +101,7 @@
     },
     "configure": {
       "enable-pre-moderation": "Habilitar pre-moderación",
+      "enable-pre-moderation-text": "tracundeme!",
```
okbel: english =>"Moderators must approve any comment before it is published"
spanish => "Los moderadores deben aprobar cualquier comentario antes de que sea publicado."

davidgljay: Thanks!

gabelula (approved):

PR #167, opened 2016-12-13
View char count

davidgljay:
## What does this PR do?

Lets a user see the character count while they are typing and prevents them from posting if they are over the limit. Pivotal story: https://www.pivotaltracker.com/story/show/134712161

## How do I test this PR?

1. Set a character count in the admin panel. `http://localhost:3000/admin/configure`
2. Write a comment on a stream. You should see an indicator displaying the number of characters you have remaining above the "post" button.
3. Exceed the character count. The indicator should turn red and you should be unable to post.


kgardnr (approved):

riley:
emojis count as two characters 🤔

davidgljay:
Checked w/ the design team, they say that doublecounting emojis is fine for now.

riley:
excellent. lgtm 💯

PR #171, opened 2016-12-14
Global comment stream close setting

impronunciable:
It adds a global setting for autoclosing comments. It can be configured in the admin.

Also when going to the comment stream as an admin you can see when the stream will close.


wyattjoh:
Do we have a way to turn on/off auto closing?

client/coral-admin/src/translations.json
```diff
@@ -117,6 +121,11 @@
       "community": "Comunidad",
       "closed-comments-desc": "Escribe un mensaje para cuando los comentarios se encuentran cerrados",
       "closed-comments-label": "Escribe un mensaje...",
+      "never": "Nunca",
```
gabelula, line 124: Is the string "never" missing from the english strings?

gabelula (changes requested):

client/coral-admin/src/translations.json
impronunciable: we are not using it

impronunciable:
@wyattjoh you can leave it blank

impronunciable:
@gabelula removed that translation and merged with master

gabelula (approved):

PR #173, opened 2016-12-15
allow non-admin folks to list their comments

riley:
## What does this PR do?

Kim got excited (thanks for the vote of confidence) and merged #163 while I was fixing this. In that PR, you could view your comments, but only if you were an admin (derp). This PR allows non-admin users to view their _own_ comments, but an admin can still view any user's comments.

## How do I test this PR?

- create a non-admin user
- leave a couple of comments
- click Settings and view a list of your comments


routes/api/comments/index.js
```diff
@@ -18,6 +18,18 @@ router.get('/', authorization.needed('admin'), (req, res, next) => {
     user_id = null
   } = req.query;
 
+  // everything on this route requires admin privileges besides listing comments for owner of said comments
```
gabelula, line 21: Can we add a tests for this?  Right now we have one that returns all the comments when the admin user is injected.

gabelula (approved):

PR #176, opened 2016-12-15
add tests for comment stream for non-admins

riley:
## What does this PR do?

Exactly what it says on the tin.

## How do I test this PR?

- `npm test`

### DIFF
--- a/tests/routes/api/comments/index.js
+++ b/tests/routes/api/comments/index.js
@@ -92,7 +92,7 @@ describe('/api/v1/comments', () => {
         .then(res => {
           expect(res).to.have.status(200);
           expect(res.body.comments).to.have.length(2);
-          expect(res.body.comments[0]).to.have.property('author_id', '456');
+          expect(res.body.comments[1]).to.have.property('author_id', '456');
         });
     });
 
--- a/tests/routes/api/comments/index.js
+++ b/tests/routes/api/comments/index.js
@@ -92,7 +92,7 @@ describe('/api/v1/comments', () => {
         .then(res => {
           expect(res).to.have.status(200);
           expect(res.body.comments).to.have.length(2);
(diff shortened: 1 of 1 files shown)


tests/routes/api/comments/index.js
```diff
@@ -85,6 +85,29 @@ describe('/api/v1/comments', () => {
       ]);
     });
 
+    it('should return only the owner’s comments if the user is not an admin', () => {
+      return chai.request(app)
+        .get('/api/v1/comments?user_id=456')
+        .set(passport.inject({id: '456', roles: []}))
```
gabelula, line 91: You don't have to add the roles field if it is empty.

tests/routes/api/comments/index.js
```diff
@@ -85,6 +85,29 @@ describe('/api/v1/comments', () => {
...
+        .then(res => {
+          expect(res).to.have.status(200);
+          expect(res.body.comments).to.have.length(2);
+          expect(res.body.comments[0]).to.have.property('id', comments[1].id);
```
gabelula: Shouldn't be better to just check the user_id property and see that is the user 456 ?

gabelula (approved):
It looks good.

tests/routes/api/comments/index.js
```diff
@@ -92,7 +92,7 @@ describe('/api/v1/comments', () => {
         .then(res => {
           expect(res).to.have.status(200);
           expect(res.body.comments).to.have.length(2);
-          expect(res.body.comments[0]).to.have.property('id', comments[1].id);
+          expect(res.body.comments[0]).to.have.property('author_id', '456');
```
gabelula: Can we add 

expect(res.body.comments[1]).to.have.property('author_id', '456');

?

riley: yeah, that makes more sense

PR #179, opened 2016-12-16
my comments design tweaks

riley:
## What does this PR do?

more closely match the designs of the My Comments feature

## How do I test this PR?

- click the settings tab to view your comments
- removed extraneous All Comments h2
- comments should be sorted chronologically
- comment count should be accurate instead of hardcoded
- add border between comments


kgardnr:
:100:

kgardnr (approved):

PR #180, opened 2016-12-16
Suspect wordlist

riley:
## What does this PR do?

This PR adds the ability to manage a "suspect words" list. Words that are not outright banned, but will be highlighted for the mods in the queue. Words can be managed in a tag manager style, and I've updated the Banned Words component as well. I added a bunch of words, so if @okbel or @gabelula could take a look at the coral-admin translations, I would appreciate it.

## How do I test this PR?

- `npm install`
- delete the settings in your db and restart the app `db.settings.remove({})` since the wordlist has changed its shape.
- go to the Configure Page > Banned & Suspect Words
- type some naughty words and hit Enter or Tab to add them
- make a list of `naughty, words, separated, by, commas` and paste it in. They should be automagically parsed.
- click Save
- go back to the Moderation Queue and view comments with those words. They should be highlighted in yellow.


client/coral-admin/src/actions/settings.js
```diff
@@ -21,10 +23,16 @@ export const fetchSettings = () => dispatch => {
     });
 };
 
+// for updating top-level settings
 export const updateSettings = settings => {
   return {type: SETTINGS_UPDATED, settings};
 };
 
+// this is a nested property, so it needs a special action.
+export const updateWordlist = (listName, list) => {
```
davidgljay, line 32: Should we make this more generic, something like `concatSettingsArray`? Also, why are banned and suspect wordlists in a nested object rather than being in the root of settings? (Since nested objects generally cause headaches.)

client/coral-admin/src/reducers/settings.js
```diff
@@ -28,6 +34,15 @@ const updateSettings = (state, action) => {
   return s.set('settings', settings);
 };
 
+const updateWordlist = (state, action) => {
+  const wordlist = state
+    .get('settings')
+    .get('wordlist')
+    .merge({[action.listName]: action.list});
+  const settings = state.get('settings').merge({wordlist});
+  return state.set('settings', settings);
```
davidgljay: For future reference you can use `setIn`, to avoid some of this nested multiline hairpulling.

davidgljay (commented):
One structural question, but no change requests.

models/setting.js
```diff
@@ -3,6 +3,13 @@ const Schema = mongoose.Schema;
 const _ = require('lodash');
 const cache = require('../services/cache');
 
+const WordlistSchema = new Schema({
```
davidgljay, line 6: Same comment as earlier, but why are these stored in a nested object?

client/coral-admin/src/reducers/settings.js
```diff
@@ -28,6 +34,15 @@ const updateSettings = (state, action) => {
   return s.set('settings', settings);
 };
 
+const updateWordlist = (state, action) => {
+  const wordlist = state
+    .get('settings')
+    .get('wordlist')
+    .merge({[action.listName]: action.list});
```
okbel: You can use setIn method here. This would be

```js
.setIn(['action', 'listname'], action.list)
```

okbel (changes requested):

client/coral-admin/src/actions/comments.js
```diff
@@ -1,10 +1,11 @@
 import coralApi from '../../../coral-framework/helpers/response';
-import * as commentActions from '../constants/comments';
+import * as commentTypes from '../constants/comments';
+import * as actionTypes from '../constants/actions';
 
 // Get comments to fill each of the three lists on the mod queue
 export const fetchModerationQueueComments = () => {
   return dispatch => {
-    dispatch({type: commentActions.COMMENTS_MODERATION_QUEUE_FETCH});
+    dispatch({type: commentTypes.COMMENTS_MODERATION_QUEUE_FETCH});
```
okbel: Add the `_REQUEST` post fix when you start a request

okbel (changes requested):

client/coral-admin/src/actions/settings.js
jde: Good question. @wyattjoh, what is the benefit of having the nested word lists that require custom code?

wyattjoh: The benefit here is grouping of concerns. This is the parallel of normalizing the database if we were using a SQL style database. By grouping like data together, we reduce duplication of namespaces and I think, aids in focusing the data models to their use cases and simplifying the API.

Imagine if we had a endpoint specifically for updating only the wordlist data, we could simple expose an `PUT /api/settings/wordlist` endpoint rather than having to rely on a `PATCH /api/settings`.

I understand that some additional code may be required on the front end in order to reduce the data to a form that is compatible with the front end data stores, but I don't think that it is unwarranted considering the vast implications that it leads to when we consider other nested objects, like `metadata`.

If we later decided to change our datastore to something like postgres, then would it not make sense to contain wordlist data in a table called `settings_wordlist`? Then the outgoing data would have to take the form of an embedded document to ensure Retrieve/Update consistency.

To me, it is essential to establish name-spacing with subdocuments.

RE: @jde, @davidgljay, @riley

riley: This makes sense, esp with the `metadata` stuff in the future. I'd just say let's keep it somewhat sane and not have stuff like:

```
settings = {
  top_level_prop: {
    sub_prop: [{ // array item that we have to update by index
      please_dont_do_this: String,
      for_the_love_of_all_that_is_holy: Number
   }]
  }
}
```

jde: Good thinking @wyattjoh. List here is basically a data type. I love the ability to create granular api controls around adding/removing/counting etc... 'words'. 

Also +1 to snake_case_novella_variable_names

okbel (approved):

PR #182, opened 2016-12-16
Admin asset list

davidgljay:
## What does this PR do?

Creates a list where admins can review assets with comment streams.

## How do I test this PR?

1. Create a few streams by visiting `http://localhost:300/assets/some-random-string`
2. Visit the admin panel.
3. Click "Streams" tab.
4. You should see a list of streams.
5. Sort, filter and search should work as expected.
6. Pagination should work as expected if more than 25 streams are present.

## Note
There is an unrelated bug causing publication dates to be incorrect.


riley:
If I close an article in the toggle thing, it doesn't flip over to show "closed" until I reload the page.

riley:
if I click Sort By any of the filters, the list disappears.

riley:
typing in the search box makes the articles disappear also

riley:
clicking Next on the pagination thing at the bottom takes me to the next page (I think?) and I can't get back to see the articles. I'd probably just disable Next if there is no next page

davidgljay:
Bug fixed and `Next` and `Prev` buttons removed.

client/coral-admin/src/containers/Streams/Streams.js
```diff
@@ -0,0 +1,182 @@
...
+      prevState.timer = setTimeout(() => {
+        fetchAssets(0, limit, this.state.search, this.state.sort, this.state.filter);
+      }, 350);
+      return prevState;
+    });
+  }
+
+  renderDate = (date) => {
+    const d = new Date(date);
+    return `${d.getMonth() + 1}/${d.getDate()}/${d.getFullYear()}`;
```
riley, line 54: is there any way to detect a non-english language and display the dates in non-american style?

client/coral-admin/src/containers/Streams/Streams.js
```diff
@@ -0,0 +1,182 @@
...
+  onSettingChange = (setting) => (e) => {
+    let options = this.state;
+    this.setState({[setting]: e.target.value});
+    options[setting] = e.target.value;
+    this.props.fetchAssets(0, limit, options.search, options.sort, options.filter);
+  }
+
+  onSearchChange = (e) => {
+    this.setState({search: e.target.value});
+    this.setState((prevState) => {
```
riley, line 41: is there anyway to combine these two `this.setState` statements?

client/coral-admin/src/containers/Streams/Streams.js
```diff
@@ -0,0 +1,182 @@
...
+    this.setState({page});
+    const {search, sort, filter} = this.state;
+    this.props.fetchAssets((page - 1) * limit, limit, search, sort, filter);
+  }
+
+  render () {
+    const {search, sort, filter} = this.state;
+    const {assets} = this.props;
+
+    console.log(assets);
```
riley: remove this log

riley (changes requested):
looking great, everything works. just a couple of optimizations if possible

client/coral-admin/src/containers/Streams/Streams.js
gabelula: In that case date format needs to be saved as part of the Settings.

davidgljay: I thought about this, but it seemed out of scope for this story. It involves an unrelated setting and associated UI.

davidgljay:
Recommended updates implemented.

riley (approved):
awesome! 💯

PR #184, opened 2016-12-16
Add flag option

davidgljay:
## What does this PR do?

Adds an option to explain a flag (Pivotal story https://www.pivotaltracker.com/n/projects/1863625/stories/135985387)

## How do I test this PR?

1. Report a comment.
2. Click any reason for the report.
3. You should see a text field asking you to provide an optional reason for the flag.
4. Enter a reason.
5. This reason should be sent to the server and stored in the database.


models/action.js
```diff
@@ -15,6 +15,7 @@ const ActionSchema = new Schema({
   user_id: String,
   field: String, // Used when an action references a particular field of an object. (e.g. a flag on a username or bio)
   detail: String, // Describes the reason for an action (e.g. 'Username is offensive')
+  description: String, // Describes the reason for the action in more detail (e.g. A description of why the comment is offensive.)
```
wyattjoh: It would make more sense to me to use words like:

```
field -> metadata.field // Describes the field for which the action is related to.
detail -> metadata.reason // Describes the reason that the action was taken (if any)
description -> metadata.comment // Describes with more words why the action was taken in the user's own words
```

Essentially moving these pieces into a new `metadata` field.

wyattjoh (changes requested):

models/action.js
gabelula: Can we use metadata.notes or metadata.details instead of metadata.comment ? It seems more appropiate not to use the word 'comment' in this context.

davidgljay:
Agreed that comment could be confusing as it has another meaning in the
system.

models/action.js
gabelula: And not sure if 'field' should be part of the metadata field or not. It is related to the item that is being flagged.

wyattjoh: To me, `comment` is what a user would add versus what a machine would add. Action's aren't only used for flags, hence why `field` should be moved into the `metadata` field as it doesn't make sense for actions like `liking`.

wyattjoh: Additionally, `details` tells me nothing about what that field contains or who made it. `comment` is clear as it's the "comment" related to the flagging action.

davidgljay: I wouldn't agree that thatxs clear. 'comment' seems like it should reference an item of type comment stored elsewhere in the system. It would be clearer to use a different word.

wyattjoh: Then possibly a `message` field? My point is to try to distinguish this field from a system set field.

davidgljay: `message` works for me.

jde: +1 on making field optional (as many actions will be on items themselves) and putting all optional fields into the `metadata` object.

jde: Also, +1 to using `metadata.notes` instead of `metadata.comment`. Using comment here is likely to confuse devs thinking that comment may refer to a comment's id or something.

`notes` is a good generic name for fields populated by user input to embellish something else.

wyattjoh (approved):

riley:
works for me! great work.

PR #188, opened 2016-12-19
E2e tests

okbel:
# E2E tests
This PR adds coverage for the following **tests**:

* **Commenter**
	* Login
	- Post a comment 
	* Likes a comment
	* Flag a comment
	* Flag a username
	* Gets Permalink
	* Visits Permalink

- **Moderator** 
	* Login 	
	
- **Admin** 
	* Login
	- Approve Comment
	- Reject Comment	
	* Ban User
	
- **Visitor**
	* Tries to like a comment
	* Tries to flag a comment 
	- Tries to flag a username	
	* Signs up
	
_Note: The `pree2e`  script will create 3 users: a Commenter, a Moderator, and an Admin_

## Getting Started 
We are using **Nightwatch** and **Selenium**. In order to run e2e tests you will need Java.

### Selenium Server Setup

Selenium Server is a Java application which Nightwatch uses to connect to the various browsers.

You will need to have the Java Development Kit (JDK) installed. [Java SE Development Kit 8 - Downloads](http://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html)

The minimum required version is 7.You can check this by running  `java -version`

### Folder Structure

```js
e2e/
  ├── pages
  |   ├── adminPage.js
  |   └── embedStreamPage.js
  ├── reports
  ├── tests
  |   ├── Admin
  |   ├── Commenter
  |   ├── Moderator
  |   └── Visitor

```

#### Pages
Here we will have all the selectors and commands for a Page

#### Reports
The folder that Nightwatch will use after running the tests

#### Tests
Within `tests` folder we have 4 Folders and a couple of files.
`Admin`, `Commenter`, `Moderator`, `Visitor`  contains all the group tests based on the user role and their actions.

This can be really useful for running group tests.

##  Run the tests

Run e2e tests
```bash
npm run e2e
```
	

### Adding an Integration Environment

```js
{
...
  “test_settings” : {
    “default” : {
      “launch_url” : “http://localhost”,
      “globals” : {
        “myGlobalVar” : “some value”,
        “otherGlobal” : “some other value”
      }
    },
    “integration” : {
      “launch_url” : “http://staging.host”,
      “globals” : {
        “myGlobalVar” : “other value”
      }
    }
  }
}
```
Running an integration environment 

```bash
nightwatch —env integration
```

Running an integration environment with npm scripts

```bash
npm run e2e -- —env integration
```

### Chrome Options
[List of Chromium Command Line Switches «  Peter Beverloo](http://peter.sh/experiments/chromium-command-line-switches/) 

## Tags
You'll notice that each test file starts with tags. This is useful to selectively target tests to run.

_i.e nightwatch --tag login will only run login tests tagged with login_
```js
export default {
  '@tags': ['login'],
  'Test': browser => {
    [...]
  }
}
```

Source: http://nightwatchjs.org/guide#test-tags	

	
### Notes

- We are not using mocks or models for e2e tests.
- Theres no need to do this in a test db because we can run `--env integration` in our CI enviroment and this will not get pushed to production. This can be tested in isolation.
- Yarn is very good and super quick

## Bugs and Concerns
- Pre-Moderation got disabled at some point
- Flagging a username is returning a 404 (Test failed but I removed the last confirmation until we fix that)
- I post a comment with premod disabled. I toggle it back on and the comment disappears. Is that expected?
- Sometimes comment list in the admin side doesn’t show anything. Maybe some async condition taking longer
	

@coralproject/frontend


.eslintrc.json
```diff
@@ -4,6 +4,16 @@
...
+    "ecmaFeatures": {
+      "sourceType": "module"
+    },
+    "parserOptions": {
+      "sourceType": "module",
+      "ecmaFeatures": {
+        "jsx": true,
+        "experimentalObjectRestSpread": true
+      }
+    },
```
wyattjoh: This is the wrong place to have these options. The root of the project is the `node` specific `.eslintrc.json` file while the files in `/client`, `/tests`, and `/tests/client` have their specific overrides. This applies the module syntax across the entire node codebase.

package.json
```diff
@@ -11,8 +11,8 @@
     "lint-fix": "./node_modules/.bin/eslint bin/* . --fix",
     "test": "NODE_ENV=test ./node_modules/.bin/mocha --compilers js:babel-core/register tests/helpers/*.js --require ignore-styles --recursive tests",
     "test-watch": "NODE_ENV=test ./node_modules/.bin/mocha --compilers js:babel-core/register --recursive -w tests",
-    "pree2e": "NODE_ENV=test ./scripts/pree2e.sh",
-    "e2e": "NODE_ENV=test ./node_modules/.bin/nightwatch",
+    "pree2e": "scripts/pree2e.sh",
```
wyattjoh: The application (the `npm start`) must be started with the test environment otherwise the environment variables will not be set.

package.json
```diff
@@ -11,8 +11,8 @@
...
+    "e2e": "NODE_ENV=test nightwatch",
```
wyattjoh, line 15: Is there a reason that we are preferring the global versus the local installation of nightwatch?

scripts/pree2e.sh
```diff
@@ -1,7 +1,13 @@
 #!/bin/bash
 
 # install selenium
-../node_modules/selenium-standalone/bin/selenium-standalone install
+selenium-standalone install
 
-# start the app server
-npm start &
+# Creating Admin Test User
+{ echo admin@test.com; echo test; echo test; echo Admin Test User; echo admin;} | dotenv ./bin/cli-users create
```
wyattjoh, line 7: This isn't maintainable like this. We'll be adding the setup screen soon which will provide a GUI for creating the first admin user. This is too fragile.

tests/e2e/tests/01_EmbedStreamTest.js
```diff
@@ -0,0 +1,44 @@
+export default {
```
wyattjoh: We shouldn't be mixing the import styles on FE JS and BE JS. I'd prefer to use the `module.exports = ` syntax here.

wyattjoh (changes requested):

package.json
okbel: Within the script task local wins. This will look for nightwatch within the local node_modules, and after that goes to global. So there's no need to keep using `./node_modules/` prefix

scripts/pree2e.sh
okbel: Agreed. But this is the only way to set a role for now. When we have that GUI we can make the refactor.

tests/e2e/tests/01_EmbedStreamTest.js
okbel: We are using `export default` everywhere within the `client` folder

.eslintrc.json
okbel: This is legacy code. This can be deleted.

wyattjoh: This code says: "the entire project should use ecma modules", but we don't. We only use the module import syntax inside the `client` + `tests/client` directories. Therefore it should be able to get removed.

tests/e2e/tests/01_EmbedStreamTest.js
wyattjoh: And we are using `module.exports = ` everywhere in the server side code, hence it would make sense to me to stay consistent rather than mixing import syntaxes.

package.json
okbel: 👍  perfect. Updated!

wyattjoh (approved):

PR #192, opened 2016-12-20
Making font sizes relative

davidgljay:
## What does this PR do?

Makes font sizes relative. This is good practice, and will make it easier to customize font sizes in the future.

## How do I test this PR?

1. Font sizes should appear normal.


riley (approved):
lgtm 👍

PR #195, opened 2016-12-21
Added `NPM_CONFIG_PRODUCTION: false` for heroku deploys

wyattjoh:
## What does this PR do?

Disabled production only dep installs for Heroku.


jde:
Closing without merging as per @wyattjoh.

PR #197, opened 2016-12-21
Adds a very simple INSTALL file for developers.

gabelula:
## What does this PR do?

Adds INSTALL.md file.

## How do I test this PR?

Read the file!


INSTALL.md
```diff
@@ -0,0 +1,85 @@
...
+	export TALK_MONGO_URL=mongodb://localhost/talk
+	export TALK_REDIS_URL=redis://redis/talk
+	export TALK_SESSION_SECRET=somethingsecret
+	export TALK_FACEBOOK_APP_ID=fbidapp_youneeditforfbconnect
+	export TALK_FACEBOOK_APP_SECRET=fbappsecret_youneeditforfbconnect
+	export TALK_ROOT_URL=http://localhost:5000
+	export TALK_SMTP_USERNAME=coralproject
+	export TALK_SMTP_PASSWORD=smtppassword
+	export TALK_SMTP_HOST=smtp.example.net
+	export TALK_SMTP_PROVIDER=Example
```
wyattjoh: We should just link to the README.md instead of providing this block.

INSTALL.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Local development
+
+`
+npm start
+`
+
+It will run in http://localhost:5000
```
wyattjoh: Should reference the port from the configuration instead:

```
npm start
```

Will start the application at `http://localhost:$PORT` where `$PORT` is defined in the configuration section.

INSTALL.md
```diff
@@ -0,0 +1,85 @@
...
+
+`
+npm test
+`
+
+`
+npm run e2e
+`
+
+## Recommendations for Deployment
```
wyattjoh: Shouldn't point out deployment solutions in the INSTALL.md doc.

INSTALL.md
```diff
@@ -0,0 +1,85 @@
...
+
+`
+git clone https://github.com/coralproject/talk
+`
+
+### Installation
+
+* cd talk
+* npm install
+* npm run build
```
wyattjoh: The build step is already added to the `postinstall` hook, shouldn't re-run it. Also, enclose in a triple backtick instead of a list.

INSTALL.md
```diff
@@ -0,0 +1,85 @@
+# BUILDING FROM SOURCE
+
+By contributing to this project you agree to the [Code of Conduct](https://coralproject.net/code-of-conduct.html).
+
+## Requirements
+
+* [Node 7.2.1](https://nodejs.org/es/download/package-manager)
```
wyattjoh: Node version is specified in the `package.json` already, no need to duplicate it here.

INSTALL.md
gabelula: Not sure. I think it is part of the setup for local development. Maybe we can remove it from the README and only have it where we have install and configuration options?

INSTALL.md
gabelula: Yes. This could be a totally separate file on deployment recommendations and howtos.

INSTALL.md
wyattjoh: The config variables here do not describe what the application needs. I think that they should be described and leave it up to the implementor to actually write the config themselves. This simply confuses the developers. All this data should live exclusively in the README, as it's the "entrypoint" to the project.

Every local install or production deploy still have the same config, and it's going to be different in each use case.

gabelula: Yes. That is a good point. I can add something about what each env variable means. 

About this variables in the README I'm not sure how much they add as there is no other instruction on how to launch the project there.

wyattjoh: This is what I imagine going into this document at this point.

> Please see README.md for a description of configuration variables required. To run the application, specify these config variables in the environment.

That should be clear to anyone familiar with what an environment variable is.

jde (approved):

PR #207, opened 2017-01-03
Fix test on Suspected words

gabelula:
## What does this PR do?

When adding a flag action the metadata field was missing. This PR adds this field metadata (that contains field and details) and fix the test.

## How do I test this PR?

- npm test


wyattjoh (approved):
Nice catch!

impronunciable (approved):

PR #211, opened 2017-01-04
Refactor, change test to include CSRF and add a new test

gabelula:
## What does this PR do?

- refactor a few lines based on Dan's review on a previous PR
- change test on client item Actions to include CSRF (inititate state with it in the test).
- add test on creating comment to check that it fails if CSRF not present

## How do I test this PR?

- npm test


wyattjoh:
I think that with https://github.com/coralproject/talk/pull/215 getting merged we should close this one.

PR #212, opened 2017-01-04
Moves the packages to dependencies.

gabelula:
## What does this PR do?

Moves some devDependencies to dependencies in package.json


riley (approved):
lgtm

PR #214, opened 2017-01-05
Bad usernames

riley:
## What does this PR do?

Validates a username at creation time for profanity and acceptable characters.

There are 2 types of errors:
1. profanity
2. no special characters
3. length can be anything greater than 0

All errors are handled on the backend, and only item 2 above is checked on the client.

Also, I added a package-level Errors file, so please create your custom errors there and import them.

## How do I test this PR?

- attempt to register a new user.
- register a username with a curse word in `models/naughty-words.json` (english only so far)
- `shit` should fail, as well as `s_h_i_t`
- `fluffy_bunny_!@#%$%^%$` should fail.
- special character errors should show up just below the display name input field, but other errors from the server will show up in the red box above.


models/naughty-words.json
```diff
@@ -0,0 +1,444 @@
+[
```
wyattjoh: We shouldn't include this wordlist in the source control. It's one thing to provide a facility for users to specify words that they don't want used in their community and another to broadly scope words that are completely valid in circumstances.

I had reviewed with the team an idea like this before when my first pass against the wordlist started, and I don't think that including this is a good idea.

models/user.js
```diff
@@ -296,56 +297,104 @@ UserService.createLocalUsers = (users) => {
...
+    return Promise.reject(ErrMissingDisplay);
+  }
+
+  if (!onlyLettersNumbersUnderscore.test(displayName)) {
+    const ErrSpecialChars = new Error('NO_SPECIAL_CHARACTERS');
+    ErrSpecialChars.status = 400;
+    return Promise.reject(ErrSpecialChars);
+  }
+
+  const hasBadWords = naughtyWords.some(badWord => {
```
wyattjoh: This should use the wordlist service instead of using it here.

models/user.js
```diff
@@ -296,56 +297,104 @@ UserService.createLocalUsers = (users) => {
...
+
+    // test the word AFTER the interspersed _ characters are removed.
+    return new RegExp(badWord, 'ig').test(displayName.replace(/_/g, ''));
+  });
+
+  if (hasBadWords) {
+    const ErrProfanity = new Error('PROFANITY_ERROR');
```
wyattjoh: This error is already defined in the wordlist service.

models/user.js
```diff
@@ -296,56 +297,104 @@ UserService.createLocalUsers = (users) => {
...
  * @param  {String}   email       email of the new user
  * @param  {String}   password    plaintext password of the new user
  * @param  {String}   displayName name of the display user
  * @param  {Function} done        callback
  */
 UserService.createLocalUser = (email, password, displayName) => {
+
   if (!email) {
-    return Promise.reject('email is required');
+    const ErrMissingEmail = new Error('EMAIL_REQUIRED');
```
wyattjoh: The `message` component of an error is designed to be a human readable name, not a machined "code"

yarn.lock
```diff
@@ -0,0 +1,7333 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
wyattjoh: `yarn.lock` should not be committed to source control until issues have been resolved with the yarn tool as it blocks production deploys.

wyattjoh (changes requested):

models/user.js
riley: so before, we were just throwing away the `message` part of the error. I updated the response helper on the frontend to pass the message through. It's a string constant like that so that I can look up the code and translate it into Spanish or whatever.

yarn.lock
riley: dammit, that was an accident. Can we just add it to the `.gitignore` file or will that screw up the deploy process?

models/user.js
wyattjoh: Is it possible to instead do something like this @riley?

```js
const ErrMissingEmail = new Error('Email required');
ErrMissingEmail.translation_key = 'EMAIL_REQUIRED';
ErrMissingEmail.code = 400;
```

That way we know that the `translation_key` is the translation key, versus just a machine message.

yarn.lock
wyattjoh: If you remove the `yarn.lock` file and add it to the project's `.gitignore` we can use that going forward until we're ready to migrate to yarn.

TLDR: Yes, do it, it won't mess up deploy.

client/coral-framework/helpers/response.js
```diff
@@ -37,7 +37,14 @@ const handleResp = res => {
   if (res.status === 401) {
     throw new Error('Not Authorized to make this request');
   } else if (res.status > 399) {
-    throw new Error('Error! Status ', res.status);
+    return res.json().then(err => {
+      console.log(err);
```
wyattjoh: Can we pull this console.log out?

routes/api/account/index.js
```diff
@@ -139,3 +139,4 @@ router.put('/settings', authorization.needed(), (req, res, next) => {
 });
 
 module.exports = router;
+module.exports.ErrPasswordTooShort = ErrPasswordTooShort;
```
wyattjoh: Can we just use the shared errors file now?

wyattjoh (changes requested):

models/user.js
```diff
@@ -310,56 +312,84 @@ UserService.createLocalUsers = (users) => {
...
+
   if (!email) {
-    return Promise.reject('email is required');
+
+    return Promise.reject(errors.ErrMissingEmail);
   }
 
-  email = email.toLowerCase();
+  email = email.toLowerCase().trim();
+  displayName = displayName.toLowerCase().trim();
```
wyattjoh, line 349: I think that these transformations should occur on the handler of the endpoints and not as a part of the service..

services/wordlist.js
```diff
@@ -166,6 +167,28 @@ class Wordlist {
...
+   * check potential username for banned words, special characters
+   */
+  static displayNameCheck(displayName) {
+    const wl = new Wordlist();
+    return wl.load()
+      .then(() => {
+        displayName = displayName.replace(/_/g, '');
+
+        // test each word, and fail if we find a match
+        const hasBadWords = wl.lists.banned.some(word => {
```
wyattjoh: Is it possible to leverage the existing `.match` function at all?

models/user.js
riley: I always feel like it's messy to implement business logic in route handlers, especially if I want to re-use the `createLocalUser` somewhere else and that logic only belongs to `createLocalUser`. I'd say a function should validate its input.

services/wordlist.js
riley: let me look into that

wyattjoh (approved):

PR #215, opened 2017-01-06
Email Confirmation and Fixes

wyattjoh:
## What does this PR do?

As always, these PR's start small and grow.. big.

This covers a few outstanding issues, simplifies lots of flows.

1. This mainly adds support for email verification requirements to the backend.
    - Sends emails with token when enabled
    - Tests to verify functionality
2. Changes the way that CSRF tokens are issued, consumed, and managed to simplify the flow.
    - Removed CSRF tokens from the body, moved to header
    - Disabled CSRF protection on unit tests (still enabled on E2E tests)
3. Session updates
    - Fixes issue with logout destroying entire session instead of only the user passport instance 
4. Endpoint conformity
    - Endpoints were adjusted to conform to outlined contributing guidelines for PUT endpoints, namely, only returning a 204
    - Endpoints that conflicted with `/user/:user_id` wildcard routes were moved to a new `/account` endpoint.

## How do I test this PR?

1. Build the client code via `npm run build`
2. Run unit tests via `npm test`
3. Run the server via `npm start` and perform a visual walkthrough of features to ensure compatibility.


client/coral-framework/helpers/response.js
```diff
@@ -2,19 +2,28 @@ export const base = '/api/v1';
...
   };
+
   const options = Object.assign({}, defaultOptions, inputOptions);
 
-  // Add CSRF field to each POST.
-  if (options.method.toLowerCase() === 'post' && options._csrf) {
-    options.body._csrf = options._csrf;
+  if (options._csrf) {
+    switch (options.method.toLowerCase()) {
+    case 'post':
```
gabelula, line 21: It is not adding it in the POST Case.

docs/swagger.yaml
```diff
@@ -583,7 +583,7 @@ paths:
           description: The user that has been created.
           schema:
             $ref: '#/definitions/User'
-  /users/update-password:
+  /account/password/reset:
```
gabelula, line 586: I really like this change on API.

client/coral-framework/helpers/response.js
wyattjoh: In JS, the case's are "fall through" which means any case not terminated with a `break` will follow down until it finds one, so this catches all 'post', 'put', and 'delete' cases and adds the headers.

gabelula (changes requested):
Can we add a test to check that it gets Forbidden if CSRF is not being send?

wyattjoh:
@gabelula: Testing if the CSRF token is being sent or not really lies to the responsibility of the client code, and represents a full E2E test case. The unit tests in my opinion is not a good place to evaluate this particular situation. We adding a token in in unit test mode with a value of `UNIT_TESTS ` here:

https://github.com/coralproject/talk/pull/215/files#diff-0364f57fbff2fabbe941ed20c328ef1aR86

So I'd rather see us add a test to the client code ensuring that we test the `coralApi` method indeed sends this token in the event it is rendered on a page where the meta tag is present containing the CSRF token, which is the preferable way of handling CSRF tokens for AJAX/RESTful requests.

gabelula (approved):

Issue #148, opened 2016-12-09
With Firefox, cannot click on comment box after login modal has been closed


jde:
### Expected behavior

Clicking in the comment box anytime should give me a cursor and let me type!

### Actual behavior

In firefox, I cannot bring focus to the comment dialog after the login box has opened and closed.

### Steps to reproduce behavior

Open a stream with firefox
click login
do the login thing
after the modal closes click on the comment box
note that it doesn't give you a cursor or let you type

jde:
Firefox shows an error in the element tree that is not reported in the Console:

![firefox-textarea-error](https://cloud.githubusercontent.com/assets/259957/22033605/30bd311e-dcb7-11e6-8e5f-0f57d2d39f93.png)

PR #225, opened 2017-01-11
Fix firefox modal bug (Issue 148)

davidgljay:
## What does this PR do?

Fixes a bug where comments cannot be entered on Firefox after logging in. (https://github.com/coralproject/talk/issues/148)

## How do I test this PR?

1) Log in using firefox.
2) Click the comment box. You should be able to enter text.


client/coral-sign-in/containers/SignInContainer.js
```diff
@@ -119,8 +119,7 @@ class SignInContainer extends Component {
     }
   }
 
-  handleSignIn(e) {
-    e.preventDefault();
```
okbel, line 123: Careful removing an `e.preventDefault()` because the event bubles up and it might refreshes the page or scrolls to top in Chrome or Safari

davidgljay: It looks like e.preventDefault() messes with `onmousedown` in firefox in a way that's causing this problem. It's a good practice, but if it's breaking firefox then we should be without it.

riley:
When I click the signin button, the comments disappear in FF.

davidgljay:
@riley Looks like this is due to an unrelated styling issue in FF. I've added a commit to this PR to address that issue as well.

jde:
I'm still seeing the bug in this branch using Firefox 50.1.0. 

Here's the behavior:

cant-click-in-comment-area

client/coral-sign-in/containers/SignInContainer.js
riley: so this line needs to stay in, since I was able to get the form params into the url accidentally since there's no `action` attribute on the form. I'm going to keep plugging away on a fix for this that doesn't involve removing this line.

jde:
Still cannot access the textarea after last round of updates.

jde:
Closing due to incorrect approach. Regrouping and going to start in on a strategy involving the placement of the <dialog> element in the future.

PR #226, opened 2017-01-11
Added dev-start and --config flag

wyattjoh:
## What does this PR do?

Adds a new `dev-start` script which runs the server with `nodemon` and specifies a `.env` configuration file with the new `-c, --config` flag.

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -1,16 +1,12 @@
 #!/usr/bin/env node
 
-// Perform rewrites to the runtime environment variables based on the contents
-// of the process.env.REWRITE_ENV if it exists. This is done here as it is the
-// entrypoint for the entire application.
-require('env-rewrite').rewrite();
-
 /**
  * Module dependencies.
  */
 
 const program = require('commander');
 const pkg = require('../package.json');
+const dotenv = require('dotenv');
 
 //==============================================================================
 // Setting up the program command line arguments.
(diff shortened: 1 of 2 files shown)


bin/cli
```diff
@@ -1,23 +1,38 @@
...
-// entrypoint for the entire application.
-require('env-rewrite').rewrite();
-
 /**
  * Module dependencies.
  */
 
 const program = require('commander');
 const pkg = require('../package.json');
+const dotenv = require('dotenv');
```
davidgljay, line 9: We can have a smaller footprint if we just `"dev-start": "source .env && "nodemon --exec \"npm start\""` and skip changing ./bin/cli, unless this implementation has other benefits on your end.

davidgljay (changes requested):

jde:
While we're introducing new scripts, thoughts on renaming the jobs to make them a little more explicit? For example: `dev-start-server` and `dev-build-client`

bin/cli
wyattjoh: I think it does offer benift as it removes the syntax of the:

```
export NAME=value
```

which to me shouldn't exist in a config file. The config flag, if you noticed, specifically loads the config from a named file, this way a user could specify another file to load config from instead.

wyattjoh:
`start` is already affiliated with the "start the server", so I don't think that naming of `dev-start-server` makes sense in this use case, as we're also starting the job runner with this job.

bin/cli
davidgljay: It looks like the `dotenv` documentation considers `export` optional in the `.env` files that `dotenv` sources, if it's preferable from a devops side to use local environment variable files for development of the format:
```
NAME=value
```
rather than 
```
export NAME=value
```
then all this looks good. Approving.

(Parotting back what I'm hearing just to confirm that I understand.)

davidgljay (approved):

PR #227, opened 2017-01-11
Fix flagged comment queue bug

davidgljay:
## What does this PR do?

Fixes a bug where flagged comments do not appear in the moderation queue. (https://www.pivotaltracker.com/n/projects/1863625/stories/137404915)
This bug was caused by an inconsistency in `action_types`, they were plural (`comments`) on the frontend and singular (`comment`) in `Comment.findIdsByActionType`.

## How do I test this PR?

1. Flag a comment.
2. It should appear in the flagged comment queue.

## Note
If you approve a comment (for example, because premod is on) and then flag it it will not appear in the flagged queue. This is by design.

### DIFF
--- a/models/comment.js
+++ b/models/comment.js
@@ -201,7 +201,7 @@ CommentSchema.statics.findByActionType = (action_type) => Action
  * @return {Promise}
  */
 CommentSchema.statics.findIdsByActionType = (action_type) => Action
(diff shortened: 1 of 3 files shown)


riley:
works for me! nice catch 💯

riley (approved):

PR #228, opened 2017-01-12
moderation queue routes

riley:
## What does this PR do?

The Moderation Queue endpoint was overloaded, along with the `/routes/api/comment` endpoint necessitating backflips on the front end, creating several bugs resulting in weird behavior. This PR creates a couple new endpoints, along with modifying the `/queue/comments/pending` endpoint to only return pending comments. Now, the admin panel loads `/queue/comments/rejected` and `/queue/comments/flagged` instead of hitting the comments endpoint.

### Also
In another PR, I'm going to hide the Pending tab when
- there are 0 pending comments in the system AND
- pre-mod is disabled globally

## How do I test this PR?

- the Pending Queue should load pending comments
- when you tab between Pending, Rejected, and Flagged the server is re-queried. (wow!)
- tests should pass

### DIFF
--- a/client/coral-admin/src/actions/comments.js
+++ b/client/coral-admin/src/actions/comments.js
@@ -55,6 +55,10 @@ export const fetchFlaggedQueue = () => {
     dispatch({type: commentTypes.COMMENTS_MODERATION_QUEUE_FETCH_REQUEST});
 
     return coralApi('/queue/comments/flagged')
+      .then(comments => {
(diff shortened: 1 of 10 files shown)


wyattjoh (approved):
I'd like to see https://github.com/coralproject/talk/pull/228/files#diff-2cf0c1a21a4a3e78253dc09d1adf840aR10 eventually moved in to a service function, but for now this works perfectly =D

PR #231, opened 2017-01-12
Removed 'rejected' and 'approved' comments from flagged queue.

davidgljay:
## What does this PR do?

The flagged comments queue was showing all flagged comments, including those that had already been accepted or rejected. This PR removes those comments from the queue.

## How do I test this PR?

1. Turn Premod off.
2. Create a comment.
3. Flag it.
4. The flagged comment should appear in the admin panel under "flagged."
5. Accept or reject it.
6. It should disappear from the queue.


gabelula (approved):
This works for now but we should add an issue to remember to filter this in the backend and not in the frontend.

gabelula:
@davidgljay  Can you add an issue about the work to be done to filter in the BE instead of the FE the status of the comments?

gabelula:
Also, it seems that we can flag approved comments, in which case the flagged queue will have comments that were already approved (unless we reset the status of the comment each time somebody flag it).

@IAmSamHankins  @losowsky  @egrdina ?

losowsky:
@gabelula that's correct. 'Flagged' is only for comments that haven't yet been read and approved by a moderator. But the Flag option still stays on those comments, because people might want to Flag it as "I don't like it" - also their flagging it tells us they're not a reliable flagger, which would be useful information for Trust.

PR #233, opened 2017-01-13
don't show unpublished comments in My Comments

riley:
## What does this PR do?

Updates the `Comment.findByUserId` method to only return published comments unless you're an admin

## How do I test this PR?

- make a non-admin user
- make two comments. approve one of the comments with your other admin account
- go to the non-admin profile page. only the approved comment should show up
- rejected comments should similarly not show up.


davidgljay (approved):
Looks good! I can still see my unpublished comments if I'm an admin, but it looks like that's by design. Nice work.

PR #235, opened 2017-01-13
Updating docs for .env functionality

jde:
## What does this PR do?

Adds documentation for the new .env environment configuration option.

## How do I test this PR?

Read the readme.md. Does it give you enough to understand this option?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,8 @@ Facebook Login enabled app.
 - `TALK_SMTP_HOST` (*required for email*) - SMTP host url with format `smtp.domain.com`.
 - `TALK_SMTP_PORT` (*required for email*) - SMTP port.
 
-During development mode, you can use the command `npm run dev-start` which will load environment variables from a file named .env in the present working directory.
+Refer to the wiki page on [Configuration Loading](https://github.com/coralproject/talk/wiki/Configuration-Loading) for
+alternative methods of loading configuration during development.
 
 ### License
 
--- a/README.md
+++ b/README.md
@@ -17,8 +17,7 @@ To launch a Talk server of your own from your browser without any need to muck a
 
 ### Configuration
 
-The Talk application looks for the following configuration values either as
-environment variables or via an `.env` file:
+The Talk application looks for the following configuration values either as environment variables:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,30 +17,32 @@ To launch a Talk server of your own from your browser without any need to muck a
 
 ### Configuration
 
-The Talk application requires specific configuration options to be available
-inside the environment in order to run, those variables are listed here:
+The Talk application looks for the following configuration values either as
```
wyattjoh: This is not the case. We should not advertise in the "Configuration" section of the README that the application runs with a `.env` file because this is a development option only, and does not follow best practices for loading the configuration into the environment. We should only include these details in the developer documentation rather than the standard docs where I would assume someone would look if they are trying to deploy the application.

Additionally, the configuration option `-c` will load **any specified file** rather than specifically a `.env` file. The `.env` file only loads with the `dev-start` npm script.

wyattjoh (changes requested):

README.md
```diff
@@ -17,30 +17,32 @@ To launch a Talk server of your own from your browser without any need to muck a
...
+- `TALK_SMTP_PASSWORD` (*required for email*) - password for the SMTP provider you are using.
+- `TALK_SMTP_HOST` (*required for email*) - SMTP host url with format `smtp.domain.com`.
+- `TALK_SMTP_PORT` (*required for email*) - SMTP port.
 
-### Install from Source
+We recommend using environment variables for configuration in staging and
+production environments.
 
-If you want to run Talk in development mode from source (without docker) you can read the [INSTALL file](INSTALL.md).
+For more information about configuration via `.env`, please [visit our wiki](https://github.com/coralproject/talk/wiki/Configuration-via-.env).
```
wyattjoh: Same thing here, we shouldn't be documenting it like this.

Rather, provide:

> During development mode, you can use the command `npm run dev-start` which will load environment variables from a file named `.env` in the present working directory. Refer to the Configuration section on which config options to provide.

wyattjoh (approved):

jde:
Thanks for updating the wiki. Agree on keeping the param list in one place and linking back to it.

PR #237, opened 2017-01-13
do not render tab content unless they're active

riley:
## What does this PR do?

I'm not sure how my updates to the queue endpoints ever worked. This should fix the fixes.

## How do I test this PR?

- click around in the Moderation Queue tabs. everything should work.


davidgljay (approved):
Once I click the "Rejected" tab it no longer shows items in the other tabs, otherwise looks good.

davidgljay (changes requested):
See above, hit the wrong button.

davidgljay (approved):

PR #244, opened 2017-01-17
My Comment bug fix

riley:
## What does this PR do?

if the existing asset was not in state, rendering the My Comments stream would fail.

## How do I test this PR?

- create a comment on an asset `localhost:3000/interesting/article`
- go to another asset e.g. `localhost:3000/some/random/asset`
- click Settings.
- you should be able to see your comments


kgardnr (approved):
LGTM 👍

PR #246, opened 2017-01-17
fix my comments

riley:
## What does this PR do?

does a null check when My Comments are loaded. This will probably work on your machine, but create a different bug on staging. My plan is to merge this and then test further on staging.

## How do I test this PR?

- load comment stream
- view your comments on the Settings tab
- the number of comments on the first tab should not change
- clicking back to the stream should not make the existing comments disappear.


jde:
Works locally for me. Merging to test on staging to resolve high visibility issues. Will open another ticket to identify the differences between environments.

jde (approved):

PR #249, opened 2017-01-18
E2E environment fix

davidgljay:
## What does this PR do?

Allows `npm run e2e` to execute without any special environment variables.

## How do I test this PR?

1. Make sure a dev server is NOT already running locally
2. `npm run e2e`
3. Test should run
4. Test should pass
5. `ps`
6. There should not be a server running in the background.

# Note

There is some ugliness here in the poste2e script. I need to start a server and detach it from the terminal in order to have something to run tests against, and there's some ugliness at the end to find and kill that script. Thoughts on better ways to handle this are welcome.


davidgljay:
Updated to graphQL architecture, all e2e tests are passing locally. Ready for review.

wyattjoh (approved):

davidgljay:
Fixed a few more e2e errors, ready for merge.

PR #251, opened 2017-01-24
Graphql for the comment feed

jde:
## What does this PR do?

Integrates GraphQL into the codebase and uses it for the comment stream.

## How do I test this PR?

This PR should be transparent to the end user. All features of past prs should be present in this one.


jde:
Edge cases and bugs to address before product/QA.

- [x]  Settings tab still missing id (@okbel)
- [x]  Closed streams stay opened, even when closed in the admin panel (@okbel)
- [x]  Page doesn't refresh to show newly posted comment (@riley)
- [x]  Permalinks create a new asset instead of jumping down the page (@riley)
- [x]  Replies to replies should actually be replies to the top level comment (as per product) (@riley)
- [x]  Usernames (with mouseover bios) are missing from comments (@davidgjay)
- [x]  'ago' times are not updating (@davidgjay)
- [ ]  Add sortOrder=[ascending/descending] to comment/child queries (@wyatt)

app.js
```diff
@@ -49,7 +51,7 @@ const session_opts = {
   name: 'talk.sid',
   cookie: {
     secure: false,
-    maxAge: 18000000, // 30 minutes for expiry.
+    maxAge: 36000000, // 1 hour for expiry.
```
riley, line 54: this number is not 1 hour, it's 10 hours. Is that what we want?

riley (approved):

app.js
jde: out of scope for this pr

PR #257, opened 2017-01-25
Cleanup and more GraphQL features 

okbel:
## What does this PR do?
- Adds Fragments Usage
- My Comment History functionality with GraphQL
- Items deprecated (reducers, actions, total cleanup)
- Asset Settings clean up and updated
- Open Stream Status working - instant update / working
- Configure tab -  instant update / working

## How do I test this PR?

- Check that Talk is working as expected.


client/coral-framework/actions/asset.js
```diff
@@ -10,30 +10,30 @@ export const fetchAssetRequest = () => ({type: actions.FETCH_ASSET_REQUEST});
 export const fetchAssetSuccess = asset => ({type: actions.FETCH_ASSET_SUCCESS, asset});
 export const fetchAssetFailure = error => ({type: actions.FETCH_ASSET_FAILURE, error});
 
-const updateConfigRequest = () => ({type: actions.UPDATE_CONFIG_REQUEST});
-const updateConfigSuccess = config => ({type: actions.UPDATE_CONFIG_SUCCESS, config});
-const updateConfigFailure = () => ({type: actions.UPDATE_CONFIG_FAILURE});
+const updateAssetSettingsRequest = () => ({type: actions.UPDATE_ASSET_SETTINGS_REQUEST});
```
jde, line 13: Great initiative to keep naming conventions clean. Things like this make all the difference toward readability.

okbel:
Ready to be tested :)

client/coral-framework/graphql/queries/myCommentHistory.graphql
```diff
@@ -0,0 +1,9 @@
+query myCommentHistory {
```
jde, line 1: What is the advantage of breaking these into their own files? It seems to be that having one file per query containing the fragment and the wrapper would be encapsulate each query/mutation cleanly and minimize the number of imports/files.

client/coral-ui/components/Checkbox.js
```diff
@@ -1,9 +1,9 @@
 import React from 'react';
 import styles from './Checkbox.css';
 
-export default ({name, cStyle = 'base', onChange, label, className, info, checked = 'false'}) => (
+export default ({name, cStyle = 'base', onChange, label, className, info, ...props}) => (
```
jde, line 4: Why pass in all ...props here? We have seen this pattern cause unexpected problems when things are changed upstream.

jde (changes requested):

client/coral-framework/graphql/queries/myCommentHistory.graphql
okbel: Are you referring to the fact that the `fragments` are in another folder? That's because `fragments` should be unique. That's why we need to import them everytime we need them. Also, I think that breaking the queries into different files is cleaner and clearer. 

Another thing is that: Fragments are no longer suported using `createFragment` https://github.com/apollostack/react-apollo/issues/435 (I created this because the Docs are not updated)



We no longer can send the fragment like this.

```
import commentView from '../fragments/commentView.graphql';

export const queryStream = graphql(STREAM_QUERY, {
  options: () => ({
    fragments: [commentView],
    variables: {
      asset_url: url
    }
  })
});
```

client/coral-ui/components/Checkbox.js
okbel: It's not the same that we discussed with Riley and you the other day.

The reason behind using `...props` here it's because if I want to use `defaultValue` or `value` (meaning that I want an Uncontrolled or a Controlled Component) I need to use one of them. And that is the prop that will finally get passed to the input. 

This is a UI component. I should be able to pass as many custom props (attributes) as needed. It's not the same as the Components that we have for Stream and such. Those are not UI Components, meaning no custom props passed.

client/coral-framework/graphql/queries/myCommentHistory.graphql
jde: Makes sense.

jde (approved):

PR #268, opened 2017-01-27
Graph Pagination

wyattjoh:
## What does this PR do?

- Refines edges to support pagination
- Adds new counting edges
- Adds new shared DataLoader cache class

### About

Using the following queries, we can see that using the `query AssetQuery`, we can get the initial stream. Future calls can use the `query LoadCommentCounts` to load counts based on the current view, whereas `query LoadMoreComments` can be used to load more based on a specific query.

**Note that the `$cursor` is a `Date` scalar type.**

```graphql
fragment commentView on Comment {
  id
  body
  created_at
}

query AssetQuery($asset_url: String!, $limit: Int = 5, $sort: SORT_ORDER) {
  asset(url: $asset_url) {
    id
    commentCount
    comments(sort: $sort, limit: $limit) {
      ...commentView
      replyCount
      replies(limit: 3) {
        ...commentView
      }
    }
  }
}

query LoadCommentCounts($asset_id: ID, $limit: Int = 5, $sort: SORT_ORDER) {
  asset(id: $asset_id) {
    id
    commentCount
    comments(sort: $sort, limit: $limit) {
      id
      replyCount
    }
  }
}

query LoadMoreComments($limit: Int = 5, $cursor: Date, $parent_id: ID, $asset_id: ID, $sort: SORT_ORDER) {
  new_top_level_comments: comments(query: {limit: $limit, cursor: $cursor, parent_id: $parent_id, asset_id: $asset_id, sort: $sort}) {
    ...commentView
  }
}
```

Example variables:

```json
{
  "limit": 5,
  "asset_url": "http://localhost:3001/",
  "asset_id": "671a53bb-ae7a-4d2d-98b0-5cd3a34d6eb7",
  "cursor": "2017-01-25T23:51:08.071Z",
  "sort": "REVERSE_CHRONOLOGICAL",
  "parent_id": null
}
```

### DIFF
--- a/graph/loaders/comments.js
+++ b/graph/loaders/comments.js
@@ -1,85 +1,134 @@
-const DataLoader = require('dataloader');
-
 const util = require('./util');
 
-const ActionModel = require('../../models/action');
 const CommentModel = require('../../models/comment');
-const CommentsService = require('../../services/comments');
 
 /**
- * Retrieves comments by an array of asset id's, results are returned in reverse
- * chronological order.
- * @param {Array} ids array of ids to lookup
+ * Returns the comment count for all comments that are public based on their
+ * asset ids.
+ * @param {Object}        context     graph context
+ * @param {Array<String>} asset_ids   the ids of assets for which there are
+ *                                    comments that we want to get
  */
-const genCommentsByAssetID = (context, ids) => {
-  return CommentModel.find({
-    asset_id: {
-      $in: ids
+const getCountsByAssetID = (context, asset_ids) => {
+  return CommentModel.aggregate([
+    {
+      $match: {
+        asset_id: {
+          $in: asset_ids
+        },
+        status: {
+          $in: [null, 'ACCEPTED']
+        },
+        parent_id: null
+      }
     },
-    parent_id: null,
-    status: {
-      $in: [null, 'ACCEPTED']
+    {
+      $group: {
+        _id: '$asset_id',
+        count: {
+          $sum: 1
+        }
+      }
     }
-  })
-  .sort({created_at: -1})
-  .then(util.arrayJoinBy(ids, 'asset_id'));
+  ])
+  .then(util.singleJoinBy(asset_ids, '_id'))
+  .then((results) => results.map((result) => result ? result.count : 0));
 };
 
 /**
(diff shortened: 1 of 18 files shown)


graph/loaders/comments.js
```diff
@@ -1,85 +1,134 @@
...
-      $in: ids
+const getCountsByAssetID = (context, asset_ids) => {
+  return CommentModel.aggregate([
+    {
+      $match: {
+        asset_id: {
+          $in: asset_ids
+        },
+        status: {
+          $in: [null, 'ACCEPTED']
```
jde, line 20: This works, but in the future we may want to abstract the concept of 'publicly view-able comments' into a function that returns or a constant that holds this array. I can see this array being hard coded in several places.

jde:
Love the solutions. Very clean and the public endpoints scale nicely to 2+ MM comments.

graph/loaders/comments.js
```diff
@@ -1,85 +1,134 @@
...
 
-  // TODO: remove when we move the enum over to the uppercase.
-  if (status) {
-    status = status.toLowerCase();
+  // Only administrators can search for comments with statuses that are not
+  // `null`, or `'ACCEPTED'`.
+  if (user != null && user.hasRoles('ADMIN') && statuses) {
+    comments = comments.where({
+      status: {
+        $in: statuses
```
jde, line 84: Does this query load all the comments in the db that have these statuses? If so, will this cause memory problems on a server with large numbers of comments?

jde (changes requested):
More clarification than changes (see above). The functionality looks solid and scales seamlessly to my test db of 2.3 million comments.

graph/loaders/comments.js
wyattjoh: If you notice, there are many different groupings applied here. These are called via the resolvers and are never done so without an asset_id being added to the query at least. This way of composing a query is more flexible as it doesn't actually execute the search until the function exits.

jde: Ah, I see that this is not indeed a query, but appending a where clause. Makes sense. I like the pattern of composing a query one step at a time.

jde (approved):

okbel (approved):

PR #270, opened 2017-01-30
Fixing bug where timeAgo cannot interpret dates from mongoose properly.

davidgljay:
## What does this PR do?

Fixes an issue where many older comments appear as "Just now"

## How do I test this PR?

1. Find a comment stream where you have commented at a range of times in the past.
2. The older comments should not appear as "Just now."


gabelula (approved):
It looks good to me.

PR #274, opened 2017-01-31
Verify Email upon signup

riley:
## What does this PR do?

- adds a setting in the admin to enable email verification for new users
- email is sent to new users upon sign up
- clicking the link will confirm your account
- if the user got distracted upon signing up the first time, or lost the email, they can request a new one after trying to sign in with the right credentials

## How do I test this PR?

- turn on Require Email Confirmation in the admin
- create a new user
- click the confirm link in your email
- if you try to request an email again if you're already verified, it should bounce you
- try to log in again. it should work.


davidgljay:
After verifying my e-mail I'm redirected to `http://localhost:3000/admin/undefined`. Is this intentional?

client/coral-sign-in/containers/SignInContainer.js
```diff
@@ -71,6 +75,17 @@ class SignInContainer extends Component {
     });
   }
 
+  handleChangeEmail(e) {
+    const {value} = e.target;
+    console.log('handleChangeEmail', value);
```
davidgljay: This should be removed.

client/coral-sign-in/containers/SignInContainer.js
```diff
@@ -71,6 +75,17 @@ class SignInContainer extends Component {
...
+    this.setState({emailToBeResent: value});
+  }
+
+  handleResendConfirmation(e) {
+    e.preventDefault();
+    this.props.requestConfirmEmail(this.state.emailToBeResent);
```
davidgljay: This doesn't seem to give the user a confirmation or close the dialog. It should probably do one or both.

davidgljay (changes requested):
Remove a console.log and address issue with resent confirmation e-mail dialog not closing.

wyattjoh (commented):
Should handle the state where the referer isn't defined or isn't a url first, possibly create a validateReferer method to the SettingsService and call out to it?

services/users.js
```diff
@@ -538,40 +538,51 @@ module.exports = class UsersService {
    * @param  {String} email The email that we are needing to get confirmed.
    * @return {Promise}
    */
-  static createEmailConfirmToken(userID, email) {
+  static createEmailConfirmToken(userID = null, email, referer) {
```
wyattjoh: Should default to something here if referer isn't provided.

routes/api/users/index.js
```diff
@@ -113,10 +116,12 @@ const SendEmailConfirmation = (app, userID, email) => UsersService
     });
   });
 
+// create a local user.
 router.post('/', (req, res, next) => {
   const {
     email,
     password,
+    redirectUri,
```
wyattjoh: Should source this from the `Referrer` header, not a payload option.

wyattjoh (changes requested):

routes/api/users/index.js
riley: whoops, good catch. I removed that from somewhere else when I was moving things around and forgot to re-add it here.

wyattjoh (approved):

davidgljay (approved):

PR #276, opened 2017-01-31
Addressing console errors in asset search.

davidgljay:
## What does this PR do?

Addresses a bug in asset search where the search field did not update and errors appeared in the console. This was caused by a non-persisting react event.

## How do I test this PR?

1. Go to http://localhost:3000/admin/streams
2. Enter a search term.
3. No errors should appear in the console
4. The stream list should update as you type.


riley (changes requested):
would it be at all possible to make this a debounce instead of just a setTimeout? it feels laggy since it's always waiting at least 350 ms  https://lodash.com/docs/4.17.2#debounce

riley (approved):

PR #277, opened 2017-01-31
Adds a screen to create username after facebook signup

gabelula:
## What does this PR do?

- Adds a dialog window after singup to create/modify display Name.

## How do I test this PR?

- Signup with facebook
- Check that the dialog is there
- Change your display name
- Check that it was changed.

Story: https://www.pivotaltracker.com/story/show/134624635


riley (changes requested):
you have to show the error for when a username has profanity in it.

screen shot 2017-02-01 at 12 12 13 pm

Also have to show the error when there are special characters in the displayName

screen shot 2017-02-01 at 12 13 52 pm

gabelula:
Done. Ready to review.

client/coral-framework/actions/auth.js
```diff
@@ -8,6 +8,28 @@ import coralApi, {base} from '../helpers/response';
...
+      dispatch(createDisplayNameSuccess());
+      dispatch(hideCreateDisplayNameDialog());
+      dispatch(updateDisplayName(formData.displayName));
+    })
+    .catch(error => {
+      dispatch(createDisplayNameFailure(lang.t(`error.${error.message}`)));
+    });
+};
+
+const createDisplayNameSuccess = () => ({type: actions.CREATEDISPLAYNAME_SUCCESS});
```
riley: would you mind moving these two actions up to line 12 so that they're grouped in the code with the other fetch actions?

client/coral-framework/components/CreateDisplayNameDialog.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
riley: These components probably make more sense to live in the `coral-plugin-sign-in` folder since that's where a lot of similar code is from.

client/coral-framework/components/FormField.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import styles from './styles.css';
+
+const FormField = ({className, showErrors = false, errorMsg, label, ...props}) => (
```
riley: Is it possible to use the `FormField` from `coral-ui`?

client/coral-framework/containers/ChangeDisplayNameContainer.js
```diff
@@ -0,0 +1,139 @@
+import React, {Component} from 'react';
```
riley: This could move to `coral-plugin-sign-in` along with the styles too

riley (changes requested):
the code looks good, just some organizational stuff requested

client/coral-framework/components/FormField.js
riley: you could probably get away with deleting this altogether if you used the `coral-ui` one

davidgljay:
Language question: is it a username or a display name? We should choose one and stick to it.

riley (approved):
lgtm 💯

PR #282, opened 2017-02-01
Hide Pre-Mod tab sometimes

riley:
## What does this PR do?

if Premoderation is enabled globally or there are pre-mod comments, then show the Pre-Mod tab. Otherwise, hide it.

## How do I test this PR?

- enable pre-moderation globally
- make a comment and do not moderate it
- Pre-mod tab should be there
- disable pre-moderation globally
- Pre-mod tab should still be there
- moderate the outstanding comment
- POOF! the Pre-Mod tab is gone.


jde:
Works like a charm. Product / QA approved!

client/coral-admin/src/actions/comments.js
```diff
@@ -14,30 +14,30 @@ export const fetchModerationQueueComments = () => {
     dispatch({type: commentTypes.COMMENTS_MODERATION_QUEUE_FETCH_REQUEST});
 
     return Promise.all([
-      coralApi('/queue/comments/pending'),
+      coralApi('/queue/comments/premod'),
       coralApi('/queue/users/pending'),
       coralApi('/queue/comments/rejected'),
```
okbel, line 19: I'm wondering why are we making 3 calls just to get the total comments :thinking: @jde @riley

okbel:
Works perfect! My only concern is why are we making 3 calls on the initial render, we can make only one and then update each one by endpoint.

I guess we can improve this in the future!

@jde @riley

okbel (approved):

client/coral-admin/src/actions/comments.js
jde: Good thought. We have stories coming up (pagination, more data in the queues, loading counts from count edges of the graph) that will refactor this heavily. Let's address this issue in a future pr.

Ultimately, the current mod queues

jde:
@riley The Pending tab is not hiding on staging despite POST mod status and nothing in the queue:

https://talk-stg.coralproject.net/admin

Please take a look and see why the hiding isn't being triggered.  Thanks!

client/coral-admin/src/actions/comments.js
okbel: 👍 Great!

PR #283, opened 2017-02-02
Username banning flow

davidgljay:
## What does this PR do?

Lets moderators force users to update their usernames. Also removes bios to reduce moderation complexity (at Greg's suggestion.)

## How do I test this PR?

1. Flag a username.
2. Go to the moderation queue. You should see a flagged username.
3. Click the red X.
4. You should see a series of dialogs about suspending the user (new language is pending from design.)
5. Click through these dialogs. The user should receive an e-mail.
6. Log in as the newly banned user, you should see a message and a prompt to update your username.
7. Try entering an invalid username (e.g. one with a space), you should receive an error message.
8. Try entering a username already listed in the system, you should receive an error message.
9. Try entering a valid username. The page should reload and you should be logged in normally.
10. Return to the admin panel. The new username should appear as flagged.

## Note
1. The page reloading at the end of the username update flow is so we don't have to worry about updating state to change the users' status in the auth store. This use case happens pretty rarely, and I want to minimize the amount of custom action and reducer code dedicated to it.
2. As per @wyatt's recommendation that we don't handle account functions with graphql I am updating the username through a REST endpoint.


routes/api/account/index.js
```diff
@@ -115,19 +115,18 @@ router.put('/password/reset', (req, res, next) => {
...
-    .catch((err) => {
-      next(err);
+    .catch(error => {
+      if (error.code === 11000) {
+        next(errors.ErrDisplayTaken);
+      } else {
+        next(error);
+      }
     });
 });
```
wyattjoh, line 131: Why wasn't this added as a mutation instead of a reset endpoint?

routes/api/users/index.js
```diff
@@ -60,6 +60,12 @@ router.post('/:user_id/status', authorization.needed('ADMIN'), (req, res, next)
     .catch(next);
 });
 
+router.post('/:user_id/username-enable', authorization.needed('ADMIN'), (req, res, next) => {
+  UsersService
+    .toggleNameEdit(req.params.user_id, true)
+    .then(() => {
+      res.status(204).end();
+
 router.post('/:user_id/displayname', authorization.needed(), (req, res, next) => {
```
wyattjoh: Going forward we should not add any more REST routes unless they pertain to auth.

services/users.js
```diff
@@ -349,8 +349,15 @@ module.exports = class UsersService {
...
-    return UserModel.update({id}, {$set: {status}});
+    
+    return UserModel.findOne({id})
+      .then((user) => {
+        if (user.status === 'APPROVED' && status === 'PENDING') {
+          return Promise.resolve();
+        } else {
+          return UserModel.update({id}, {$set: {status}});
+        }
+      });
```
wyattjoh: Better to replace this with an atomic operation:

```js
return UserModel.update({
  id,
  status: {
    $ne: 'APPROVED'
  }
}, {
  $set: {status}
});
```

services/users.js
```diff
@@ -644,4 +651,36 @@ module.exports = class UsersService {
...
+    return UserModel.findOne({id})
+      .then((user) => {
+        return user.canEditName ?
+        UserModel.update({id}, {
+          $set: {
+            displayName: displayName.toLowerCase(),
+            canEditName: false,
+            status: 'PENDING'
+          }
+        })
```
wyattjoh: Same as above:

```js
return UserModel.update({
  id,
  canEditName: true
}, {
  $set: {
    displayName: displayName.toLowerCase(),
    canEditName: false,
    status: 'PENDING'
  }
});
```

This will not return an error if they aren't allowed to do this, but I think this is a safer tradeoff.

wyattjoh (dismissed):
Some comments on the database actions.

routes/api/account/index.js
davidgljay: I'm going to switch this over in a seperate PR to avoid further merge conflicts piling up.

kgardnr (approved):

PR #287, opened 2017-02-03
Default asset url

jde:
## What does this PR do?

Defaults the asset_url to the pym generated parentUrl if asset_url isn't explicitly passed.

## How do I test this PR?

Embed this stream without an asset_url. The stream should reflect the parent's url.


jde:
I realized the bug was elsewhere. PEBKAC.  Closing.

PR #288, opened 2017-02-03
Futureproof Embed Code

gobengo:
Story: https://www.pivotaltracker.com/story/show/138858105
Implementation Plan: https://github.com/gobengo/talk/wiki/Embed-Code-for-Talk-Stream

New embed code like:
```
<script src="http://your-talk-installation/Coral.js" onload="
      Coral.Talk.render(document.getElementById('some-div'), {
        talk: 'http://your-talk-installation'
      })
"></script>
```

## What does this PR do?

* Add client/coral-embed-api module. It's meant to be included in an HTML page and define window.Coral
* Serve coral-embed-api at /Coral.js
* Update / handler (article view) to use new embed format
* Update Talk Admin embed code section to use new embed format

## How do I test this PR?

* The embed at '/' should load normally
* In the Admin -> Configure -> Embed Code pane, you should see a new format (containing 'Coral.Talk.render...'). And you should be able to copy paste that into a jsbin and see the embed load


app.js
```diff
@@ -34,6 +34,7 @@ app.use(helmet({
 }));
 app.use(bodyParser.json());
 app.use('/client', express.static(path.join(__dirname, 'dist')));
+app.get('/Coral.js', (req, res) => res.sendFile(path.join(__dirname, 'client/coral-embed-api/index.js')));
```
gobengo: I defer to you on exact best URL, though I personally prefer the aesthetics of this.

I definitely recommend using a URL that doesn't depend on the structure of `dist/` or `client/` (as the current /client/ route does), as those details will change over time, and you don't want embed codes to have to change.

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
...
+  // This function should return value of window.Coral
+  let pym = requirePym();
+  let Coral = {};
+  let Talk = Coral.Talk = {};
+
+  /**
+   * Render a Talk stream
+   * @param {HTMLElement} el - Element to render the stream in
+   * @param {Object} opts - Configuration options for talk
+   * @param {String} opts.talk - Talk base URL
```
gobengo: It may be desirable to not require this here, but it's needed to build the iframe URL. It *could* be added to this script at build time (or render of the .js time), but right now this is just a static script so it needs to be provided by user at time of use.

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
...
+   * @param {String} [opts.title] - Title of Stream (rendered in iframe)
+   * @param {String} [opts.asset] - parent Asset ID or URL. Comments in the
+   * stream will replies to this asset
+   */
+  Talk.render = function (el, opts) {
+    console.log('Coral.Talk.render()', el, opts);
```
gobengo: oops I'll remove this.

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
+/* eslint-disable no-var */
+/**
+ * @file Exposes the main 'window.Coral.Talk' API that developers can use to
```
gobengo: @jde I forgot what you told me to call this module so made something up. Remind me if you want me to change it.

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
...
+      encodeURIComponent(asset)
+    ].join('');
+    return iframeUrl;
+  }
+
+  // Set up postMessage listeners/handlers on the pymParent
+  // e.g. to resize the iframe, and navigate the host page
+  function configurePymParent(pymParent, assetUrl) {
+    let notificationOffset = 200;
+    let ready = false;
```
gobengo: I shouldn't use 'const' and 'let' in this file since it doesn't get babel-d, and oldish browsers won't like those keywords. I'll remove these.

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
...
+
+    // Resize parent iframe height when child height changes
+    pymParent.onMessage('height', function(height) {
+
+      // TODO: In local testing, this is firing nonstop. Maybe there's a bug on the inside?
```
gobengo: Heads up, I noticed this while devving and may be worth creating a task in pivotal to investigate. Basically the 'height' callback is just constantly executing when talk is on a page. And because it updates the DOM each time it's going to have a big negative effect on page performance.

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
...
+      pymParent.sendMessage('position', position);
+    });
+
+    // Tell child when parent's DOMContentLoaded
+    pymParent.onMessage('childReady', function () {
+      var interval = setInterval(function () {
+        if (ready) {
+          window.clearInterval(interval);
+
+          // @todo - It's weird to me that this is sent here in addition to the iframe URL. Could it just be in one place?
```
gobengo: I'm curious about this too. The asset URL is sent to the child both in the iframe URL and via this message. Are both required? Could that be changed to just one to avoid out-of-sync bugs later?

client/coral-embed-api/index.js
```diff
@@ -0,0 +1,164 @@
...
+    let pym;
+
+    // fake AMD `define` so that the pym.js copypasta doesn't create a global
+    function define(createPym) {
+      pym = createPym();
+    }
+    define.amd = true;
+
+    /* eslint-disable */
+    /* ! pym.js - v1.1.2 - 2016-10-25 */
```
gobengo: I think if we don't further process this file, the LICENSE of pym.js may require me to copy paste it into this module.
https://github.com/nprapps/pym.js/blob/master/LICENSE

client/coral-embed-api/index.js
jde: Let's call this `coral-embed` as this is the common entry point for `coral-embed-*`.

app.js
jde: My first thought was to add a simply copy operation to the webpack file to move this static file into `/dist/` so it can be served via our static file route `/client/coral.js`. Although now that you mention it, keeping this explicit route will allow us ultimate flexibility of pathing this specific file to various locations / versions.

client/coral-embed/index.js
```diff
@@ -0,0 +1,163 @@
+/* eslint-disable no-var */
+/**
+ * @file Exposes the main 'window.Coral.Talk' API that developers can use to
```
gobengo: I think if we don't further process this file, the LICENSE of pym.js may require me to copy paste it into this module.
https://github.com/nprapps/pym.js/blob/master/LICENSE

client/coral-embed/index.js
```diff
@@ -0,0 +1,163 @@
...
+  // This function should return value of window.Coral
+  let pym = requirePym();
+  let Coral = {};
+  let Talk = Coral.Talk = {};
+
+  /**
+   * Render a Talk stream
+   * @param {HTMLElement} el - Element to render the stream in
+   * @param {Object} opts - Configuration options for talk
+   * @param {String} opts.talk - Talk base URL
```
gobengo, line 40: It may be desirable to not require this here, but it's needed to build the iframe URL. It could be added to this script at build time (or render of the .js time), but right now this is just a static script so it needs to be provided by user at time of use.

client/coral-embed-api/index.js
gobengo: done

client/coral-embed-api/index.js
jde: Good thought.

client/coral-embed/index.js
```diff
@@ -0,0 +1,163 @@
...
+  // Set up postMessage listeners/handlers on the pymParent
+  // e.g. to resize the iframe, and navigate the host page
+  function configurePymParent(pymParent, assetUrl) {
+    let notificationOffset = 200;
+    let ready = false;
+
+    // Resize parent iframe height when child height changes
+    pymParent.onMessage('height', function(height) {
+
+      // TODO: In local testing, this is firing nonstop. Maybe there's a bug on the inside?
```
gobengo, line 98: Heads up, I noticed this while devving and may be worth creating a task in pivotal to investigate. Basically the 'height' callback is just constantly executing when talk is on a page. And because it updates the DOM each time it's going to have a big negative effect on page performance.

client/coral-embed/index.js
```diff
@@ -0,0 +1,163 @@
...
+      pymParent.sendMessage('position', position);
+    });
+
+    // Tell child when parent's DOMContentLoaded
+    pymParent.onMessage('childReady', function () {
+      var interval = setInterval(function () {
+        if (ready) {
+          window.clearInterval(interval);
+
+          // @todo - It's weird to me that this is sent here in addition to the iframe URL. Could it just be in one place?
```
gobengo, line 120: I'm curious about this too. The asset URL is sent to the child both in the iframe URL and via this message. Are both required? Could that be changed to just one to avoid out-of-sync bugs later?

client/coral-embed/index.js
```diff
@@ -0,0 +1,163 @@
...
+    let pym;
+
+    // fake AMD `define` so that the pym.js copypasta doesn't create a global
+    function define(createPym) {
+      pym = createPym();
+    }
+    define.amd = true;
+
+    /* eslint-disable */
+    /* ! pym.js - v1.1.2 - 2016-10-25 */
```
gobengo: I think if we don't further process this file to do it automatically, the LICENSE of pym.js may require me to copy paste it into this module.
https://github.com/nprapps/pym.js/blob/master/LICENSE

client/coral-embed-api/index.js
jde: As far as I now, only the url param `asset_url` is used. If it is sent elsewhere we should investigate and clean up.

client/coral-embed-api/index.js
jde: Let's put it in to err on the side of compliance.

jde (changes requested):

client/coral-embed/index.js
```diff
@@ -0,0 +1,163 @@
...
+    opts = opts || {};
+
+    // @todo infer this URL without explicit user input (if possible, may have to be added at build/render time of this script)
+    if (! opts.talk) {
+      throw new Error('Coral.Talk.render() expects opts.talk as the Talk Base URL');
+    }
+
+    // ensure el has an id, as pym can't directly accept the HTMLElement
+    if ( ! el.id) {el.id = `_${  String(Math.random())}`;}
+    let asset = opts.asset || window.location;
```
wyattjoh: We have to support "asset_id + asset_url" or just "asset_url". This might require further expansion here.

app.js
```diff
@@ -34,6 +34,7 @@ app.use(helmet({
 }));
 app.use(bodyParser.json());
 app.use('/client', express.static(path.join(__dirname, 'dist')));
+app.get('/Coral.js', (req, res) => res.sendFile(path.join(__dirname, 'client/coral-embed/index.js')));
```
wyattjoh: I'd rather be more clear here and rename this to be: `/embed.js`.

wyattjoh (changes requested):

app.js
jde: Please update the route to `/coral-latest.js`. This will allow us to potentially host multiple versions in a semantic way.

client/coral-embed/index.js
jde: At the moment `coral-embed-stream` only looks at `asset_url`. We will, in the future, want to extend this to also accept `asset_id`.

client/coral-embed/index.js
gobengo: done

app.js
jde: +1 to `/embed.js`.

client/coral-embed/index.js
gobengo: I added it to bottom of file

jde: Let's add the license to err on the side of compliance.

wyattjoh: We should be able to add a step to webpack to build in licenses into the embed file.

jde:
Please update the Embed link in our admin dashboard to reflect the new way of things:

https://github.com/coralproject/talk/blob/master/client/coral-admin/src/containers/Configure/EmbedLink.js#L28

gobengo:
@jde I think I did already (I edited an EmbedLink.js).

app.js
gobengo: done

jde (approved):
This is fantastic work. Thanks!

wyattjoh (approved):

PR #294, opened 2017-02-07
fix a warning in the console when you're signed out

riley:
## What does this PR do?

I accidentally a proptype.

## How do I test this PR?

- be signed out in the comment stream
- observe the lack of errors in the console
- revel.


jde (changes requested):
I'm still seeing the Auth error on that branch: 
unspecified

Is this the error you were referring to?

riley:
@jde no, you should be seeing this on `master` which was an oversight on my part. I think the not logged in error is just going to be there since the response from auth is a 401.
screen shot 2017-02-07 at 10 59 27 am

jde (approved):

jde:
Ok, confirmed that warning is gone.

PR #295, opened 2017-02-07
Story 138524853 domain whitelist

gabelula:
## What does this PR do?

- Adds a whiltelist domains attribute to the Settings model.
- Adds a service to be able to work with the domains (match and add them).
- Adds the FE to Admin/Configure to be able to add and remove domains.
- Blocks domains that do not match when creating the Asset. (Asset service, findOrCreateByUrl)
- Add tests for the service

## How do I test this PR?

- Remove localhost Domain in Admin/Configure
- Remove the asset from the Asset collection in the DB
- Go to localhost and check that it does not work (It is throwing an error right now as we still do not resolve the errors in the loader/resolver and do not have a response when creating restricted asset url)

Something that is missing is a message for the user when embedding the asset that says that is not permitted. 

This is the story: https://www.pivotaltracker.com/story/show/138524853


jde (approved):
Looks good!

PR #296, opened 2017-02-07
Using GraphQL and Apollo in Talk Admin.

okbel:
This is the minimal configuration to have Apollo running in Talk Admin.

We will go over this configuration tomorrow morning :)


riley (approved):
lgtm. thanks for setting this up 💯

PR #298, opened 2017-02-09
Bugfix 9000

jde:
## What does this PR do?

Fixes embedding error that broke our faux article when using ids.

## How do I test this PR?

You should be able to embed streams both with default urls and via ids.


gabelula (approved):
It looks good.

PR #299, opened 2017-02-09
Inital rename of displayName -> username

wyattjoh:
## What does this PR do?

- Renames displayName -> username
- **Allows uppercase characters in usernames**, fixed lowercase index issue

## How do I test this PR?

Run the following migration if you have not already done so:
```
db.users.find({}).forEach(function(user) {
  db.users.update({id: user.id}, {$set: {
    username: user.displayName,
    lowercaseUsername: user.displayName
  }});
});
```

- Run e2e
- Run tests


gabelula (changes requested):
It looks good but tests are failing.

gabelula (approved):
Tests work now.

PR #300, opened 2017-02-10
Tag comments made by Staff with 'STAFF' tag automatically.

gabelula:
## What does this PR do?

- add tags to the Comment model
- automatically add 'STAFF' tag to the Comment when author is ADMIN or MODERATOR
- adds tag plugin to show a label to the comment when is staff 

## How do I test this PR?

- have an admin user and a non-admin user
- make comments with both of them
- check that only the comments made by the admin have the tag 'Staff' in the stream

Pivotal Story: https://www.pivotaltracker.com/story/show/137818425


wyattjoh (approved):
I moved the tag creation out of the comments service which also removed a round trip to the DB, so it should be good to go =D looking great!

PR #305, opened 2017-02-13
add a non-breaking space in between username and pubdate

riley:
## What does this PR do?

There was no space between the author name and the published date on a comment. Fixed.

## How do I test this PR?

- gaze upon any comment
- see that there is a space in between the author name and "a day ago" or whatever


gabelula:
Shouldn't this be made via CSS ?

riley:
that's a good question. That would probably give more predictable behavior. brb

kgardnr (approved):

PR #306, opened 2017-02-13
Load more comments

davidgljay:
## What does this PR do?

Loads only 10 comments initially and adds a "load more" button to load more in batches of 10.

## How do I test this PR?

1. Go to a comment stream with less than 10 comments.
2. Load More button should not be visible.
3. Comment until there are > 10 comments.
4. Load more button should be visible.
5. Clicking the button should load more comments.


riley:
the big thing that I noticed was if there are 10 comments on an article and you make the 11th, it hides the first comment, and you can only get it back when you click View more. I spoke with the designers and it seems more intuitive to just add them to the stack instead of hitting the server for a comment that was visible only a second ago.

davidgljay:
@riley PropTypes don't do anything on stateless components (though we could keep them there purely for devleoper reference.)

davidgljay:
@riley I'd like to fix the issue you mentioned in a seperate PR. Right now we refetch after posting a comment, we should inject that comment directly through updateQuery. That'll avoid a number of cases where post a comment or reply makes things disappear.

davidgljay:
Closed on accident!

riley:
@davidgljay `propTypes` work on stateless components. Try setting one to `isRequired` and then not including it. It doesn't break, but it'll give a warning.

davidgljay:
@riley Really?? That's fantastic. I remember researching it earlier and that wasn't the case. Good to know.

riley:
I think earlier versions of React didn't when they first rolled out stateless components.

scripts/pree2e.sh
```diff
@@ -6,6 +6,9 @@ set -e
 # install selenium
 selenium-standalone install --config=./selenium.config.js
 
+# Clear db
+mongo test --eval "db.dropDatabase()"
```
riley, line 10: 💯 oonz oonz oonz

client/coral-embed-stream/src/Embed.js
```diff
@@ -158,6 +159,11 @@ class Embed extends Component {
               clearNotification={this.props.clearNotification}
               notification={{text: null}}
             />
+          <LoadMore
+            asset_id={asset.id}
```
okbel: We are using CamelCase convention. This would be `assetId`.

test/e2e/tests/EmbedStreamTests.js
```diff
@@ -119,48 +119,51 @@ module.exports = {
...
-          .assert.containsText('#coral-notif', 'moderation team');
-        done();
-      })
-      .catch((err) => {
-        console.log(err);
-        done();
-      });
-    });
-  },
+  // Commenting out this test, it fails if the notification is below the fold, which
```
okbel, line 123: We will get back and review how to approach to this test after the Test Meeting

client/coral-embed-stream/src/Embed.js
```diff
@@ -158,6 +159,11 @@ class Embed extends Component {
...
+            comments={asset.comments}
+            moreComments={asset.commentCount > asset.comments.length}
```
okbel: Isn't the `commentCount` the total comment count? This will be always greater than the length of the comments array if the limit is 5.

okbel (changes requested):

client/coral-embed-stream/src/Embed.js
jde: This is a bit confusing, but stems from a product decision. The call was only to include top level comments in the commentCount we show in the tab and not count children.

So, the variable name is misleading, but the confusion stems from our comment counting methodology.

okbel: 👍

test/e2e/tests/EmbedStreamTests.js
davidgljay: Once we are clearing the DB between tests this wont't be an issue.

client/coral-embed-stream/src/Embed.js
davidgljay: @okbel Once you hit "load more" the coments array can be greater than 5, so this is essentially saying "is the commentCount greater than the number of comments currently loaded?"

okbel (approved):

PR #310, opened 2017-02-14
fix Configure in admin

riley:
## What does this PR do?

Something happened where the shape of the settings state passed to the Config object was broken. This reverts the changes.

## How do I test this PR?

- go to Configure
- update settings. it should work
- add some banned words. it should work
- add some whitelisted domains. it should work

### DIFF
--- a/client/coral-admin/src/reducers/settings.js
+++ b/client/coral-admin/src/reducers/settings.js
@@ -2,14 +2,12 @@ import {Map, List} from 'immutable';
 import * as actions from '../actions/settings';
 
 const initialState = Map({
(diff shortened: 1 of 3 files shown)


client/coral-admin/src/reducers/settings.js
```diff
@@ -2,14 +2,12 @@ import {Map, List} from 'immutable';
 import * as actions from '../actions/settings';
 
 const initialState = Map({
-  settings: Map({
```
okbel, line 5: 👍

okbel (approved):

PR #314, opened 2017-02-15
Story 139595043 Question Box to Comment Stream

gabelula:
## What does this PR do?

- Adds the option of save a question box to the settings of the comment stream
- The question box has an enable field AND a content field.
- It adds the question box to the top of the comment stream if it is enable in the settings for the stream.

## How do I test this PR?

- Get an admin user
- Enable the question box in the settings for the comment stream.
- Add the question 
- Go check that the question is at the top of the comment stream.

Story: https://www.pivotaltracker.com/story/show/139595043


gabelula:
Only thing missing is the icon in the question box. That can be added later. This is ready for review.

jde:
I get the following error when loading a stream:

unspecified-1

jde (approved):

PR #316, opened 2017-02-15
Apollo update on post comment

davidgljay:
## What does this PR do?

Displays new comments and replies using Apollo's updateQueries function rather than using refetch. Also implements optimistic updating for comments and replies.

## How do I test this PR?

1. Post a new comment.
2. Post a new reply.
3. Both should appear right away.


client/coral-framework/graphql/mutations/index.js
```diff
@@ -10,16 +10,59 @@ export const postComment = graphql(POST_COMMENT, {
...
           parent_id
+        },
+        optimisticResponse: {
+          createComment: {
+            comment: {
+              user: {
+                id: ownProps.auth.user.id,
+                name: ownProps.auth.user.username
+              },
+              created_at: new Date(),
```
riley: I think this needs to be a string? the dates coming back from the json are ISO strings.

client/coral-framework/graphql/mutations/index.js
```diff
@@ -10,16 +10,59 @@ export const postComment = graphql(POST_COMMENT, {
...
+              status: null,
+              id: `${Date.now()}_temp_id`
+            }
+          }
+        },
+        updateQueries: {
+          AssetQuery: (oldData, {mutationResult:{data:{createComment:{comment}}}}) =>
+
+            // If posting a reply
+            parent_id ? {
```
riley: I love es6 as much as the next guy, but can we get explicit return statements, especially for returning a ternary of objects? I don't want to abuse the arrow syntax.

client/coral-framework/graphql/mutations/index.js
```diff
@@ -10,16 +10,59 @@ export const postComment = graphql(POST_COMMENT, {
...
+              ...oldData,
+              asset: {
+                ...oldData.asset,
+                comments: oldData.asset.comments.map((oldComment) =>
+                  oldComment.id === parent_id
```
riley: same thing here.

client/coral-framework/graphql/mutations/index.js
```diff
@@ -10,16 +10,59 @@ export const postComment = graphql(POST_COMMENT, {
...
+                name: ownProps.auth.user.username
+              },
+              created_at: new Date(),
+              body,
+              parent_id,
+              asset_id,
+              action_summaries: [],
+              tags: [],
+              status: null,
+              id: `${Date.now()}_temp_id`
```
riley, line 35: since this whole object is initialized here, does that mean that every time the Comment schema changes we're going to have to remember to update this?

client/coral-framework/graphql/mutations/index.js
```diff
@@ -10,16 +10,59 @@ export const postComment = graphql(POST_COMMENT, {
...
-      return mutate({
+  props: ({ownProps, mutate}) => ({
+    postItem: ({asset_id, body, parent_id}) =>
+      mutate({
         variables: {
           asset_id,
           body,
           parent_id
+        },
+        optimisticResponse: {
```
riley, line 21: if premod is `POST` the comment still shows up in the feed, but is gone when you refresh. Is this the expected product behavior?

riley (changes requested):
i got this warning in the console after I posted a comment.

screen shot 2017-02-16 at 3 45 53 pm

client/coral-framework/graphql/mutations/index.js
jde: This is a good question that hits at the heart of optimistic updates.

davidgljay: Yup. This is the response we expect to get back from the server. Unless we have some way to generate that automatically, but that also seems brittle.

client/coral-framework/graphql/mutations/index.js
davidgljay: You mean when premod is PRE? Handling that case..

client/coral-framework/graphql/mutations/index.js
riley: we could probably just mark it as `//TODO think of a more solid way to do this` and keep this in the PR for now.

davidgljay:
Re: the console warning: I can't find any documentation referencing it, and adding `addTypeName:true` to the apollo client doesn't fix it.

client/coral-framework/graphql/mutations/index.js
riley: er, yes.

riley (approved):

PR #317, opened 2017-02-15
most flagged assets widget

riley:
## What does this PR do?

Creates a new page `/dashboard` in admin that will contain widgets with information about 🔥  articles and comments.

## How do I test this PR?

- flag a couple of comments in the stream
- go to `/dashboard` in the admin panel within 5 minutes
- there will be two widgets; only the one on the left should have a table in it
- there should be stats for the number of flags and number of comments, as well as a link to the asset


jde (approved):
Looks good!. Empty states, more formatting, etc... to be performed once we have all the widgets into place.

PR #318, opened 2017-02-16
Added count clearing for status changes

wyattjoh:
## What does this PR do?

Resolves issue where comment counts aren't cleared when updated.

## How do I test this PR?

1. Create a comment
2. Reload the page to reload the comment count
3. Flag the comment
4. Go into the admin, reject the comment
5. Go back to the stream, notice that the stream count is correct


graph/mutators/comment.js
```diff
@@ -169,16 +169,26 @@ const createPublicComment = (context, commentInput) => {
...
 
-const setCommentStatus = ({comment}, {id, status}) => {
-  return CommentsService.setStatus(id, status)
-    .then(res => res);
+const setCommentStatus = ({loaders: {Comments}}, {id, status}) => {
+  return CommentsService
+    .setStatus(id, status)
+    .then((comment) => {
+
+      // If the loaders are present, clear the caches for these values because we
```
jde, line 177: This is a good pattern so long as we have a relatively small number of places to clear these caches. As the number of cache clearing triggers grows, let's consider refactoring this cache layer into a service that exposes a tight api.

jde (approved):

PR #321, opened 2017-02-16
Counts on Mod Queues

riley:
## What does this PR do?

- adds a badge to the moderation queue tabs showing the count (will have to be updated when pagination lands in mod queue)
- makes the first tab Dashboard, per @IAmSamHankins 

## How do I test this PR?

- view the moderation queue
- there should be numbers next to the names of the queues


client/coral-admin/src/containers/ModerationQueue/ModerationContainer.js
```diff
@@ -70,8 +70,10 @@ class ModerationContainer extends Component {
       <div>
         <ModerationHeader asset={asset} />
         <ModerationMenu
-          activeTab={activeTab}
           asset={asset}
+          premodCount={data.premod.length}
```
okbel: Not this kind of count. This needs to come back from the server. Otherwise this will mess up the pagination.

okbel (changes requested):

client/coral-admin/src/containers/ModerationQueue/ModerationContainer.js
jde: +1, the front end isn't going to hold all of the things in the queue so we can't base the count on fe data.

riley: i know, but this pr is just UI. we'll pass different ints here when pagination is added (see PR description).

wyattjoh: The data is now available via the commentCount edge, observe:

```graphql
commentCount(query: {statuses: [NONE, PREMOD], action_type: FLAG})
```

okbel (approved):

PR #324, opened 2017-02-16
fix bug when likes or flags were 0 for an article

riley:
## What does this PR do?

The `FlagActionAssetSummary` or `LikeActionAssetSummary` is undefined when there are 0 of the respective actions. This PR adds a safety-check for potentially missing data.

## How do I test this PR?

- have an asset with either 0 likes or 0 flags.
- view the Dashboard.
- it should not crash.


gabelula (approved):
It looks good.

PR #325, opened 2017-02-17
From displayName to username everywhere

gabelula:
## What does this PR do?

- It renames displayname in other places that were missing.
- It also adds an error key for when not authorized to edit username.


wyattjoh (approved):

jde (changes requested):
I updated a dozen or so more places in the translation files, but there are still a number of places in the code that use displayName. 

unspecified-2

gabelula:
Line 106 of services/passport.js has to remain the same because it is the field we are requestion from facebook.  The function findOrCreateExternalUser is getting this fields from the profile that we are getting from FB.

I can change the translation key.

jde:
When I sign up for a new account using a new email, username and password, I get this error:

unspecified-3

jde (approved):

PR #332, opened 2017-02-21
Dashboard: Articles with most likes, articles with most flags

riley:
## What does this PR do?

Have 3 widgets on the dashboard:
1. Most flagged assets
2. Most liked assets
3. ~Most liked comments~

# Open Qs

- The Most Liked Comments list is basically just a `ModerationQueue` component since the idea was to be able to moderate directly from this view. This is kinda weird since post-moderation, the comment would not disappear from the list, since it's still returned by the "most liked" query. Do we want to have a different Comment view altogether? @egrdina @IAmSamHankins @losowsky thoughts?
- this was quite a battle to get Redux and Apollo to play nicely together. There's probably a better way to structure this code (I had to do some contortions to get the metric query to work with a rolling time window). It might just be that I worked on this for too long and I'm missing something obvious.

## How do I test this PR?

- Like or flag a comment
- it should show up in the list on the dashboard.


client/coral-admin/src/actions/install.js
```diff
@@ -75,8 +75,7 @@ export const submitUser = () => (dispatch, getState) => {
     dispatch(installRequest());
     coralApi('/setup', {method: 'POST', body: data})
       .then(result => {
-        console.log(result);
-        dispatch(installSuccess());
+        dispatch(installSuccess(result));
```
okbel, line 78: Why are we passing the result here?

client/coral-admin/src/graphql/queries/index.js
```diff
@@ -1,23 +1,7 @@
...
-    return {
-      variables: {
-        sort: 'FLAG',
-        from: fiveMinutesAgo.toISOString(),
-        to: new Date().toISOString()
-      }
-    };
-  }
-});
+import METRICS from './metrics.graphql';
```
okbel: We are using a naming convetion for queries, `metrics. graphql ` should be `metricsQuery.graphql` Can we rename it?

client/coral-admin/src/graphql/queries/metrics.graphql
```diff
@@ -0,0 +1,38 @@
+fragment metrics on Asset {
```
okbel: We should add this fragment in the fragment folder, so we can reuse it

client/coral-admin/src/graphql/queries/metrics.graphql
```diff
@@ -0,0 +1,38 @@
...
+}
+
+query Metrics ($from: Date!, $to: Date!) {
+  assetsByFlag: assetMetrics(from: $from, to: $to, sort: FLAG) {
+    ...metrics
+  }
+  assetsByLike: assetMetrics(from: $from, to: $to, sort: LIKE) {
+    ...metrics
+  }
+#  mostLikedComments: commentMetrics(from: $from, to: $to, sort: LIKE) {
```
okbel: Do we still need this commented part?

client/coral-admin/src/reducers/settings.js
```diff
@@ -1,11 +1,20 @@
...
+const DASHBOARD_WINDOW_MINUTES = 5;
+let then = new Date();
+then.setMinutes(then.getMinutes() - DASHBOARD_WINDOW_MINUTES);
+
 const initialState = Map({
   wordlist: Map({
     banned: List(),
     suspect: List()
   }),
+  dashboardWindowStart: then.toISOString(),
```
okbel, line 18: :thinking: We can leave this here for now if it doesn't get bigger than that. But I think `Dashboard` will become big enough to have its own reducer, like we do with Moderation for example.

okbel: Also, `settings` are for the asset settings.

okbel (changes requested):

client/coral-admin/src/actions/install.js
riley: so lint will pass. we were logging `result` before, but I deleted the log.

client/coral-admin/src/reducers/settings.js
riley: This settings reducer reflects the state of `models/setting.js` which is for the entire app, afaik. Agreed though on having a separate dashboard reducer eventually. I'll add a comment

client/coral-admin/src/actions/install.js
riley: wyatt added a lint rule so that we don't leave `console.log` in browser code.

okbel (approved):

PR #333, opened 2017-02-21
Cleaned up some routes

wyattjoh:
## What does this PR do?

- Cleans up some routes to standardize

### DIFF
--- a/routes/api/account/index.js
+++ b/routes/api/account/index.js
@@ -102,15 +102,13 @@ router.put('/password/reset', (req, res, next) => {
 
   UsersService
     .verifyPasswordResetToken(token)
-    .then(user => {
+    .then((user) => {
       return UsersService.changePassword(user.id, password);
     })
     .then(() => {
       res.status(204).end();
     })
-    .catch(error => {
-      console.error(error);
-
+    .catch(() => {
       next(authorization.ErrNotAuthorized);
     });
 });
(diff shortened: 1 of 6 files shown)


routes/api/account/index.js
```diff
@@ -121,12 +119,8 @@ router.put('/username', authorization.needed(), (req, res, next) => {
...
       res.status(204).end();
     })
-    .catch(error => {
-      if (error.code === 11000) {
-        next(errors.ErrUsernameTaken);
-      } else {
-        next(error);
-      }
+    .catch((err) => {
+      next(err);
```
gabelula, line 123: Where are we managing the type of error now? Does this work to return the right error in FE?

routes/api/assets/index.js
```diff
@@ -79,7 +79,8 @@ router.get('/:asset_id', (req, res, next) => {
     .findById(req.params.asset_id)
     .then((asset) => {
       if (!asset) {
-        return res.status(404).end();
+        res.status(404).end();
+        return;
```
gabelula, line 83: Why this change?

routes/api/account/index.js
gabelula: I see that it is resolved in the user service.

routes/api/assets/index.js
wyattjoh: It wasn't documented what `res.status(404).end();`, so I didn't feel it necessary for it to return it's result through the promise chain.

gabelula (approved):
Fine if nobody is using the results of res.status(404).end() from that endpoint.

PR #334, opened 2017-02-22
Story 140080313 modal Username 

gabelula:
## What does this PR do?

It changes the design for the modal window that changes the username after facebook signup

This is the story:
https://www.pivotaltracker.com/story/show/140080313

## How do I test this PR?

- Sign up into facebook
- Look at the new modal window.


client/coral-sign-in/components/FakeComment.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
davidgljay, line 1: I believe it is, but why is an extra component necessary here? It seems like it cold be a pain to keep this up to date with the main comment component.

client/coral-sign-in/components/FakeComment.js
```diff
@@ -0,0 +1,66 @@
...
+        <PubDate created_at={created_at} />
+        <Content body={body} />
+          <div className="commentActionsLeft">
+              <div className={`${'coral-plugin-likes'}-container`}>
+                <button className={`${'coral-plugin-likes'}-button`}>
+                  <span className={`${'coral-plugin-likes'}-button-text`}>{lang.t('like')}</span>
+                  <i className={`${'coral-plugin-likes'}-icon material-icons`}
+                    aria-hidden={true}>thumb_up</i>
+                </button>
+              </div>;
```
davidgljay: This semicolon is showing up as text, it should probably be removed.

davidgljay (changes requested):

client/coral-sign-in/components/FakeComment.js
gabelula: Yes. I though about that but if we are using Comment as a fake comment here we need to refactor all the other components as it is impossible to disable the actions that way.

routes/api/auth/index.js
```diff
@@ -70,20 +70,15 @@ const HandleAuthPopupCallback = (req, res, next) => (err, user) => {
...
-      });
-    });
+  // Perform the login of the user!
+  req.logIn(user, (err) => {
+    if (err) {
+      return res.render('auth-callback', {err: JSON.stringify(err), data: null});
+    }
+
+    // We logged in the user! Let's send back the user data.
+    res.render('auth-callback', {err: null, data: JSON.stringify(user), changeusername: true});
```
wyattjoh: It is not evident that this is a sign in versus a sign up, so passing the `changeusername: true` doesn't exactly make sense here.

wyattjoh (changes requested):

routes/api/auth/index.js
gabelula: yes! Good catch. This was left from before and I forgot to remove it.

wyattjoh (approved):

client/coral-sign-in/components/FakeComment.js
davidgljay: Fair enough!

davidgljay (approved):

client/coral-sign-in/components/FakeComment.js
```diff
@@ -0,0 +1,66 @@
...
+  constructor (props) {
+    super(props);
+  }
+
+  render () {
+    const {username, created_at, body} = this.props;
+
+    return (
+      <div
+        className={`comment ${styles.Comment}`}
```
gobengo, line 24: @gabelula I believe that this made a lot of the e2e tests fail. But folks might not have seen it. A lot of the '@element's in embedPageStream start their selectors with '.comment ...'. This matches the first one on the page, which this Dialog ends up being.

idk how the e2e tests do when they all run together, since they don't each clean up after themselves, but you can reproduce by running just this test on an empty db.

```
npm run pree2e && NODE_ENV=test ./node_modules/.bin/nightwatch --test test/e2e/tests/Commenter/LikeCommentTest.js
```

I'm working on a commit that will make the selectors specific to comments in the Embed Stream. I need it to make my e2e tests pass.

gobengo: https://github.com/coralproject/talk/pull/350/commits/f8479a86612cce1ff2d1a479871454b1e7abc4dc

PR #335, opened 2017-02-22
Fixed issue related to action create on suspect word match

wyattjoh:
## What does this PR do?

- Fixed structure of action creation for suspect word matched comments


riley (approved):
good catch 💯

PR #336, opened 2017-02-22
View New Comments

davidgljay:
## What does this PR do?

Polls for comments that have been posted since the user loaded the page. When these comments exist, the user is given a "View X new comments" button, which loads these comments.

## How do I test this PR?

1. Load the comment stream.
2. Load it again in a different tab.
3. Post a new comment in the second tab.
4. The first tab should have a "View 1 new comment" button within 20 seconds.
5. Click the button.
6. The new comment should load.


riley (approved):
very nice work! great implementation 💯

PR #338, opened 2017-02-22
show login page instead of communicorn :'(

riley:
## What does this PR do?

Redirect to the `/login` page instead of showing the Communicorn.

## How do I test this PR?

- logout
- try to view the admin
- login
- view dashboard
- logout
- login as non-admin user
- fail 😭


davidgljay:
lgtm. Note: there is no way for the admin to log in if they have authed through Facebook (though with our current admin login flow this shouldn't be possible.)

davidgljay (approved):

PR #345, opened 2017-02-24
Prevents "0" from appearing on empty comment streams

davidgljay:
## What does this PR do?

Fixes a bug wherin "0" appears on empty comment streams and removes an unused refetch function from the Comment component.

## How do I test this PR?

1. Start a new asset with 0 comments.
2. The number "0" should not appear underneath the comment box.


kgardnr (approved):
:100:

gabelula (approved):

PR #349, opened 2017-02-27
Routes Production mode

wyattjoh:
## What does this PR do?

- Moves development routes (/, /assets*) to only get mounted in development mode
- Adjusted embed script to be a part of the webpack build process

## How do I test this PR?

- Pull, test, and attempt to access test routes while `NODE_ENV` is `production` or `development`


kgardnr (approved):
LGTM :100:

PR #350, opened 2017-02-28
Best Comments

gobengo:
* Story: https://www.pivotaltracker.com/n/projects/1863625/stories/138617379
* pre-implementation notes that @jde gave me goahead on. https://gist.github.com/gobengo/12cab171d4706509d345c8f33a71d9f2

## What does this PR do?

* Adds a button in the stream to tag comments as "BEST"
* Best comments get a tag label for all to see
* Fixes e2e tests

best comment in action

code
* Add addCommentTag,removeCommentTag root mutations
* Add CommentsService#removeTag
* Test CommentsService#addTag and ensure same tag.name can't be added twice
* style changes to make sure that material-icons look good outside of the comment actions, multiple TagLabels are spaced reasonably, many comment actions can be in the footer on narrow screens without ugly wrapping
* Added class to allow selecting '.comment's only inside the Embed > Stream. And changed the embedStreamPage nightwatch element selectors to use it. This made e2e tests pass for me. See https://github.com/coralproject/talk/pull/334/files#r103414119

## How do I test this PR?

* Log in as a moderator and go to the embed
* Click the heart action
* the heart fills in, and a heart label is added at the top of the comment
* Refresh. Make sure everything persisted okay
* Click the filled in heart to toggle it off if you want (and refresh to verify), but then toggle back on
* Log out, the action to toggle best tag is gone because you're not a moderator. The <3 tag is still there for all to see though.

## Business Logic Changes

* Any MODERATOR or ADMIN can modify any allowed tag. This means, e.g. that a moderator can (by submitting graphql queries) remove the 'staff' tag from their comments. Is this okay? That seems both acceptable auth logic and, even if not, very low risk/harm that a moderator will ever do it.

## Questions For Product/Design

* [x] I added a class that would add `cursor: pointer` css to action buttons, but left it off of this PR for now. Can I style all the action buttons so the cursor changes to show it's clickable?
  * resolution: Sam said I could make all the action buttons cursor: pointer. So I did.
* [x] I used the heart icons instead of 'Tag as Best' (etc) copy because the latter was kind of long. What copy would you like? Or do you like the ❤️ ?
* [ ] Same for the indicator of best comment? Should it be a tag label like it is? Should it say "BEST" or something?
* [ ] Do you want any 'saving' indicator when a moderator clicks the BestButton? Right now I just disable the button while it's sending the request (to prevent double clicks). The slight change in style at least gives you a hint that something is happening.
* [x] What should happen if you click the BestButton and the request fails (e.g. your wifi goes out or the backend is down)? Right now there is no error message. It'll just behave like it didn't save.
  * resolution: Sam suggested I show the user a notification using the Notificatoin/SnackBar, so I wired that up.


.eslintrc.json
```diff
@@ -3,6 +3,7 @@
         "es6": true,
         "node": true
     },
+    "parser": "babel-eslint",
```
wyattjoh: It was a architecture decision not to use babel on the server side code, we should avoid linting with this parser.

graph/mutators/comment.js
```diff
@@ -187,11 +187,33 @@ const setCommentStatus = ({loaders: {Comments}}, {id, status}) => {
...
 };
 
+/**
+ * Adds a tag to a Comment
+ * @param {String} id          identifier of the comment  (uuid)
+ * @param {String} tag     name of the tag
+ */
+const addCommentTag = ({user, loaders: {Comments}}, {id, tag}) => {
+  return CommentsService.addTag(id, tag, user.id)
+    .then(() => CommentsService.findById( id ));
```
wyattjoh: Adding a comment tag should not return a comment, instead it should return nothing, or the tag created. This makes it more difficult to compose mutators later on when the focus is deemphasized on the definition of the mutation.

graph/mutators/comment.js
```diff
@@ -187,11 +187,33 @@ const setCommentStatus = ({loaders: {Comments}}, {id, status}) => {
...
+};
+
+/**
+ * Removes a tag from a Comment
+ * @param {String} id          identifier of the comment  (uuid)
+ * @param {String} tag     name of the tag
+ */
+const removeCommentTag = ({user, loaders: {Comments}}, {id, tag}) => {
+  return CommentsService.removeTag(id, tag)
+    .then(() => CommentsService.findById( id ));
```
wyattjoh: Adding a comment tag should not return a comment, instead it should return nothing, or the tag created. This makes it more difficult to compose mutators later on when the focus is deemphasized on the definition of the mutation.

graph/resolvers/root_mutation.js
```diff
@@ -48,7 +48,13 @@ const RootMutation = {
   },
   setCommentStatus(_, {id, status}, {mutators: {Comment}}) {
     return wrapResponse(null)(Comment.setCommentStatus({id, status}));
-  }
+  },
+  addCommentTag(_, {id, tag}, {mutators: {Comment}}) {
+    return wrapResponse('comment')(Comment.addCommentTag({id, tag}));
+  },
+  removeCommentTag(_, {id, tag}, {mutators: {Comment}}) {
+    return wrapResponse('comment')(Comment.removeCommentTag({id, tag}));
```
wyattjoh: As with above, we shouldn't have this return a "comment", rather null as above when performing a mutation on the comment itself.

graph/typeDefs.graphql
```diff
@@ -631,6 +631,20 @@ type SetCommentStatusResponse implements Response {
...
+type AddCommentTagResponse implements Response {
+  # An array of errors relating to the mutation that occured.
+  comment: Comment
+  errors: [UserError]  
+}
+
+# Response to removeCommentTag mutation
+type RemoveCommentTagResponse implements Response {
+  # An array of errors relating to the mutation that occured.
+  comment: Comment
```
wyattjoh, line 654: Same as above, should not return comment.

services/comments.js
```diff
@@ -44,29 +45,87 @@ module.exports = class CommentsService {
...
+                  throw new Error(`Can't add tag to comment. There is no comment with id ${id}`);
+                }
+                throw new Error(`Can't add tag ${name} to comment. Comment already has that tag.`);
+              });
+        case 1:
+
+            // tag added
+          return;
+        default:
+          console.warn('CommentsService#addTag modified multiple comments, but it should only have modified 1. '
```
wyattjoh: This is really an impossible situation due to the unique index on the id field, and the fact that that the update was not made with `{multi: true}`. I'm hesitant to add defensive code for this.

services/comments.js
```diff
@@ -44,29 +45,87 @@ module.exports = class CommentsService {
...
       return Promise.reject(new Error('tag not allowed'));
     }
 
-    return CommentModel.update({
+    const filter = {
       id,
-      tags: {
-        $ne: {
-          name
+      tags: {$not: {$elemMatch: {name}}}
```
wyattjoh: Should just use the previous query here, since `$elemMatch` provides only a single query, it is not necessary, see https://docs.mongodb.com/manual/reference/operator/query/elemMatch/#single-query-condition

services/comments.js
```diff
@@ -44,29 +45,87 @@ module.exports = class CommentsService {
...
+        created_at: new Date()
+      }}
+    };
+    return CommentModel.update(filter, update)
+      .then(({nModified}) => {
+        switch (nModified) {
+        case 0:
+
+            // either the tag was already there, or the comment doesn't exist with that id...
+          return this.findById(id)
```
wyattjoh: Can not use `this.findById` as all these functions do not provide context. This operation should not query the database at all, this is mixing of database queries which complicates the service interface. I do not expect that `addTag` would query for a comment, this would introduce a performance bottleneck that would be difficult to diagnose.

services/comments.js
```diff
@@ -44,29 +45,87 @@ module.exports = class CommentsService {
...
+  /**
+   * Removes a tag from a comment
+   * @throws if the tag is not on the comment
+   * @param {String} id the id of the comment to tag
+   * @param {String} name the name of the tag to add
+   */
+  static removeTag(id, name) {
+    const filter = {
+      id,
+      tags: {$elemMatch: {name}}
```
wyattjoh: Same as comment above, should not use `$elemMatch` with a single condition.

services/comments.js
```diff
@@ -44,29 +45,87 @@ module.exports = class CommentsService {
...
+    };
+    const update = {$pull: {tags: {name}}};
+    return CommentModel.update(filter, update)
+      .then(({nModified}) => {
+        switch (nModified) {
+        case 0:
+
+            // either the tag was already there, or the comment doesn't exist with that id...
+          return this.findById(id)
```
wyattjoh: Same as above, should not be querying for comments, it complicates the interface.

test/graph/mutations/addCommentTag.js
```diff
@@ -0,0 +1,62 @@
...
+const schema = require('../../../graph/schema');
+const Context = require('../../../graph/context');
+const UserModel = require('../../../models/user');
+const SettingsService = require('../../../services/settings');
+const CommentsService = require('../../../services/comments');
+
+describe('graph.mutations.addCommentTag', () => {
+  let comment;
+  beforeEach(async () => {
+    SettingsService.init();
```
wyattjoh: This returns a promise, should also be awaited (correct me if I'm wrong).

test/graph/mutations/removeCommentTag.js
```diff
@@ -0,0 +1,69 @@
...
+const schema = require('../../../graph/schema');
+const Context = require('../../../graph/context');
+const UserModel = require('../../../models/user');
+const SettingsService = require('../../../services/settings');
+const CommentsService = require('../../../services/comments');
+
+describe('graph.mutations.removeCommentTag', () => {
+  let comment;
+  beforeEach(async () => {
+    SettingsService.init();
```
wyattjoh: Same as above, should be await'ed?

test/mocha.opts
```diff
@@ -2,6 +2,7 @@ test/helpers/*.js
 test
 --compilers js:babel-core/register
 --require ignore-styles
+--require babel-polyfill
```
wyattjoh: Should not include babel-polyfill for tests as our server side code should never require any form of transpiling.

wyattjoh (changes requested):

graph/mutators/comment.js
gobengo: @wyattjoh Alright. I did that originally, then put this here as the quickest way of Apollo being able to update the Comments cache to reflect the changed tag.

How do you recommend I have the client update its representation of the Comment? Originally I looked to see if there was a single comment query I could make and use with refetchQueries, but I didn't see one. And it seems heavy to totally refetch the whole comment stream of the asset.

Should I use [updateQueries](http://dev.apollodata.com/react/cache-updates.html#updateQueries) and manually update streamQuery, loadMore, and myCommentHistory queries that fetch comments?

Is there a way to update the cached comment directly?

services/comments.js
```diff
@@ -4,7 +4,8 @@ const ActionModel = require('../models/action');
 const ActionsService = require('./actions');
 
 const ALLOWED_TAGS = [
-  {name: 'STAFF'}
+  'STAFF',
+  'BEST',
```
wyattjoh: This decision was made as we wanted it such that eventually we wanted to support tags being added dynamically, hence the structure of the array of objects instead of the array of strings.

gobengo: I wil revert.

services/comments.js
gobengo: The previous query allowed a tag with the same name to be added twice. But I will update the query to use the method described in your link.

graph/mutators/comment.js
wyattjoh: I'm unsure of the best approach to this for now, possibly we can adjust the way that the query for the comment is completed, but it should definitely not live in the mutators folder. Possibly moving them to the root_mutators.js file in the resolvers folder?

services/comments.js
wyattjoh: If the `$ne` resulted in a false positive match, then we should be able to adjust the single query, if not, then we can defer to the `$elemMatch`, thanks for the persistence.

test/graph/mutations/addCommentTag.js
gobengo: You're right. Great catch.

.eslintrc.json
gobengo: eslint didn't like async functions without this. But it looks like a better way of configuring that is the following, which I'll change to.

```
    "parserOptions": {
        "ecmaVersion": 2017
    },
```

graph/mutators/comment.js
gobengo: @wyattjoh What do you think of if I change CommentsQuery to allow specifying ids[]? The internal comments loader already supports providing ids, its just that the typedef for CommentsQuery doesn't have it?

Then, the frontend can addCommentTag, get no comment back, but then just re-fetch the comment by ID by CommentsQuery(ids=[id]).

Sound okay?

wyattjoh: We already have a `comment(id: ID!): Comment` edge being added in another branch, I'll update here when I verify that it has been pushed :)

gobengo: @wyattjoh Will you please link me to that branch, if it's pushed?

wyattjoh: Hey @gobengo, it should be on https://github.com/coralproject/talk/tree/fix-permalink within a few minutes.

wyattjoh: It's available now @gobengo

test/mocha.opts
gobengo: @wyattjoh Things are already being transpiled because of the `--compilers` flag above, right? The tests don't just test server side code, they test client side code that use e.g. `import` which no version of node currently does.

I had to add this because that compiler is transpiling async functions to:
```
/Users/ben/dev/talk/test/graph/mutations/addCommentTag.js:20
  beforeEach(_asyncToGenerator(regeneratorRuntime.mark(function _callee() {
                               ^

ReferenceError: regeneratorRuntime is not def
```

So I would like to leave this in as it's unrelated to server side could requiring any form of transpiling. *All* the test code is already being transpiled and this just makes that transpilation not break things that *would* work if they weren't transpiled.

If you see this and still want me to remove this, say so and I will remove all use of async functions that necessitated it. But also... async functions are (IMO) great for readability compared to .thennables and now part of standard node 7.6.

wyattjoh: I really, really want to use the `async`/`await` style as because (as you mentioned) it is supported in node 7.6. The inclusion of the `--compilers js:babel-core/register` flag doesn't quite perform transpilation beyond some stylistic rewriting as necessary to make the imports work for the client side code, but it doesn't modify the global parameters.

That flag was ported over simply because the tests required them. I would for the moment, unfortunately, recommend removing async/await until we can find a way to tell babel to only apply to the client mocha tests instead of all tests, hence the problem you're facing.

I didn't realize that this would be a deeper issue when I saw async/await and would have recommended removing until we can rectify.

Sorry for the confusion!

wyattjoh: See https://babeljs.io/docs/usage/babel-register/ describing how the register compiler works if you're interested in a read!

services/comments.js
gobengo: This worked:
```
    const filter = {
      id,
      'tags.name': { $ne: name },
    };
```

services/comments.js
gobengo: I'm going to remove the use of findById per your general advice of removing the extra query.

But just so I can understand: What do you mean by "all these functions do not provide context."? Do you mean `this`? If so, `this` is preserved because arrow functions. If you mean graphql context or something, pardon my mistake.

Like I said though, irrelevant to this review because I'll make the bigger change you're suggesting to move extra queries after update alltogether.

wyattjoh: As we were in a static method, what I meant is that it is out of context, referring to `this` implies an instance of the model rather than the static model method `CommentsService .findById`, hence I recommend if we were to use it, we should instead use `CommentsService.findById` as there's no way to mistake what we are referring to.

test/mocha.opts
gobengo: @wyattjoh 

> I would for the moment, unfortunately, recommend removing async/await until we can find a way to tell babel to only apply to the client mocha tests instead of all tests

Where there's a will, there's a way! This commit would only enable transform-async-to-generator when babel bundles code in the client/ dir. What do you think?

https://github.com/gobengo/talk/commit/405dac292c1da4a0b6e1cfce81eac501de62eef0

services/comments.js
gobengo: Oh I missed where this was before. Notably this only happens in the error condition, and its just to provide the most useful error message possible. Otherwise it's not clear exactly why the update failed. I will remove this query though and just throw an ambiguous error.

graph/typeDefs.graphql
gobengo: As we discussed on Slack, I've left this in this api definition, but lifted the findByIds to power it as high as possible.

Once the permalink PR lands with its [comment query](https://github.com/coralproject/talk/pull/352/files#diff-12b7fbc29e00b65c215fd45aea1ba81bR484), it'll be easier to just have the client re-fetch the comment via that query, and remove the utility of returning it here. But let's make that improvement to a later PR.

graph/resolvers/root_mutation.js
```diff
@@ -50,10 +51,10 @@ const RootMutation = {
     return wrapResponse(null)(Comment.setCommentStatus({id, status}));
   },
   addCommentTag(_, {id, tag}, {mutators: {Comment}}) {
-    return wrapResponse('comment')(Comment.addCommentTag({id, tag}));
+    return wrapResponse('comment')(Comment.addCommentTag({id, tag}).then(() => CommentsService.findById(id)));
```
wyattjoh: Even better, you can actually do this:

```js
addCommentTag(_, {id, tag}, {mutators: {Comment}, loaders: {Comments}) {
  return wrapResponse('comment')(Comment.addCommentTag({id, tag}).then(() => Comments.get.load(id)));
}
```

Not sure, but you may need to wait till that other PR lands with that loader in it.

wyattjoh (approved):

okbel:
@gobengo Could you quickly solve the merge conflicts so we merge this?

okbel (approved):

PR #351, opened 2017-02-28
Fixing order bug with mod queue sorting.

davidgljay:
## What does this PR do?

Stops "newest first" and "oldest first" from being reversed.

## How do I test this PR?

1. Go to the moderation queue.
2. "Oldest first" and "newest first" should order as expected.


kgardnr (approved):
LGTM 💯

PR #353, opened 2017-02-28
Refetching on login and logout.

davidgljay:
## What does this PR do?

Loads user state on login and removes it on logout.

## How do I test this PR?

1. Log in to a comment thread and like some stuff.
2. Log out.
3. Likes should disappear immediately. 
4. Log back in.
5. Likes should reappear immediately.

## Note
This is accomplished through refetching the stream, which means that we lose "load more" state. If you are 30 comments down, click "like" and log in the comment you just liked will disappear. Both retrieving the user's likes AND keeping the "load more" state of the stream will be significantly harder, we should decide on a tradeoff for now.


kgardnr:
You know what, let's wait until @jde is back next week to review this with him and talk about the best way forward. I'm down to approve this for now and circle back with him next week.

kgardnr (approved):

PR #360, opened 2017-03-01
Comments with suspect words should not show up as "Reported" by the user who posted them.

davidgljay:
## What does this PR do?

Prevents comments with suspect words from being marked as "Reported" by the user who posted them. Relating to https://www.pivotaltracker.com/n/projects/1863625/stories/140721235 

## How do I test this PR?

1. Leave a comment with a suspect word.
2. The "Reported" flag should still be black.


kgardnr (approved):

graph/mutators/action.js
```diff
@@ -16,7 +16,7 @@ const createAction = ({user = {}}, {item_id, item_type, action_type, group_id, m
   return ActionsService.insertUserAction({
     item_id,
     item_type,
-    user_id: user.id,
+    user_id: group_id === 'Matched suspect word filter' ? null : user.id,
```
wyattjoh: We can handle this before this point, I'll submit a commit asap.

wyattjoh (changes requested):

wyattjoh (approved):

PR #369, opened 2017-03-02
use something besides the reply key

riley:
## What does this PR do?

There were a bunch of warnings in the console when you had too many replies. Hopefully this should help.

## How do I test this PR?

- have several replies to one comment. There shouldn't be warnings in the console


davidgljay:
Wait, I'm confused. If two replies with the same ID are being displayed wouldn't we want to know that? It seems like a useful error. If we're getting a non-unique key error when displaying a lot of replies then it seems like we should investigate why rather than muting it.

riley:
Hm, @davidgljay you're probably right. I just created this when we were rolling out the beta and I wanted to suppress some warnings when people would be looking. This probably isn't helpful. I'll close this for now.

PR #374, opened 2017-03-02
Bump version number

kgardnr:


kgardnr:
Closing, we will update to 1.0.1 when we do our next release.

PR #383, opened 2017-03-03
Premod links

davidgljay:
## What does this PR do?

Re-enables pre-moderation of comments with links.

## How do I test this PR?

1. Enable link pre-moderation in the global or stream-level configuration (note that stream configuration always overrides global.)
2. Post a comment with a valid URL in it.
3. Comment should go to premod.

## Note
Comment highlighting has not yet been re-enabled. This is more complicated than first anticipated, here's why:

We need to highlight both suspect words and comment URLs. We have two different components for handling these two tasks, but they don't nest. When we put the suspect words `Highlighter` component inside the highlight-links `Linkify` component the links stop working. In order to get both working we'll need to build a custom component that does both, which involves some complex string manipulation and lower-level react stuff. It's not too complicated, but probably out of scope for this bug.


riley (approved):
looks good to me 💯

PR #385, opened 2017-03-06
Moved to correct route

wyattjoh:
## What does this PR do?

- Fixes regression related to api service route being mounted on the wrong router

## How do I test this PR?

- Start up the application in dev mode (not production)
- Hit `/` and see the development stream


davidgljay (approved):
lgtm

PR #387, opened 2017-03-06
move notifications to parent

riley:
## What does this PR do?

We were having issues with the notifications not always showing up or showing up in weird places. This is due to how `position: fixed` interacts with the parent page in different browsers. This PR skirts the problem by moving the notification view code up to the parent embed script, so the notification is just a DOM element that is drawn on the parent. pym is used to send messages.

Pertinent Chrome WontFix bug:
https://bugs.chromium.org/p/chromium/issues/detail?id=20574
http://stackoverflow.com/questions/8304148/positionfixed-inside-of-an-iframe#answer-22072777

## Caveats
The styles for the notification cannot live in `default.css` since those are the styles for the iframe. I ended up inlining the styles in js, **which is gross**. If anyone has better ideas, please let me know and I'll update the PR.

## How do I test this PR?

- post a comment when mod: `PRE`. View the notification
- change settings in the embed on the Configure Stream tab. notifications should appear


davidgljay (changes requested):
Notifications appear when mod='PRE', but not when mod='POST' (as per testing instructions).

riley:
Whoops, @davidgljay that was a typo. It's supposed to be `PRE`

client/coral-framework/reducers/notification.js
```diff
@@ -1,24 +0,0 @@
...
-import {Map} from 'immutable';
-
-const initialState = Map({
-  text: '',
-  type: '',
-  position: 400
-});
-
-export default (state = initialState, action) => {
-  switch (action.type) {
```
okbel, line 11: Why did we delete the reducer?

okbel (changes requested):

client/coral-framework/reducers/notification.js
riley: it's not being used anywhere. I deleted the `<Notification>` component, so nothing was reading its state.

okbel: Hm. It used to be... but not with this refactor.

davidgljay (approved):

okbel (approved):

PR #390, opened 2017-03-06
Fix password reset view render path

wyattjoh:
## What does this PR do?

- Fixes path for resetting password


riley (approved):
good catch! 💯

PR #392, opened 2017-03-07
Load More button in mod queue

davidgljay:
## What does this PR do?

Adds a load more button to the mod queue

## How do I test this PR?

1. Put more than 10 comments in the premod, flagged, or rejected queue in the admin panel.
2. Scroll to the bottom, there should be a load more button.
3. Click the button, more comments should load.
4. When all comments are loaded the button should no longer be visible.


client/coral-admin/src/containers/ModerationQueue/components/LoadMore.js
```diff
@@ -0,0 +1,22 @@
...
+            cursor: comments[comments.length - 1].created_at,
+            sort,
+            tab,
+            asset_id: assetId
+          })}>
+        Load More
+      </Button>
+    }
+  </div>;
+
```
riley, line 21: could we get some `propTypes` please?

riley (changes requested):
just add some comments and `propTypes`, but otherwise this looks good to go

riley (approved):

PR #393, opened 2017-03-07
rejected comment optimize bug

riley:
## What does this PR do?

There was some weirdness before such that when you posted a comment that ended up being rejected because of a banned word, it would linger in your (and only your) stream due to our optimistic posting logic. This would be confusing for a user.

I changed it so that the comment disappears out of your stream, but while its "pending" it's blurred out and you can't interact with it.

## How do I test this PR?

- post any comment with mod set to `POST`. the comment should flash blurred for a second and then resolve
- post a comment with a banned word in it. it should flash blurred and then disappear.


davidgljay (approved):
lgtm

PR #394, opened 2017-03-07
Preventing admin panel refresh on accept and reject

davidgljay:
## What does this PR do?

Stops the queue from refreshing when accepting or rejecting a comment in the moderation panel.

## How do I test this PR?

1. Accept or reject a comment.
2. The admin panel should not refresh.
3. The comment should disappear from the stream.
4. The appropriate count should update.
5. If rejecting a comment, the comment should appear in the rejected queue and the rejected count should increase.


riley (approved):
works for me. very nice 💯

PR #396, opened 2017-03-07
fix some firefox display things

riley:
## What does this PR do?

- Fixes an issue where the Stream open/closed dropdown would render funny in Firefox.
- Fixes an issue where the Dashboard table headers would expand like crazy in FF

Please merge this one before #398 

## How do I test this PR?

- admin Stream dropdowns should render correctly
- tables should render correctly when there are no likes or flags


davidgljay:
Arrow seems lower than it should be in Firefox. (Though this may be unrelated.)

screen shot 2017-03-08 at 3 24 07 pm

davidgljay (commented):
Looks good!

davidgljay (approved):
Looks good!

PR #398, opened 2017-03-07
Auto-update Dashboard

riley:
## What does this PR do?

updates some styles and adds auto-update to the Dashboard.

## How do I test this PR?

- view the dashboard. there should be a little timer at the top that counts down from 5 minutes
- after 5 minutes, the timer should reset and the dashboard reloads

### DIFF
--- a/client/coral-admin/src/containers/Dashboard/Dashboard.css
+++ b/client/coral-admin/src/containers/Dashboard/Dashboard.css
@@ -7,3 +7,28 @@
   font-size: 1.5rem;
   font-weight: bold;
 }
+
+.autoUpdate {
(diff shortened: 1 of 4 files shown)


davidgljay (approved):
Looks good, though the design feels too strong. It jumps out as the most important thing on the page (color, movement, etc), when it feels like it could be pretty low in the information hierarchy, but that's not a reason to hold up the PR.

riley:
Please wait to merge this one after #396

riley:
agreed @davidgljay I think it's going to be something that always grabs your attention, which might not be good when you're going to be looking at it all day every day.

PR #401, opened 2017-03-08
incorrectly using dispatch now that we do pym directly

riley:
## What does this PR do?

the new way we do notifications doesn't actually return an action, so `dispatch` is confused. This removes the warning.

## How do I test this PR?

- go to embed stream
- enter a banned word to trigger a notification
- there should not be errors in the console


davidgljay (approved):
Works for me.

PR #403, opened 2017-03-08
Hiding rejected note on ban dialog for rejected comments.

davidgljay:
## What does this PR do?

Removes the phrase "Note: Banning this user will also place this comment in the Rejected queue." from the ban user dialog when banning a user from an already rejected comment.

## How do I test this PR?

1. Ban a user from a flagged or premod comment. Note about comment being rejected should appear.
2. Ban user from a rejected comment. Note about comment being rejected should not appear.


riley (approved):
lgtm 👍

PR #404, opened 2017-03-09
Scrolling to keep selected comments in view.

davidgljay:
## What does this PR do?

When paging with keyboard shortcuts in the admin panel automatically scrolls the page to keep selected comments in view. 

## How do I test this PR?

1. Use 'j' and 'k' to page through the admin panel.
2. Page should scroll so that the selected comment is always visible.


client/coral-admin/src/containers/ModerationQueue/ModerationContainer.js
```diff
@@ -90,6 +91,15 @@ class ModerationContainer extends Component {
     key.unbind('t');
   }
 
+  componentDidUpdate(_, prevState) {
+
+    // If paging through using keybaord shortcuts, scroll the page to keep the selected
+    // comment in view.
+    if (prevState.selectedIndex !== this.state.selectedIndex) {
+      document.querySelector(`.${styles.selected}`).scrollIntoViewIfNeeded();
```
riley: While `scrollIntoViewIfNeeded()` is admittedly totally awesome and I didn't know about it, [the docs](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoViewIfNeeded) say it's not on a standards track and doesn't work most places. We might just have to write some code to check the position in the window.

riley (changes requested):
I don't think we can get away with this yet :)

riley (approved):

PR #416, opened 2017-03-15
correctly handle error when not logged in

riley:
## What does this PR do?

we were trying to access the user object when no one was logged in. This PR adds a handler for that scenario.

## How do I test this PR?

- go to the admin while logged out
- there should be no errors in the console


kgardnr (approved):
:100:!

PR #417, opened 2017-03-15
move CountdownTimer to its own component

riley:
## What does this PR do?

the countdown was just a timer running on the dashboard, which triggered a re-render every second. This PR creates a separate CountdownTimer component so we're not spamming the container.

I also added a fix for a random-ish APOLLO_QUERY_ERROR that may have been due to a rendering error.

## How do I test this PR?

- click a button
- view the cat
- see the cat meow


kgardnr (approved):
Can't reproduce the error locally anymore, so I'll merge to staging and confirm there

PR #418, opened 2017-03-16
Added new pageData action/reducer

wyattjoh:
## What does this PR do?

- Moves recaptcha client keys into new pageData store
- Integrates LayoutContainer with the new pageData store


riley:
I know this might be a bit low-brow, but what if we just put constants like this right on the `window` object? That way we don't have to pass it all around "the react way" which is a lot of moving parts.

I think you were going for the ability to add more fields to data in the future, but we could just create a single global `app` object or something that holds all of our constants, like so:

```javascript
<script>
window.app = {};
<% Object.keys(data).forEach(key => { %>
  window.app[key] = data[key];
<% }) %>
</script>
```
or even simpler;

```javascript
<script>
<% Object.keys(data).forEach(key => { %>
window[key] = data[key]; // it's probably fine to munge window, we control the entire admin environment.
<% }) %>
</script>
```

and we could just do
```javascript
<Recaptcha
  sitekey={window.TALK_RECAPTCHA_PUBLIC}
  render='explicit'
  theme='dark'
  onloadCallback={this.onRecaptchaLoad} />
```

which I don't think would be a huge sin. let me know what you think @wyattjoh @jde @okbel

wyattjoh:
The whole point I think @riley is that the `pageData` contained all of the globals, hence if we used that as the single point of info for the globals, we wouldn't have to change much at all (because it's all loaded for redux in the LayoutContainer. If everyone else feels we should just bind this to `window`, that's fair, easy fix :)

jde:
I agree that we should not pass individual values around but we shouldn't bind our app to window. It is easier to do this in the short run but creates bindings and divides our application state between the redux store and a random object outside of the scope of the application.

I like the single object that @wyattjoh set up to pass all configuration available. It keeps all data in the store so it can be inspected along with the other data and doesn't bind us to a browser context.

jde (changes requested):
What do you think about formalizing the term _config_ for this level of data. This will create consistency throughout with two levels of data that drive the app:

* settings - db stored values that are often editable through the gui, and
* config - values loaded in from the environment

wyattjoh:
I can make that change @jde. Good point.

jde (approved):

riley:
sounds good to me. there's only one change i would make to this pr, it's just joining some prop definitions together. I'll submit a change

PR #421, opened 2017-03-17
Server Graph Plugins

wyattjoh:
## What does this PR do?

- Adds initial support for the graph to have injectable plugins

## How do I test this PR?

- Review the documentation on plugins `PLUGINS.md`
- Attempt to create a plugin
- Check `/admin/docs`, and `/api/v1/graph/iql` to see if new plugins were added


jde:
This is a great model for injecting new functionality into the Graph. 

Would it make sense to add optional `startup` and `shutdown` methods that execute on server startup and shutdown? For example, if we want an atomic plugin that _only_ adds a Slack object to context, this is best done exactly once when the server starts. This could also be useful for a graph plugin that establishes a connection to a foreign service. `startup` would open the connection once, while `shutdown` could be used to close it.

wyattjoh:
The startup/shutdown hooks can be done outside the plugin infrastructure. The context plugins must be loaded per-request, never per-application. That introduces scoping issues.

jde:
Without startup/shutdown, how would we, for example, build a plugin that opens a connection to another datasource, writes some subset of requests to it, then closes the connection on server shutdown?

Would namespacing solve the scoping issues? Plugins writing to `context.plugins.[pluginname]` would prevent collisions. This could be by convention at first, as I understand it's not practical to prevent abuse here.

jde:
Also, this is probably out of scope for this pr, but Is it possible to apply pre/post hooks on existing resolvers this iteration?

jde:
I see that pre/post hooks and resolvers can be created in Plugins and applied to core (or any) types.

jde:
After digging deeper into the implementation and talking offline, I realize that each context function is initialized on a _per request_ basis. Thees context functions return apis, which then become available to _all requests._ No server start/shutdown necessary, as we want to be able to reason about each request separately when providing that context object's api for that particular request. Very clean and powerful.

jde (approved):

PR #427, opened 2017-03-20
Comment Stream Bugfix

riley:
## What does this PR do?

Let's say that you're viewing a stream with a single top level comment. Elsewhere, someone makes one or more replies to that comment. The View Replies button appears for you and you click it, excited to see the freshest conversation happening about [Chinese Crested dogs](http://www.vetstreet.com/dogs/chinese-crested), but nothing happens. But with this PR, you'll be able to enjoy comments about ugly dogs on the information superhighway!

## How do I test this PR?

- have two tabs open (maybe with different accounts) to the same article
- reply a couple times to a comment, or to a reply WITH NO PRE-EXISTING COMMENTS
- switch to the second tab and wait for the View Replies button to appear.
- click it. replies should appear.

### DIFF
--- a/client/coral-embed-stream/src/Comment.js
+++ b/client/coral-embed-stream/src/Comment.js
@@ -156,8 +156,6 @@ class Comment extends React.Component {
           : null }
         <PubDate created_at={comment.created_at} />
 
-        <p>parent_id {comment.parent ? comment.parent.id : 'no parent'}</p>
-        <p>id: {comment.id}</p>
         <Content body={comment.body} />
           <div className="commentActionsLeft comment__action-container">
             <ActionButton>
--- a/client/coral-embed-stream/src/Comment.js
+++ b/client/coral-embed-stream/src/Comment.js
@@ -155,6 +155,9 @@ class Comment extends React.Component {
           ? <TagLabel><BestIndicator /></TagLabel>
           : null }
         <PubDate created_at={comment.created_at} />
+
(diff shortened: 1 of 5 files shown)


client/coral-embed-stream/src/Comment.js
```diff
@@ -232,7 +233,7 @@ class Comment extends React.Component {
               removeCommentTag={removeCommentTag}
               showSignInDialog={showSignInDialog}
               reactKey={reply.id}
-              key={`${reply.id}:${depth}`}
+              key={reply.id}
```
okbel, line 236: Can a reply comment have the same id as a parent Comment?

riley: it can't. I thought so too at first, but the problem was that loadMore would return replies that might already be loaded. Further, even if a comment is a reply to a reply, the deepest nested reply's `parent.id` is the top-level comment.

okbel:
**1)** This button says 3 but when I click it only shows one, the new one. Is this expected?

screen shot 2017-03-21 at 3 08 06 pm


--------


**2)** This button (the one at the bottom) is not working.
screen shot 2017-03-21 at 3 06 45 pm

client/coral-embed-stream/src/Comment.js
okbel: 👍

okbel (dismissed):

jde (approved):

PR #431, opened 2017-03-22
remove express dep from mailer service

riley:
## What does this PR do?

Initially I set up the mailer service with a dependency injection using the express app to render the email templates. This meant that you needed a request object to send an email. No more!

I updated the mailer to just load the templates as strings, and use `_.template` which works the same way.

## How do I test this PR?

- sign up for a new user
- you should get an email confirmation that works.
- maybe try to reset your password and make sure that still works too


gabelula (approved):
It works great! Thanks!

services/mailer.js
```diff
@@ -13,6 +16,20 @@ if (smtpRequiredProps.some(prop => !process.env[prop])) {
...
+    throw err;
+  }
+
+  files.forEach(file => {
+    fs.readFile(path.join(__dirname, 'email', file), 'utf8', (err, data) => {
+      templateStrings[file] = _.template(data);
+    });
+  });
+});
+
```
wyattjoh, line 32: Should have provided live-reloading for development, and production loads them all in once. Might have also been better to abstract the access to the object to a function factory to do it for you:

```js
const template = {};
if (process.env.NODE_ENV === 'production') {

  // Returns the data loaded at application start.
  template.data = fs.readdirSync(path.join(__dirname, 'email')).reduce((data, filename) => {

    // Split the filename.
    let splitFileName = filename.split('.');

    // Remove the 'ejs' extension.
    splitFileName.splice(splitFileName.length - 1, 1);

    const [name, format] = splitFileName;

    if (!(name in data)) {
      data[name] = {};
    }

    data[name][format] = _.template(fs.readFileSync(filename, 'utf8'));
  }, {});

  // Renders a template based on the request name.
  template.render = (name, format = 'txt', context) => template.data[name][format](context);
} else {
  template.render = (name, format = 'txt', context) => new Promise((resolve, reject) => {
    fs.readFile(path.join(__dirname, 'email', [name, format, 'ejs'].join('.')), (err, file) => {
      if (err) {
        return reject(err);
      }

      return resolve(_.template(file)(context));
    });
  });
}
```

Which would make it consistent with the rest of the development behaviour around templates.

riley: 👍

PR #432, opened 2017-03-22
My Comments / Profile Tab UX

okbel:
https://www.pivotaltracker.com/story/show/140774191


okbel:
@gabelula Ready to be tested! Thank you!

jde (dismissed):
The "Story: [title]" link should cause the parent page to go to the article url. 

Currently it points directly to the embed stream page. This will work to just refresh the iframe. If the comment happens to be on the article that the user is on now, this is not a problem. When clicking on a different article's comment, however, this will result in showing the comment on the wrong article.

jde (approved):

PR #435, opened 2017-03-22
redirect to the parent url when resetting your password

riley:
## What does this PR do?

When you reset your password from within the stream, you're redirected to that page upon resetting your password.

## How do I test this PR?

- trigger a reset from the stream
- be redirected after following the email and resetting.


services/users.js
```diff
@@ -498,6 +498,7 @@ module.exports = class UsersService {
         const payload = {
           jti: uuid.v4(),
           email,
+          loc,
```
wyattjoh, line 517: We should do domain validation to ensure this location is whitelisted before we just encode it.

wyattjoh (changes requested):

services/users.js
riley: great catch!

wyattjoh (approved):

PR #438, opened 2017-03-23
always show moderation link to admins

riley:
## What does this PR do?

show moderation link even on closed streams.

## How do I test this PR?

- view an open stream
- Moderation link should appear
- close the stream, moderation link should still be there
- be logged in with a non-admin user. moderation link should not be there


kgardnr (approved):
:100:

PR #441, opened 2017-03-24
Auto close comment toggle

riley:
## What does this PR do?

Default to not auto-closing a comment stream after 2 weeks.

## How do I test this PR?

- create a new asset
- check the db to make sure there's no `closedAt` date set on the record.


services/assets.js
```diff
@@ -57,12 +57,24 @@ module.exports = class AssetsService {
...
+      if (settings.autoCloseStream) {
+        setOnInsert.closedAt = new Date(Date.now() + settings.closedTimeout * 1000);
+      }
+
       if (!whitelisted) {
         return Promise.reject(errors.ErrInvalidAssetURL);
       } else {
         return AssetModel.findOneAndUpdate({url}, {url}, {
 
+          $setOnInsert: setOnInsert,
```
wyattjoh: This isn't valid, you need to include this in the second parameter to `findOneAndUpdate`, not in the options (this third parameter)

wyattjoh (changes requested):

kgardnr:
This LGTM! Just want to check if @wyattjoh's good with it and we can demo and merge today.

wyattjoh (approved):

PR #445, opened 2017-03-27
hot-load email templates during development

riley:
## What does this PR do?

This PR makes it so that email templates are cached on production, but are loaded upon request during development to make things a bit easier.

## How do I test this PR?

- request a password reset link, the email should still have content.


services/mailer.js
```diff
@@ -17,18 +17,42 @@ if (smtpRequiredProps.some(prop => !process.env[prop])) {
...
 
-  files.forEach(file => {
-    fs.readFile(path.join(__dirname, 'email', file), 'utf8', (err, data) => {
-      templateStrings[file] = _.template(data);
+      return resolve(_.template(file)(context));
     });
   });
-});
+} else {
+  templates.data = fs.readdirSync(path.join(__dirname, 'email')).reduce((data, filename) => {
```
jde: Forking the code for development / production in this manner could introduce bugs / maintenance issues that are impossible to detect in staging. 

Please refactor the code that loads/parses the file into a single function that is run in all environment. This way there will be one single method for loading templates regardless of the environment, so bugs will appear in dev/staging. If you'd like to add a convenience method in dev for reloading, the dev only code should trigger that same template loading code to ensure it is getting the files the same way production will.

jde (dismissed):

services/mailer.js
wyattjoh: @jde I made this suggestion instead of the previous version as I felt that it matched the way that the existing express template loading worked.

wyattjoh (approved):

PR #446, opened 2017-03-28
Change permalink behavior

riley:
## What does this PR do?

Show only the Permalink and its parent comment (if applicable) instead of the entire comment stream. Also add a `View all comments` button that takes you back to the regular view before being able to make a comment. This PR sort of ended up being "implement Permalinks fully".

### Also fixed in this PR:
- `View n more replies` button now appears on a permalink page
- replies now appear when viewing a permalinked top-level comment
- the permalink popup now appears in the correct place for top-level comments, long comments
- `View n more replies` correctly loads more replies and displays them
- if a top-level comment is linked, only that comment will be highlighted, not all its replies as well
- polling now happens on the permalink page

### Not in this PR
- polling for more replies on a permalink page

## How do I test this PR?

- go to a comment stream and get a permalink
- on the permalink page, you should see the comment's parent comment and all its children
- if you click the View all comments button, it'll go back to the regular view.
- when you go back to the regular view, the url of the parent page and the iframe should change.

Also:

- create a permalink from a top-level comment, a reply and a long comment to make sure everything still works


kgardnr (approved):

client/coral-framework/constants/asset.js
```diff
@@ -9,3 +9,5 @@ export const UPDATE_ASSET_SETTINGS_FAILURE = 'UPDATE_ASSET_SETTINGS_FAILURE';
 export const OPEN_COMMENTS = 'OPEN_COMMENTS';
 export const CLOSE_COMMENTS = 'CLOSE_COMMENTS';
 export const UPDATE_COUNT_CACHE = 'UPDATE_COUNT_CACHE';
+
+export const VIEW_ALL_COMMENTS = 'VIEW_ALL_COMMENTS';
```
gabelula, line 13: What are you doing with VIEW_ALL_COMMENTS ? I do not see it being used.

riley: I just added the hook, since I could see us doing something with it eventually. Redux complains if you don't return an action from an action generator http://localhost:3000/assets/title/p#24420e5b-1c1f-4d6b-979b-67a25bddc5b0

client/coral-embed-stream/src/Embed.js
```diff
@@ -124,6 +124,16 @@ class Embed extends Component {
             <Tab>{lang.t('MY_COMMENTS')}</Tab>
             <Tab restricted={!isAdmin}>Configure Stream</Tab>
           </TabBar>
+          {
+            highlightedComment &&
+            <Button
```
okbel, line 152: This should be an isolated component. We can refator this when we refactor the Embed.

okbel (dismissed):

client/coral-embed-stream/src/Embed.js
riley: I agree. I'll refactor it

okbel:
While testing we found a bug - The tooltip goes to the top of the comment: 

screen shot 2017-03-29 at 4 35 21 pm

client/coral-embed-stream/src/Embed.js
riley: leaving myself a note here. we decided to leave this little refactor until the dust settles from the plugin additions.

PR #450, opened 2017-03-29
Remove duplicate material.css

alinhle:
## What does this PR do?

- Removes duplicate `material.css`. The css in the correct theme is already added here: https://github.com/coralproject/talk/blob/598fcc4cb0fa7f55185a037e4eff618d8f92c8fb/views/admin.ejs#L24

## How do I test this PR?

- Go to the admin area and everything should still look the same

### DIFF
--- a/client/coral-admin/src/index.js
+++ b/client/coral-admin/src/index.js
@@ -7,7 +7,6 @@ import store from './services/store';
 
 import App from './components/App';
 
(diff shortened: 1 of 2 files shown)


okbel:
@cvle  Could you also modify the one in the template with a normal one? That one has a theme on it.

okbel (approved):

PR #451, opened 2017-03-29
Fixes issues with wordlist and usernames

wyattjoh:
## What does this PR do?

- Ensure that the wordlist can't contain emoji's
- Ensure that username checks user the correct tokenizer

### DIFF
--- a/services/wordlist.js
+++ b/services/wordlist.js
@@ -2,8 +2,12 @@ const debug = require('debug')('talk:services:wordlist');
 const _ = require('lodash');
 const natural = require('natural');
 const tokenizer = new natural.WordTokenizer();
+const nameTokenizer = new natural.RegexpTokenizer({pattern: /\_/});
 const SettingsService = require('./settings');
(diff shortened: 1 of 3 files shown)


services/wordlist.js
```diff
@@ -2,9 +2,13 @@ const debug = require('debug')('talk:services:wordlist');
 const _ = require('lodash');
 const natural = require('natural');
 const tokenizer = new natural.WordTokenizer();
+const nameTokenizer = new natural.RegexpTokenizer({pattern: /\_/});
 const SettingsService = require('./settings');
 const Errors = require('../errors');
 
+// REGEX to prevent emoji's from entering the wordlist.
```
riley, line 9: Is this just a character range? I thought they kept adding emojis all the time? or is `\udfff` the upper limit?

wyattjoh: There are series of ranges there, copied from lodash's implementation really.

kgardnr (approved):

PR #453, opened 2017-03-30
Upgrade React

alinhle:
## What does this PR do?
- Upgrades React to newest 15.4.2
- Allows us to debug and profile performance using `Chrome Timeline`. See https://facebook.github.io/react/blog/2016/11/16/react-v15.4.0.html#profiling-components-with-chrome-timeline

## How do I test this PR?
- Everything should run as before
- To test the timeline feature, append `?react_perf` to the URL and record using the `Chrome Timeline` tool


riley (approved):
everything seems to work for me. tests pass

PR #455, opened 2017-03-30
Plugin Architecture and Respect Button Plugin

okbel:
# Plugin Architecture
This pull request adds:

- New Plugin Architecture
- New Plugins:
    - Coral Respect Button

## How to build plugins
Every plugin will live in the `/plugins` folder. 
Each plugin must have an `index.js` file and two folders `client` and `server`.

### The Client Folder
Inside the `client` folder we must add an `index.js` file that exports our Plugin.

The client folder will be plugged-in inside talk based on the configuration that we provided in our `index.js`

First we must specify our slots.

#### Slots
Slots are the components that we will plug-in into Talk’s Frontend

For example:

```js
import RespectButton from './containers/RespectButton';

export default {
  slots: {
    commentDetail: [RespectButton],
  }
};
```

Here I’m specifying that the RespectButton Component will take place within the `commentDetail` in Talk.

`commentDetail` it’s a specific place in the CommentStream. It means that it will be embedded inside de comment detail.

We can add as many component’s as we need and in multiple places!

#### Reducers and Actions : Redux

Talk is powered by Redux and our plugins can too! Our plugins can have their own reducers and actions.

```js
import RespectButton from './containers/RespectButton';
import actions from './actions';
import reducer from './reducer';

export default {
  slots: {
    commentDetail: [RespectButton],
  },
  reducer
};
```

#### Styles and CSS
Talk uses CSS Modules. This basically means that you can also add your CSS Module to your plugin without colliding with the rest of Talk!

##### My Component 
```js
import styles from './style.css';

class MyCoralPlugin extends Component {
  render() {
    return (
      <div className={styles.plugin}>
	  <button>My Button</button>
      </div>
    );
  }
}
````

#### My styles `style.css`

```css
.plugin button{
  background: coral;
  border-radius: 3px;
}
````

### The server folder and the index file
Read more about the `/server` and how to extend Talk here.
[talk/PLUGINS.md at master · coralproject/talk · GitHub](https://github.com/coralproject/talk/blob/master/PLUGINS.md)

### CODE REVIEW

File: .eslintrc.json [hunk 1]
```diff
@@ -4,6 +4,7 @@
         "node": true
     },
     "extends": "eslint:recommended",
+    "parser": "babel-eslint",
```
wyattjoh: Modifications to the project wide `.eslintrc.json` file will affect server side code too. We should not be applying linting rules to the server that are reflective of client side concerns, such as `parserOptions.sourceType = module`, which the server does not support.

client/coral-embed-stream/src/Comment.js
```diff
@@ -141,22 +142,32 @@ class Comment extends React.Component {
...
-          author={comment.user}/>
-        { isStaff(comment.tags)
-          ? <TagLabel>Staff</TagLabel>
-          : null }
 
-        { commentIsBest(comment)
-          ? <TagLabel><BestIndicator /></TagLabel>
-          : null }
-        <PubDate created_at={comment.created_at} />
+        <div id="Comment.InfoBar">
```
alinhle: Why is the `id` needed?

plugins/coral-plugin-love/client/config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "name": "Coral Plugin Love",
+  "slot": "Comment.Detail"
```
alinhle: I'm not sure about the naming convention. I'm not strongly against it, I think I personally prefer `camelCase` like `comment.detail` and `comment.infoBar`. Let's see what everyone thinks.

plugins/coral-plugin-respect/client/style.css
```diff
@@ -0,0 +1,27 @@
+.Respect {
```
alinhle: Why naming the class with a initial capital letter?

plugins/coral-plugin-love/client/config.json
okbel: Comment is a Class and a Component, that's the reason behind it. Not strongly against it either. We can change it.

plugins/coral-plugin-respect/client/style.css
```diff
@@ -0,0 +1,27 @@
...
+    display: inline-block;
+    button {
```
alinhle: This should be a top-level class.

plugins/coral-plugin-respect/client/style.css
```diff
@@ -0,0 +1,27 @@
...
+        border: none;
+        font-size: inherit;
+
+        &:hover {
+          color: #d6ab00;
+          cursor: pointer;
+         }
+    }
+
+    .respected {
```
alinhle: This should be a top-level class.

plugins/coral-plugin-respect/client/style.css
```diff
@@ -0,0 +1,27 @@
...
+        color: #ffcc00;
+        i {
```
alinhle: This should be a top-level class.

plugins/coral-plugin-respect/client/style.css
```diff
@@ -0,0 +1,27 @@
...
+          color: #161616;
+        }
+    }
+
+    i {
```
alinhle: This should be a top-level class.

plugins/coral-plugin-respect/client/style.css
alinhle: The reasons for this is, that we want to isolate the styles for this Component and don't leak out styles that might affect other components.

alinhle:
Note: Sometimes we use the term `mutators` and sometimes `mutations` for the same thing. We should pick one and use it consequently.

plugins/coral-plugin-love/client/config.json
jde: I like the idea of sticking to a simple convention like camelCase unless there's an already established convention in the community. I know that the variables themselves are traditionally capitalized, but these strings are grey area. I would probably default to camel case here to keep the concept simple.

.gitignore
```diff
@@ -16,5 +16,5 @@ coverage/
 .tags1
 
 # remove plugin folders
-plugins
-plugins.json
+# plugins
+# plugins.json
```
wyattjoh: We should not exclude these from git, instead you should write as such:

```
plugins
!plugins/core-coral-plugin-1
!plugins/core-coral-plugin-2
```

client/coral-framework/helpers/plugins.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import merge from 'lodash/merge';
+import flatten from 'lodash/flatten';
+import plugins from 'pluginsConfig';
```
riley, line 4: Why does this need an alias? Can't we just import the `plugins.json` file?

client/coral-framework/index.js
```diff
@@ -1,15 +1,14 @@
...
+export default {
   pym,
+  Slot,
   I18n,
   store,
-  authActions,
-  assetActions,
-  notificationActions
+  actions,
+  ...actions
```
riley, line 15: This is going to confuse people. Could you add a comment here to explain what's going on?

client/coral-framework/components/Slot.js
```diff
@@ -0,0 +1,15 @@
...
+  render() {
+    const {fill, ...rest} = this.props;
+    return (
+      <div>
+        {getSlotElements(fill, rest)}
+      </div>
+    );
+  }
+}
+
```
riley, line 14: could we get some `PropTypes` here? Also, this component would be a great place to add a ton of comments to help guide contributors trying to create their own plugin.

client/coral-framework/loaders/plugins-loader.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Executes `source` to retrieve plugins configuration
```
riley, line 2: what is `source`?

client/coral-framework/loaders/plugins-loader.js
```diff
@@ -0,0 +1,31 @@
...
+  return config.client.map(x => typeof x === 'string' ? x : Object.keys(x)[0]);
+}
+
+module.exports = function(source) {
+  this.cacheable();
+  const config = this.exec(source, this.resourcePath);
+  const plugins = getPluginList(config);
+
+  const list = [];
+  plugins.forEach(plugin => {
```
riley: this could just be a `map` on the plugins list.

client/coral-framework/utils/index.js
```diff
@@ -0,0 +1,2 @@
+export const getActionSummary = (type, comment) => comment.action_summaries
```
riley: this is a long expression without an explanation. probably need a comment here.

plugins/coral-plugin-respect/client/.eslintrc.json
```diff
@@ -0,0 +1,23 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true,
+    "mocha": true
+  },
+  "parserOptions": {
```
riley, line 7: I think this needs `ecmaVersion: 8` to work properly with async/await keywords.

plugins/coral-plugin-respect/client/components/RespectButton.js
```diff
@@ -0,0 +1,61 @@
...
+
+export default class RespectButton extends React.Component {
+
+  handleClick = () => {
+    const {postRespect, showSignInDialog, deleteAction, commentId} = this.props;
+    const {me, comment} = this.props.data;
+
+    const respect = comment.action_summaries[0];
+    const respected = (respect && respect.current_user);
+
```
riley, line 19: need some `PropTypes` all up in here.

plugins/coral-plugin-respect/client/components/RespectButton.js
```diff
@@ -0,0 +1,61 @@
...
+    // If the current user does not exist, trigger sign in dialog.
+    if (!me) {
+      const offset = document.getElementById(`c_${commentId}`).getBoundingClientRect().top - 75;
```
riley, line 22: this probably needs to be wrapped in a `try / catch` since we're calling a method after querying the DOM.

test/server/graph/context.js
```diff
@@ -1,8 +1,8 @@
 const expect = require('chai').expect;
 
-const User = require('../../models/user');
-const Context = require('../../graph/context');
-const errors = require('../../errors');
+const User = require('../../../models/user');
```
riley, line 3: this might be dumb, but is there a way to not have all these horrible relative paths in require statements in node?

riley (changes requested):

plugins/coral-plugin-respect/client/.eslintrc.json
okbel: That's specific for each plugin.

riley: oh right. carry on.

client/coral-framework/index.js
okbel: This is a hack that we are going to address in another PR. Right now this is needed. (It's about our legacy code)

client/coral-framework/loaders/plugins-loader.js
okbel: The plugins path

client/coral-framework/utils/index.js
okbel: 👍

client/coral-framework/loaders/plugins-loader.js
alinhle: The output of the `loader` before this one, which currently is the `json-loader`.

client/coral-framework/utils/index.js
alinhle: Not used.

client/coral-framework/loaders/plugins-loader.js
alinhle: Not sure what you mean. The result is javascript code btw. for anyone who never wrote a webpack loader.

riley: sorry, I didn't mean a `Map`, I meant `Array.prototype.map()`

client/coral-framework/index.js
riley: maybe just a `//TODO` then for now

client/coral-framework/helpers/plugins.js
okbel: We are using an alias because otherwise we would end up with `../../../../`etc

wyattjoh (dismissed):
Putting this here to stop merge before some validation fixes get in

riley (approved):

marcosnils:
this is awesome, congrats! :tada: . No official docs for the client side plugins besides this PR comments?

okbel:
@marcosnils We are working on the FE docs, examples with simple to complex plugins (with uptimistic updates) and also tutorials! 🎉

client/coral-framework/index.js
alinhle: I don't think this is used anywhere bel, correct me if I'm wrong.

client/coral-framework/utils/index.js
```diff
@@ -0,0 +1,7 @@
+/**
+* getActionSummary
+* retrieves the action summary based on the type and the comment
+*/
+
+export const getActionSummary = (type, comment) =>
```
alinhle, line 6: This is also not used anywhere, as we rely on graphQL to get the correct type.

test/server/graph/context.js
alinhle: It's possible to use https://www.npmjs.com/package/babel-plugin-module-alias or https://www.npmjs.com/package/babel-plugin-webpack-alias which would require `babel-node`.

PR #459, opened 2017-04-03
Added support for a plugin which disables lazy assets

wyattjoh:
## What does this PR do?

- Ensures that embed.js supports disabling lazy asset loading

Here's the plugin code that will disable lazy asset creation:

```js
module.exports = {
  resolvers: {
    RootQuery: {
      asset(_, query, {loaders: {Assets}}) {
        return Assets.getByID.load(query.id);
      }
    }
  }
};
```

This simply overrides the asset resolver for the `RootQuery`. These types of operations are made possible by our extremely simple resolver system! The old asset resolver looked like this:

```js
  asset(_, query, {loaders: {Assets}}) {
    if (query.id) {
      return Assets.getByID.load(query.id);
    }

    return Assets.getByURL(query.url);
  }
```

So by simply overriding it, we replace core behaviour!


jde (approved):
Works as expected. When implementing the functionality, however, we should consider the implications of of overwriting a core resolver. Updates to the core resolver will need to be reflected in this plugin.

PR #464, opened 2017-04-05
do not load google fonts in embed

riley:
## What does this PR do?

We were loading Google fonts just for the Lato font. I removed it so the default stream is more "default" looking.

### Also
We're still actually hitting `fonts.googleapis.com` for the mdl icons. Do we want to figure out a way to build these into the app somehow on a case by case basis?

## How do I test this PR?

- view the stream
- it should be in Helvetica if you're on a mac, and Verdana probably if not.


kgardnr:
Oh good point about MDL. @okbel / @cvle / @jde what are your thoughts? Aren't we mostly using Coral-UI at this point anyways?

jde (approved):

PR #466, opened 2017-04-05
only refetch the comments if you're an admin

riley:
## What does this PR do?

if you're a regular user, it would be annoying to have the entire comment stream refresh if you just changed your settings on another tab. Before, when you returned to the stream, all the replies were collapsed. This PR makes it so that this is only true for admins.

## How do I test this PR?

- be logged in as a non-admin user
- view a stream with lots of comments and replies
- expand some old comments and replies
- click to settings
- click back. you should still see the expanded comments as before.

### DIFF
--- a/client/coral-embed-stream/src/Embed.js
+++ b/client/coral-embed-stream/src/Embed.js
@@ -41,9 +41,10 @@ class Embed extends Component {
   state = {activeTab: 0, showSignInDialog: false, activeReplyBox: ''};
 
   changeTab = (tab) => {
+    const {isAdmin} = this.props.auth;
 
     // Everytime the comes from another tab, the Stream needs to be updated.
-    if (tab === 0) {
+    if (tab === 0 && isAdmin) {
       this.props.data.refetch();
     }
 


jde (changes requested):
Let's play with this sharing screens to work through options tomorrow.

Admins seeing this behavior will think the application is buggy, as they will loose their place in the page. Could we easily attach the reload to the Save Settings action? Ultimately, with proper state management, the refresh shouldn't be necessary.

alinhle:
We need to think about the refetching thing on a broader architectural angle. Also the refetching mechanism itself is not the cause for the collapsing comments. It's how the loading indicator is currently implemented. It seems so that anytime anything in the streams query tree is updated, the stream gets destroyed and recreated at the moment due to the loading indicator. This is absolutely unnecessary. This issue currently overlaps with plugin development to the point that I cannot proceed unless I fix the loading indicator issue. So I might work on this first today because it's blocking me. Also there is an issue with changing settings where we also need to look at it from an architectural perspective as it seems to me it currently bypasses apollo, though I didn't take a deep look yet.

alinhle:
Here is the PR that solves the collapsing issue: https://github.com/coralproject/talk/pull/467

jde:
@cvle Good catch on the loading indicator forcing the whole stream to redraw. This PR addresses a different issue.

On first load we only show 10 comments and some of their children. A user may "load more' several times. If they then switch tabs and switch back, they "loose their place" and start over from the initial 10 loads. 

This is a good iterative fix.

jde (approved):

alinhle:
@jde The thing you are pointing out, doesn't seem to be related for me.

PR #467, opened 2017-04-06
Fix collapsing comment tree

alinhle:
## What does this PR do?
- Do not destroy DOM tree when loading data.

## How do I test this PR?
- Any action that updates the query should not show the loading indicator anymore (which destroys and recreates the DOM tree): e.g. switching tabs.


jde (approved):
Excellent catch. This will prevent re-rendering on each tab action.

PR #468, opened 2017-04-06
Fix graph not importing plugin resolvers

alinhle:
## What does this PR do?
- Includes plugin resolvers into the graph


alinhle:
Also the server didn't output any warning that the resolver was not available, which took some time to debug. This is still an issue.

wyattjoh:
Dupe of https://github.com/coralproject/talk/pull/465

PR #477, opened 2017-04-10
Ignore User

gobengo:
## What does this PR do?

* Implements story [IGNORE ANOTHER USER](https://www.pivotaltracker.com/n/projects/1863625/stories/141803099)
  * There is now a little menu in the top right of Comments authored by users other than the logged in user. Click it to start the Ignore User flow.
  * Asset->comments, Comment->replies, and several count edges in the graph can now be personalized to take into account ignored users. @jde and I discussed a bit how to do this before I went forth, but we both acknowledged its a bit tricky. So if any reviewer has good ideas for altering my first draft here, I welcome the feedback.

* Implements story [MANAGE IGNORED USERS](https://www.pivotaltracker.com/n/projects/1863625/stories/141803209)
  * Rename "Settings" tab to "My profile"
  * Adds "Ignored Users" section if you are ignoring users, which allows you to stop ignoring users

gif of this in action

## How do I test this PR?

Ignoring another user

1. Load the embed
2. Click on the little chevron to open the new menu in the top right of Comments *that aren't authored by you*
3. Follow the mini wizard to ignore that user
4. Once you ignore the user, all their comments are replace with a tombstone until the data for the stream is next fetched (like refresh, or switching to another tab and then back)
5. After the stream is next fetched (e.g. refresh), you no longer see the ignored users comments, nor do you see any tombstones

Managing ignored users

1. Click the 'My profile' tab to manage ignored users
2. If you aren't ignoring any users, there is no (empty) "Ignored Users" section here, don't worry about it
3. If you *are* ignoring users, there is an "Ignored Users" section listing all users you've ignored.
4. Click "Stop Ignoring" on one
5. Switch back to the comment stream tab
6. You can see that users comments again


graph/typeDefs.graphql
```diff
@@ -140,6 +140,9 @@ input CommentsQuery {
 
   # Sort the results by created_at.
   sort: SORT_ORDER = REVERSE_CHRONOLOGICAL
+
+  # Exclude comments ignored by this user ID
+  notIgnoredBy: String
```
wyattjoh: This should only ever be the current user, why not inherit this from the context instead? It always contains the current logged in user (unless the user is not logged in)

gobengo: I wanted some sort of option in the query so that it could be optional (i.e. a client could still request all comments including the ignored ones if some feature requires it).

But I could still allow for that with this being a boolean property instead of actually having to pass in the User ID. I think I did it this way just to generally make the query explicit or to allow for future changes to allow the query to be an array of user IDs (e.g. if the requesting user was an admin). But it's not obvious we'll ever need that.

Want me to change it? e.g. to an `excludeIgnored` boolean?

wyattjoh: I think for our purposes, it would keep the interface clean.

gobengo: done!

client/coral-embed-stream/src/TopRightMenu.js
```diff
@@ -0,0 +1,156 @@
...
+            cancel={reset}
+            ignoreUser={ignoreUserAndCloseMenuAndNotifyOnError}
+            />
+        </div>
+      </Toggleable>
+    );        
+  }
+}
+
+class IgnoreUserWizard extends React.Component {
```
okbel: Can we break this class into another file? We have one class per file.

gobengo: done!

client/coral-embed-stream/src/IgnoredCommentTombstone.js
```diff
@@ -0,0 +1,18 @@
...
+const IgnoredCommentTombstone = () => (
+  <div>
+    <hr aria-hidden={true} />
+    <p style={{
+      backgroundColor: '#F0F0F0',
+      textAlign: 'center',
+      padding: '1em',
+      color: '#3E4F71',
+    }}>
+      This comment is hidden because you ignored this user.
```
gabelula: It needs translation.

gabelula (changes requested):

client/coral-embed-stream/src/IgnoredCommentTombstone.js
kgardnr: @gabelula could you add this translation? Thanks!

gabelula: The PR needs to be finish to allow the translation to be in.

gobengo: @gabelula @kgardnr I made it use translations. Do either of you know spanish very well? If so, feel free to suggest a better translation, as I just used google translate.

gabelula: This comment is hidden because you ignored this user.

translates to

Este comentario está escondido porque has ignorado al usuario.

gabelula (approved):

wyattjoh:
@gobengo, just taking another look at the ignore user pr, would it not be cleaner to have the `excludeIgnored` option enabled as a user preference stored on the user object? That way the underlying queries would not need to pass `excludeIgnored` into each and every graph query and instead it’s just resolved at the loader layer? Let me know if that makes sense.

It just feels like we have lots of new code introduced into the resolvers, and I want to keep those files lite, to reduce branching

client/coral-embed-stream/src/IgnoredCommentTombstone.js
gobengo: @gabelula Thanks, I changed it to use that spanish translation.

gobengo:
@wyattjoh That may result in less code in those files, but it's different logic than I was going for, because the pivotal stories didn't say to make it a user preference. When @jde and I chatted before I implemented, he okayed me adding these options to the query itself (as opposed to user preference), so that's what I did. Ofc I can change it, but would kinda want to hear from @jde first since its different than what we originally discussed.

Personally, I tend to prefer explicit queries instead of implicit behavior based on context, even if it results in more verbose code.

wyattjoh:
That makes sense to me. I was trying to invision the situation where a user would not want to have ignored user's, ignored. Thinking on how Twitter implements it, it is done without explicitly asking for it. This just means that in the future, queries involving any user generated content will have to have a `excludeIgnored` query parameter, which I thought dirtied the interface and made it more difficult to implement.

I understand your + @jde's reasoning however.

PR #479, opened 2017-04-11
Hooks and Context for Comment Plugins And Off-Topic Plugin

okbel:
# Hooks and Context for Comment Plugins And Off-Topic Plugin

Summary:
* Context for Comment Plugin
* Hooks for Comment Plugin
	* registerHook (plugin context)
	* unregisterHook  (plugin context)
* CommentBox 
	* refactor
	* new Actions, and  reducer (plugin like)
	* postButton gets disabled when there’s no text.
* Inline Slots
* Off Topic Plugin
* createComentInput
	* mutation `createComent`  can now receive tags. (And More)

##  Context for Comment Plugin
All the Plugins injected within `commentInfoBar` and `commentDetail` (The slots within the Comment Component)
will have access to the `comment`.  The context can be accessed through `this.props`

#### Context:
* registerHook 
* unregisterHook
* updateComment
* comment 

## Hooks for Comment Plugin
The plugins injected in the **`CommentBox`** such as `commentBoxDetail` (The slot next to the **Post** button) will inherit through props  `registerHook` and `unregisterHook`

Available hook types:
* `postSubmit`  :  To perform actions after submitting the comment
* `preSubmit` :  To perform actions before submitting the comment

### Register Hooks
`registerHook` is a function that takes, `hookType`,  and the `hook` function
and returns the hookData

Usage:
```js
      this.addCommentTagHook = this.props.registerHook('postSubmit', (data) => {
        const {comment} = data.createComment;
        this.props.addCommentTag({
          id: comment.id,
          tag: 'OFF_TOPIC'
        });
      })
```

### Unregister Hooks
`unregisterHook` is a function that takes the hookData that comes back from registering a hook.
```js
  this.props.unregisterHook(this.addCommentTagHook);
```

## Inline Slots
We now can pass `inline`  to our slots. 
For example our actions within the `commentDetail` need to be inline in order to be next to the other components.

## Off Topic Plugin
Components: 
* Off Topic Checkbox
* Off Topic Tag

### Off Topic Checkbox 
* React Class Component  
* Dispatches actions

### Off Topic Tag 
* Stateless Component  

## To developers:
Keep the structure of `coral-plugin-offtopic` in mind since we are going to be using this as default for the rest of our plugins.
Notice how we _can_ enhance components using the mutations that the `coral-framework` provide. Avoid pollution inside components as much as possible components.
Keep your components dry, one class per file and pull the logic away if possible. 
We have support for `reducers`and `actions`, so have fun!


plugins/coral-plugin-offtopic/client/components/styles.css
```diff
@@ -0,0 +1,18 @@
+.offTopic {
+  height: 100%;
+}
+
+.offTopic label {
```
alinhle: Can you make this a `.offTopicLabel` class?

plugins/coral-plugin-offtopic/client/components/OffTopicCheckbox.js
```diff
@@ -0,0 +1,42 @@
...
+          addTag: () => {}
+        }
+      })
+    }
+  }
+
+  render() {
+    return (
+      <div className={styles.offTopic}>
+        <label>
```
alinhle: Use class directly on label to style it.

okbel: Sure!

okbel: Done.

okbel: Althought... I'd like to see a good reason for this. I think this adds to much pollution to the render function.

client/coral-plugin-commentbox/CommentBox.js
```diff
@@ -68,6 +101,15 @@ class CommentBox extends Component {
     this.setState({body: ''});
   }
 
+  addCommentHooks = (hooks = {}) => {
+    if (typeof hooks === 'object') {
```
alinhle: Using objects that gets merged into a single one will result in naming conflicts. Maybe a list would be better?

okbel: The reason behind using names is because I'd want to overide them or modify them or all them from other hooks.

alinhle: You want to influence the behavior of other plugins?

okbel: See the example. If multiple plugins adds hooks. And you want to toggle yours off, how would you approach that (in a friendly way) using lists?

alinhle: One example is to return a callback to unregister :
```js
const unregisterHook = registerCommentHook(data => {...});
// To unregister you would do
unregisterHook();
```

okbel: I think it's smart but it's also a bit confusing. But I agree that we could provide `register` and `unregister` methods.

plugins/coral-plugin-offtopic/client/components/OffTopicCheckbox.js
alinhle: I wouldn't call it pollution, more like information about your intent.
It makes it easier for someone else to look at it and knows what it does.
Also a developer expects that Components are isolated, so if someone
would modify this class and adds another Component into the render, he will 
be surprised why all of his labels have a different styling.

client/coral-plugin-commentbox/CommentBox.js
alinhle: Yeah maybe it can be a bit confusing. `register` and `unregister` is a good idea.

plugins/coral-plugin-offtopic/client/components/OffTopicCheckbox.js
okbel: Agreed. But I wouldn't enforce using multiple classes when you only have one tag inside that Component.

client/coral-plugin-commentbox/CommentBox.js
alinhle: Hm the latest change, didn't solve the name conflict problem.

alinhle: I imagine sth like this:
```js
this.hookId = registerHook(type, callback);
unregisterHook(this.hookId);
```
Also I suggest an unregister at `componentWillUnmount`

okbel: +1 to add the hook to`componentWillUnmount`. 

But I don't see the naming as a confict or issue. Also, I don't see the point of making a list, maybe around performance but it will be tiny. 🤔 

Having a restricted API could lead to issues in the future, that we are not going to experiment, but plugin developers might.

I'll move forward with the list approach, but also keeping in mind that we might need to revisit this again.

okbel: That implementation triggers a lot of issues when a user tries to add the same hook again. I won't move forward with it.

I will replace the API for registerHook to be similar to unregister.

okbel: Issues: 
- What happen if I want to invalid a hook from another plugin.

graph/typeDefs.graphql
```diff
@@ -593,6 +593,28 @@ input CreateLikeInput {
   item_type: ACTION_ITEM_TYPE!
 }
 
+
+
+enum TAG_TYPE {
+  OFF_TOPIC
+}
```
wyattjoh: Should this be in the plugin code??

wyattjoh (changes requested):

graph/typeDefs.graphql
okbel: That's correct! Thanks!

okbel: Updated :)

wyattjoh (approved):

alinhle:
I didn't run it locally, but code looks good!

PR #485, opened 2017-04-11
Clicking '# comments' tab refresh all the comments back

gabelula:
## What does this PR do?

It resolves story https://www.pivotaltracker.com/story/show/143520139

It adds an action to the tab "# Comments". Now it is clickable and bring back all the comments.

## How do I test this PR?

- Go to a permalink
- Click on the tab "# comments"
- Check that you get all the comments back


client/coral-embed-stream/src/Embed.js
```diff
@@ -144,7 +144,13 @@ class Embed extends Component {
       <div style={expandForLogin}>
         <div className="commentStream">
           <TabBar onChange={this.changeTab} activeTab={activeTab}>
-            <Tab><Count count={asset.totalCommentCount}/></Tab>
+            <Tab><Count
+              count={asset.totalCommentCount}
+              onClick={()=> {
+                this.props.viewAllComments();
+                this.props.data.refetch();
```
wyattjoh: Can you move this to a class method instead?

wyattjoh (changes requested):

client/coral-embed-stream/src/Embed.js
okbel: Agreed.

client/coral-plugin-comment-count/CommentCount.js
```diff
@@ -3,8 +3,8 @@ import {I18n} from '../coral-framework';
 import translations from './translations.json';
 const name = 'coral-plugin-comment-count';
 
-const CommentCount = ({count}) => {
-  return <div className={`${name}-text`}>
+const CommentCount = ({count, onClick}) => {
+  return <div className={`${name}-text`} onClick={() => onClick()}>
```
okbel: change the name to `handleClick` and pass it without wrapping it into another function if possible

client/coral-embed-stream/src/Embed.js
gabelula: Agree. Done.

client/coral-plugin-comment-count/CommentCount.js
gabelula: Done.

client/coral-embed-stream/src/Embed.js
```diff
@@ -36,9 +36,18 @@ import HighlightedComment from './Comment';
...
-class Embed extends Component {
+class Embed extends React.Component {
 
-  state = {activeTab: 0, showSignInDialog: false, activeReplyBox: ''};
+  constructor(props) {
+    super(props);
+    this.state = {
+      activeTab: 0,
+      showSignInDialog:
+      false, activeReplyBox: ''
```
alinhle: `false` seems to be on the wrong line

client/coral-embed-stream/src/Embed.js
```diff
@@ -36,9 +36,18 @@ import HighlightedComment from './Comment';
...
+    };
+
+    this.handleClick = this.handleClick.bind(this);
```
alinhle: See below

client/coral-embed-stream/src/Embed.js
```diff
@@ -114,6 +123,11 @@ class Embed extends Component {
     }
   }
 
+  handleClick() {
```
alinhle: You can just do `handleClick = () => {` to have it automatically bound to `this`.

client/coral-embed-stream/src/Embed.js
```diff
@@ -144,7 +158,10 @@ class Embed extends Component {
       <div style={expandForLogin}>
         <div className="commentStream">
           <TabBar onChange={this.changeTab} activeTab={activeTab}>
-            <Tab><Count count={asset.totalCommentCount}/></Tab>
+            <Tab><Count
+              count={asset.totalCommentCount}
+              handleClick={this.handleClick}/>
```
alinhle: I think the `refetching` logic should be in `changeTab` and not on a clickable `Count`.
There is already a logic that was wrongly changed to only `refetch` for admins. 
You should remove the admin check and add your logic to show all comments.

client/coral-plugin-comment-count/CommentCount.js
```diff
@@ -3,8 +3,8 @@ import {I18n} from '../coral-framework';
 import translations from './translations.json';
 const name = 'coral-plugin-comment-count';
 
-const CommentCount = ({count}) => {
-  return <div className={`${name}-text`}>
+const CommentCount = ({count, handleClick}) => {
```
alinhle: See above

client/coral-embed-stream/src/Embed.js
gabelula: That is refactoring of the code, can we leave that for a refactoring story/PR ?

gabelula: The issue is that this logic is also used by the Show all comments button. 

And I agree about the issue on isAdmin only refetching data. Let's work that out in other story.

alinhle: No it's perfect there. The `isAdmin` was a merge mistake. It should never have been there. 
Just remove the  `isAdmin` and put the `his.props.viewAllComments();` over `refetch` and everything should work.

alinhle: Also we already have  `changeTab` for tab changing logic and we shouldn't spread the logic into different callbacks.

gabelula: Yes. I see that. The issue is that each time it comes from other tab by click '# comments' it will refetch twice.

wyattjoh (approved):

client/coral-plugin-comment-count/CommentCount.js
alinhle: You still need to remove the `handleClick` stuff

alinhle (changes requested):

client/coral-embed-stream/src/Embed.js
```diff
@@ -114,6 +121,11 @@ class Embed extends Component {
     }
   }
 
+  handleClick = () => {
+    this.props.viewAllComments();
+    this.props.data.refetch();
+  }
```
alinhle: This also needs to be removed

client/coral-embed-stream/src/Embed.js
alinhle: And the `handleClick` here :-)

alinhle (approved):

alinhle:
I removed my merge block (because I will go to sleep), so once you applied the last suggestions, this is good to go.

PR #489, opened 2017-04-12
Double quotes should be single

kgardnr:


wyattjoh (approved):
jason approved.

kgardnr:
@wyattjoh <3

PR #490, opened 2017-04-12
Fix load more buttons

alinhle:
## What does this PR do?

- Restore `View X New Comments` functionality.
- Fix bug falsely displaying the `View More Comments` button.

## How do I test this PR?

- Check it out in an asset with a lot of comments to see if the  `View More Comments` Button at the bottom works correctly.

- Post a comment in another browser tab and wait for the `View X New Comments` to appear.

- Check use case where both buttons are displayed

### DIFF
--- a/graph/resolvers/asset.js
+++ b/graph/resolvers/asset.js
@@ -1,4 +1,11 @@
 const Asset = {
+  lastComment({id}, _, {loaders: {Comments}}) {
+    return Comments.getByQuery({
+      asset_id: id,
+      limit: 1,
+      parent_id: null
+    }).then(data => data[0]);
+  },
   recentComments({id}, _, {loaders: {Comments}}) {
     return Comments.genRecentComments.load(id);
   },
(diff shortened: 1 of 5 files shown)


graph/resolvers/asset.js
```diff
@@ -1,4 +1,11 @@
 const Asset = {
+  lastComment({id}, _, {loaders: {Comments}}) {
+    return Comments.getByQuery({
+      asset_id: id,
+      limit: 1,
+      parent_id: null
+    }).then(data => data[0]);
```
wyattjoh, line 7: Should check that we have at least one result here.

wyattjoh: Also, is there a reason why we can't just use the separate comment query edge instead? Most of these params are exposed here.

wyattjoh (changes requested):

graph/resolvers/asset.js
alinhle: 1. I checked that and the graphql code is handling this well, it basically returns undefined and graphql translate that to `null`. Should I still do a check? :-D

alinhle: 2. Oh yes we could, but semantically this is really nice on the client. Otherwise I'll get a data property `lastComment` and it's an array with 1 comment instead of just an object.

kgardnr (approved):
:100:!

wyattjoh (approved):

PR #493, opened 2017-04-12
Configure stream settings 

gabelula:
… query.

## What does this PR do?

Story Bug: https://www.pivotaltracker.com/story/show/143520971

- moves settings into the state
- adds premodlinks to the query

## How do I test this PR?

1) 
- Go to the 'Configure Stream' tab of the asset
- Start writing a question
- Wait
- Check that it does not disappear when the query happens

2) 
- Go to the 'Configure Stream' tab of the asset
- Enable or Disable Pre moderation links and Apply 
- Refresh
- Check that is still there what you changed.


client/coral-configure/containers/ConfigureStreamContainer.js
```diff
@@ -53,18 +53,20 @@ class ConfigureStreamContainer extends Component {
 
   handleChange (e) {
     if (e.target && e.target.id === 'qboxenable') {
-      this.props.asset.settings.questionBoxEnable = e.target.checked;
+      this.state.settings.questionBoxEnable = e.target.checked;
     }
+    if (e.target && e.target.id === 'qboxcontent') {
+      this.state.settings.questionBoxContent = e.target.value;
+    }
+    if (e.target && e.target.id === 'qboxcontent') {
```
okbel: These two if's are the same?

okbel: These*

alinhle (changes requested):
~Problem with not reflecting the changed state:~
- ~We are not using apollo for the settings mutations and that's the reason why the state does not get updated. You can solve this by porting the settings mutations to apollo or instead as a temporary solution just do a refetch after applying the settings.~

client/coral-configure/containers/ConfigureStreamContainer.js
```diff
@@ -53,18 +53,20 @@ class ConfigureStreamContainer extends Component {
 
   handleChange (e) {
     if (e.target && e.target.id === 'qboxenable') {
-      this.props.asset.settings.questionBoxEnable = e.target.checked;
+      this.state.settings.questionBoxEnable = e.target.checked;
```
alinhle: Use `setState` to change the state. Never directly manipulate the state in React.

gabelula: I do not want to re-render the page each time I change this settings. That is what happens if I use setState.

gabelula:
The problem was not that the state is not being updated but that the settings are being re-fetch each time and overwriting what people may be changing while still did not hit apply.

client/coral-configure/containers/ConfigureStreamContainer.js
alinhle: You can calculate the new state before passing it to `setState`.

alinhle:
@gabelula ~This only applies to issue no 1:~

> 1) 
> - Go to the 'Configure Stream' tab of the asset
> - Start writing a question
> - Wait
> - Check that it does not disappear when the query happens

~But issue no 2 is because the new state is not reflected in Apollo as I mentioned above~

> 2)
> - Go to the 'Configure Stream' tab of the asset
> - Enable or Disable Pre moderation links and Apply 
> - Refresh
> - Check that is still there what you changed.

client/coral-configure/containers/ConfigureStreamContainer.js
alinhle: One reason is that React expects exclusive access to the state. To avoid any hard to debug bugs in the future and to maintain compatibility with future React versions, we should only mutate the state using `setState`. 

Also we want to treat state as an immutable object, to be able to apply performance optimisations to the component. In this example, you can construct a new settings object using `settings = {...this.state.settings, questonBoxEnable: true};`. Also because we won't need `premodLinksEnable` nor `questionBoxContent `, I would just avoid using the `settings` object and just put `questonBoxEnable` on the state.

alinhle:
Sorry for the back and forth, I got confused with another issue. It's perfectly fine to manage _dirty state_ here.

alinhle (approved):

alinhle:
Approved after adding a little todo comment :-)

client/coral-configure/containers/ConfigureStreamContainer.js
```diff
@@ -49,23 +49,29 @@ class ConfigureStreamContainer extends Component {
...
         });
       }, 300);
+
+      // this.props.loadAsset(this.props.data.asset);
     }
   }
 
   handleChange (e) {
+
+    // TO DO: Don’t directly manipulate state and make state change immutable.
```
alinhle: The convention is to write `TO DO` together like `TODO`. Most IDE supports detecting todos out of the box if you follow this convention.

PR #495, opened 2017-04-13
Bug on closing the stream

gabelula:
## What does this PR do?

- Adds closedAt to the state
- Check that the settings closedTimeout for the assets exists

## How do I test this PR?

- Go to configure tab for the stream
- Check that closing and opening stream is happening.


riley (approved):
good catch! works for me. 💯

PR #498, opened 2017-04-14
Docs update

wyattjoh:
## What does this PR do?

- Updates documentation regarding plugin management and install

## How do I test this PR?

- Review the documentation!


INSTALL.md
```diff
@@ -2,43 +2,203 @@
...
+There are some runtime requirements for running Talk for Docker:
+
+- [Docker](https://www.docker.com/) v1.13.0 or later
+- [Docker Compose](https://docs.docker.com/compose/) v1.10.0 or later
+
+_Please be sure to check the versions of these requirements. Incorrect versions
+of these may lead to unexpected errors!_
+
+### Installing
+
```
gabelula, line 57: Should we say that they have to create this file and feed it to the docker compose?

gabelula: And should also have a second title '## Installing from Docker' ?

gabelula (changes requested):

INSTALL.md
wyattjoh: For anyone familiar with docker-compose, this is already known and standard. Any wording beyond what is described here for the docker based setup doesn't make any sense to duplicate here, the only reason this much was described for the docker compose based install was we wanted to ensure that users knew how to select different versions of the command to run in order to scale differently and how the container exposes port 5000 as the base port.

wyattjoh: Also, the title format is consistent I think, as we have a super header of:

> Installing From Docker

As well as:

> Installing From Source

With other "Installing" subheaders under that.

INSTALL.md
```diff
@@ -2,43 +2,203 @@
...
+    restart: always
+    ports:
+      - "5000:5000"
+    depends_on:
+      - mongo
+      - redis
+    environment:
+      - TALK_MONGO_URL=mongodb://mongo/talk
+      - TALK_REDIS_URL=redis://redis
+  talk-jobs:
```
jde, line 118: This is excellent

jde (approved):

gabelula (approved):
It looks really good!

PR #502, opened 2017-04-17
All comments queue

riley:
## What does this PR do?

Adds a tab so you can see all the comments as they come in.

Also in this PR is the work for adding approved/rejected highlighted buttons depending on comment status. #530 

## How do I test this PR?

- view the admin
- there should be a tab that has the most recent comments, regardless of status.


gabelula:
I see the tab but I do not see any comment there. I see there is a comment in pre-mod but not in all. Attachign screenshots.

screen shot 2017-04-21 at 2 46 16 pm
screen shot 2017-04-21 at 2 46 13 pm

gabelula (changes requested):
I do not see all the comments. Are this only for the approved/rejected ones?

gabelula (approved):
It works.

PR #506, opened 2017-04-19
Fix load more in premod queue

alinhle:
## What does this PR do?

- Fix the load more button in pre-mod queue.

## How do I test this PR?

- Go to the pre-mode queue and load more.


riley (approved):
excellent! lgtm 💯

PR #509, opened 2017-04-19
Correctly handle action_summaries for respect

alinhle:
## What does this PR do?

- Correct handling of `action_summaries` in the respect plugin. The reason for this was because I had a wrong assumption on how `inline fragments` works in `graphQL`.

## How do I test this PR?

- Like, respect and flag a comment.
- Unrespect comment


riley:
This is breaking when `comment.action_summaries` is undefined. Creating a fix now.

riley:
reviewing this PR exposed a bunch of other places where we were using the same logic. I've moved all the fixes to https://github.com/coralproject/talk/pull/512 and am closing this one.

PR #511, opened 2017-04-20
Added --force to rebuild native modules

wyattjoh:
## What does this PR do?

Adds a `--force` flag to the yarn install process so that when installing dependancies, we ensure that native modules are rebuilt.

## How do I test this PR?

1. Build images locally: `./scripts/docker.sh`
2. Create a new folder (elsewhere on your harddrive) and create a Dockerfile in it with the following content:

```Dockerfile
FROM coralproject/talk:latest-onbuild
```

3. Build the docker image: `docker build -t coralproject/talk:pr-test .`
4. Run the image `docker run --rm -ti coralproject/talk:pr-test bash`
5. Run the following `node models/user.js`

There should be an error (specifically about not connecting to the database), but you should no longer see an error related to `bcrypt` not being installed.

This issue is related to an upstream issue with yarn: https://github.com/yarnpkg/yarn/issues/756


jde (changes requested):
Following the steps in the pr leads to a bcrypt error for me:

```
...
tmp :) ]$ docker run --rm -ti coralproject/talk:pr-test bash
root@b72f8b54aaff:/usr/src/app# node models/user.js
module.js:472
    throw err;
    ^

Error: Cannot find module '/usr/src/app/node_modules/bcrypt/lib/binding/bcrypt_lib.node'
    at Function.Module._resolveFilename (module.js:470:15)
    at Function.Module._load (module.js:418:25)
...
```

jde (approved):

PR #512, opened 2017-04-20
Action summary fix - Part Deux

riley:
## What does this PR do?

We had boilerplate all over the app for filtering by type out of the `comment.action_summaries` array. Several places in the code assume that

0. `comment.action_summaries` exists
1. The desired action summary will be the 0th element in the array
2. the `filter`ed ActionSummary array will be `length <= 1`

Sadly, these assumptions are false. This leads to all sorts of fun behavior such as likes triggering respects, counts being wrong, etc. This PR fixes and adds some helper utilities that @okbel started the other day.

This PR should fix:
- riley's flag detail PR that made assumptions about flag arrays
- LikeButton, RespectButton and FlagButton madness

## How do I test this PR?

- like, flag and respect with multiple comments and multiple accounts. behavior should make sense
- flag a comment with different accounts and different types. all the queues should show up and flag details should appear.


alinhle (changes requested):
As far as I know `comment.action_summaries` is never `null` so assumption no. 0 is correct (though it is not reflected in the graph schema). We need to talk this over.

alinhle:
The assumption that `comment.actions` is never `null` though is false, which should caused the issues you experienced with the ` flag detail`.

graph/typeDefs.graphql
```diff
@@ -440,7 +440,7 @@ type Asset {
 
   # Summary of all Actions against all entities associated with the Asset.
   # (likes, flags, etc.). Requires the `ADMIN` role.
-  action_summaries: [AssetActionSummary]
+  action_summaries: [AssetActionSummary!]
```
alinhle, line 443: Is this correct @wyattjoh? And also the other changes in the schema.

client/coral-embed-stream/src/Comment.js
```diff
@@ -124,9 +123,16 @@ class Comment extends React.Component {
       commentIsIgnored,
     } = this.props;
 
-    const like = getActionSummary('LikeActionSummary', comment);
-    const flag = getActionSummary('FlagActionSummary', comment);
-    const dontagree = getActionSummary('DontAgreeActionSummary', comment);
+    const likeSummary = getActionSummary('LikeActionSummary', comment);
+    const flagSummary = getActionSummary('FlagActionSummary', comment);
+    const dontagreeSummary = getActionSummary('DontAgreeActionSummary', comment);
```
alinhle: I don't agree that dontagree is not in camel case :-D But that's a very minor thing ^^

alinhle (approved):

PR #517, opened 2017-04-21
Fixed bug with status changes in the graph

wyattjoh:
## What does this PR do?

Previously, when a comment was approved/rejected, status changes were not pushed into history. This resolves that.

## How do I test this PR?

- Create a comment
- Report it
- Reject/Approve it
- Check in the db, notice that the `status_history` field has those updates in it


gabelula (approved):
It looks good.

PR #518, opened 2017-04-21
Added fix for flag actions

wyattjoh:
## What does this PR do?

When an action is created by the system, the flag `user_id` is null, because the system made it. Therefore, when a action is loaded when the user is not found, instead return undefined instead of trying to use the dataloader.


gabelula (approved):
It looks good. Locally I do not see the error we had with that flag.

PR #520, opened 2017-04-21
add a little keyboard shortcut hint widget

riley:
## What does this PR do?

in the moderation queue, there was no way to know that there were keyboard shortcuts. Now there's a little hint thing that can be dismissed letting you know that there's a wealth of shortcuts at your fingertips!

## How do I test this PR?

- view the moderation queue
- there should be a little shortcut thing there
- dismiss it
- reload the page; it should no longer be there


gabelula:
There is nothing that tells me how to have that box up again. Is that ok?

riley:
@IAmSamHankins should we add something that will allow you to bring up the hint box again?

theysamtheirs:
@riley @gabelula i would say it isn't needed. the pattern typically involves someone clicking "dismiss" because they longer want to see it/hints. we are also keeping the keyboard shortcuts menu in the "cog" in the upper right for user reference.

gabelula (approved):
It looks good to me.

alinhle (changes requested):
Can we put the `ui state` into `redux`? I wouldn't want to manage another layer of `state` in the `localStorage`. 

The best was for me is to sync the `localStorage` inside `redux`, e.g:

Load `initialState` from `localStorage` and during an `action` update both the `redux` and the `localStorage`.

alinhle:
That way we keep our `redux` store authoritative for all ui state and we are flexible with `actions` and other `redux` related advantages.

alinhle (approved):
Awesome @riley! This can be merged, after someone tested it.

okbel:
💯

okbel:
Let's merge when the CI ends testing.

PR #522, opened 2017-04-21
Get tags into a separate structure that can be applied to other models

gabelula:
## What does this PR do?

PT Story: https://www.pivotaltracker.com/story/show/142993479

Gets tags out of the Comment Model and into its own. We still need to have something in the FE to show the tags only to intended roles, look on how the system of decisions on who can apply them will work, add tags to the User Model.

## How do I test this PR?

- Everything should be working as before. We only changed on how the tags are being implemented in the backend.


wyattjoh (changes requested):
Tags should live on the models themselves, as it would act like metadata instead of splitting it into a separate collection. I would advise adjusting this to instead behave like our status_history behaves where we just apply atomic operations to the documents to add new tags and remove them.

I think this introduces too many new factors.

models/comment.js
```diff
@@ -89,6 +92,15 @@ const CommentSchema = new Schema({
   }
 });
 
+// Add the indexes on the comment tag.
+CommentSchema.index({
+  'id': 1,
+  'tags.id': 1
```
wyattjoh: The tags id should not be a part of this index, it is then coupling the id of the comment with the id of the tag, which doesn't make sense.

models/user.js
```diff
@@ -16,9 +17,32 @@ const USER_STATUS = [
...
+//   },
+//
+//   created_at: {
+//     type: Date,
+//     default: Date
+//   }
+// }, {
+//   _id: false
+// });
+
```
wyattjoh: Should remove this commented out code.

services/comments.js
```diff
@@ -49,70 +53,77 @@ module.exports = class CommentsService {
...
+        .then(({nModified}) => {
+          switch (nModified) {
+          case 0:
+            return Promise.reject(errors.ErrNoCommentFound);
+          case 1:
+            return;
+          default:
+          }
+        });
+    });
```
wyattjoh: There's a few issues here, I'll push some changes that resolve some issues with promises + query paths.

services/comments.js
```diff
@@ -49,70 +53,77 @@ module.exports = class CommentsService {
...
+      $pull: {
+        tags: {
+          id: tag_id
         }
-      });
+      }
+    }
+    )
+    .then(({nModified}) => {
+      switch(nModified) {
```
wyattjoh: If you are just performing an update, then just use the update instead of the `findOneAndUpdate`. Also, there is a `nMatched` that should be used to check to see if the query found something.

test/server/graph/loaders/metrics.js
```diff
@@ -23,11 +23,11 @@ describe('graph.loaders.Metrics', () => {
 
     describe('different comment states', () => {
 
-      beforeEach(() => CommentModel.create([
-        {id: '1', body: 'a new comment!'},
-        {id: '2', body: 'a new comment!'},
-        {id: '3', body: 'a new comment!'}
-      ]));
+      beforeEach(() =>[
+        CommentModel.create( {id: '1', body: 'a new comment!'}),
```
wyattjoh: An extra space here should be removed.

test/server/graph/mutations/removeCommentTag.js
```diff
@@ -34,34 +34,53 @@ describe('graph.mutations.removeCommentTag', () => {
...
+    await CommentsService.addTag(comment.id, 'BEST', user.id);
     const response = await graphql(schema, query, {}, context, {id: comment.id, tag: 'BEST'});
     if (response.errors && response.errors.length) {
       console.error(response.errors);
     }
     expect(response.errors).to.be.empty;
     expect(response.data.removeCommentTag.errors).to.be.null;
-    expect(response.data.removeCommentTag.comment.tags).to.deep.equal([]);
+
+    CommentsService.findById(response.data.removeCommentTag.comment.id)
```
wyattjoh: Should use `await` here, it escapes context by not calling return/await and we won't capture the error.

test/server/graph/mutations/removeCommentTag.js
```diff
@@ -34,34 +34,53 @@ describe('graph.mutations.removeCommentTag', () => {
...
-      it(userDescription, async function () {
-        const context = new Context({user});
 
-        // add a tag first
-        await CommentsService.addTag(comment.id, 'BEST');
-        const response = await graphql(schema, query, {}, context, {id: comment.id, tag: 'BEST'});
-        if (response.errors && response.errors.length) {
-          console.error(response.errors);
+    // allow the tag in the settings
+    SettingModel.findOneAndUpdate({id: 1}, {
```
wyattjoh: All database operations should be in a `beforeEach/before/it` block, otherwise it's not scoped properly for tests.

test/server/services/comments.js
```diff
@@ -224,20 +224,24 @@ describe('services.CommentsService', () => {
...
+      expect(tags[0].added_by).to.equal(userId);
+      expect(tags[0].created_at).to.be.an.instanceof(Date);
     });
     it('can\'t add a tag to comment id that doesn\'t exist', async () => {
       const commentId = 'fakenews';
       const tagName = 'BEST';
       const userId = users[0].id;
-      await expect(CommentsService.addTag(commentId, tagName, userId)).to.be.rejected;
+
+      CommentsService.addTag(commentId, tagName, userId)
```
wyattjoh: Same as before, need `await/return` to resolve.

wyattjoh (changes requested):

wyattjoh (dismissed):
After performing a review of the backend, I've made some alterations to the schema to resolve some issues with the workflow. The frontend should be adjusted to match the new returns and graph structure in the `typeDefs.graphql` file.

jde:
The permissions logic and tagLink data looks right. I can see how this separates the concept of tags cleanly for both tags that we want to store in a list as well as ad hoc tags!

The add/remove mutations, however, are still tightly bound to comments. Looking into the functions, the only thing that makes this comment specific is which service is used to add the tag.

I'd like to change addCommentTag, etc.. to addTag, etc... and pass in which model we want to add the tag to. This should realize our flexible tagging architecture for all product needs going forward.

graph/mutators/comment.js
```diff
@@ -209,21 +211,139 @@ const setCommentStatus = ({user, loaders: {Comments}}, {id, status}) => {
...
 
 /**
- * Adds a tag to a Comment
- * @param {String} id          identifier of the comment  (uuid)
- * @param {String} tag     name of the tag
+ * Adds a tag to a Comment.
  */
-const addCommentTag = ({user, loaders: {Comments}}, {id, tag}) => {
-  return CommentsService.addTag(id, tag, user.id);
+const addCommentTag = async ({user, loaders: {Assets, Comments, Settings}}, {id, asset_id, name}) => {
```
jde: Change addCommentTag to addTag. Accept input concerning which model to add the tag to.

jde: My recommendation would be to create a thin getServiceFromString function that takes a string and returns a the service object. We can then check to make sure that service contains `addTag()` and use that interchangeably below.

graph/mutators/comment.js
```diff
@@ -209,21 +211,139 @@ const setCommentStatus = ({user, loaders: {Comments}}, {id, status}) => {
...
+  }
+
+  // Load the settings from the dataloader, it will be cached if it was
+  // retrieved already.
+  let settings = await AssetsService.rectifySettings(AssetsService.findById(asset_id));
+
+  // Try to find the tag in the asset tags. This will contain the permission
+  // information.
+  let tag = settings.tags.find((globalTag) => {
+    return globalTag.name === name && globalTag.models.include('COMMENTS');
```
jde: `COMMENTS` here would become dynamic based on input from the query variables

graph/mutators/comment.js
```diff
@@ -209,21 +211,139 @@ const setCommentStatus = ({user, loaders: {Comments}}, {id, status}) => {
...
+  };
+
+  // If the tag was found, we need to ensure that the current user can indeed
+  // set this tag on the comment.
+  if (tag) {
+    
+    // If the tag has roles defined, and the current user has at least one of
+    // the required roles, then add the tag without checking for ownership.
+    if (tag.permissions && tag.permissions.roles && tag.permissions.roles.some((role) => user.roles.include(role))) {
+      return CommentsService.addTag(id, tagLink, false);
```
jde: `CommentsService` here and below would be the service returned by `getServiceFromString` at the outset of this function.

graph/mutators/comment.js
```diff
@@ -209,21 +211,139 @@ const setCommentStatus = ({user, loaders: {Comments}}, {id, status}) => {
...
 
 /**
- * Removes a tag from a Comment
- * @param {String} id          identifier of the comment  (uuid)
- * @param {String} tag     name of the tag
+ * Removes a tag from a Comment.
  */
-const removeCommentTag = ({user, loaders: {Comments}}, {id, tag}) => {
-  return CommentsService.removeTag(id, tag);
+const removeCommentTag = async ({user, loaders: {Comments}}, {id, asset_id, name}) => {
```
jde: See notes from addCommentTag.

graph/resolvers/root_mutation.js
```diff
@@ -29,12 +28,12 @@ const RootMutation = {
   setCommentStatus(_, {id, status}, {mutators: {Comment}}) {
     return wrapResponse(null)(Comment.setCommentStatus({id, status}));
   },
-  addCommentTag(_, {id, tag}, {mutators: {Comment}}) {
-    return wrapResponse('comment')(Comment.addCommentTag({id, tag}).then(() => CommentsService.findById(id)));
-  },
-  removeCommentTag(_, {id, tag}, {mutators: {Comment}}) {
-    return wrapResponse('comment')(Comment.removeCommentTag({id, tag}).then(() => CommentsService.findById(id)));
+  addCommentTag(_, {id, asset_id, name}, {mutators: {Comment}}) {
```
jde: `addCommentTag` becomes `addTag`, passes through `model` as part of input

graph/typeDefs.graphql
```diff
@@ -52,14 +52,30 @@ type User {
   status: USER_STATUS
 }
```
jde, line 92: We can create something like this:

```
enum TaggableModels {
  COMMENT,
  USER,
  ASSET
}
```

To control which models can receive comments.

graph/typeDefs.graphql
```diff
@@ -746,10 +767,10 @@ type RootMutation {
   setCommentStatus(id: ID!, status: COMMENT_STATUS!): SetCommentStatusResponse
 
   # Add tag to comment.
-  addCommentTag(id: ID!, tag: String!): AddCommentTagResponse
+  addCommentTag(id: ID!, asset_id: ID!, name: String!): AddCommentTagResponse
```
jde: See above recommendations to allow this query to apply to any `TaggabeModel`.

services/comments.js
```diff
@@ -43,76 +31,53 @@ module.exports = class CommentsService {
...
-   * @param {String} name the name of the tag to add
-   * @param {String} assigned_by the user id for the user who added the tag
-   */
-  static addTag(id, name, assigned_by) {
-
-    // Disabling allowed tags until we are able to extend them
-    // if (ALLOWED_TAGS.find((t) => t.name === name) == null) {
-    //   return Promise.reject(new Error('tag not allowed'));
-    // }
+  static addTag(id, tagLink, verifyOwnership = false) {
```
jde: This function and `removeTag` live in `services/tags.js` and be imported and appended to all services for 'taggable" models.

jde (changes requested):
This review contains a plan to introduce a concept "taggable" that encapsulates the tagging logic in a way that can be applied to any model.

graph/resolvers/util.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Decorates the typeResolver with the tags field.
+ */
+const decorateWithTags = (typeResolver) => {
+  typeResolver.tags = ({tags = []}, _, {user}) => {
+    if (user && user.hasRoles('ADMIN')) {
```
jde: This should be ADMIN || MODERATOR atm. Are we going to a user.canDo('seeTags') type of scheme to centralize all role logic on the backend or does it make the most sense to explicitly hardcode it here?

jde (changes requested):

graph/resolvers/util.js
wyattjoh: Makes sense to me, will implement those changes.

jde (approved):

jde:
Back end implementation looks good to me.

errors.js
```diff
@@ -75,6 +75,18 @@ const ErrMissingToken = new APIError('token is required', {
   status: 400
 });
 
+// ErrNoCommentFound is returned when trying to add a tag to a comment that does not exist.
+const ErrNoCommentFound = new APIError('comment does not exist', {
+  translation_key: 'COMMENT_NOT_FOUND',
+  status: 400
+});
+
+// ErrNoCommentFound is returned when trying to add a tag to a comment that does not exist.
```
alinhle: Looks like the copy and paste devil was here ;-)

okbel:
@wyattjoh Frontend functionality ready! Please check, best and staff tag.

jde (approved):

kgardnr:
Okay @wyattjoh @okbel @jde, we can merge this once the conflicts are resolved!

client/coral-framework/graphql/mutations.js
```diff
@@ -173,39 +173,91 @@ export const withDeleteAction = withMutation(
...
+  `;
+
+export const withAddTag = withMutation(
   gql`
-    mutation AddCommentTag($id: ID!, $tag: String!) {
-      addCommentTag(id:$id, tag:$tag) {
-        ...AddCommentTagResponse
+    mutation AddTag($id: ID!, $asset_id: ID!, $name: String!) {
+      addTag(tag: {name: $name, id: $id, item_type: COMMENTS, asset_id: $asset_id}) {
+          errors {
```
alinhle: Here should be `...AddTagResponse` to make it extendable.

client/coral-framework/graphql/mutations.js
```diff
@@ -173,39 +173,91 @@ export const withDeleteAction = withMutation(
...
   });
 
-export const withRemoveCommentTag = withMutation(
+export const withRemoveTag = withMutation(
   gql`
-    mutation RemoveCommentTag($id: ID!, $tag: String!) {
-      removeCommentTag(id:$id, tag:$tag) {
-        ...RemoveCommentTagResponse
+    mutation RemoveTag($id: ID!, $asset_id: ID!, $name: String!) {
+      removeTag(tag: {name: $name, id: $id, item_type: COMMENTS, asset_id: $asset_id}) {
```
alinhle, line 232: Here should be `...RemoveTagResponse` to make it extendable.

alinhle (changes requested):
- `Response` interface has been changed. The plugins that defines new responses must follow this or the break (like reactions).
- graphql framework has some wrong assumptions that need to be removed.

alinhle (approved):
GTG after wyatt approves!

PR #523, opened 2017-04-23
Frontend Plugins DOCS

okbel:
## What does this PR do?
Adds DOCS for Building Frontend Plugins


kgardnr (approved):
LGTM!!! Well done @okbel!

docs/frontend/PLUGINS.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+### Import Actions from Talk
+We can easily trigger `Talk` actions in our plugin Components.
+
+```js
+import React, {Component} from 'react';
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+import {addTag, removeTag} from 'coral-plugin-commentbox/actions';
```
alinhle: At one point we'll need to move this to another path right? Then plugins depending on this will break.

alinhle: Or will this be part of our public API as it is?

alinhle: If not, we should add a comment somewhere pointing out that things are still under development and are subjected to change.

jde: Sorry for the delayed response to this. Agreed that we cannot expose core functionality to plugin developers without a layer that will allow us to make updates without breaking plugins. I'm working on an update that will clarify.

PR #530, opened 2017-04-25
have active state on action buttons depending on comment status

riley:
## What does this PR do?

In the moderation queue, we want to have the buttons pre-highlighted as a visual indicator that they've already been rejected/approved.

## How do I test this PR?

- view the rejected queue
- the Reject button should be red

## note
there is currently no way to view an Approved comment. If that doesn't work for some reason while I'm building the Approved queue, I'll just fix it there.


riley:
I'm going to merge this into the All Comments PR #502

PR #531, opened 2017-04-25
shortcuts list should be accessible from the cog

riley:
## What does this PR do?

I forgot to put a link to the shortcut modal in the cog like @IAmSamHankins had asked. Now it works!

## How do I test this PR?

- go to the admin
- click the cog in the upper right
- there should be a menu option for Shortcuts
- it should display the shortcuts modal


kgardnr (approved):

kgardnr:
@riley @IAmSamHankins this is a little confusing, because the menu item exists but it only does anything on the Moderate tab.

PR #533, opened 2017-04-25
Only shows the Loading component when there is no search value

gabelula:
## What does this PR do?

removes Loading when there is a search value in admin/Community/People

## How do I test this PR?

- go to admin/community/people
- search something
- check that there is no 'flashing load comments' thingie coming up


kgardnr:
@gabelula there's still a weird flashing happening, but it's much improved than before. We can demo for design tomorrow and get their feedback!

kgardnr (approved):

PR #535, opened 2017-04-25
position the reporting popup in a more reliable way

riley:
## What does this PR do?

There was some weirdness with how the comment reporting popup looked on mobile
download

I changed the way the popup is positioned to be more like how the permalink popup is positioned so that it becomes more reliable at different widths.

## How do I test this PR?

- report a comment on mobile
- the popup shouldn't go off the screen


alinhle:
This still doesn't look right (in Chrome).
screen shot 2017-04-26 at 21 30 28

alinhle:
Also is this the design for a wider screen as well?
Doesn't look so nice for me, there is a lot of whitespace.

screen shot 2017-04-26 at 21 30 05

alinhle (changes requested):
See above

riley:
I'll figure that out @cvle. Not sure about the whitespace thing though.

alinhle:
@riley Also shadow is cut off on the right. I think you should ask for an ok from the design team for this change.

alinhle:
As a reference, it looked like this before: 
screen shot 2017-04-26 at 21 37 33

alinhle (approved):

PR #542, opened 2017-04-27
Only query once during initial render

alinhle:
## What does this PR do?
- Don't refetch on initial render

## How do I test this PR?
- Check that only one query is sent to the server


jde (changes requested):
This reduces the initial EmbedQuery count by one, but I'm still seeing it called twice on the initial load locally.

jde (approved):

Issue #551, opened 2017-04-28
Build broken when plugins.json is an empty object


wyattjoh:
### Expected behaviour

- Displays the stream

### Actual behaviour

- Displays the following error:

![screen shot 2017-04-28 at 8 38 14 am](https://cloud.githubusercontent.com/assets/633002/25533762/e3eec82a-2bee-11e7-9510-7aa11b6cf1e0.png)

### Steps to reproduce behaviour

- Create a `plugins.json` file with contents `{}`
- Build bundle
- Run server + view stream

wyattjoh:
This is also the same beheviour when the `plugins.json` file contains: `{"client": []}`

PR #552, opened 2017-04-28
Ret empty string when encountering undefined fragment

alinhle:
## What does this PR do?
- Fixes https://github.com/coralproject/talk/issues/551

## How do I test this PR?
- Run with empty plugins object.


wyattjoh (approved):
Resolved the problem for me

PR #555, opened 2017-04-28
Fix reply count polling

alinhle:
## What does this PR do?
- Fix reply count polling.
- Fix issue when posting a reply did not increment `replyCount` which also hides the fact that there might be a __single__ new reply that can be loaded.

### Bug 1
Use two tabs and send replies on one, and check that the other receives a `view x replies` message.

### Bug 2
Use two tabs and send ___one___ reply to the other, check that it says `view x replies` on the other tab. Reply before clicking on the button and see how the number of total replies increments by one and the button does not dissappear.


kgardnr (approved):
:100:!!! Way better!!

jde (approved):

PR #557, opened 2017-05-01
load more comments to moderate if available

riley:
## What does this PR do?

This is a little enhancement for the moderators. If you're in a queue, and you're approving/rejecting, and then you reach the end of the visible comments. If there were more comments available, you'd have to click the Load More button. With this PR, this all happens automatically.

## How do I test this PR?

- have some pre-moderated comments in the queue.
- in another tab, make another comment or two.
- now go back to the mod queue and take care of all the visible comments
- your new(er) comments should load automatically.


kgardnr (approved):

alinhle (dismissed):
I get this error when trying this branch: 
screen shot 2017-05-03 at 01 22 17

client/coral-admin/src/graphql/queries/index.js
```diff
@@ -62,13 +63,21 @@ export const loadMore = (fetchMore) => ({limit, cursor, sort, tab, asset_id}) =>
...
+      const updatedData = {
         ...oldData,
         [tab]: [
           ...oldData[tab],
           ...comments
         ]
       };
+
+      // if we're not in the all tab, put the new comments in there anyway
+      // this way the acceptComment and rejectComment mutations don't break
```
alinhle: I think this comments should be more like: `Always put comments into the all tab to maintain a consistent state.`. The breaking mutations are a result of an inconsistent state.

alinhle: Going through the code again I realised, that adding this code here is the wrong approach, and tries to work around the fact that `acceptComment` and `rejectComment` have some issues and need to be implemented differently.

client/coral-admin/src/graphql/mutations/index.js
```diff
@@ -54,7 +54,20 @@ export const setCommentStatus = graphql(SET_COMMENT_STATUS, {
...
+            }
+
+            if (!comment) {
+              comment = oldData.flagged.find(c => c.id === commentId);
+            }
+
+            if (!comment) {
+              comment = oldData.rejected.find(c => c.id === commentId);
+            }
+
```
alinhle: More pretty (sth like this): 

```js
const views = ['all', 'premod', 'flagged', 'rejected'];
const findCommentInQuery = (data, id) => views
  .reduce((comment, tab) => comment ? comment : data[tab].find(c => c.id === id), null)

```

riley: I don't know if that's more pretty. it's definitely more "es6", but much harder to parse.

alinhle: It should be fast. But most importantly for me is to reduce duplicated code :-O

PR #558, opened 2017-05-02
Update docker-compose sample configs

TrevorNT:
## What does this PR do?

Updates README.md and INSTALL.md with information on running the docker-compose samples by noting the requirement of `NODE_ENV` in README.md and by adding `NODE_ENV`, `TALK_SESSION_SECRET`, `TALK_ROOT_URL`, and `TALK_PLUGINS_JSON` (which when set to {} disables Facebook integration) to the sample configurations in INSTALL.md.

## How do I test this PR?

Do `docker-compose up` on both of the new sample configurations.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -72,6 +72,11 @@ services:
     environment:
       - TALK_MONGO_URL=mongodb://mongo/talk
       - TALK_REDIS_URL=redis://redis
+      - TALK_SESSION_SECRET=this_is_a_secret_string
+      - TALK_ROOT_URL=http://localhost
+      - TALK_FACEBOOK_APP_ID=none
+      - TALK_FACEBOOK_APP_SECRET=none
+      - NODE_ENV=development
   mongo:
     image: mongo:3.2
     restart: always
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -74,8 +74,7 @@ services:
       - TALK_REDIS_URL=redis://redis
       - TALK_SESSION_SECRET=this_is_a_secret_string
       - TALK_ROOT_URL=http://localhost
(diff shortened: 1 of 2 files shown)


wyattjoh:
It is already noted in the `INSTALL.md` documentation below the snippets that:

> At this stage, you should refer to the README.md for configuration variables that are specific to your installation.

I don't feel comfortable providing all the configuration (like the session secret) is a safe design for our documentation. I don't want to suggest using "keyboard cat" as the session secret as express (for nodejs) does because then you're left with something like this: https://github.com/search?utf8=✓&q=keyboard+cat&type=Code (Yes, there are a few examples there, but in 1.2M results, I guarantee that there is uses in production) where many people just simply don't take the time to change from the default.

TrevorNT:
Thanks for the feedback.  I get what you're saying about the secret (and yeah, that one could be left out).  I do still have a few concerns though:

1. It's still not mentioned that NODE_ENV is a required env var (leave it out and CSRF breaks).
2. It's not mentioned anywhere that the Facebook plugin initializes unless you provide a TALK_PLUGINS_JSON without it in there; otherwise, the app _appears_ to require TALK_FACEBOOK env vars, and quits without them (which had me very confused when I initially tried to start the app).

Would you accept a PR that fixed just those two in the docs?

wyattjoh:
The NODE_ENV is actually specified as production via: https://github.com/coralproject/talk/blob/master/Dockerfile#L24

It is also the default functionality of our application to include these plugins by default, these plugins which are defined in the `plugins` folder do provide documentation on setup (if necessary). I will patch the initial install docs to reiterate this fact to ensure that developers know where to look.

TrevorNT:
That helps.  Thank you!

PR #559, opened 2017-05-02
Config via Embed - Add external config passed into the script tag

okbel:
# Config via Embed
 Add external config passed into the script tag  

Usage:

```html
    <script src="/embed.js" async onload="
      Coral.Talk.render(document.getElementById('coralStreamEmbed'), {
        talk: '/',
        asset_url: '<%= asset_url ? asset_url : '' %>',
        asset_id: '<%= asset_id ? asset_id : '' %>',
        plugin_config: {
          test: 'data',
          debug: false
         // Pass custom data to your plugins
        }
       // Pass all your custom data
      })
    "></script>
```

The external config will live in the redux store `config`.

## Debug plugins
Still experimental. But you can pass a boolean to debug plugins. It will paint all the slots.
We will add another PR with more!

```js
  plugin_config: {
    debug: true
  }
```


alinhle (changes requested):
If I remember correctly our goal is to have everything in camelCase in the front-end. 

I didn't dig deeply into the changes yet. Will do it tomorrow.

client/coral-framework/components/Slot.js
```diff
@@ -1,16 +1,22 @@
...
 
-export default function Slot ({fill, inline = false, ...rest}) {
+function Slot ({fill, inline = false, plugin_config: {debug, config}, ...rest}) {
   return (
-    <div className={cn({[styles.inline]: inline})}>
-      {getSlotElements(fill, rest)}
+    <div className={cn({[styles.inline]: inline, [styles.debug]: debug})}>
+      {getSlotElements(fill, {config, ...rest})}
     </div>
   );
```
alinhle: It looks like you are passing `plugin_config.config` into `SlotElements`? It doesn't look right. Also `debug` is not passed into `SlotElements`.

client/coral-framework/components/Slot.js
```diff
@@ -1,16 +1,22 @@
 import React from 'react';
 import cn from 'classnames';
 import styles from './Slot.css';
+import {connect} from 'react-redux';
 import {getSlotElements} from 'coral-framework/helpers/plugins';
 
-export default function Slot ({fill, inline = false, ...rest}) {
+function Slot ({fill, inline = false, plugin_config: {debug, config}, ...rest}) {
```
alinhle: What's the `debug` for?

client/coral-embed-stream/src/containers/Embed.js
```diff
@@ -8,31 +8,41 @@ import renderComponent from 'recompose/renderComponent';
...
 const {fetchAssetSuccess} = assetActions;
 
 class EmbedContainer extends React.Component {
-
   componentDidMount() {
     pym.sendMessage('childReady');
+
+    pym.onMessage('config', config => {
+      this.props.addExternalConfig(JSON.parse(config));
+    });
```
alinhle: This is too late. The config must be available before any Components are rendered. Also how do you plan on applying the configs like `asset_id` to our components/query/existing state?

client/coral-framework/components/Slot.js
alinhle: Just saw it in the description. Seems like there have been some thought put into this. Would love to hear about it :-)

okbel: 👍

client/coral-framework/components/Slot.js
okbel: I don't want debug to be passed. Debug is for  `[styles.debug]: debug` this right now. And it highlights all the slots.

client/coral-embed-stream/src/actions/embed.js
```diff
@@ -8,3 +8,7 @@ export const setActiveTab = (tab) => (dispatch, getState) => {
   }
 };
 
+export const addExternalConfig = externalConfig => ({
```
alinhle: This looks duplicated and not used.

okbel: 👍

client/coral-embed-stream/src/containers/Embed.js
```diff
@@ -94,25 +92,24 @@ const mapStateToProps = state => ({
...
+  );
 
 export default compose(
   connect(mapStateToProps, mapDispatchToProps),
-  branch(
-    props => !props.auth.checkedInitialLogin,
-    renderComponent(Spinner),
-  ),
-  withQuery,
+  branch(props => !props.auth.checkedInitialLogin && props.config, renderComponent(Spinner)),
```
alinhle, line 114: Where does `props.config` come from? I think you forgot to add it in `mapStateToProps`.

client/coral-framework/components/Slot.js
```diff
@@ -1,16 +1,22 @@
...
+import {connect} from 'react-redux';
 import {getSlotElements} from 'coral-framework/helpers/plugins';
 
-export default function Slot ({fill, inline = false, ...rest}) {
+function Slot ({fill, inline = false, plugin_config: {debug, ...config}, ...rest}) {
   return (
-    <div className={cn({[styles.inline]: inline})}>
-      {getSlotElements(fill, rest)}
+    <div className={cn({[styles.inline]: inline, [styles.debug]: debug})}>
+      {getSlotElements(fill, {...config, ...rest})}
```
alinhle: I think this config part shouldn't have the `...`, this would essentially merge the data defined in config with the extra props passed to the slot, while letting the props from the slot to overwrite any config variables.

client/coral-framework/components/Slot.js
```diff
@@ -1,16 +1,22 @@
 import React from 'react';
 import cn from 'classnames';
 import styles from './Slot.css';
+import {connect} from 'react-redux';
 import {getSlotElements} from 'coral-framework/helpers/plugins';
 
-export default function Slot ({fill, inline = false, ...rest}) {
+function Slot ({fill, inline = false, plugin_config: {debug, ...config}, ...rest}) {
```
alinhle: A little thought: Any particular reasons why you hide the `debug` from the plugins? We could just pass it to the slots and whether they want to use it or not is up to the plugins. I think this would be more straight forward. Like "everything we pass into `plugin_config` is available to plugins" instead of "everything we pass into `plugin_config` is available to plugins except `debug`".

client/coral-framework/components/Slot.js
okbel: Right. We can put it at the end. Or pass it as an special reserved property.

client/coral-framework/components/Slot.js
```diff
@@ -4,10 +4,10 @@ import styles from './Slot.css';
 import {connect} from 'react-redux';
 import {getSlotElements} from 'coral-framework/helpers/plugins';
 
-function Slot ({fill, inline = false, plugin_config: {debug, ...config}, ...rest}) {
+function Slot ({fill, inline = false, plugin_config: {...config}, ...rest}) {
```
alinhle: you can simplify this to 
```js
function Slot ({fill, inline = false, plugin_config: config, ...rest}) {
```

client/coral-framework/components/Slot.js
```diff
@@ -4,10 +4,10 @@ import styles from './Slot.css';
...
   return (
-    <div className={cn({[styles.inline]: inline, [styles.debug]: debug})}>
-      {getSlotElements(fill, {...config, ...rest})}
+    <div className={cn({[styles.inline]: inline, [styles.debug]: config.debug})}>
+      {getSlotElements(fill, {...rest, ...config})}
```
alinhle: Now `plugin_config` can essentially overwrite any props. I believe you mean sth this
```js
{getSlotElements(fill, {...rest, config})}
```

plugins/coral-plugin-offtopic/client/components/OffTopicTag.js
```diff
@@ -7,6 +7,7 @@ const isOffTopic = (tags) => {
 
 export default (props) => (
   <span>
+    {console.log(props)}
```
alinhle: Don't forget to remove this.

okbel:
@cvle Updated!

alinhle (approved):

PR #561, opened 2017-05-03
Removed LikeAction from core typeName interface resolver

wyattjoh:
## What does this PR do?

- Removes `LikeAction` from the root `Action` resolver, this is added via a hook in the like plugin.

## How do I test this PR?

- Have a database containing likes
- Disable like plugin
- Rebuild static assets `yarn build`
- Restart server `CTRL-C` + `yarn dev-start`
- Load the moderation panel
- Notice now that the stream can load!


riley (approved):
great find! 💯

PR #563, opened 2017-05-04
Edit comment

gobengo:
## What does this PR do?
Implements "Edit Comment" story: https://www.pivotaltracker.com/story/show/142533447
Fixes bug where stream wouldn't reflect disabled Limit Comment Length setting: https://www.pivotaltracker.com/story/show/144654193

## How do I test this PR?

gif of editing a comment


graph/mutators/comment.js
```diff
@@ -219,20 +220,62 @@ const addCommentTag = ({user, loaders: {Comments}}, {id, tag}) => {
...
 
+/**
+ * Edit a Comment
+ * @param {String} id         identifier of the comment  (uuid)
+ * @param {Object} edit       describes how to edit the comment
+ * @param {String} edit.body  the new Comment body
+ */
+const editComment = async ({user, loaders: {Comments}}, {id, edit}) => {
+  const {body} = edit;
+  const comment = await CommentsService.findById(id);
```
wyattjoh: The edit operation should be able to get completed atomically. It would be sufficient enough to instead perform a query that looked like:

```js
const lastValidExpiryDate = new Date((new Date()).getTime() - EDIT_WINDOW_MS);
db.comments.update({
  id,
  author_id: user.id,
  created_at: {
    $gt: lastValidExpiryDate
  }
}, {
  $set: {
    body,
    status,
  },
  $push: {
    body_history: {
      body,
      created_at: new Date(),
  },
  status_history: {
    type: status,
    created_at: new Date(),
  }
});
```

Where after that you can inspect the nMatched + nModified to respond with the correct error.

wyattjoh (changes requested):

client/coral-plugin-commentbox/CommentBox.js
```diff
@@ -1,28 +1,174 @@
...
+  if (status === 'REJECTED') {
+    addNotification('error', lang.t('comment-post-banned-word'));
+  } else if (status === 'PREMOD') {
+    addNotification('success', lang.t('comment-post-notif-premod'));
+  }
+};
+
+/**
+ * Common UI for Creating or Editing a Comment
+ */
```
alinhle: I think this component should be put in an extra file.

client/coral-embed-stream/src/components/EditableCommentContent.js
```diff
@@ -0,0 +1,204 @@
...
+import {Icon} from 'coral-ui';
+import I18n from 'coral-framework/modules/i18n/i18n';
+import translations from 'coral-framework/translations';
+const lang = new I18n(translations);
+
+const getEditableUntilDate = (comment) => {
+  const editing = comment && comment.editing;
+  const editableUntil = editing && editing.editableUntil && new Date(Date.parse(editing.editableUntil));
+  return editableUntil;  
+};
```
alinhle: This is duplicated logic, we should share the same logic e.g. in a utils file.

client/coral-embed-stream/src/components/EditableCommentContent.js
```diff
@@ -0,0 +1,204 @@
...
+        />
+      </div>
+    );
+  }
+}
+
+/**
+ * Countdown the number of seconds until a given Date
+ */
+class CountdownSeconds extends React.Component {
```
alinhle: This should be in an extra file.

client/coral-embed-stream/src/containers/Embed.js
```diff
@@ -84,6 +85,9 @@ export const withQuery = graphql(EMBED_QUERY, {
       hasComment: commentId !== '',
       excludeIgnored: Boolean(auth && auth.user && auth.user.id),
     },
+    reducer: (previousResult, action, variables) => {
+      return reduceEditCommentActionsToUpdateStreamQuery(previousResult, action, variables);
```
alinhle: So far we have all our logic in mutations. I'm not sure if we should have some of our logic that modifies the query in mutations and some in the query reducer. What do you think?

client/coral-plugin-commentbox/CommentBox.js
```diff
@@ -37,7 +183,7 @@ class CommentBox extends React.Component {
     let comment = {
       asset_id: assetId,
       parent_id: parentId,
-      body: this.state.body,
+      body: body,
```
alinhle: Can shortened to: `body,`

client/coral-embed-stream/src/containers/Embed.js
gobengo: Doing it this way is what felt most right after reading http://dev.apollodata.com/react/cache-updates.html .

it says

> We recommend using the reducer option, except when there’s a good reason to use updateQueries instead (eg. if it would make your app much easier to understand and maintain).

graph/mutators/comment.js
gobengo: @wyattjoh I'm glad you can help me optimize this. Right now that findById is also needed to look up that comment's asset_id to apply the appropriate (e.g. bad word) filtering before deciding whether the update should even be made. Can your proposal accomodate this and still save a mongo query?

wyattjoh: Sure! You would have to adjust the mutation for editing a comment to include the asset id in the query params, that would ensure that it was possible to retrieve the data you need. It would update as so:

```js
const lastValidExpiryDate = new Date((new Date()).getTime() - EDIT_WINDOW_MS);
db.comments.update({
  id,
  asset_id,
  author_id: user.id,
  created_at: {
    $gt: lastValidExpiryDate
  }
}, {
  $set: {
    body,
    status,
  },
  $push: {
    body_history: {
      body,
      created_at: new Date()
    },
    status_history: {
      type: status,
      created_at: new Date(),
    }
  }
});
```

gobengo: @wyattjoh Ok I will add asset_id to the graph api and do something like this so that the happy path involves only one mongo write.

gobengo:
#review

gobengo:
@wyattjoh it says you still request changes. I pushed since your last comment. Any other changes you want?

services/comments.js
```diff
@@ -51,50 +51,76 @@ module.exports = class CommentsService {
...
+        },
+      }
+    );
     switch (nModified) {
-    case 0:
-      throw new Error(`Couldn't edit comment. There is no Comment with id "${id}"`);
+    case 0: {
+
+          // disambiguate possible error cases
+      const comment = await this.findById(id);
```
wyattjoh: It shouldn't be necessary to provide this verbose level of errors. This error is already resolved by the frontend, and the edit is already protected via the backend. I don't feel like we should waste another call to the database to resolve a possible convoluted error situation.

services/comments.js
```diff
@@ -51,50 +51,76 @@ module.exports = class CommentsService {
...
+        throw Object.assign(new Error('Comment not found'), {
+          name: 'CommentNotFound'
+        });         
+      } else if (comment.author_id !== editor) {
+        throw Object.assign(new Error('You aren\'t allowed to edit that comment'), {
+          name: 'NotAuthorizedToEdit'
+        });
+      } else if (( ! ignoreEditWindow) && editWindowExpired(comment)) {
+        throw Object.assign(new Error('Edit window is over.'), {
+          name: 'EditWindowExpired'
```
wyattjoh: All errors should be registered in the `errors.js` file to support translations.

services/comments.js
gobengo: > This error is already resolved by the frontend

This is only true insofar as coral-embed-stream is the only 'frontend' that wants to talk with the CommentsService. Right now that's mostly true. But if this project is successful, there will also be:

* Alternative web clients for customers who want different design/ux/etc
* Native mobile app clients
* Forks who want to add features using the services directly

Developing alternative clients (and even other talk features that want to use this service method) is significantly easier with useful error messages. 

The 'cost' of this call right now is a couple milliseconds in a very rare case.

The 'cost' of any other user of CommentService trying to figure out what to do with an ambiguous error is several orders of magnitude more. Let's say one of those use cases involves a requirement that the end-user can see a message explaining that their edit failed because the time window was up (as opposed to another reason). Then the 'cost' to that user also involves reintroducing this code anyway (and having this same discussion with you).

I've never read an API Design Guide that didn't include a principle like "Return useful error messages". I'm all for ensuring an optimized happy path, but removing this part is optimizing for theory and fractions of a penny of CPU time at the expense of real humans.

wyattjoh: From my perspective, any path that does not aggressively minimize the number of queries being executed against the database is one that needs reconsideration.

The 'cost' may be a few ms per request, but consider a high traffic site where that number can easily explode to a significant load against the database server.

Ideally, when faced with any one function, we should keep these to an absolute minimum, justifying each and every external service call. We should definitely return useful error messages, but not at the expense of slowing down the system, that's why I've always advocated use of defensive database queries that negate the need for non-atomic style operations which would fail spectacularly with a non-transaction based database like Mongo.

Instead of using an `update` operation, mongoose (through the underlying MongoDB driver) provides a `findOneAndUpdate` which would reduce the query cost down to 1 rather than 2.

wyattjoh (approved):

client/coral-framework/graphql/mutations/index.js
```diff
@@ -171,3 +172,22 @@ export const stopIgnoringUser = graphql(STOP_IGNORING_USER, {
...
+    return {
+      editComment: (id, asset_id, edit) => {
+        return mutate({
+          variables: {
+            id,
+            asset_id,
+            edit,
+          },
+          refetchQueries: [
+            'EmbedQuery'
```
alinhle: Why are you refetching the `EmbedQuery`? We don't want to rely on refetching for mutations.

alinhle (changes requested):

client/coral-framework/graphql/mutations/index.js
gobengo: @cvle If you edit a comment that's APPROVED, and that sets the state back to something like PREMOD or REJECTED, then the result of EmbedQuery will no longer include the edited comment. Without this line, the edited (and now non-public) comment won't be removed from UI.

> We don't want to rely on refetching for mutations.

Why not? What's a better tool for the job?

services/comments.js
gobengo: > Instead of using an update operation, mongoose (through the underlying MongoDB driver) provides a findOneAndUpdate which would reduce the query cost down to 1 rather than 2.

I want this if possible! I looked at findOneAndUpdate as well, but I didn't think it would help here. Tell me if I'm wrong. Because the filter we're using is filtering on all of { id, asset_id, author_id, created_at }, if any one of those conditions doesn't match, findOneAndModify wouldn't return a document. So to figure out *which* of those conditions didn't match, you'd still need to do a find by { id } to determine what's wrong.

client/coral-framework/graphql/mutations/index.js
alinhle: You'd have to update the query using the mutation result and the optimistic response. Refetching is not only ineffective (extra round trip, server load, and slow ui response) but also removes any pagination data that was fetched from the server and merged into the query.

client/coral-plugin-commentbox/CommentBox.js
```diff
@@ -53,12 +71,12 @@ class CommentBox extends React.Component {
...
+        notifyForNewCommentStatus(addNotification, postedComment.status);
+
         if (postedComment.status === 'REJECTED') {
-          addNotification('error', lang.t('comment-post-banned-word'));
-          !isReply && setCommentCountCache(commentCountCache);
+          !isReply && setCommentCountCache(assetId, commentCountCache);
         } else if (postedComment.status === 'PREMOD') {
-          addNotification('success', lang.t('comment-post-notif-premod'));
-          !isReply && setCommentCountCache(commentCountCache);
+          !isReply && setCommentCountCache(assetId, commentCountCache);
```
alinhle: image
I have an error when posting a comment and premod is on.
Why was these two lines changed? Is it a merge error?

client/coral-framework/graphql/mutations/index.js
alinhle: Aren't you already using a reducer? What is it doing? Refetch essentially resets and rewrites everything in the query cache.

client/coral-plugin-commentbox/CommentBox.js
gobengo: Yes! I missed this in the big ol merge with coral-embed-stream refactor. Thanks for catching it!

client/coral-framework/graphql/mutations/index.js
gobengo: @cvle Alright I updated to use reducer instead of refetching EmbedQuery.

alinhle (approved):

PR #564, opened 2017-05-04
User detail ui

riley:
## How do I test this PR?

- click a button
- view the cat
- see the cat meow


client/coral-admin/src/containers/ModerationQueue/UserDetail.js
```diff
@@ -0,0 +1,60 @@
...
+  render () {
+    const {data, hideUserDetail} = this.props;
+
+    if (!('user' in data)) {
+      return null;
+    }
+
+    const {user} = data;
+    const localProfile = user.profiles.find(p => p.provider === 'local');
+    const facebookProfile = user.profiles.find(p => p.provider === 'facebook');
```
wyattjoh: This should not reference plugins content. We should instead utilize slots to inject Facebook details into the user detail pane.

wyattjoh (changes requested):

kgardnr (approved):

wyattjoh (approved):

PR #565, opened 2017-05-04
Don't redo actions on comments

riley:
## What does this PR do?

If a comment is approved, don't allow the front-end to re-approve it. Same for rejected comments. The surface area of this PR includes the approve/reject action buttons in the queues, the keyboard shortcuts, and auto-rejecting a comment when you ban a user.

## How do I test this PR?

- go to the queue and click the Approve button on an `ACCEPTED` comment. nothing should happen. same for the rejected queue and a `REJECTED` comment.
- try to use `t` in any queue on an approved comment. nothing.
- use `r` on an already rejected comment
- ban a user from the Rejected queue. There should be a query for setting the status of the user, but not a following one for rejecting the attached comment.


alinhle (approved):
Looks good

PR #567, opened 2017-05-05
Extendable GraphQL framework

alinhle:
## What does this PR do?

Implements a framework that exposes a similar API like `react-apollo` but allows most aspects to be extended by our core and by plugins.

- Implements a `registry` service to register fragments, mutation options, and query options.
- Fragments defined in `coral-framework` are automatically registered in the `registry`.
- Fragments, mutation and query options defined in `plugins` are automatically registered in the `registry`.
- Exposes HOCs `withMutation` and `withQueries` to be used instead of `graphql`. The HOCs accept the same signature as `graphql` and also applies registered mutation options and query options.
- Extends our plugins' `index.js` to accept the following configuration: 
```js
{
  fragments: {
    CreateCommentResponse: gql`
      fragment CoralRandomEmoji_CreateCommentResponse on CreateCommentResponse {
        comment {
          emoji
          replies {
            emoji
          }
        }
      }`,
  },
  mutations: {
    // state is the current redux state, which is sometimes 
    // necessary to fill the optimistic response.
    PostComment: ({variables, state}) => ({
      optimisticResponse: {
        CreateComment: {
          emoji: '',
        },
      },
      refetchQueries: [],
      updateQueries: {
        EmbedQuery: (previous, data) => {
          return previous;
        },
      },
      update: (proxy, result) => {
      },
    })
  },
  queries: {
    EmbedQuery: {
      reducer: (previousResult, action, variables)  => {
        return previousResult;
      },
    },
  }
```
- Our core can also register such a config object by using `registerConfig` from the `registry` service.
- ~~We can now allow any `graphql document` to be extended by using fragments from the `registry`:~~
- Any registered fragments in the registery will be included in `queries` and `mutations` if it matches a fragment name.
```js
gql`
  mutation PostComment($comment: CreateCommentInput!) {
    createComment(comment: $comment) {
      ...CreateCommentResponse
    }
  }
`;
```

## Notes / Caveats

- Query, Mutation and Fragment names must be globally unique (not a new requirement).
- Mutation HOCs are defined globally like before. Embeds and plugins add their logic to the mutations using the new framework.
- ~~Extending a `graphql document` inside a plugin is currently not supported due to circular dependency issues~~
- ~~The fragment registry API will change in the next iteration in order to remove the limitation above and to support `plugins in plugins`~~
- The automatic registration for plugins ~~will likely~~ _might_ be removed ~~with above changes~~ _later_

## How do I test this PR?

Check if changes doesn't break any existing feature.


alinhle:
Do you think `registry` and `registerConfig` are appropriate names? Other ideas?

riley:
is the registry only handling plugin stuff? maybe it should just be `pluginRegistry`

alinhle:
No it handles everything related to extending our graphql. I could call it `graphRegistry` instead if noone think that this is too verbose.

jde (changes requested):
This is fantastic! three naming updates that came up in review:

registry -> graphqlRegistry
graphqlRegistry.register* -> graphqlRegistry.add*
graphqlRegistry.registerConfig -> graphqlRegistry.add

alinhle:
Done :-)

jde (approved):
Approved. This is a really impressive thing. Amazing work @cvle!

PR #569, opened 2017-05-05
Apply button overlapped text

riley:
## What does this PR do?

The settings Apply button was being positioned with css transforms and `postition: absolute`. This takes it out of the css flow and does weird things with text. `float: right;` is just fine and will give the desired behavior.

## How do I test this PR?

- view the stream settings on mobile.
- the Apply button no longer obscures the description.


kgardnr (approved):
Awesome.

PR #571, opened 2017-05-08
Coral Reactions and Coral Plugin Love

okbel:
# Coral Reactions
This PR adds a better API for creating Reactions.
Adds `coral-plugin-love`

## Building Reactions
This PR introduces `withReactions` HOC that abstracts all the logic behind handling reactions.

A minimal reaction plugin should look like this:
```js
import React from 'react';
import {withReaction} from 'coral-plugin-api';

class LoveButton extends React.Component {
  handleClick = () => {
    const {
      postReaction,
      deleteReaction,
      alreadyReacted
    } = this.props;

    if (alreadyReacted()) {
      deleteReaction();
    } else {
      postReaction();
    }
  };

  render() {
    const {count} = this.props;
    return (
      <button onClick={this.handleClick}>
        <span>Love</span>
        <span>{count > 0 && count}</span>
      </button>
    );
  }
}

export default withReaction('love')(LoveButton);
```

Yes. That's it.

## Translations, Styles and handling User behaviour
Please check the source code within `coral-plugin-love`. That's a complete example for building Reactions.

## Coral Plugin Love
Allows you to love comments ❤️ 
out

## Test this branch
- Pull the branch
- Go to `plugins.json` and add `coral-plugin-love` both in `client` and `server`
- Run it
- Check that the love reaction is working as expected


client/coral-framework/hocs/withReaction.js
```diff
@@ -0,0 +1,238 @@
...
+      );
+
+      const count = getTotalActionCount(
+        `${capitalize(reaction)}ActionSummary`,
+        comment
+      );
+
+      const withReactionProps = {reactionSummary, count};
+
+      return <WrappedComponent {...this.props} {...withReactionProps} />;
```
jde: I recommend giving each wrapped building block component a unique name, such as `WrappedReactionComponent`. This would give us the ability to quickly, visually distinguish between "building blocks" by looking at their wrapped components.

plugins/coral-plugin-love/client/LoveButton.js
```diff
@@ -0,0 +1,52 @@
...
+    const {
+      postReaction,
+      deleteReaction,
+      showSignInDialog,
+      reactionSummary
+    } = this.props;
+    const {root: {me}, comment} = this.props;
+
+    // If the current user does not exist, trigger sign in dialog.
+    if (!me) {
```
jde, line 20: This is great logic to point to as _recipes_ for all sorts of front end components. My first thought was to hide these in the building block, but after thinking about it I agree to leave them here to allow more customization by plugin authors. Great design choice.

jde (changes requested):
Excellent implementation! My only comments are about naming conventions. 

I'd like to make the name of the folder that we import these building blocks from very explicit so that people coming to the code for the first time will intuitively find them.

What does the /hocs/ folder name mean?

Thoughts on the term `building blocks`?

This is our opportunity to make the most commonly used import paths for all of our plugins super concise and readable.

I would also recommend moving the folder to the top level so the import statements read:

```
  const withSomething = import('/[a clear name tbd]/withSomething'); 
```

plugins/coral-plugin-love/client/LoveButton.js
```diff
@@ -0,0 +1,52 @@
...
+      showSignInDialog();
+      return;
+    }
+
+    // If the current user is banned, do nothing.
+    if (me.status === 'BANNED') {
+      return;
+    }
+
+    if (reactionSummary) {
```
jde: I'd make this a bit more declarative by creating a helper function:

```
  if (alreadyReacted) {...
```

This will allow us to change the reactionSummary architecture without breaking this concept.

okbel: Hm, but that object contains the actual reaction summary. It's not a boolean.

jde: True, but by exposing that implementation detail to plugin developers we are making it so that we can never change it without breaking plugins.

If, on the otherhand, we pass in a funciton that looks like this:

```
  const allreadyReacted = () => {
    return reactionSummary ? true : false;
  }

```

We pass that logic into the higher order component so we can change it in the future.

Also, it isn't intuitive to check for a truthy state of reactionSummary to determine whether or not a user has already taken an action. Creating a clearly named function makes it more declarative even if it does nothing but return whether or not there's a reaction summary.

plugins/coral-plugin-love/client/LoveButton.js
```diff
@@ -0,0 +1,52 @@
...
+  };
+
+  render() {
+    const {count, reactionSummary} = this.props;
+    return (
+      <button
+        className={`${styles.button} ${reactionSummary ? styles.loved : ''}`}
+        onClick={this.handleClick}
+      >
+        <span>{lang.t(reactionSummary ? 'loved' : 'love')}</span>
```
jde: Same comment here as above. This would be more declarative / clear and future proof:

```
<span>{lang.t(alreadyReacted() ? 'loved' : 'love')}</span>
```

plugins/coral-plugin-love/client/LoveButton.js
alinhle: There is no need to turn this into a function, the HOC fully controls how the props are assembled and passed to the component.

client/coral-framework/hocs/withReaction.js
alinhle: @jde What you mean is the display name of the resulting component (right?), we should totally do this and it needs to be defined in the `displayName` static variable, see our other HOCs that we already have in the framework. @okbel

plugins/coral-plugin-love/client/LoveButton.js
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import {Icon} from 'coral-ui';
```
alinhle, line 2: @jde Do you consider this public api? What is our policy. Maybe only things that are explicitly documented in a doc file in our repository is public api?

plugins/coral-plugin-love/client/LoveButton.js
```diff
@@ -0,0 +1,52 @@
...
+import styles from './styles.css';
+import {I18n} from 'coral-framework';
```
cvle: @jde see above

alinhle, line 4: See above. (Also is our I18n code stable?)

plugins/coral-plugin-love/server/typeDefs.graphql
```diff
@@ -0,0 +1,70 @@
+enum ACTION_TYPE {
```
cvle: @jde Every name we introduce here in the schema is global. 2 plugins that happen to use the same name anywhere will not work with each other. I leave it up to you, if we need a strategy now or later.

plugins/coral-plugin-love/server/typeDefs.graphql
```diff
@@ -0,0 +1,70 @@
...
+
+    # Represents a Love.
+    LOVE
```
alinhle, line 4: @jde Every name we introduce here in the schema is global. 2 plugins that happen to use the same name anywhere will not work with each other. I leave it up to you, if we need a strategy now or later.

plugins/coral-plugin-love/client/LoveButton.js
alinhle: This essentially exposes the result of our query from the graph. We could provide more abstracted props that does not relate to our graph, but on the other hand our graph is fully exposed on the server side anyway, so I don't really see a point. (high prio for stabilising our graph schema)

alinhle:
`hocs` means Higher-Order-Components and is a technical term which is defined and all HOCs have more or less the same api.

Personally I am not a big fan of the term `building blocks` :-P

plugins/coral-plugin-love/client/LoveButton.js
jde: Agreed that a function isn't necessary here. An `alreadyReacted` boolean passed in would do.

jde:
Finishing List:

1 - Make name of the wrapper component unique, use `displayName` as suggested by @kiwi (https://github.com/coralproject/talk/pull/571#discussion_r115562297)

2 - Pass in `alreadyReacted` boolean variable from hoc. Use this for logic in place of relying on truthy reactonSummary

3 - Consider a distinct path to put the hocs that we are exposing as our stable plugin API. I'm not in love with `building blocks` but I'd like to find a clear name for these things and a permanent path for them to live. Our path should reflect the fact that these particular hocs (and no others) constitute our plugin API.

plugins/coral-plugin-love/server/typeDefs.graphql
jde: The strategy that I've been recommending is to use the name of the plugin in all globally scoped places. In this case, this plugin works with an action type of LOVE. If another plugin also uses LOVE as an action type there may be unexpected consequences. 

To be super strict, we could insist that this kind of thing is written 
```
enum ACTION_TYPE {

  CORAL-PLUGIN-LOVE
}
```

This is more verbose and gets ugly when writing graphql types but does self-document all the way to the database level as the strings in the action show that it came from a plugin.

jde: Ultimately my call is to not include the "CORAL-PLUGIN" everywhere. I tried doing that here: https://github.com/coralproject/talk-plugin-mentioned-in-story/blob/server-side-initial-dev/server/typeDefs.graphql#L45 and found that things are much clearer when the concept is modeled.

jde (approved):

PR #580, opened 2017-05-11
remove ability to un-flag a comment

riley:
## What does this PR do?

We don't want to allow a user to un-report something they've already reported.

## How do I test this PR?

- report a comment
- try report again, nothing should happen even after a reload

### DIFF
--- a/client/coral-plugin-flags/FlagButton.js
+++ b/client/coral-plugin-flags/FlagButton.js
@@ -15,8 +15,7 @@ class FlagButton extends Component {
     message: '',
     step: 0,
     posted: false,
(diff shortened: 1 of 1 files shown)


kgardnr (changes requested):
Hey @riley, this works perfectly for reporting a comment.

However, when you report a username, the button doesn't turn to "Reported", but the comment does go to the Flagged Usernames queue. Can you take a look?

kgardnr (approved):

kgardnr:
This is a bug in master; I've created an issue for it: https://www.pivotaltracker.com/story/show/145283169

PR #503, opened 2017-04-18
Embed Stream refactor

alinhle:
This is a first step to add support for extendable queries.

## What does this PR do?
- Move `components` into a separate folder.
- Add `reducers`, `actions` and `constants` to the embed in order to manage UI state
- Split `Embed ` component into a `presentational component` and a `container`
- Put `activeReplyBox` into `Redux`
- Simplify `countCache` as `commentCountCache` and put it in the local embed reducer.
- Remove stale props from `ConfigureStreamContainer` and `ProfileContainer`
- Let `ConfigureStreamContainer` and `ProfileContainer` retrieve their own props from store instead from `Embed`.

## How do I test this PR?

- Check that everything in the embed still works as before.

PR #596, opened 2017-05-17
Ignored user tombstones work again

gobengo:
## What does this PR do?

Help with https://www.pivotaltracker.com/n/projects/1863625/stories/145546647

#503 made it so the ignoreUser mutation refetched EmbedQuery. This broke the behavior of "Ignore User" where, once you ignore a user using the menu on their comment, all their comments are supposed to re-render as 'tombstones'.

It also caused this behavior that the bug describes:
> When User A ignores User B on the second page, the page reloads, and User A is viewing the first page only. When User A ignores a user, the same amount of comments they loaded should remain accessible.

This PR makes it so ignoreUser does not refetch the EmbedQuery.

## How do I test this PR?

* Ignore a user
* That comment and all others by that user should render as tombstones like the gif on #477


client/coral-embed-stream/src/containers/Stream.js
```diff
@@ -113,6 +113,7 @@ class StreamContainer extends React.Component {
   componentDidMount() {
     if (this.props.previousTab) {
       this.props.data.refetch();
+      this.props.setCommentCountCache(0);
```
gobengo: I need advice on this.

Here's the issue: Let's say you have ignored a user with a several recent comments. Refresh. The stream loads, and EmbedQuery and the comment counts queries aren't factoring in the comments from the ignored user.

Now go to the 'My profile' tab. Stop ignoring that user. Now switch back to the 'Comments' tab, which triggers the 'refetch()' above this line. Now asset.comments will include the comments by the previously ignored user. *However*, nothing has reset the commentCountCache, while asset.commentCount has gone up (by the amount of comments by the previously ignored user). This causes NewCount to think there are new comments, and thus render.

So... is this the best place to reset the comment count cache? Since EmbedQuery (asset.comments) and commentCountCache are so tightly coupled, is there any way we can make it so that whenever EmbedQuery is re-fetched, we also reset the commentCountCache?

gobengo: @cvle any advice? Or who else knows a lot about setCommentCountCache?

alinhle: I think `refetch` is a promise and you can set the commentCount to the correct value when the promise returns the new result. The whole `commentCount` thing will be gone as soon as we have an alternative subscription based implementation.

client/coral-embed-stream/src/graphql/index.js
```diff
@@ -122,19 +122,43 @@ const extension = {
...
+        }
+      }
     }),
-    StopIgnoringUser: () => ({
-
-      // TODO: don't rely on refetching.
+    StopIgnoringUser: ({variables}) => ({
       refetchQueries: [
-        'EmbedQuery', 'EmbedStreamProfileQuery',
+        'EmbedQuery',
```
alinhle: I think the `EmbedQuery` refetch here is not needed. Whenever we change to the `comments` tab, a refetch is triggered automatically.

gobengo: done

client/coral-embed-stream/src/containers/Stream.js
gobengo: done

client/coral-embed-stream/src/containers/Stream.js
```diff
@@ -112,8 +112,8 @@ class StreamContainer extends React.Component {
 
   componentDidMount() {
     if (this.props.previousTab) {
-      this.props.data.refetch();
-      this.props.setCommentCountCache(0);
+      this.props.data.refetch()
+        .then(() => this.props.setCommentCountCache(0));
```
cvle: Why do you set it to `0`? Does this work? I think it should be set to `asset.commentCount` from the new query result. Do you have access to it inside this callback?

cvle: The commentCountCache basically should save the `asset.commentCount` from the last full query, which is then used to determine if new comments have arrive when the our count polling returns a higher commentCount.

client/coral-embed-stream/src/containers/Stream.js
```diff
@@ -112,7 +112,8 @@ class StreamContainer extends React.Component {
 
   componentDidMount() {
     if (this.props.previousTab) {
-      this.props.data.refetch();
+      this.props.data.refetch()
+        .then(() => this.props.setCommentCountCache(0));
```
alinhle: Hm. Why do you set it to `0`?  Does it work? I think it should be set to `asset.commentCount` from the new query result. Do you have access to it inside this callback?

alinhle: The `commentCountCache` basically should cache `asset.commentCount` from the last full query, which is then used to determine if new comments have arrived when the count polling returns a higher commentCount.

gobengo: @cvle wrt setting to 0... that was silly. Yes, it did 'work' insofar as it hid the NewCount, but probably would have caused other badness. I think I read too fast and thought the commentCountCache reflected the gap between what the client knows about and the server has. Silly me, and good catch. I've updated to set it to the new asset.commentCount from the refetch.

> The commentCountCache basically should cache asset.commentCount from the last full query,

This is the type of invariant I was looking for. If this is the case, that we always want to reset commentCountCache to asset.comments whenever EmbedQuery is successful, then what do you think about handling this in the reducer that manages commentCountCache?

Could do so by adding the following to the [stream reducer](https://github.com/coralproject/talk/blob/master/client/coral-embed-stream/src/reducers/stream.js#L42)

```javascript
case 'APOLLO_QUERY_RESULT':
  switch (action.operationName) {
    case 'EmbedQuery':
      const commentCount = action.result && action.result.data && action.result.data.asset && action.result.data.asset.commentCount
      if (typeof commentCount === 'number') {
        return {
          ...state,
          commentCountCache: commentCount,
        }
      }
      break;
    default:
      break;
  }
```

gobengo:
https://github.com/coralproject/talk/pull/596#discussion_r117311438

client/coral-embed-stream/src/containers/Stream.js
alinhle: Great idea @gobengo, I think this should work, then we can also remove the logic from here: https://github.com/coralproject/talk/blob/0dfec6e619b622f767a01e17d6038a37ad8bd0c3/client/coral-embed-stream/src/containers/Embed.js#L40

alinhle:
Sounds good, see https://github.com/coralproject/talk/pull/596#discussion_r117316256. Go for it if this is straightforward.

kgardnr:
Amazing @gobengo! 💯 Thank you!

kgardnr (approved):

PR #598, opened 2017-05-17
Edit Comment Window Length Setting

gobengo:
## What does this PR do?

Implements story [EDIT SETTINGS | As an admin, I want to set a timeframe for commenters to edit their comment](https://www.pivotaltracker.com/story/show/144459753)

## How do I test this PR?

* Visit /admin/configure as an admin
* Use left nav to go to "Moderation Settings" (we should really make this permalinkable)
* Scroll to bottom, see 
* Change the edit comment window
* Click "Save Changes" (you may have to scroll up)
* Visit comment stream, leave a comment
* Click to edit your comment.
* Ensure that the edit countdown timer is counting down from the number of seconds you just configured


services/comments.js
```diff
@@ -53,8 +53,10 @@ module.exports = class CommentsService {
 
     // Establish the edit window (if it exists) and add the condition to the
     // original query.
-    const lastEditableCommentCreatedAt = new Date((new Date()).getTime() - EDIT_WINDOW_MS);
+    let lastEditableCommentCreatedAt;
     if (!ignoreEditWindow) {
+      const editWindowMs = (await SettingsService.retrieve()).editCommentWindowLength;
```
wyattjoh: Could change to a restructuring syntax:

```
const {editCommentWindowLength: editWindowMs} = await SettingsService.retrieve();
```

client/coral-admin/src/containers/Configure/ModerationSettings.js
```diff
@@ -72,6 +78,27 @@ const ModerationSettings = ({settings, updateSettings, onChangeWordlist}) => {
         bannedWords={settings.wordlist.banned}
         suspectWords={settings.wordlist.suspect}
         onChangeWordlist={onChangeWordlist} />
+
+      {/* Edit Comment Timeframe */}
+      <Card className={`${styles.configSetting}`}>
```
alinhle: You can remove the backticks completely.

client/coral-embed-stream/src/graphql/index.js
```diff
@@ -200,7 +200,11 @@ const extension = {
       variables: {id, edit},
     }) => ({
       updateQueries: {
-        EmbedQuery: (previousData, {mutationResult: {data: {editComment: {comment: {status}}}}}) => {
+        EmbedQuery: (previousData, {mutationResult: {data: {editComment: {comment, errors}}}}) => {
+          if (errors && errors.length) {
```
alinhle: We're going to streamline error handling soon, that makes this check here unnecessary.
Can you add a todo to revisit this after streamlining error handling?
You can also remove this check for now.

gobengo: I added the todo comment.

I disagree that it's a good idea to remove the error checking *before* you do something that makes it unnecessary. If I did that, it would allow the UI to inaccurately reflect a failed ignoreUser and potentially confuse the end-user. When these lines are made unnecessary, then it makes sense to remove them, but not before.

kgardnr (approved):

PR #603, opened 2017-05-22
Coral plugin Auth

okbel:
## What does this PR do?
 - Abstracts Auth flow as a plugin `coral-plugin-auth`
 - Adds `embed`, `stream`, `login` slots

## Testing this PR
- Check that every auth path works as expected:
  - Login Flow
  - Sign Up Flow
  - Sign In Flow
  - Forgot Password
  - Change Username, etc.


kgardnr (approved):

client/coral-admin/src/actions/auth.js
```diff
@@ -52,7 +52,9 @@ const forgotPassowordFailure = () => ({
 
 export const requestPasswordReset = (email) => (dispatch) => {
   dispatch(forgotPassowordRequest(email));
-  return coralApi('/account/password/reset', {method: 'POST', body: {email}})
+  const redirectUri = location.href;
```
riley, line 55: I think the `location.href` would be the url of the iframe here.

plugins/coral-plugin-auth/client/components/SignInButton.js
```diff
@@ -0,0 +1,24 @@
...
+import {Button} from 'coral-ui';
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+import {showSignInDialog} from 'coral-framework/actions/auth';
+
+const SignInButton = ({loggedIn, showSignInDialog}) => (
+  <div>
+    {!loggedIn
+      ? <Button id="coralSignInButton" onClick={showSignInDialog} full>
+          Sign in to comment
```
riley, line 11: this should probably be translated while we're in here

plugins/coral-plugin-auth/client/components/SignUpContent.js
```diff
@@ -1,39 +1,21 @@
...
 
-  constructor (props) {
-    super(props);
-    this.successfulSignup = false;
+    this.state = {
+      successfulSignup: false
+    };
   }
 
-  static propTypes = {
```
riley, line 15: why did we delete all the propTypes?

jde (approved):

client/coral-admin/src/actions/auth.js
okbel: It’s fine. We need the host there and the popup it’s a permitted host.

plugins/coral-plugin-auth/client/components/SignUpContent.js
okbel: Restoring them

client/coral-admin/src/actions/auth.js
riley: 👍

PR #604, opened 2017-05-22
Support response errors in the graphql framework

alinhle:
## What does this PR do?
- Add support for response errors to the graphql framework
  - Prevent execution of query reducers, and mutation updaters when there are response errors.
  - Fail mutation promises with a custom error `ResponseErrors` that returns an error message and an array of `ResponseError` at `ResponseErrors.errors`.
  - `ResponseErrors` and `ResponseError` implementation:
```js
class ResponseErrors extends Error {
  constructor(errors) {
    super(`Response Errors ${JSON.stringify(errors)}`);
    this.errors = errors.map((e) => new ResponseError(e));
  }
}

class ResponseError {
  constructor(error) {
    Object.assign(this, error);
  }

  translate(...args) {
    return lang.t(`error.${this.translation_key}`, ...args);
  }
}
```

## How do I test this PR?

- Login to a suspended user.
- Perform an unauthorized action like respecting.
- See that the mutation is rolled back successfully and that an error is logged to the console.


client/coral-framework/hocs/withMutation.js
```diff
@@ -6,7 +6,25 @@ import flatten from 'lodash/flatten';
 import isEmpty from 'lodash/isEmpty';
 import {getMutationOptions, resolveFragments} from 'coral-framework/services/graphqlRegistry';
 import {store} from 'coral-framework/services/store';
-import {getDefinitionName} from '../utils';
+import {getDefinitionName, getResponseErrors} from '../utils';
+import t from 'coral-framework/services/i18n';
+
+class ResponseErrors extends Error {
+  constructor(errors) {
+    super(`Response Errors ${JSON.stringify(errors)}`);
```
riley, line 14: I think calling `JSON.stringify()` on an error just evaluates to `"{}"`

client/coral-framework/hocs/withMutation.js
```diff
@@ -6,7 +6,25 @@ import flatten from 'lodash/flatten';
...
+    this.errors = errors.map((e) => new ResponseError(e));
+  }
+}
+
+class ResponseError {
+  constructor(error) {
+    Object.assign(this, error);
```
riley, line 21: I've always wondered. Does our transpiler evaluate `Object.assign` into something that IE can use? `Object.assign` is not supported in any version.

Counterpoint: do we care? total IE usage is down to 3 or 4%

client/coral-framework/hocs/withMutation.js
```diff
@@ -6,7 +6,25 @@ import flatten from 'lodash/flatten';
...
+  }
+
+  translate(...args) {
+    return t(`error.${this.translation_key}`, ...args);
```
riley, line 25: what does this do if the `translation_key` is missing?

client/coral-framework/hocs/withQuery.js
```diff
@@ -1,7 +1,14 @@
 import * as React from 'react';
 import {graphql} from 'react-apollo';
 import {getQueryOptions, resolveFragments} from 'coral-framework/services/graphqlRegistry';
-import {getDefinitionName, separateDataAndRoot} from '../utils';
+import {getDefinitionName, separateDataAndRoot, getResponseErrors} from '../utils';
+
+const withSkipOnErrors = (reducer) => (prev, action, ...rest) => {
+  if (action.type === 'APOLLO_MUTATION_RESULT' && getResponseErrors(action.result)) {
```
riley, line 7: does apollo expose constants that we could import instead of using a string constant? probably not a big deal.

riley (changes requested):
just a couple of questions. otherwise looks good.

client/coral-framework/hocs/withMutation.js
alinhle: These are our custom errors returned from the graph, so you actually get a nice string out of that.

client/coral-framework/hocs/withMutation.js
alinhle: We use `babel-polyfill` which gives us a full es6 environment.

client/coral-framework/hocs/withQuery.js
alinhle: No it doesn't.

client/coral-framework/hocs/withMutation.js
alinhle: These are errors from our graph schema and all of them must provide a `translation_key`.

riley (approved):

PR #616, opened 2017-05-29
Use correct redirect url

alinhle:
## What does this PR do?
- All redirect urls were redirecting the user to an invalid location with an unusable application state.
- Fix redirect urls to redirect to the original article.
- Affected workflows: Forgot password, Resend verification email, and verify email.

image

## How do I test this PR?

- Register user with required email verification


jde (approved):
Excellent catch.

PR #619, opened 2017-05-30
allow people to be downgraded to regular commenter

riley:
## What does this PR do?

the list of acceptable roles did not include `""` which is just a regular commenter. You could be created as such, but not be downgraded to that.

## How do I test this PR?

- change a user that has `ADMINISTRATOR`, `MODERATOR` or `STAFF` to the `.` role in admin
- it should stick


alinhle:
This is already fixed in master after the GraphQL refactor.

PR #621, opened 2017-05-31
Update README with Trust info

kgardnr:
Introducing the first features of Trust to Talk!


losowsky (changes requested):
"to automatically do special things based on intelligence built into the system."

change to "to automate certain actions based on previous user behavior."

gabelula (approved):

PR #624, opened 2017-05-31
Look for tokens passed through embed and handle them

jde:
## What does this PR do?

Allows auth tokens to be passed in through the embed string and used in all requests.

## How do I test this PR?

- Generate a token
- Pass it into the embed via the authToken param
- If the server accepts the token, you should be logged in like magic

### DIFF
--- a/client/coral-embed-stream/src/actions/config.js
+++ b/client/coral-embed-stream/src/actions/config.js
@@ -4,13 +4,13 @@ import {handleAuthToken} from 'coral-framework/actions/auth';
 export const addExternalConfig = (config) => {
 
   /**
-   * If an authToken is present and not falsey, handle it.
+   * If an auth_token is present and not falsey, handle it.
    *
-   * This allows the host page to pass in an authToken in order to control
+   * This allows the host page to pass in an auth_token in order to control
    * the Talk session.
    */
-  if (config.authToken) {
-    handleAuthToken(config.authToken);
+  if (config.auth_token) {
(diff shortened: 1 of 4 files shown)


views/article.ejs
```diff
@@ -29,6 +29,7 @@
         talk: '/',
         asset_url: '<%= asset_url ? asset_url : '' %>',
         asset_id: '<%= asset_id ? asset_id : '' %>',
+        authToken: '',
```
alinhle: Should this be camelCase or snake_case?

client/coral-embed-stream/src/actions/config.js
```diff
@@ -1,6 +1,20 @@
...
+export const addExternalConfig = (config) => {
+
+  /**
+   * If an authToken is present and not falsey, handle it.
+   *
+   * This allows the host page to pass in an authToken in order to control
+   * the Talk session.
+   */
+  if (config.authToken) {
+    handleAuthToken(config.authToken);
```
alinhle: To conform with the rest of the application, this should be dispatched.

alinhle (changes requested):
I think we should avoid persisting the third party token. This token is generated and passed by the host site and usually it is already persistent somewhere, where the host site has access. This avoids having the token persisted at several places which is good for security. I'm  open for discussion.

jde:
Closing in favor of approach here: https://github.com/coralproject/talk/pull/638

PR #628, opened 2017-06-01
Fix translation bug in CommentCount

alinhle:
## What does this PR do?
- Fix translation bug on `CommentCount` when having only 1 comment.


kgardnr (approved):
Yes! I'm so glad you changed it `comment_singular`, much clearer.

PR #629, opened 2017-06-01
Increase/Decrease totalCommentCount after mutation

alinhle:
## What does this PR do?
- Increase `totalCommentCount` when posting a comment
- Decrease `totalCommentCount` when comment is removed from stream (e.g. due to an edit).

## How do I test this PR?

- Post a comment -> See that the comment counter on the tab increases
- Edit a comment with premod enabled -> See that the comment counter on the tab decreases


kgardnr (approved):
Yesss! And we are also now calculating comments + replies = total comment count. This was a request from users. Two birds, one stone.

alinhle:
We say "two flies with one swat" in Germany :-D

PR #637, opened 2017-06-02
I18n emails

gabelula:
Translate emails. I will check with @cvle on a few things.


routes/api/users/index.js
```diff
@@ -109,12 +109,12 @@ const SendEmailConfirmation = (app, userID, email, referer) => UsersService
...
     return mailer.sendSimple({
       template: 'email-confirm',              // needed to know which template to render!
-      locals: {                                     // specifies the template locals.
+      locals: {                               // specifies the template locals.
         token,
         rootURL: ROOT_URL,
         email
       },
-      subject: 'Email Confirmation',
+      subject: 'email.confirm.subject',
```
alinhle: This should be `t('email.confirm.subject')`

services/i18n.js
```diff
@@ -0,0 +1,94 @@
...
+// default language
+let defaultLanguage = 'en';
+let language = defaultLanguage;
+const translations = Object.assign(en, es);
+
+/**
+* guess language setting based on http headers
+*/
+const guessLanguage = (request) => {
+
```
cvle: Rather than developing all this stuff by ourselves. I suggest using an established library like this one https://github.com/jshttp/accepts. This library is also utilised by `express` and  `koa`.

services/i18n.js
```diff
@@ -0,0 +1,94 @@
...
+// default language
+let defaultLanguage = 'en';
+let language = defaultLanguage;
+const translations = Object.assign(en, es);
+
+/**
+* guess language setting based on http headers
+*/
+const guessLanguage = (request) => {
```
alinhle: Rather than developing all this stuff by ourselves. I suggest using an established library like this one https://github.com/jshttp/accepts. This library is also utilised by express and koa.

services/mailer.js
```diff
@@ -87,7 +89,9 @@ const mailer = module.exports = {
     }
 
     // Prefix the subject with `[Talk]`.
-    subject = `[Talk] ${subject}`;
+    subject = `[Talk] {t(${subject})}`;
```
alinhle: We won't support a dynamic use of `t` in the future. The subject should be translated before passing it to this method. See my comment above.

services/i18n.js
gabelula: ok. Changing it.

gabelula:
Ready for final review and merge.

kgardnr:
Hey @gabelula should we translate all the /es/ templates into Spanish as part of this work?

gabelula:
What you mean?

kgardnr (approved):

services/i18n.js
```diff
@@ -0,0 +1,53 @@
...
+ * Exposes a service object to allow translations.
+ * @type {Object}
+ */
+const i18n = {
+
+  /**
+   * Create the new Task kue.
+   */
+  init(req) {
+    const lang = accepts(req).language(languages);
```
cvle: This doesn't choose the client's preferred language, see: https://github.com/jshttp/accepts#languagelanguages. `accepts(req).languages()` should be used instead.

services/i18n.js
```diff
@@ -0,0 +1,53 @@
...
+
+    if (has(translations[language], key)) {
+
+      let translation = get(translations[language], key);
+
+      // replace any {n} with the arguments passed to this method
+      replacements.forEach((str, i) => {
+        translation = translation.replace(new RegExp(`\\{${i}\\}`, 'g'), str);
+      });
+
```
alinhle, line 44: It would be great if wouldn't duplicate this logic on the server and on the client, but instead share the common part.

PR #638, opened 2017-06-02
Use config auth_token if present on http headers

jde:
# What does this PR do?

Allows auth tokens to be passed in through the embed string and used in all requests.

## How do I test this PR?

- Generate a token
- Pass it into the embed via the authToken param
- If the server accepts the token, you should be logged in like magic


alinhle:
This is good to review and test!

client/coral-framework/helpers/request.js
```diff
@@ -14,12 +38,10 @@ const buildOptions = (inputOptions = {}) => {
...
 
-  if (!bowser.safari && !bowser.ios) {
-    let authorization = Storage.getItem('token');
-
-    if (authorization) {
-      options.headers.Authorization = `Bearer ${authorization}`;
-    }
+  // Apply authToken header
+  let authToken = getAuthToken();
+  if (authToken) {
```
wyattjoh: Should change:

```diff
- if (authToken) {
+ if (authToken !== null) {
```

wyattjoh (approved):

PR #641, opened 2017-06-05
Use same lifespan for cookies as for tokens

alinhle:
## What does this PR do?
- Uses the same lifespan for safari login cookies as for jwt tokens

## How do I test this PR?

- Login with Safari
- Inspect cookie in `storage` tab and check the expiry time to match the JWT token expiry time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,6 @@
     "node-emoji": "^1.5.1",
     "node-fetch": "^1.6.3",
     "nodemailer": "^2.6.4",
-    "parse-duration": "^0.1.1",
     "passport": "^0.3.2",
     "passport-jwt": "^2.2.1",
     "passport-local": "^1.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -92,6 +92,7 @@
     "minimist": "^1.2.0",
     "mongoose": "^4.9.8",
     "morgan": "^1.8.1",
+    "ms": "^2.0.0",
```
wyattjoh, line 95: We already have the https://www.npmjs.com/package/parse-duration package included, no need to import this new package.

wyattjoh (changes requested):

package.json
alinhle: This is the same package used by the JWT token generator. It ensures that the same logic is applied when parsing the duration. There are some subtle parsing differences with `parse-duration` and the `ms`.

alinhle: We can either preparse the duration using `parse-duration` before passing it to the JWT Token Generator. Or port our codebase to `ms`.

alinhle: Which one do you prefer?

wyattjoh (approved):

PR #644, opened 2017-06-05
Add supplied classes to tabbar/tabs

StephanieDClark:
## What does this PR do?
Support additional classes added to tabbar, tab.  

 - This will allow classes to be applied in stylesheets at a detailed level; currently, we can only apply classes to the dynamic classnames. 

 - this will also support reuse of TabBar componenet.

## How do I test this PR?
See that classes are supplied on tabbar and tab components 

- click a button
- view the cat
- see the cat meow


riley:
I made some changes on a corresponding branch https://github.com/coralproject/talk/pull/648 so I'll go ahead and close this one. More class hooks to come!

PR #649, opened 2017-06-06
add classnames to RespectButton component

ebirving:
## What does this PR do?
Adds class names to the respect button component, mirroring the structure and conventions used in the like button.

## How do I test this PR?
- In the embed stream, with `coral-plugin-respect` plugin enabled, inspect the Respect button
- Every html element should have a classname that begins with `coral-plugin-respect`


riley (approved):
looks great! 💯

PR #651, opened 2017-06-06
add semantic classname to active tab

ebirving:
## What does this PR do?
Adds a non-hashed class to active `Tab` components to allow styling without wildcard selectors

## How do I test this PR?

- Examine the active tab in the embed stream
- It should have class `active-tab` in addition to other class names


client/coral-ui/components/Tab.js
```diff
@@ -4,7 +4,7 @@ import styles from './Tab.css';
 export default ({children, tabId, active, onTabClick, cStyle = 'base', ...props}) => (
   <li
     key={tabId}
-    className={`${active ? styles[`${cStyle}--active`] : ''} tab ${props.className}`}
+    className={`${active ? `${styles[`${cStyle}--active`]} tab-active` : ''} tab ${props.className}`}
```
riley: The only thing that I would suggest is that we should make it slightly more specific with `coral-tab-active` in case there are other tabs at some point. If we add that, this is good to merge 👍

riley (changes requested):

ebirving:
@riley :+1:

riley (approved):
great! 💯

PR #652, opened 2017-06-06
Css override classes + tab bugfix

riley:
## What does this PR do?

Adds classes for UserBox area and fixes a bug where the active tab was not visually displaying as active.

## How do I test this PR?
- the user box now has a class `talk-stream-auth-userbox`
- when you click between tabs in the stream, there is visual feedback that you've changed tabs


kgardnr (approved):
LGTM!

PR #654, opened 2017-06-06
Off Topic Features

okbel:
# Off Topic Features

This PR adds:
- Adds `coral-plugin-viewing-options`
- Actions and Reducers for Comment Component
- Add/Remove classNames based on conditions. i.e tags
- Fixes issues with plugins
- Adds architecture `container-component` for plugins
- Uses `isReply` and `depth` for high-level comments off-topic
- A new slot on the stream called `streamBox` on the top of the comment stream.
- Adds a stable architecture: `alpha` `beta` and current API for plugin devs. (More docs will come)
- Adds camelCase for plugin reducers 🎉 

## Coral Plugin Viewings Options
 `coral-plugin-viewing-options` is a plugin that enables a dropdown on the top of the comment stream to embed filters from other plugins.

This plugin adds a new slot called `viewingOptions`.

Now `coral-plugin-offtopic` depends on it if you want to filter the stream.


alinhle (changes requested):
Just skipped over it and noticed that you connected a few components to the store. We should follow the best practises and only connect stores to containers. Also to make things simple it is not always necessary to create a new container but instead use an existing one like the `Stream` container and pass the props down. 

_Note: Currently you can't connect the `Comment` Component to the store because it would break some stuff in the live update feature. You can just pass it down using the `Stream` container as mentioned above, which would also be a lot more performant_

alinhle:
The original intention of the `stream` reducer was to include all state affecting the stream including its comments. I suggest reusing the `stream` reducer instead of creating a new one for comments.

cvle:
I suggest not connecting the store to the comment because that will have a noticeable performance impact.

okbel:
Great! Well, I added all the changes. 

The idea behind not using the containers-component structure in one of the plugins was to ease up the learning curve while building plugins. But the `off-topic` plugin got a bit complex due to it's features and it makes sense to add a containers-components structure. ✅ 

I also avoided reconnecting the store again for each comment, and now we pass store info from it's parent. The Stream Component. ✅ 

I think that the classes for the comment should remain in the comment Reducer since this is info strictly related to the comments component and not to the stream itself.

client/coral-embed-stream/src/containers/Stream.js
```diff
@@ -323,6 +323,7 @@ const mapStateToProps = (state) => ({
   assetUrl: state.stream.assetUrl,
   activeTab: state.embed.activeTab,
   previousTab: state.embed.previousTab,
+  classNames: state.comment.classNames
```
alinhle: I suggest using a name like `commentClassNames` which is more clear on the purpose of this property, to not confuse with classnames on the `Stream` container itself.

client/coral-embed-stream/src/components/Stream.js
```diff
@@ -108,11 +108,17 @@ class Stream extends React.Component {
...
+
+        {loggedIn && (
           <ModerationLink
             assetId={asset.id}
             isAdmin={can(user, 'MODERATE_COMMENTS')}
-          />}
+          />
+        )}
+
+        <div className="streamBox">
```
alinhle: Any reason why you use a global css class name? If this should be part of our CSS override API than consult Riley and David for the current naming schema that they've been working on lately. Otherwise use css modules.

client/coral-embed-stream/src/components/Stream.js
```diff
@@ -140,7 +146,7 @@ class Stream extends React.Component {
               maxCharCount={asset.settings.charCount}
               editComment={this.props.editComment}
             />
-          : <div>
+          : <div className="commentStreamContainer">
```
alinhle: See above

client/coral-embed-stream/style/default.css
```diff
@@ -485,3 +494,7 @@ button.comment__action-button[disabled],
     visibility: visible;
   }
 }
+
+.offTopicComment {
```
alinhle: This should be part of the `off-topic` plugin not the `embed-stream`.

client/coral-framework/helpers/plugins.js
```diff
@@ -11,7 +11,7 @@ import {loadTranslations} from 'coral-framework/services/i18n';
 export const pluginReducers = merge(
   ...plugins
     .filter((o) => o.module.reducer)
-    .map((o) => ({...o.module.reducer}))
+    .map((o) => ({[o.module.reducer.name] : o.module.reducer}))
```
alinhle: We should change this to include plugin reducers only into its own namespace using its own name as a key.

plugins/coral-plugin-offtopic/client/components/OffTopicFilter.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import styles from './styles.css';
+
+export default class OffTopicFilter extends React.Component {
+
+  tag = 'OFF_TOPIC';
+  className = 'offTopicComment';
```
alinhle: If this is supposed to be a global css classname that can be styled by users, consult david or riley for current naming convetions.

plugins/coral-plugin-offtopic/client/containers/OffTopicCheckbox.js
```diff
@@ -0,0 +1,11 @@
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+import {addTag, removeTag} from 'coral-plugin-commentbox/actions';
```
alinhle: We need to provide this actions through the `coral-plugin-api` in a folder called `alpha` or `beta` if we are more confident that it doesn't change as discussed in a previous arch meeting.

plugins/coral-plugin-offtopic/client/containers/OffTopicFilter.js
```diff
@@ -0,0 +1,27 @@
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+import {toggleCheckbox} from '../actions';
+import {
+  addClassName,
+  removeClassName
+} from 'coral-embed-stream/src/actions/comment';
```
alinhle: We need to provide this actions through the `coral-plugin-api` in a folder called `alpha` or `beta` if we are more confident that it doesn't change as discussed in a previous arch meeting.

plugins/coral-plugin-offtopic/client/containers/OffTopicFilter.js
```diff
@@ -0,0 +1,27 @@
...
+import OffTopicFilter from '../components/OffTopicFilter';
+import {closeViewingOptions} from 'coral-embed-stream/src/actions/stream';
```
alinhle: We need to provide this actions through the `coral-plugin-api` in a folder called `alpha` or `beta` if we are more confident that it doesn't change as discussed in a previous arch meeting.

plugins/coral-plugin-viewing-options/client/components/ViewingOptions.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import cn from 'classnames';
+import {Icon} from 'coral-ui';
+import styles from './ViewingOptions.css';
+import Slot from 'coral-framework/components/Slot';
```
alinhle: We need to provide this through the `coral-plugin-api` in a folder called `alpha` or `beta` if we are more confident that it doesn't change.

plugins/coral-plugin-viewing-options/client/containers/ViewingOptions.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+import ViewingOptions from '../components/ViewingOptions';
+import {openViewingOptions, closeViewingOptions} from 'coral-embed-stream/src/actions/stream';
```
alinhle: We need to provide this actions through the `coral-plugin-api` in a folder called `alpha` or `beta` if we are more confident that it doesn't change.

alinhle:
There is something wrong with the mutations updators. The query breaks and reloads when I post a comment if off topics on.

client/coral-framework/helpers/plugins.js
alinhle: sth like: `.map((o) => ({[o.plugin] : o.module.reducer}))`

plugins/coral-plugin-offtopic/client/containers/OffTopicCheckbox.js
```diff
@@ -0,0 +1,11 @@
...
+import OffTopicCheckbox from '../components/OffTopicCheckbox';
+
+const mapStateToProps = ({commentBox}) => ({commentBox});
```
alinhle: We should use `selectors` that we export inside our `coral-plugin-api` to decouple the internal redux state from plugins.

plugins/coral-plugin-viewing-options/client/containers/ViewingOptions.js
```diff
@@ -0,0 +1,12 @@
...
+
+const mapStateToProps = ({stream}) => ({open: stream.viewingOption.open});
```
alinhle: We should use `selectors` that we export inside our `coral-plugin-api` to decouple the internal redux state from plugins.

plugins/coral-plugin-offtopic/client/constants.js
```diff
@@ -0,0 +1 @@
+export const TOGGLE_CHECKBOX = 'TOGGLE_CHECKBOX';
```
alinhle: These constant value needs a global prefix so it doesn't collide with other plugins and the core. This should be something like `CORAL_OFFTOPIC_TOOGLE_CHECKBOX`.

client/coral-embed-stream/src/containers/Stream.js
okbel: Ok. SGTM

alinhle:
Another thing I found was that linting needs to be enabled for `plugins/coral-plugin-viewing-options`

client/coral-embed-stream/src/components/Stream.js
okbel: This is a key place and developers want to style that portion of the Stream.

client/coral-embed-stream/src/components/Stream.js
okbel: Same

client/coral-embed-stream/style/default.css
okbel: 👍

client/coral-embed-stream/src/components/Stream.js
alinhle: Than it needs a sync on the current naming schema with riley + david.

client/coral-embed-stream/src/components/Stream.js
alinhle: See above.

client/coral-framework/helpers/plugins.js
okbel: Hmm. I did that at the beginning but I don't think it's right. First, I think we should leave the name of the function as is, just like we handle our reducers in the rest of the app. And second, what happen if I want to use more reducers?

plugins/coral-plugin-offtopic/client/components/OffTopicFilter.js
okbel: Yeah. Well this is a specific request where people using this plugin would like to add a new className. That's one of the features of this PR.

client/coral-framework/helpers/plugins.js
alinhle: You can always nest reducers. We don't want to run into the situation were plugins accidentally overwrites another reducer.

plugins/coral-plugin-offtopic/client/components/OffTopicFilter.js
alinhle: Yes, just need a sync with riley + david for the current naming scheme. They've been working on that.

plugins/coral-plugin-offtopic/client/constants.js
alinhle: the variable name can stay the same, but the value needs to be unique.

plugins/coral-plugin-offtopic/client/containers/OffTopicCheckbox.js
alinhle: This would ensure future evolution of talk, where we want to have multiple embeds with connected store, and some store restructuring will be required.

plugins/coral-plugin-viewing-options/client/containers/ViewingOptions.js
alinhle: This would ensure future evolution of talk, where we want to have multiple embeds with connected store, and some store restructuring will be required.

client/coral-embed-stream/src/components/Stream.js
okbel: @jde @riley

client/coral-embed-stream/src/components/Stream.js
okbel: @jde @riley

client/coral-framework/components/Slot.js
```diff
@@ -4,9 +4,9 @@ import styles from './Slot.css';
 import {connect} from 'react-redux';
 import {getSlotElements} from 'coral-framework/helpers/plugins';
 
-function Slot ({fill, inline = false, plugin_config: config, ...rest}) {
+function Slot ({fill, inline = false, right = false, plugin_config: config, ...rest}) {
```
alinhle: Not sure if we should include all these styling into the Slot itself. If we continue this pattern someday we would have `inline, right, left, top, down, center, fixed, absolute`. Maybe just allow the `style` and `className` property to add the styling to this component from a parent instead of doing it here?

plugins/coral-plugin-offtopic/client/components/OffTopicFilter.js
okbel: @jde  @riley

riley: for global classnames, we've been doing something like `coral-plugin-off-topic-comment`. Trying to keep it simple.

client/coral-embed-stream/src/components/Stream.js
riley: we're going with something more along the lines of `talk-stream-wrapper-box`. Look to examples here https://github.com/coralproject/talk/blob/master/client/coral-embed-stream/src/components/Embed.js#L45

plugins/coral-plugin-offtopic/client/components/OffTopicFilter.js
okbel: 👍

client/coral-embed-stream/src/components/Stream.js
okbel: 👍

okbel:
@cvle Ready to be checked :)

alinhle (approved):
Approving all except my commits as usual.

kgardnr (approved):

PR #658, opened 2017-06-07
Pass onAuthChange callback into embed

jde:
## What does this PR do?

Allows the host page to provide a callback that fires on the parent page when the Talk auth state changes. 

## How do I test this PR?

- embed a talk stream with an onAuthChange callback:
```
<div id="coral_talk_8938213609030934"></div>
<script src="http://localhost:5000/embed.js" async onload="
  Coral.Talk.render(document.getElementById('coral_talk_8938213609030934'), {
    talk: 'http://localhost:5000',
    onAuthChange: function(auth) {
      console.log(auth);
    }
  });
"></script>
```
- log in through Talk
- see the auth state logged in the console from the parent page

### DIFF
--- a/client/coral-framework/actions/auth.js
+++ b/client/coral-framework/actions/auth.js
@@ -285,6 +285,7 @@ export const logout = () => (dispatch) => {
     resetWebsocket();
 
     dispatch({type: actions.LOGOUT});
+    pym.sendMessage('coral-logout');
   });
 };
 
@@ -305,7 +306,6 @@ export const checkLogin = () => (dispatch) => {
   dispatch(checkLoginRequest());
   coralApi('/auth')
     .then((result) => {
-      pym.sendMessage('coral-check-login', JSON.stringify(result));
 
       if (!result.user) {
         if (!bowser.safari && !bowser.ios) {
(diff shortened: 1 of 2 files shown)


client/coral-framework/actions/auth.js
```diff
@@ -315,6 +316,9 @@ export const checkLogin = () => (dispatch) => {
     .catch((error) => {
       console.error(error);
       dispatch(checkLoginFailure(`${error.translation_key}`));
+    })
+    .then(() => {
+      pym.sendMessage('checkLogin', JSON.stringify(getState().auth));
```
alinhle: Why not passing the result of the rest call? The redux state of `auth` currently also contains state of the frontend interface related to auth, which I don't think we should expose.

alinhle:
What’s the use case for this?

meddy:
Hello @cvle , the use case is dynamically showing content based on the auth state of the embed app.

jde:
Hey @meddy, sorry for the delay. This is good to go.

Before we can merge your branch into master you will need to update your branch with our master. Please let us know when you are doing this so we can coordinate accepting the PR.

Thanks!

alinhle (approved):

alinhle:
Awesome @meddy, thanks for your contribution!

PR #669, opened 2017-06-08
Subscription fixes

wyattjoh:
## What does this PR do?

Changes the way that subscriptions receive the auth state.


wyattjoh:
Please use the `coral-framework/services/client` `resetWebsocket` method to reset the web socket when the auth state changes (the token is different or for safari, we have confirmed a new user).

alinhle (approved):
This is really awesome!

PR #674, opened 2017-06-12
Live reactions

alinhle:
## What does this PR do?
- Implement live subscriptions
- Reconnect web socket without delay when auth data changes.
- Move `withReaction` into `plugin-api/beta/client/hocs`.
- Encapsulate server config into `getReactionConfig` in `plugin-api/beta/server`.
- Upgrade `graphql-subscriptions`, `subscriptions-transport-ws` to latest.
- Implement all reactions (`love`, `respect`, and `like`) using the `plugin-api`.

I had to find a workaround to reconnect a websocket after a login/logout to supply the new auth data, see: https://github.com/apollographql/subscriptions-transport-ws/issues/171.

You might encounter warnings on the server console about invalid auth tokens. The reason for this is that the first subscription never actually gets unsubscribed until a fix for this gets released: https://github.com/apollographql/subscriptions-transport-ws/issues/180


alinhle:
I changed the prop `alreadyReacted` in the `withReaction` hoc to a simple boolean. I couldn't imagine any use case in which this should be a callback/function. The hoc shouldn't require any external variables passed in as arguments to determine `alreadyReacted`.

okbel:
Wooo 🎉 Amazing. 

PS: Please, if we need to extend the docs let's update them in another PR.

okbel (approved):

PR #675, opened 2017-06-12
Highlighting Links and New plugin: coral-plugin-comment-content

okbel:
# Highlighting Links and New plugin: coral-plugin-comment-content

## What does this PR do?
This PR adds `coral-plugin-comment-content`.

- Enables full customization of the comment body/content.
- Enables highlighting for links in comments.
- Adds default components for Slots
- Replaces coral-ui path, by `plugin-api/beta/client/components/ui`
- Switch reaction plugins to components/ui
- Only create Slot elements once
- Support Class Components as defaultComponent

## How to test this PR?

- Add `coral-plugin-comment-content` to you plugins.json within `client`
- Post a comment with a link on it.
- Click it
- Check that everything is working as expected.

_Note: We will be migrating all the `coral-plugin-*` in the `client` folder to plugins._


alinhle:
Interesting, I just realised that what we discussed last time doesn't work in React, which was using Components in a nested way to allow multiple plugins altering the body. We should have a quick talk about this with @jde to see if this is enough to satisfy the requirements.

alinhle (changes requested):
I found several cases where link detection failed:
image.

I suggest rather than rolling our own solution here, we should use a tiny library like ~https://github.com/alexcorvi/anchorme.js~ to do it for us in combination with ~`dangerouslySetInnerHTML`~.

alinhle:
Just realised that `dangerouslySetInnerHTML` is a very bad idea security wise... we can use a solution like https://github.com/tasti/react-linkify that uses a tokenizer from markdown-it (https://github.com/markdown-it/linkify-it) to do it.

okbel:
ready to test @cvle :)

plugins/coral-plugin-comment-content/client/components/CommentContent.js
```diff
@@ -0,0 +1,24 @@
...
+const name = 'coral-plugin-comment-content';
+
+const CommentContent = ({comment}) => {
+  const textbreaks = comment.body.split('\n');
+  return <div className={`${name}-text`}>
+    {
+      textbreaks.map((line, i) => {
+        return (
+          <span key={i} className={`${name}-line`}>
+            <Linkify>
```
alinhle: Missing something like `<Linkify properties={{target: '_parent'}} />`, so the links work inside the iframe.

alinhle:
Awesome! We'll talk with @jde if this satisfies the requirements, other than that the code looks great!

jde:
When I run the server without `coral-plugin-comment-content`, however, the body of the comments isn't rendered:

screenshot 2017-06-14 10 39 01

If no plugin is present for this slot the content should be rendered as is from state.

alinhle (approved):
I added a few small changes:
- Switch reaction plugins to components/ui
- Only create Slot elements once
- Support Class Components as defaultComponent

PR #678, opened 2017-06-13
Robust live reactions

alinhle:
## What does this PR do?
- Prevent sending a reaction mutation, when another one is already running.
- Only publish live updates when a reaction was actually created or deleted.

## How do I test this PR?

- Speed clicking on a reaction should only cause one mutation.
- Posting a reaction to a comment, that was already reacted on should not update the counter for other users.
- Removing a reaction from a comment, that was already removed should not update the counter for other users.

### DIFF
--- a/errors.js
+++ b/errors.js
@@ -104,6 +104,20 @@ class ErrAuthentication extends APIError {
   }
 }
 
+/**
+ * ErrAlreadyExists is returned when an attempt to create a resource failed due to an existing one.
+ */
+class ErrAlreadyExists extends APIError {
+  constructor(existing = null) {
+    super('authentication error occured', {
+      translation_key: 'ALREADY_EXISTS',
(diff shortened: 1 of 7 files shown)


errors.js
```diff
@@ -104,6 +104,20 @@ class ErrAuthentication extends APIError {
...
 }
 
+/**
+ * ErrAlreadyExists is returned when an attempt to create a resource failed due to an existing one.
+ */
+class ErrAlreadyExists extends APIError {
+  constructor(existing = null) {
+    super('authentication error occured', {
+      translation_key: 'ALREADY_EXISTS',
+      status: 400
```
wyattjoh: This is really better as a 409 (Conflict).

alinhle: True

jde (approved):

PR #679, opened 2017-06-13
style updates to flagged user names queue and updated translations

theysamtheirs:
## What does this PR do?

Made style updates for "Flagged Usernames" queue on Community page.

Note:  There are two outstanding issues here that are outside of my skillz:
- Update the "ban user button" to be the "actions" (aka suspend/ban user) button (cc @cvle)
- Usernames that are flagged as "Offensive" are not returning the string to show "offensive" as the reason (was able to replicate on staging as well).  (cc @gabelula) 

## How do I test this PR?

- click a button
- view the cat
- see the cat meow

### DIFF
--- a/client/coral-admin/src/routes/Community/components/Community.css
+++ b/client/coral-admin/src/routes/Community/components/Community.css
@@ -175,7 +175,7 @@
      right: 0;
      height: 100%;
      top: 0;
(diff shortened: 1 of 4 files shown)


jde (approved):
Looking sharp!

PR #680, opened 2017-06-14
Search stories

okbel:
# Search stories
Enables Search Stories by clicking the title of the story in the Moderate Section on the Admin Side

## What does this PR do?
- Search Stories
  - Add “Search Results”
  - Handle Empty States
  - Remove “All Streams” “Select Streams”
  - Search on Enter


okbel:
@IAmSamHankins Ready to be tested, Sama!

kgardnr (approved):
I'm good if @IAmSamHankins is too!! 💯

client/coral-admin/src/routes/Moderation/containers/Moderation.js
```diff
@@ -208,7 +212,7 @@ const withModQueueQuery = withQuery(gql`
     }) {
       ...CoralAdmin_Moderation_CommentConnection
     }
-    assets: assets {
+    assets: assets(query: {}) {
```
alinhle: This breaks current moderation on the majority of assets, because it assumes the whole list of assets. 
The proposed changes would limit the results to only 10 assets.

image

graph/loaders/assets.js
```diff
@@ -20,6 +20,17 @@ const genAssetsByID = (context, ids) => AssetModel.find({
 }).then(util.singleJoinBy(ids, 'id'));
 
 /**
+ * [getAssetsByQuery description]
+ * @param  {Object} context  the context of the request
+ * @param  {String} value    text string to search agains the documents
+ * @param  {Number} limit    limit the number of results
+ * @return {Promise}         resolves the assets
+ */
+const getAssetsByQuery = (context, value, limit) => {
```
alinhle: I suggest putting the query variables into an object, so that it reflects what was defined in the schema and would also be more in line with code that has already been written.

e.g. `const getAssetsByQuery = (context, query) => {`

graph/loaders/assets.js
```diff
@@ -65,6 +76,7 @@ module.exports = (context) => ({
     // this operation create a new asset if one isn't found.
     getByURL: (url) => findOrCreateAssetByURL(context, url),
 
+    search: (value, limit) => getAssetsByQuery(context, value, limit),
```
alinhle: Same as above, instead of putting the query variables into separate arguments, just pass a query object.

graph/resolvers/root_query.js
```diff
@@ -6,12 +6,14 @@ const {
...
-  assets(_, args, {loaders: {Assets}, user}) {
+  assets(_, {query}, {loaders: {Assets}, user}) {
     if (user == null || !user.can(SEARCH_ASSETS)) {
       return null;
     }
 
-    return Assets.getAll.load();
+    const {value = '', limit} = query;
+
+    return Assets.search(value, limit);
```
alinhle: It'll be simpler to just pass the whole query object e.g. `Assets.search(query)`

graph/typeDefs.graphql
```diff
@@ -626,7 +634,7 @@ type RootQuery {
   comment(id: ID!): Comment
 
   # All assets. Requires the `ADMIN` role.
-  assets: [Asset]
+  assets(query: AssetsQuery!): [Asset]
```
alinhle: As we currently allow querying all assets without any query parameters, I think `query` should be optional instead.

alinhle:
_Removed this as it referred to an outdated commit, oops sorry_

client/coral-admin/src/routes/Moderation/containers/Moderation.js
```diff
@@ -212,7 +212,7 @@ const withModQueueQuery = withQuery(gql`
     }) {
       ...CoralAdmin_Moderation_CommentConnection
     }
-    assets: assets(query: {}) {
+    assets: assets(query: {limit: 10}) {
```
alinhle: The ModerationQueue expects all assets, limiting it to 10 will break it.

alinhle: It would be a good idea though to refactor the ModerationQueue to not need the whole asset array.

alinhle: Something like this would do for now: `assets: assets()`

okbel: That would throw a syntax error.

alinhle: Does it? Even when all arguments are defined optional? There must be some easier way to specify it :thinking: than `assets: assets(query: null)`.

okbel: screen shot 2017-06-16 at 3 44 53 pm

That one too :(

client/coral-admin/src/routes/Moderation/containers/Moderation.js
alinhle: @okbel This also breaks, because the AssetsQuery type has a default limit of 10 ;-) 
I played around in a GraphQL playground and I think you can just do `assets {`.

okbel: `assets {`?

alinhle (changes requested):
See above review, most importantly making sure that the ModQueue will not break with these changes.

client/coral-admin/src/routes/Moderation/containers/Moderation.js
alinhle: Yeah either `assets {` or `assets: assets(query: null)` or `assets: assets(query: {limit: 0})` are possible choices in the order of preference :-)

alinhle (approved):

Issue #702, opened 2017-06-22
Selecting "Moderate →" shows no comments for that story and loses the tab


ALawliet:
Not sure if intended, but this seemed off to me.

### Steps to reproduce behavior
Start with comments that have not been moderated (neither Approved nor Rejected).

**Path 1 ✅**
- open search dropdown
- select story "Coral Talk"
- notice it still shows those unmoderated comments, and the tab underline indicating we're at the "All" tab (expected behavior)

(start over by opening the search dropdown and clicking "Moderate comments on All Stories")

**Path 2 ❌**
- select "Moderate →" next to "Story: Coral Talk" of the comment feed
- notice it shows no comments even though they have not been moderated yet and no tab is selected
- click "All" tab to confirm that indeed no comments have been moderated

### Expected behavior
I would expect these two paths to moderate the comments for a story to be the same, still showing the comments and having "All" tab.

### Actual behavior
**Path 2** seems to lose the comments and the tab. I certainly do not deserve a ☕️  because I haven't actually moderated any comments yet. :woman_shrugging: 

![buggu](https://user-images.githubusercontent.com/10943083/27444798-7809b3fe-5745-11e7-8952-cf99a391c273.gif)

ALawliet:
Fixed in: https://github.com/coralproject/talk/pull/703 https://github.com/coralproject/talk/pull/707

PR #703, opened 2017-06-22
Fix Moderate → link by setting to "All" for that story

ALawliet:
## What does this PR do?
This looked like a bug to me: https://github.com/coralproject/talk/issues/702, so here's the simple fix if that's the case.

## How do I test this PR?
Click Moderate → link for a story and notice it no longer blanks out the comments and loses the tab. Notice Path 1 and Path 2 from https://github.com/coralproject/talk/issues/702 now do the same thing.

The issue just seems to have been a bad link.


alinhle (approved):
Great catch, thanks!

PR #704, opened 2017-06-22
Added support for keep alive websocket frames

wyattjoh:
## What does this PR do?

Adds a configurable web socket timeout option that will send `{"type":"ka"}` over the web socket (default 30s) to prevent load balancers from terminating the connection.

## How do I test this PR?

1. Open the page in Chrome
2. Look under Network -> WS -> /live
3. Select the "Frames" tab
4. Wait 30+ seconds, notice at least 1 `{"type":"ka"}` frame.


okbel (approved):

okbel:
💯 ❤️

PR #707, opened 2017-06-23
If not specified redirect to 'all' tab

alinhle:
## What does this PR do?
- Always redirect to `all` tab when not specified
- e.g. `http://192.168.5.5:3000/admin/moderate/1675539d-7136-4844-af08-781dc4465191` will automatically be redirected to `http://192.168.5.5:3000/admin/moderate/all/1675539d-7136-4844-af08-781dc4465191`.


kgardnr (approved):
:100:!

PR #709, opened 2017-06-23
Docs plugins quickstart

jde:
Note: comments on both style and content welcome!

## What does this PR do?

* Expands docs for plugins 
* Adds a new quickstart that walks a first time plugin author through the process

## How do I test this PR?

Either read the .md files or view them locally: https://coralproject.github.io/talk/faq.html#how-do-i-contribue-to-these-docs


wyattjoh:
Looks awesome!!!

ALawliet:
Some thoughts from the point of view of newbie who just looked at this project a few days ago and doesn’t know as much about it as you guys do. 😉

- I think the Overview should include what is a plugin and what is the point? For example, I didn’t realize something as simple as the “Like” button was a plugin until you guys told me during my interview. 😝 I think it would be beneficial if the first thing people read helped solidify the idea of plugins, their design, and purpose. 🔌
- If the Quickstart is what every new plugin developer is going to refer to in building their first plugin,  especially if they are less familiar with the project, it should be a little more comprehensive. For example, it should definitely include a client aspect, not just server, and also provide more details on properly setting up ES6/React/etc. for compatibility. Better yet, while I saw you've included .babelrc and .eslintrc, it’d be nice to have a plugin template with the basics set up that already runs (Babel/ESLint/React/(testing framework if you guys have a preference, static typing if that's added in the future), etc., shows up as a component somewhere, and does something simple on the server) that people could just fork and go to town without worrying about the setup, enabling them to focus on adding their functionality! 🍴🍰 
- Could you elaborate more on “avoid conflicts with not only the existing Talk API, but other plugins in the open source ecosystem that may be creating routes”? For example, more on the namespacing universals, and is there a list of routes we should probably avoid? 🤔
- Client API - Is there a list of slots somewhere? Maybe a pretty diagram of slot names mapping to a screenshot of where they are in the app? Where are all the places I can inject my components? I think knowing this would also help people come up with cool plugin ideas ahead of time! 🎰 
- Server API - As someone with a GraphQL background, I actually think it’d be better if the “Full Example” came before the individual parts. I was  just like, “Huh, what am I looking at here?”, until I got to the bottom and was like, “Oh those were just the fields for the main server export, I've seen all of this GraphQL stuff before, and it all makes perfect sense now”. I think seeing the big picture first helps a lot.
- That typo in “How do I contribue to these docs" hurts me. 😄

So awesome. 👍 💪 💯

kgardnr (approved):
:100:

kgardnr:
@ALawliet great catch on that typo - fixed! I'll leave it to @jde to incorporate the rest of your feedback - thank you for this awesome feedback!

jde:
Thanks for the excellent feedback @ALawliet! 

1) Agreed, an overview page worthy of the name is forthcoming!
2) Also in the works. We have been working to build our examples of all the possibilities in the `/plugins/` folder. Providing write-ups for each of the things that our plugin architecture allows is coming next. We've also discussed a scaffolding command... something akin to `cli plugins create talk-plugin-thingo .`
3) Adding that as it's own page, coming soon to a docs near you.
4) Top secret method for showing slot possibilities forthcoming... No need to stay up nights in anticipation...
5) I think the APIs will make a lot more sense as a reference once we have a few more "So you want to do x" tutorial style docs that reference them. 
6) I just slipped that typo in to see if anyone was paying attention! You win 22% of all the internets! (Seriously, thanks, that was painful!)

PR #715, opened 2017-06-26
Comment Count by Asset URL

okbel:
# Comment Count by Asset URL

## What does this PR do?
Allows you to retrieve Comment Count by Asset URL

## How do I test this PR?

This two should return the same result. (`asset_url`should correspond to `asset_id` replace them by some asset in your existing database)
```js
{
  commentCount(query:{
    asset_url: "http://localhost:3000/assets/title/new-article"
  })
}

# {
#   commentCount(query:{
#     asset_id: "8624b518-4a33-46e1-9448-baf89f373440"
#   })
# }

```

### DIFF
--- a/graph/loaders/assets.js
+++ b/graph/loaders/assets.js
@@ -71,9 +71,7 @@ const findByUrl = async (context, asset_url) => {
     throw errors.ErrInvalidAssetURL;
   }
 
-  let asset = await AssetsService.findByUrl(asset_url);
-
-  return asset;
+  return AssetsService.findByUrl(asset_url);
 };
 
 /**
--- a/graph/loaders/assets.js
+++ b/graph/loaders/assets.js
@@ -63,18 +63,33 @@ const getAssetsForMetrics = async ({loaders: {Actions, Comments}}) => {
     .then((connection) => connection.nodes);
 };
 
+const findByUrl = async (context, asset_url) => {
+
+  // Verify that the asset_url is parsable.
+  let parsed_asset_url = url.parse(asset_url);
+  if (!parsed_asset_url.protocol) {
+    throw errors.ErrInvalidAssetURL;
+  }
+
+  let asset = await AssetsService.findByUrl(asset_url);
+
+  return asset;
+};
+
 /**
  * Creates a set of loaders based on a GraphQL context.
  * @param  {Object} context the context of the GraphQL request
  * @return {Object}         object of loaders
  */
+
 module.exports = (context) => ({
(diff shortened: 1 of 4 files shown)


graph/loaders/assets.js
```diff
@@ -63,18 +63,33 @@ const getAssetsForMetrics = async ({loaders: {Actions, Comments}}) => {
...
+
+  // Verify that the asset_url is parsable.
+  let parsed_asset_url = url.parse(asset_url);
+  if (!parsed_asset_url.protocol) {
+    throw errors.ErrInvalidAssetURL;
+  }
+
+  let asset = await AssetsService.findByUrl(asset_url);
+
+  return asset;
```
wyattjoh: Maybe just:

```diff
-  let asset = await AssetsService.findByUrl(asset_url);
-
-  return asset;
+  return AssetsService.findByUrl(asset_url);
```

graph/resolvers/root_query.js
```diff
@@ -40,17 +40,26 @@ const RootQuery = {
...
     }
 
+    if (asset_url) {
+      return Assets.findByUrl(asset_url)
+        .then((asset) => {
+          const newQuery = Object.assign({}, query, {asset_id: asset.id});
+          return Comments.getCountByQuery(newQuery);
+        });
+    }
+
```
wyattjoh: Should instead be:

```js
if (asset_url) {
  query.asset_id = await Assets.findByUrl(asset_url);
}
```

wyattjoh (changes requested):
Some small changes that should make this more simple ;)

graph/resolvers/root_query.js
okbel: Maybe 

```js
if (asset_url) {
  query.asset_id = await Assets.findByUrl(asset_url).asset_id;
}
```

Also, this would't work since the CommentModel does not support `asset_url`. I can move this logic (find asset by url, return id, then retrieve comment count) to the `getCommentCountByQuery` within the comments loader.

Is that ok?

wyattjoh: You can always do:

```js
if (asset_url) {
  let asset = await Assets.findByUrl(asset_url);
  if (asset) {
    query.asset_id = asset.id;
  }
}
```

The CommentModel does indeed support `asset_id`, so that works fine.

okbel: Updated. Thank you! Let me know what you think

wyattjoh (approved):

PR #718, opened 2017-06-27
Don't alias username

alinhle:
## What does this PR do?
- Remove `username` alias. Reasoning: It is more clear and straightforward when we use the property name directly as specified in the graph schema.


wyattjoh:
I wonder what made that so easy to change what field name you got on the frontend :P GraphQL is pretty slick...

kgardnr:
@cvle @wyattjoh - quick question, didn't we have `username` because we wanted to differentiate between that and `name` which we get from Facebook?

alinhle:
@kgardnr our model and schema doesn't indicate that we have this distinction in our database.

alinhle:
Also this is just about querying our graph in the frontend. GraphQL allows renaming (aliasing) fields to a different name. As we pass the comment data around in other components and plugins, it is more straightforward and future-proof to use the same property names as defined in the GraphQL schema.

kgardnr:
Ok great @cvle!

kgardnr (approved):

PR #722, opened 2017-06-28
Allow editing to the same username in the "create username" flow

alinhle:
## What does this PR do?
- Allow editing username to the same username... "create username" workflow when registered with facebook.
- Require submitted username to be different in the "rejected username" workflow.

## How do I test this PR?

- Register a facebook account and submit the unchanged username.


kgardnr (approved):
LGTM!

PR #725, opened 2017-06-29
Updates to tab classes

StephanieDClark:
## What does this PR do?
Per conversations some time ago, just made classes correspond to other usages of "talk-xxx" rather than "coral-xxx".  Also added default classes to tab bar

## How do I test this PR?
See that tab, tab bar now have classes like "talk-tab" rather than "coral-tab".


okbel:
Looks great! Thank you!

okbel (approved):

PR #726, opened 2017-06-29
Don't create empty Slot elements

StephanieDClark:
## What does this PR do?
**Updated Slot component to:**
 - just return null if no elements.  Previously, it was always creating an empty div like "<div class></div>"  This made it difficult to determine if the slot actually had any real contents.

**Updated plugins.js getSlotElements to:**
 - check if module that is being inserted into slot has an isEnabled function, and if so, if that function returns true.
This made it possible to pass a property of "enabled:true|false" to the plugin in config.  

See examples of this in the pull for off topic & viewing options.  

## How do I test this PR?

For a Slot with no children, will now return null, rather than an empty div.

getSlotElements changes should have no effect until called for an updated plugin.


alinhle:
> This made it difficult to determine if the slot actually had any real contents.

What is the exact use case for doing this? What are you trying to achieve as the end result?

StephanieDClark:
Context:

I have a plugin streamtopper (which is added to the streamBox slot, for us), that puts out some html, and adds component <ViewingOptions> .

ViewingOptions puts out some html & whatnot, but in particular, builds a dropdown list of components that may have added themselves to the ViewingOptions slot.  So far so good.

For right now, the only thing that adds itself to the ViewingOptions dropdown is the off-topic plugin, but hypothetically, there could be 0 or many things that add themselves to the dropdown.

Use Case:
In our implementation, in particular, while the off-topic plugin is included in the build, it is not always supposed to be available on the page.  So, on most pages it is disabled, and only enabled on some (see the enable/disable configuration support for the PR I supplied yesterday)

So, for our purposes, I don't want the ViewingOptions slot to output anything at all, no list, and not itself, if no plugins have added themselves to the Slot.  

We will have other plugins that do the same thing; are enabled on some pages, not enabled on others.  The enabled: true|false is added to the plugin_config based on information about the particular page being loaded.

Intended Effect:

1.  in general, for a Slot with nothing in it, no html is returned.  So, nothing on the page.  Prior to this, html like ```<div class></div>``` was returned, at minimum.  
2. I can query getSlotElements for the slot I want to fill, and see that it is actually empty, and then choose to do nothing. e.g.,  ```
    let listItems = getSlotElements("viewingOptions", this.props);

    if (listItems && listItems.length === 0) {return (null);}```
3. getSlotElements now may check to see if the module in the slot has an "isEnabled" function, and if so, run it to see if the module is enabled.  So, if the off-topic plugin puts itself in the "ViewingOptions" slot, but is, on any page, disabled, then the component will not be returned as part of the slot.

alinhle:
Wouldn't the [:empty](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) css selector satisfy your requirement?

```css
.slotSelectorXYZ:empty {
  display: none;
}
```

StephanieDClark:
@cvle No.  In this case, I also don't want to display the parent component, ViewingOptions

alinhle:
Is your parent component that you are talking about the `streamtopper` plugin?  and is it using the logic you pointed out here?

```js
let listItems = getSlotElements("viewingOptions", this.props);

if (listItems && listItems.length === 0) {return (null);}
```

In this case, would it be an alternative to disable the `streamtopper` plugin in the config whenever all "ViewingOptions" plugins are disabled as well? Is it similar to what you are trying to achieve? 

```
plugin_config: {
  'coral-plugin-offtopic': {
    enabled: false,
  },
  'coral-plugin-streamtopper': {
    enabled: false,
    plugins: ['coral-plugin-offtopic']
  },
}
```

StephanieDClark:
@cvle Well, it would seem to be difficult to make sure that all the plugins knew about what was enabled when.

This seems, to me, like a very generic solution; a Slot can have 0 to many things in it.  If my component cares about that it can check for that.  There doesn't need to be any specific configuration of the plugins with this solution.  Also note that not all plugins currently support the enable/disable functionality.  However, this solution would still provide the functionality necessary; as long as it returns an empty component.

So, can you clarify why this proposal wouldn't work?  Or why it is unsatisfactory?

PR #729, opened 2017-06-29
Permalink by Query Param instead of Hash

okbel:
## What does this PR do?
- Uses `commentId` query param instead of hash
- Solves error while using wrong commentId

## How do I test this PR?

- Go to the Stream 
- Select a comment
- Check the permalink (you should have `talk-plugin-permalink` enabled)
- Open that in another tab
- Replace that commentId by a fake one


client/coral-embed-stream/src/components/StreamError.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import styles from './StreamError.css';
+
+export const StreamError = () => (
+  <div className={styles.streamError}>
+    An error has occured!
```
alinhle: Translation?

kgardnr:
This works great!

Question -- @okbel @cvle @jde -- now that we're using the query param, do we need to update where we reference permalinks elsewhere on Talk, e.g. on My Profile tab, and View Context link from admin?

client/coral-embed/src/index.js
```diff
@@ -190,7 +190,10 @@ Talk.render = function(el, opts) {
   // Compose the query to send down to the Talk API so it knows what to load.
   let query = {};
 
-  query.comment_id = window.location.hash.slice(1);
+  let urlParams = new URLSearchParams(window.location.search);
```
alinhle, line 194: Support for `URLSearchParams` is not good, see: http://caniuse.com/#feat=urlsearchparams.
Needs alternative or polyfill.

okbel: I checked that: Are we supporting IE or Opera Mini? /cc @kgardnr

kgardnr: Ah, we do need to support IE unfortunately @okbel

okbel: Adding that polyfill right now!

client/coral-embed-stream/src/components/StreamError.js
okbel: Yes! Adding that right now!

alinhle:
@kgardnr @okbel I think so. The `view context` links must probably be adapted.

okbel:
@kgardnr @cvle Ready to be tested!

kgardnr:
Just need to update "View Context" link on the comment cards in Admin and we're good to go!
talk_-_coral_admin

okbel:
@kgardnr I added that one too! That one should be fixed too :)

alinhle (approved):
🎉

PR #741, opened 2017-07-07
Remove CommonJS style exports from framework

alinhle:
## What does this PR do?
- Fixes `User history drawer issues`

## Issue
- Two thinks were biting us:
  - A circular dependency `coral-framework` -> `actions` -> `coral-framework` causes a moment where dependencies (e.g. `pym`) were not initialised yet and were in an intermediate `undefined` state.
  - The CommonJS style export used in some places in `coral-framework` persisted the intermediate state and exported it as `default` which was consumed by other parts of the code with the help of the `babel-plugin-add-module-exports` which adds support for this deprecated behavior.

## Solution
- Removes the deprecated CommonJS style exports from the  `coral-framework`.
- Remove the circular dependency

## How do I test this PR?
- Check that user history drawer is working (click outside behavior)


kgardnr:
This looks 💯 to me! When @okbel gives the okay, we can merge it!

okbel (approved):

wyattjoh (approved):

PR #744, opened 2017-07-07
Flush Redis before each test

wyattjoh:
## What does this PR do?

- Flushes the database before each test to resolve the outstanding bug while testing with the rate limit middleware many times (it's too fast ;))
- Changes default redis test database to `1`, which differs from the default `0`

## How do I test this PR?

- Run `yarn test`, many times, notice no errors after many tests


alinhle:
Will this interfere with my running talk dev as it also utilize redis?

kgardnr (approved):
No errors! :1st_place_medal:

PR #746, opened 2017-07-09
Create Plugins - Create plugins through the CLI

okbel:
# Create Plugins - Create plugins through the CLI

## What does this PR do?
- Allows you to create plugins through the CLI
- Fixes bug having plugins with empty config or without slots

## How do I test this PR?
- Pull the branch `git checkout create-plugin`
- Go to the console and type `./bin/cli-plugins create`
- Answer the questions _**HINT**: Choose yes on client ;)_
- It should look like this: screen shot 2017-07-09 at 12 13 36 pm

- Go to to Talk `localhost:3000`
- Refresh the page and you should see a brand new plugin on the stream.
<details>
<summary>**SPOILER**: Your new plugin: </summary>
screen shot 2017-07-09 at 12 17 07 pm
</details>


scripts/templates/plugin/index.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
wyattjoh: Possibly throw a link here for the server side plugin docs?

bin/cli-plugins
```diff
@@ -272,11 +273,129 @@ async function reconcilePluginDeps({skipLocal, skipRemote, dryRun, upgradeRemote
...
+
+  if (fs.existsSync(seedPlugin)) {
+
+    if (answers.server && answers.client) {
+
+      // This is a server-side and client-side plugin!, let's copy the template
+      fs.copySync(seedPlugin, newPluginPath);
+    } else  {
+
+      fs.copySync(seedPlugin, newPluginPath, {filter: (p) => {
```
wyattjoh, line 335: I would suggest using EJS templates instead of using the copy/sync method to give more flexibility to our templates.

okbel: 🤔 Thinking closely about it, I don't see the point of making it a template (at least for now) since we are only injecting a seed (literally copying it).

//cc @cvle thoughts?

alinhle: Hm I think having the flexibility of a template is a lot more powerful and will come in handy when we provide more customisation options.

jde:
I recommend creating the plugins in a folder parallel to the talk folder instead of inside the plugins folder. Most plugin authors should create individual repos for plugins and publish them separately, which is much more seamlessly accomplished if they are outside the talk repo.

kgardnr (approved):

PR #751, opened 2017-07-10
Talk Plugin Avatar

okbel:
## What does this PR do?
- Adds changes to use `talk-plugin-avatar`
 --- https://github.com/coralproject/talk-plugin-avatar
- Adds slot `commentAvatar`
- Adds `withFragment` to the plugin-api

https://www.pivotaltracker.com/story/show/144939589

## How do I test this PR?
- Add `talk-plugin-avatar` to the plugins.json, both client and server
- Test it out


okbel:
@cvle Please, check also https://github.com/coralproject/talk-plugin-avatar. Thank you!

alinhle:
image
Is the license here correct?

alinhle:
image
Why is this a webhook, and why is it called `user_update` when it only updates the avatar.

kgardnr:
The license should be "Apache v2.0" @okbel/@cvle :)

okbel:
@cvle @kgardnr Updated! https://github.com/coralproject/talk-plugin-avatar/blob/master/package.json

alinhle:
I think the `talk-plugin-avatar` needs to extend mutations because it adds a new field to the stream query. I haven't verified it yet.

wyattjoh:
The webhook example there @cvle is because it was meant as a demo/snippet of the implementation. The full implementation would depend on the actual login/cms system that the newsroom/org has in order to implement that as a webhook. The original story focused specifically on providing support for external providers (namely, external auth schemes). So this code could be forked and included in their own user/auth plugin.

alinhle:
image

@okbel The reason you didn't run into mutation errors, was because you didn't use your container :-)

okbel:
Updated! https://github.com/coralproject/talk-plugin-avatar/tree/master/client too.

alinhle:
image

A margin is added to the left of the comments, even when no avatar is shown. There should be no margin. Maybe we should let the plugin add the margin instead? :thinking:

okbel:
@cvle I thought we were not rendering the slot if it was empty. I will add it to the plugin.

alinhle:
@okbel we didn't add that yet, you can do that with CSS though `.slot:empty {display:none}`

okbel:
Updated! Both plugin and core

kgardnr (approved):

PR #757, opened 2017-07-11
Allow query comments and comment counts by tags

alinhle:
## What does this PR do?
- Allow querying asset comments and comment counts by tags.


graph/resolvers/asset.js
```diff
@@ -4,31 +4,32 @@ const Asset = {
...
       parent_id: null,
       excludeIgnored,
     });
   },
-  commentCount({id, commentCount}, {excludeIgnored}, {user, loaders: {Comments}}) {
-    
+  commentCount({id, commentCount}, {query: {excludeIgnored, tag}}, {user, loaders: {Comments}}) {
+
     // TODO: remove
     if (user && excludeIgnored) {
```
alinhle: This if clause is not right yet :-P

graph/resolvers/asset.js
```diff
@@ -4,31 +4,32 @@ const Asset = {
...
     if (commentCount != null) {
       return commentCount;
     }
     return Comments.parentCountByAssetID.load(id);
   },
-  totalCommentCount({id, totalCommentCount}, {excludeIgnored}, {user, loaders: {Comments}}) {
+  totalCommentCount({id, totalCommentCount}, {query: {excludeIgnored, tag}}, {user, loaders: {Comments}}) {
 
     // TODO: remove
     if (user && excludeIgnored) {
```
alinhle: Same

graph/loaders/comments.js
```diff
@@ -218,6 +218,14 @@ const getCommentCountByQuery = (context, {ids, statuses, asset_id, parent_id, au
     query = query.where({author_id});
   }
 
+  if (tag) {
+    query = query.find({
+      'tags.tag.name': {
+        $in: tag,
```
wyattjoh: I think that `tag` should be `tags`. With regards to the performance issues, I think we may need to approach that later in a performance related PR so we can define our indexes.

alinhle: Yeah my first thought was `tags` too. Someone already added that to the definitions (without implementing the loader) to the graphql and used `tag` so I thought maybe it had some good reasoning :-)

wyattjoh: Sounds good! Feel free to change it up!

wyattjoh (approved):

PR #758, opened 2017-07-11
Create cli docs

jde:
## What does this PR do?

Add docs for the cli tool

## How do I test this PR?

read 'em

### DIFF
--- a/docs/architecture-cli.md
+++ b/docs/architecture-cli.md
@@ -23,10 +23,12 @@ In a terminal, try:
 Commonly, you'll be in the `talk/` folder, in which case you can:
 
 ```
-bin cli [options] [commands] [arguments]
+bin/cli [options] [commands] [arguments]
 ```
 
-If you're a heavy cli user, consider adding the `bin` folder to your path so you can run `cli` from anywhere!
+If you're a heavy cli user, consider adding the `bin` folder to your PATH so you can run `cli` from anywhere!
+
+If you are using [our Docker environment](install-docker.html), the bin folder will already be in the PATH.
 
 ## What can I do with the cli?
 
@@ -34,8 +36,10 @@ The Talk cli ships with 'unix style' help. To access the docs, simply run the cl
 
 Let's say I wanted to figure out how to change a user's password. I'd start be seeing what the cli has for me.
 
+(Note: the following output may change, please reference at the `--help` output for your version as you use the cli.)
+
 ```
-talk :) ]$ bin/cli
+talk :) ]$ bin/cli --help
 
   Usage: cli [options] [command]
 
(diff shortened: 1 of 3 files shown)


docs/architecture-cli.md
```diff
@@ -0,0 +1,133 @@
...
+It is designed to provide a convenient way for engineers to perform key tasks without the need to muck about in the UI. It also opens the door for scripts to perform operations programmatically.
+
+Note: the cli tool requires [the Talk environment to be configured](configuration.html) either via env vars or by using a `.cli` file via `bin/cli -c .env [command] ....`
+
+## Using the cli
+
+In a terminal, try:
+
+```
+/path/to/talk/bin cli [options] [commands] [arguments]
```
wyattjoh, line 20: It's `bin/cli` not `bin cli`

wyattjoh: You can also note that in the Docker environment, `cli` is in the PATH, so you can just type `cli ...`

docs/architecture-cli.md
```diff
@@ -0,0 +1,133 @@
...
+    plugins     provides utilities for interacting with the plugin system
+    help [cmd]  display help for [cmd]
+
+  Options:
+
+    -h, --help           output usage information
+    -V, --version        output the version number
+    -c, --config [path]  Specify the configuration file to load
+    --pid [path]         Specify a path to output the current PID to
+```
```
wyattjoh, line 66: I wouldn't include the docs here for that, it can change too quickly, and instead I'd suggest people just run `--help` on the commands to discover.

wyattjoh (changes requested):

docs/architecture-cli.md
jde: Updated

docs/architecture-cli.md
jde: I'll add a note that this can change. One of the mindsets that I'm trying to develop here is that people _think in terms of the cli_. Having this verbose output here I believe will trigger folks to think *woah, there's a lot I can do* and actually try it for themselves. This is under the flag of showing off/selling what we have to people who may not have tried it yet.

wyattjoh: You convinced me :) keep it, possibly with a note that it may be out of date with their version and they should check with the `--help` flag.

jde: Done!

jde:
Ready for another review @wyattjoh, thanks!

docs/architecture-cli.md
```diff
@@ -124,7 +128,7 @@ This is accomplished by running the cli tool on any box using the mongo/redis/et
 
 The cli tool will connect directly with the install's db and redis instance(s) so ensure that your box can reach those servers on the appropriate ports.
 
-Also, _please ensure your cli and the server(s) in an environment are using the same version of Talk._ 
+Also, _please ensure your cli and the server(s) in an environment are using the same version of Talk._
```
wyattjoh: It's not actually possible to have a mismatch here, they are the same thing.

jde: If you have a box that's running cli against an environment with a bunch of talk servers it can be... Or if I were using my local talk cli against a db/redis combo in a staging environment as well.

docs/architecture-cli.md
kgardnr: I like this idea too!

kgardnr:
I think this is good to merge if @wyattjoh agrees!

wyattjoh (approved):

PR #762, opened 2017-07-13
BUG: Console error when you reply to a comment

alinhle:
## What does this PR do?
- For some reason the second `setState` caused an error in react. The solution was to just use `setState` once.


okbel:
Woohoo! :)

kgardnr (approved):

PR #769, opened 2017-07-17
Conditionally exclude a Plugin Component from a Slot

alinhle:
## What does this PR do?
- Adds a hoc `excludeIf(condition: func)` to support conditionally excluding a component from a slot depending on the current props. (available in the plugins-api). e.g. 
```es6
 const enhance = compose(
    // `withFragments` omitted.
    excludeIf((props) => props.asset.totalComments === 0).
  );
 export default enhance(BaseComponent);
```
- Adds a hoc `connect` that proxies through the `connect` from `react-redux` but also keeps the mapping information to make `excludeIf` aware of the props from redux. (available in the plugins-api)
- Adds a component called `Popup` that declaratively controls a real window popup using React. Provides events like `onFocus`, `onLoad`, `onClose`, ... to allow syncing the state back to React.
```jsx
<Popup
  href='/embed/stream/login'
  title='Login'
  features='menubar=0,resizable=0,width=500,height=550,top=20'
  open={showSignInDialog}
  focus={signInDialogFocus}
  onFocus={focusSignInDialog}
  onBlur={blurSignInDialog}
  onClose={hideSignInDialog}
/>
```
- Adds a component `IfSlotIsEmpty` and `IfSlotIsNotEmpty` that only renders its children depending on whether the given slot is empty or not. The same additional props that are supplied to the `Slot` component must also be supplied to get the correct result. E.g. 
```jsx
<IfSlotIsEmpty slot="slotName" {...slotProps} >{children}</IfSlotIsEmpty>
```
- Refactor our plugins and core to use the new API.
- Deprecate `isSlotEmpty` in `plugin-api`.


okbel (approved):

okbel:
This one is amazing!! Good job 💯  We should add this to the docs! I'll add a task for it.

PR #772, opened 2017-07-17
Generate and integrate introspection query results

wyattjoh:
## What does this PR do?

- Added new WEBPACK env var which is enabled during yarn build scripts
- Defered redis starting until listen is called
- Moved pubsub to a factory pattern init
- Async/Await'ed the routes
- Moved pubsub handle for routes into middleware
- Adjusted redis cache and job processors to have lazy connection
starting
- Disabled mongo from auto-connecting on require
- Adjusted package redis clients to act as factory singletons instead
- Generate a `introspection.json` as part of our build process.
- Auto update `introspection.json` when using `yarn dev-start`.
- Port client code to use the `IntrospectionFragmentMatcher`.

## How do I test this PR?

1. Stop mongo + redis docker containers.
2. Create a file `test_772.js` with the contents:

```js
require('./app');
```

3. Run `WEBPACK=true NODE_ENV=test node test_772.js`
4. See that the script starts/stops without issue.


services/scraper.js
```diff
@@ -12,11 +12,20 @@ const metascraper = require('metascraper');
 const scraper = {
 
   /**
-   * Create the new Task kue.
+   * Create the new Task kue singleton.
    */
-  task: new kue.Task({
-    name: 'scraper'
-  }),
+  _task: null,
```
alinhle: I think `_task` should be kept internal like:

```js
let taskInstance = null;

const scraper = {
  // ...
  get task() {
    if (taskInstance) {
      return taskInstance;
    }

    taskInstance = new kue.Task({
      name: 'scraper'
    });

    return taskInstance;
  },
  // ...
};
```

services/mailer.js
```diff
@@ -75,9 +75,18 @@ const mailer = module.exports = {
   /**
    * Create the new Task kue.
    */
-  task: new kue.Task({
-    name: 'mailer'
-  }),
+  _task: null,
```
alinhle: I think _task should be kept internal. See below.

services/mongoose.js
```diff
@@ -40,14 +41,23 @@ if (enabled('talk:db')) {
...
 
-// Connect to the Mongo instance.
-mongoose.connect(MONGO_URL, (err) => {
-  if (err) {
-    throw err;
-  }
+if (WEBPACK) {
+
+  console.warn('Not connecting to mongodb during webpack build');
+  mongoose.disconnect();
```
alinhle: So mongoose automatically connects and we have to disconnect manually when in Webpack build?

services/passport.js
```diff
@@ -9,12 +9,21 @@ const LocalStrategy = require('passport-local').Strategy;
...
 const debug = require('debug')('talk:services:passport');
-const {createClient} = require('./redis');
 const bowser = require('bowser');
 const ms = require('ms');
 
 // Create a redis client to use for authentication.
-const client = createClient();
+const {createClient} = require('./redis');
+let _client = null;
+const getClient = () => {
```
alinhle: Why not export a `getClient` from `./redis.js` and reuse it here?

```js
// redis.js
let redisInstance = null;
module.exports = {
  // ...
  getClient() {
    if (redisInstance) {
      return redisInstance;
    }
    redisInstance = createClient();
    return redisInstance;
  },
  // ...
}
```

services/users.js
```diff
@@ -32,6 +29,19 @@ const PASSWORD_RESET_JWT_SUBJECT = 'password_reset';
 // through during the salting process.
 const SALT_ROUNDS = 10;
 
+// Create a redis client to use for authentication.
+const {createClient} = require('./redis');
+let _client = null;
+const getClient = () => {
```
alinhle: See above

services/pubsub.js
```diff
@@ -2,4 +2,13 @@ const {RedisPubSub} = require('graphql-redis-subscriptions');
 
 const {connectionOptions} = require('./redis');
 
-module.exports = new RedisPubSub({connection: connectionOptions});
+let pubsub = null;
+module.exports = () => {
```
alinhle: Maybe exporting an object with a descriptive function is more clear:

```js
// pubsub.js
module.exports = {
  getInstance() {
    if (pubsub) {
      return pubsub;
     }
    pubsub = new RedisPubSub({connection: connectionOptions});
  },
}
```

So it can be consumed like this:

```js
const pubsub = require('pubsub');
const pubsubInstance = pubsub.getInstance();
```

Instead of

```js
const pubsub = require('pubsub');
const pubsubInstance = pubsub();
```
The line `pubsub = pubsub()` as used in this PR was not immediately clear to me what it meant.

services/kue.js
```diff
@@ -8,18 +8,24 @@ const kue = module.exports.kue = require('kue');
 // Note that unlike what the name createQueue suggests, it currently returns a
 // singleton Queue instance. So you can configure and use only a single Queue
 // object within your node.js process.
-const Queue = module.exports.queue = kue.createQueue({
-  redis: {
-    createClientFactory: () => redis.createClient()
-  }
-});
+let Queue = module.exports.queue;
```
alinhle: Doesn't this equal to `let Queue = undefined`? If so rather set it to `let Queue = null;` and put a comment that the instance is lazily created and assigned.

services/mongoose.js
wyattjoh: Not exactly. We don't call connect, but because we include it, it will hold the socket ready, preventing node from exiting. Calling disconnect here just ensures that the application can quit correctly.

alinhle: Would be great if you add that as a comment :-)

services/redis.js
```diff
@@ -29,21 +29,36 @@ const connectionOptions = {
...
 module.exports = {
   connectionOptions,
-  createClient() {
-    let client = redis.createClient(connectionOptions);
-
-    client.ping((err) => {
-      if (err) {
-        console.error('Can\'t ping the redis server!');
+  createClient,
+  createClientFactory: () => {
```
alinhle, line 51: I saw you use a Factory Pattern to create client instances to redis. Wouldn’t we just want to have one client instance that connects to redis, and as such a singleton pattern would be more appropiate (or written in the factory pattern style but always returns the same instance)?

alinhle (approved):

PR #773, opened 2017-07-17
Fix email address env var in docs

icco:
## What does this PR do?

Update docs

## How do I test this PR?

The variable mentioned (`TALK_SMTP_EMAIL`) doesn't exist, but `TALK_SMTP_FROM_ADDRESS ` does at https://github.com/coralproject/talk/blob/244d421eda6829280a6639875b2abb4794c47ec3/config.js#L76


wyattjoh (approved):
Great catch!

PR #779, opened 2017-07-20
Renaming Talk Plugins

okbel:
## What does this PR do?
- Renames Talk Plugins from `coral-plugin` to `talk-plugin`

## How do I test this PR?
- Test carefully and report if you find any errors! We don't want to miss anything while renaming talk plugins :)


kgardnr:
@cvle and @okbel I'm good to merge this after extensive testing. We can continue to do testing on staging as well prior to 3.0.0 release. Thoughts?

kgardnr:
Actually, I need to do some more testing before we merge fyi @okbel @cvle - sorry for the premature ok!

kgardnr (approved):

PR #787, opened 2017-07-21
Fix for long titles in the All Streams Queue

okbel:
## What does this PR do?
- Adds fix for https://www.pivotaltracker.com/story/show/148148403


kgardnr (approved):
:100:!

PR #790, opened 2017-07-24
Search Comment in an Asset

okbel:
## What does this PR do?
- Fixes this issue https://www.pivotaltracker.com/story/show/148441077
- Extends the Asset schema


alinhle (dismissed):

alinhle (approved):
I found some other places that needs to be adapted (autoscroll, query utils) and made the changes. The rest of the code looks good.

The error message when either the asset or the comment was not found could be more descriptive though:
image

PR #798, opened 2017-07-25
Live Reactions should not always expect the comment to be there.

alinhle:
## What does this PR do?
- Currently if a live update for a reactions comes in and the targeted comment has not been loaded yet in my store, I will get an error message.
- Verify that we already have the comment in the store before updating it and prevent error being logged to the console.

## How to test this PR
- Make reactions on comments that has not been loaded yet on the other browser.


okbel (approved):
Looks fine. Maybe we need to work in some helper functions to access the cache so we don't repeat ourselves.

PR #794, opened 2017-07-25
Mobile and Touch Events for Featured Tags/ Viewing Options

okbel:
## What does this PR do?
- Adds touch event for featured tags https://www.pivotaltracker.com/story/show/148441077
- Fix and adds filter Icon for Viewing Options https://www.pivotaltracker.com/story/show/149353761

PR #799, opened 2017-07-25
Prefix redux constants in plugins

alinhle:
## What does this PR do?
- Redux constants must be unique across the whole application. Prefix plugin redux constants so that a name collision is less likely to happen.


alinhle:
I figured, that we will have merge conflict with https://github.com/coralproject/talk/pull/794. So we should coordinate this^^

okbel (approved):
👍

PR #801, opened 2017-07-26
Secret Rotation + Improvements

wyattjoh:
## What does this PR do?

This extends the existing JWT system to support rotating secrets using any supported signing algorithm. This is designed to be backwards compatible with existing installations.

### DIFF
--- a/bin/cli-serve
+++ b/bin/cli-serve
@@ -90,7 +90,7 @@ async function onListening() {
   let bind = typeof addr === 'string'
     ? `pipe ${addr}`
     : `port ${addr.port}`;
-  console.log(`API Server Listening on ${bind}`);
+  debug(`API Server Listening on ${bind}`);
 }
 
 /**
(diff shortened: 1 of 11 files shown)


wyattjoh:
Note that docs for this are available on the #788 PR.

alinhle (approved):
This is an awesome job wyatt. I have one suggestion, but I don't want it to block your work, so I'm going to approve it and I let you decide whether or not to act on this suggestion:

I expected `config.js` to export something that is more or less a direct representation of the configuration values that the user passed in (+ default values).

```js
 const {
   JWT_SECRET,
   JWT_SECRETS,
   JWT_ALG,
   // ...
 } = require('../config');
```

While most value in `config` meet that expectation, unlike `JWT_SECRETS` or `JWT_ALT`,  `JWT_SECRET` has been replaced by one of the new secret classes which is the result of processing multiple `JWT_*` values.

Maybe it would be a better separation to move the _secrets initialisation_ out of the config and into another file like `secret.js` that consumes `JWT_SECRET` and  `JWT_SECRETS` from `config` and exports the secret object.

wyattjoh:
I've implemented the refactoring suggestion you've made @cvle :)

PR #813, opened 2017-07-28
New Queue

okbel:
## What does this PR do?
- Separate Queue, called New (shows if there is no Pre-Mod queue)
- "New" queue only contains comments that have not been Accepted or Rejected

`Pre-mod` = stream/system set to pre-mod on

`Reported` = user reported, suspect words, `system reported` aka trust, perspective toxicity threshold, etc.

--  If someone hits a trust threshold -- is their comment marked as `reported` `flagged` or `premod`

So the options would be:
New | Reported | Approved | Rejected | All

or

Pre-Mod | Reported | Approved | Rejected | All

## How do I test this PR?
- Create a new asset
- Switch it to premod
- Go to admin
- Switch between articles 
- Notice that Pre-Mod Tab shows up when the article has premod enabled, and New when it's disabled.

- Try to go to `http://localhost:3000/admin/moderate/new/a36b75ed-e1d0-4c8b-8905-7c8e08971730`
- Erase the `new`
- It should go to `reported` if there is reported comments, then `premod` if premod is enabled, and if not, it will default to `new`


alinhle:
Without doing a full review I noticed these two points have not been dealt with yet:
- From the story:
> If there is anything that has been "reported" then moderation opens on this tab, if not default open to "New"
- After a moderation action has been taken, by the user or by a subscription update then the comment should be removed from the queue and appear in appropiate queues (e.g. rejected should go to rejected, etc...)

okbel:
> After a moderation action has been taken, by the user or by a subscription update then the comment should be removed from the queue and appear in appropiate queues (e.g. rejected should go to rejected, etc...)

This one has been taken care of. There were some missing new queue names for the subscription.

And for this one:

> If there is anything that has been "reported" then moderation opens on this tab, if not default open to "New"

I'm comming up with a solution.

okbel:
Done! Ready :)

client/coral-admin/src/AppRouter.js
```diff
@@ -47,18 +51,19 @@ const routes = (
...
         </Route>
-        <Route path='flagged' components={Moderation}>
+        <Route path='reported' components={Moderation}>
           <Route path=':id' components={Moderation} />
         </Route>
-        <Redirect from=':id' to='all/:id' />
+
+        <Route path=':id' components={Moderation} />
+
         <IndexRedirect to='all' />
```
alinhle: This line should be removed as we don't want to automatically redirect to the "all" tab.

alinhle: Also the other IndexRedirect should be removed as well on line 18: https://github.com/coralproject/talk/pull/813/files#diff-c7143cc58e326db8b86b6cdbb68e6170R18

client/coral-admin/src/AppRouter.js
```diff
@@ -35,10 +35,14 @@ const routes = (
       {/* Moderation Routes */}
 
       <Route path='moderate' component={ModerationLayout}>
+
```
alinhle: New IndexRoute should exist here to route to `Moderation` e.g. `<IndexRoute component={Moderation} />`.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -7,6 +7,7 @@ import ModerationMenu from './ModerationMenu';
 import ModerationHeader from './ModerationHeader';
 import ModerationKeysModal from '../../../components/ModerationKeysModal';
 import StorySearch from '../containers/StorySearch';
+import {isPremod, getModPath} from '../../../utils';
 
 export default class Moderation extends Component {
```
alinhle: The routing code should not exist here but in the _container_. Here we only want to have _presentational_ related code.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -19,6 +20,17 @@ export default class Moderation extends Component {
   }
 
   componentWillMount() {
+    const {router, root: {settings, reportedCount}} = this.props;
+
+    // Will update the URL to premod or new based on the moderation of All Streams
+    // If there is a reported comments, reported queue has priority!
+    if (!router.params.id) {
```
alinhle: The conditional should be to check if we are currently at the `IndexRoute` and redirect to the correct "queue" if true.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -30,6 +42,19 @@ export default class Moderation extends Component {
...
+
+    // Will update the URL to premod or new based on the moderation of the *asset*
+    // If there is a reported comments, reported queue has priority!
+    if (route.path === ':id') {
+      const queue = isPremod(asset.settings.moderation) ? 'premod' : 'new';
+      router.push(
+        getModPath(reportedCount ? 'reported' :  queue, asset.id)
+      );
+    }
+  }
```
alinhle: This shouldn't be necessary at all.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -19,6 +20,17 @@ export default class Moderation extends Component {
...
+
+    // Will update the URL to premod or new based on the moderation of All Streams
+    // If there is a reported comments, reported queue has priority!
+    if (!router.params.id) {    
+      const queue = isPremod(settings.moderation) ? 'premod' : 'new';
+      router.push(
+        getModPath(reportedCount ? 'reported' : queue)
+      );
+    }
+
```
alinhle: This code should live in the  `container`.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -58,8 +83,14 @@ export default class Moderation extends Component {
...
   getComments = () => {
-    const {root, route} = this.props;
-    const activeTab = route.path === ':id' ? 'premod' : route.path;
+    const {root, root: {asset, settings}, router, route} = this.props;
+
+    // Grab premod from asset or from settings
+    const premod = !router.params.id ? settings.moderation : asset.settings.moderation;
+
+    const queue = isPremod(premod) ? 'premod' : 'new';
+    const activeTab = router.params.id ? queue : route.path;
```
alinhle: `activeTab` should be taken from `props` instead of recalculating it here.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -137,12 +170,14 @@ export default class Moderation extends Component {
...
           allCount={root.allCount}
+          newCount={root.newCount}
           acceptedCount={root.acceptedCount}
           premodCount={root.premodCount}
           rejectedCount={root.rejectedCount}
-          flaggedCount={root.flaggedCount}
+          reportedCount={root.reportedCount}
           selectSort={this.props.setSortOrder}
           sort={this.props.moderation.sortOrder}
+          premodEnabled={assetId ? isPremod(root.asset.settings.moderation) : isPremod(settings.moderation)}
```
alinhle: Whether or not premod is enabled should be calculated in the `container` and passed down to this components, as this should be purely presentational.

client/coral-admin/src/routes/Moderation/components/StorySearch.js
```diff
@@ -64,14 +65,15 @@ const StorySearch = (props) => {
...
+
                   return <Story
                     key={i}
                     id={story.id}
                     title={story.title}
                     createdAt={new Date(story.created_at).toISOString()}
                     open={storyOpen}
                     author={story.author}
-                    goToStory={props.goToStory}
+                    goToStory={() => props.goToStory(story.id, isPremod(story.settings.moderation))}
```
alinhle: No need to pass in `isPremod`, see below.

client/coral-admin/src/routes/Moderation/containers/StorySearch.js
```diff
@@ -57,15 +57,15 @@ class StorySearchContainer extends React.Component {
     this.props.storySearchChange(searchValue);
   }
 
-  goToStory = (id) => {
+  goToStory = (id, premod) => {
     const {router} = this.props;
-    router.push(`/admin/moderate/all/${id}`);
+    router.push(`/admin/moderate/${premod ? 'premod' : 'new'}/${id}`);
```
alinhle: This should always point to `/admin/moderate/${id}`

client/coral-admin/src/routes/Moderation/containers/Moderation.js
```diff
@@ -337,10 +350,17 @@ const withModQueueQuery = withQuery(gql`
...
+      }
     }
     allCount: commentCount(query: {
       asset_id: $asset_id
     })
+    newCount: commentCount(query: {
+      statuses: [NONE, PREMOD],
+      asset_id: $asset_id
+    })
     acceptedCount: commentCount(query: {
```
alinhle: For consistency this should also be called `approvedCount`

client/coral-admin/src/routes/Moderation/containers/Moderation.js
```diff
@@ -380,6 +401,10 @@ const withQueueCountPolling = withQuery(gql`
...
```
alinhle: For consistency this should also be called `approvedCount`

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -112,14 +142,17 @@ export default class Moderation extends Component {
     case 'all':
       activeTabCount = root.allCount;
       break;
+    case 'new':
+      activeTabCount = root.newCount;
+      break;
     case 'accepted':
       activeTabCount = root.acceptedCount;
```
alinhle: For consistency this should also be called `approvedCount`

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -112,14 +142,17 @@ export default class Moderation extends Component {
     case 'all':
       activeTabCount = root.allCount;
       break;
+    case 'new':
+      activeTabCount = root.newCount;
+      break;
     case 'accepted':
```
alinhle: This should be `approved`.

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -137,12 +170,14 @@ export default class Moderation extends Component {
         <ModerationMenu
           asset={asset}
           allCount={root.allCount}
+          newCount={root.newCount}
           acceptedCount={root.acceptedCount}
```
alinhle: For consistency this should also be called `approvedCount`

client/coral-admin/src/routes/Moderation/components/ModerationMenu.js
```diff
@@ -4,51 +4,69 @@ import styles from './styles.css';
...
-) => {
-
-  function getPath (type) {
-    return asset ? `/admin/moderate/${type}/${asset.id}` : `/admin/moderate/${type}`;
-  }
-
+const ModerationMenu = ({
+  asset = {}, 
+  allCount,
+  acceptedCount,
```
alinhle: For consistency this should also be called `approvedCount`

client/coral-admin/src/routes/Moderation/components/ModerationMenu.js
```diff
@@ -4,51 +4,69 @@ import styles from './styles.css';
...
-            <Icon name='access_time' className={styles.tabIcon} /> {t('modqueue.premod')} <CommentCount count={premodCount} />
+            <Icon name='flag' className={styles.tabIcon} /> {t('modqueue.reported')} <CommentCount count={reportedCount} />
           </Link>
           <Link
-            to={getPath('flagged')}
+            to={getModPath('approved', asset.id)}
             className={`mdl-tabs__tab ${styles.tab}`}
             activeClassName={styles.active}>
-            <Icon name='flag' className={styles.tabIcon} /> {t('modqueue.flagged')} <CommentCount count={flaggedCount} />
+            <Icon name='check' className={styles.tabIcon} /> {t('modqueue.approved')} <CommentCount count={acceptedCount} />
```
alinhle: For consistency this should also be called `approvedCount`

client/coral-admin/src/graphql/utils.js
```diff
@@ -66,21 +66,27 @@ function addCommentToQueue(root, queue, comment, sort) {
 
 function getCommentQueues(comment) {
   const queues = ['all'];
+
   if (comment.status === 'ACCEPTED') {
     queues.push('accepted');
```
alinhle: This should be `approved`.

client/coral-admin/src/graphql/utils.js
```diff
@@ -66,21 +66,27 @@ function addCommentToQueue(root, queue, comment, sort) {
...
   ) {
     queues.push('flagged');
   }
+  else if (comment.status === 'PREMOD') {
+    queues.push('premod');
+    queues.push('new');
+  } 
+  if (comment.status === 'NONE') {
+    queues.push('new');
+  }
```
alinhle: Are those changes above intentional? It means that if a comment is `flagged` it will not appear in the `premod`  queue. It essentially means if a comment gets flagged, it will be removed from the `premod`, which I don't think is good, because flags are made by users and I don't want users to remove stuff from queues.

client/coral-admin/src/graphql/utils.js
```diff
@@ -66,21 +66,27 @@ function addCommentToQueue(root, queue, comment, sort) {
...
     queues.push('accepted');
   }
   else if (comment.status === 'REJECTED') {
     queues.push('rejected');
   }
-  else if (comment.status === 'PREMOD') {
-    queues.push('premod');
-  }
-  if (
+  else if (
```
alinhle: This should remain an `if` condition instead of `else if` (see explanation above/below)

client/coral-admin/src/graphql/utils.js
```diff
@@ -66,21 +66,27 @@ function addCommentToQueue(root, queue, comment, sort) {
...
     ['NONE', 'PREMOD'].indexOf(comment.status) >= 0
     && comment.actions && comment.actions.some((a) => a.__typename === 'FlagAction')
   ) {
     queues.push('flagged');
   }
+  else if (comment.status === 'PREMOD') {
```
alinhle: This should move up to join the other `else if` conditions (see explanation above/below)

client/coral-admin/src/graphql/utils.js
```diff
@@ -66,21 +66,27 @@ function addCommentToQueue(root, queue, comment, sort) {
...
+    queues.push('premod');
+    queues.push('new');
+  } 
+  if (comment.status === 'NONE') {
```
alinhle: This should be an `else if condition` and move up to join the other `else if` conditions (see explanation above/below)

alinhle (changes requested):
See comments above

client/coral-admin/src/routes/Moderation/components/Moderation.js
okbel: This one is necessary when there is a transition that updates. i.e clicking the StorySearch results

alinhle: It's not necessary because switching routes result in a _remount_, resulting a call to `componentWillMount`.

okbel: It doesn't result in a remount. It's just updating the props :/ Checking

client/coral-admin/src/routes/Moderation/components/Moderation.js
okbel: I need this to use it on the `componentWillMount` here because in the container I don't have `root` data populated. :/

client/coral-admin/src/routes/Moderation/components/Moderation.js
alinhle: Oops I guess I was wrong, I tested this and apparently I did not 😭

kgardnr:
This is amaaaaazing @okbel!!!! If @cvle is good with the updates, let's merge it!!

alinhle (approved):
Awesome! Looks good now :-) 

_Fyi @okbel I added a couple of remaining small changes that I found_

PR #815, opened 2017-07-28
Embed Control

alinhle:
## What does this PR do?
- Return an object on `Talk.render` to control the embed (currently triggering login / logout and listening to events)
- Remove token from local store when getting a `401 Unauthorized` response.

## Example

```js
  const embed = Talk.render(document.getElementById('talkStreamEmbed'), opts);

  // trigger a login with optional token.
  embed.login(token);

  // trigger a logout.
  embed.logout();

  // listen to events (in this case all events).
  embed.on('**', function(value) {
    console.log(this.event, value);
  });
```


wyattjoh (approved):
Besides comments I've already made on Slack related to potential security issues of the event emitter designed as is, this looks good!

PR #818, opened 2017-07-31
Adds support for disabling the babel-polyfill on the parent embed.js

wyattjoh:
## What does this PR do?

Removes the polyfill from everything except the main embed.


alinhle:
Why do we need this?

wyattjoh:
@cvle on pages where they already include the babel polyfill, this breaks the entire page. The polyfill can only be included once. _Some_ news organizations already have it included on their parent pages, hence, this ensures it can run along side it as the polyfill pollutes the global namespace in order to support things.

alinhle:
@wyattjoh Using `babel-plugin-transform-runtime` is also problematic because instance methods of builtin types are not polyfilled like `myArray.some()` or `myString.startWith()` are not polyfilled. These are methods that we already use and will not work in unsupported browsers. 

The only time I can imagine conflict is when you include `coral-embed`. So instead of removing the `babel-polyfill` from every plugin and target, why not just apply it to `coral-embed`? Maybe even better is that we provide a switch `NO_POLYFILL` to disable the polyfill on `coral-embed` in case the parent page already brings the polyfills.

alinhle (approved):
Looks good! :-)

PR #824, opened 2017-08-01
Fix Safari issues for popups

alinhle:
## What does this PR do?
- Closing a popup causes tons of errors on Safari. This PR fixes it.


kgardnr (approved):
:100:!

Issue #808, opened 2017-07-27
Asset closedAt is improperly evaluated in embed stream


offensivepolitics:
### Qualifications
Asset A has a closedAt date of two weeks in the future. Asset A shows state as Open in the admin
We are using release 2.4.0

### Expected behavior
Loading the embed for Asset A should render the stream.

### Actual behavior
Loading the embed for Asset A reports the stream is closed

### Steps to reproduce behavior
1. Set the closedAt date for an asset to something in the future
2. Load the embed for the asset

### Possible fix
I believe the default was to set closedAt to NULL on creation, but at some point the auto close set closedAt to an actual date.

This line (https://github.com/coralproject/talk/blob/master/client/coral-embed-stream/src/components/Stream.js#L103) suggests that an asset with *any* value for closedAt will show up as closed, which is our experience. 
Changing this line to instead read:
`const open = asset.closedAt === null || new Date(asset.closedAt) > new Date();`
lets assets with either an empty closedAt date, or closedAt in the future, render properly.

wyattjoh:
Good catch @offensivepolitics, we should be able to resolve that client side soon.

kgardnr:
PT Ref: https://www.pivotaltracker.com/story/show/149789244

kgardnr:
PR: https://github.com/coralproject/talk/pull/827

wyattjoh:
The PR which resolved this issue was merged @offensivepolitics, re-open if this particular issue still exists.

PR #827, opened 2017-08-01
Correctly determine the status of the asset

alinhle:
## What does this PR do?
- Correctly determine the status of the asset


wyattjoh:
Suggestion @cvle, could you add a timeout if the closedAt time is defined and close the stream at that point? That way the UI could be like "oops! the commenting on this stream is now closed."

alinhle:
Done, automatic timeout closing added :-)

wyattjoh:
Resolves #808

kgardnr (approved):

PR #847, opened 2017-08-10
Remove Tag should keep remaining tags

alinhle:
## What does this PR do?
Currently removing a tag always removes all available tags. E.g. removing featured tag also removes the staff tag. This fixes this behavior to keep the remaining tags.


wyattjoh (approved):
Good catch!

PR #849, opened 2017-08-10
Pluggable modqueues

alinhle:
## What does this PR do?
- Support plugin defined mod queues by adding a field `modQueues` in the `index.js` of client side plugins

## Example

```js
modQueues: {
  newQueueKey: {
    // name
    name: 'My Queue Name',                    

    // material design icon
    icon: 'star',        

    // Filter by tags
    tags: ['MY_TAG'],                       
    
    // Filter by statuses
    statuses: ['NONE', 'PREMOD', 'ACCEPTED', 'REJECTED'],   

   // Filter by comment containing action_type
    action_type: 'FLAG',                    
  },
},
```

## How To Test
E.g. add this to the featured comments plugin:
```js
modQueues: {
  featured: {
    tags: ['FEATURED'],
    icon: 'star',
    name: 'Featured',
  },
},
```

### DIFF
--- a/client/coral-admin/src/AppRouter.js
+++ b/client/coral-admin/src/AppRouter.js
@@ -37,27 +37,11 @@ const routes = (
       <Route path='moderate' component={ModerationLayout}>
         <IndexRoute components={Moderation} />
 
-        <Route path='all' components={Moderation}>
-          <Route path=':id' components={Moderation} />
-        </Route>
-        <Route path='new' components={Moderation}>
-          <Route path=':id' components={Moderation} />
-        </Route>
-        <Route path='approved' components={Moderation}>
-          <Route path=':id' components={Moderation} />
-        </Route>
-        <Route path='premod' components={Moderation}>
-          <Route path=':id' components={Moderation} />
-        </Route>
-        <Route path='rejected' components={Moderation}>
-          <Route path=':id' components={Moderation} />
-        </Route>
-        <Route path='reported' components={Moderation}>
-          <Route path=':id' components={Moderation} />
-        </Route>
+        <Route path=':tabOrId' components={Moderation} />
 
-        <Route path=':id' components={Moderation} />
-
+        <Route path=':tab' components={Moderation}>
+          <Route path=':id' components={Moderation} />
+        </Route>
       </Route>
     </Route>
   </div>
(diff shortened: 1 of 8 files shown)


kgardnr (approved):
💯!!

leeeandroo:
Hello @cvle ,

I have a small question about this implementation, could you help me?

How can I use a translation file to use on the queue name?
I've tried to create a translations.yml file, and use it on the queue, but it's not loading. My guess is that the queue config is loaded before the translations.

Thanks,

wyattjoh:
You should be able to do something like this:

```js
import t from 'coral-framework/services/i18n';

// ...
modQueues: {
  featured: {
    tags: ['FEATURED'],
    icon: 'star',
    name: t('modqueue.featured'),
  },
},
// ...
```

As we've done with our own internal queues. Try using an existing translation first, to see if the translations are loading at all, then experiment to see if you can get your own translation loading.

leeeandroo:
Hi @wyattjoh , thanks for your answer.

Thats exactly what I have on my code. But something does not work with my own translation.

On my plugin folder `talk-plugin-test`, I have a `index.js` file with this code:

`module.exports = {  };`

And then on the `client` folder, I have the `modQueues.js
`
```
import t from 'coral-framework/services/i18n';

const modQueues = {
    test: {
        tags: ['TEST'],
        icon: 'reply',
        name: t('modqueue.test'),
    }
}

export default modQueues
```

And on my` translations.yml`:

```
en:
  modqueue:
    test: "Test"
```

And load this on` index.js`:

```
import translations from './translations.yml';
import modQueues from './modQueues';

export default {
    translations,
    modQueues
};
```
  
If I use a language key from the core locales, It works. But If I try to use my key `modqueue.test`, It's not load and I can see an error on console: `en.modqueue.test language key not set`
Also, If I try to use some translation from another plugin, It's not work too.

My guess is that maybe the `plugins.getTranslations()` is called after the `plugins.getModQueueConfigs()`.

Does it make sense for you? Maybe I'm missing something.

Thanks!

alinhle:
Hi @leeeandroo, you have assessed the situation perfectly. We're going add a feature to lazily evaluate the modqueue config from plugins, so that custom translations can be included. You can track the progress here: https://www.pivotaltracker.com/story/show/154447869.

kgardnr:
Also if this is an urgent need @leeeandroo, feel free to submit a PR for this and we will review it and merge! I think with our current roadmap, we won't get to this for several weeks FYI.

Thanks!

leeeandroo:
I would like to try this. 
Any tip about this implementation? Are there something similar on the core that I can have a look before start coding? @cvle @wyattjoh @kgardnr

PR #863, opened 2017-08-18
Front-End Optimizations

alinhle:
### Implement optimisations using `shouldComponentUpdate` and our graphql framework.
- `withFragments` now filters query data using its fragments (requires `data` property from `withQuery` as a current workaround to https://github.com/apollographql/graphql-anywhere/issues/38)
- `withFragments` now passes previous query data if not changed
- `withFragments` intelligently detect whether it is necessary to render the component.
- The passed props from `withQuery` are now only mutated when necessary
- The passed props from `withMutation` are now only mutated when necessary
- `Slot` now accepts a `queryData` property, where props from the graphql query should be passed.
- `Slot` detect if the plugin component defines fragments and only passes the props from `queryData` that also has fragments set.
- `Slot` passes all `queryData` to the plugin component if no fragments are defined on the plugin component. In this case any access to the query data properties will result in a warning during development.
- Refactored Comment so it works as a container. (See `withAnimateEnter`)
- Optimize rendering Comment for `activeReplyBox` change (redux).
- Refactored the stream tab panel into `StreamTabPanel` container and related components.
- Optimize `Slot`, `IfSlotIsEmpty`, `IfSlotIsNotEmpty` and `StreamTabPanel` to render only when necessary.
- `withTags` and `withReactions` now accept a second optional parameter `{fragments}` allowing additional fragments to be defined.
- Use `hoistStatics` for every HOC
- Integrate some existing Slots into our graphql framework.
- Some smaller refactorings.
- Adapted to changed `networkStatus` behavior of apollo.
- Fixed issues where fragments where incorrect, not defined, or incomplete (`Stream`, `Comment`, `talk-plugin-featured-comments`, `talk-plugin-offtopic`, `talk-plugin-permalink`...)
- Components using `ClickOutside` now only trigger a state change when necessary (`talk-plugin-permalink`, `talk-plugin-flags`)


kgardnr (approved):

plugins/talk-plugin-offtopic/client/containers/OffTopicTag.js
```diff
@@ -0,0 +1,15 @@
+import {gql} from 'react-apollo';
+import OffTopicTag from '../components/OffTopicTag';
+import {withFragments} from 'plugin-api/beta/client/hocs';
+
+export default withFragments({
+  comment: gql`
+    fragment TalkOfftopic_OffTopicTag_comment on Comment {
```
okbel, line 7: This `fragment` query will live in each plugin that queries the tags. Could that trigger any issue?

okbel (approved):

okbel:
This is amazing 🙌 😍

plugins/talk-plugin-offtopic/client/containers/OffTopicTag.js
alinhle: I don't foresee that this query will cause any issues. Multiple similar fragments might be added to the query, but graphql can handle this, and the fragment is needed to filter, optimise, and decouple the component.

PR #868, opened 2017-08-21
New Slot: commentAuthorName

okbel:
## What does this PR do?
- Story:   https://www.pivotaltracker.com/story/show/150178457

- Adds a new Slot `commentAuthorName`
- Removes old code
- New Recipe: Author Name Recipe: **Member Since**: https://github.com/coralproject/talk-recipes/pull/2

### Author Name Recipe: **Member Since** - Preview
screen shot 2017-08-21 at 12 15 26 pm


client/coral-embed-stream/style/default.css
```diff
@@ -238,14 +238,8 @@ body {
...
 
-.talk-plugin-author-name-text {
-  display: inline-block;
-  margin: 10px 5px 10px 0;
-  font-weight: bold;
-}
-
-.talk-plugin-author-name-bio-flag {
-  float: right;
+.talk-stream-comment-header {
```
alinhle: Avoid adding styles to `default.css` as it is deprecated.

plugins/talk-plugin-featured-comments/client/components/Comment.js
```diff
@@ -22,9 +22,14 @@ class Comment extends React.Component {
...
 
         <div className={cn(`${pluginName}-comment-username-box`)}>
-          <strong className={cn(styles.username, `${pluginName}-comment-username`)}>
-            {comment.user.username}
-          </strong>
+
+          <Slot
+            className={cn(styles.username, `${pluginName}-comment-username`)}
+            fill="commentAuthorName"
+            comment={comment}
```
alinhle: Slot Requirements:
- [x] The slot component needs the queryData `{root, comment, asset}`
- [x] The slot fragments needs to be added to the fragments tree.
- [x] Current workaround also requires `data={data}` to be passed to the Slot.
- [x] Missing default component

alinhle (changes requested):
See above

client/coral-embed-stream/style/default.css
okbel: Got it. I didn't know about this. 👍

okbel:
Changes added. 👍 

I've noticed this: 
screen shot 2017-08-21 at 2 52 49 pm

Shouldn't `withFragments` also extend the `update` with the fields that will be updated too?

//cc @cvle

alinhle:
I guess you are right, the comments fragment from the embed stream should be added to the post comment mutation.

alinhle:
@okbel I couldn't reproduce the fragments warning and these also seem like an older issue that has been already resolved. How did you get these warnings?

okbel:
I'm still getting that warning. Maybe because I'm running an older version of Apollo? @cvle

alinhle:
I added the slot to the featured comments fragments

alinhle (approved):

PR #869, opened 2017-08-21
Reply Badge - Admin Side

okbel:
## What does this PR do?
- Shows a badge if a comment is a reply on the admin side
- Adds hasParent on Comment to the graph
- Adds Badge to `coral-ui`
- Fixes bugs with flagged comments not showing badge and flagbox

https://www.pivotaltracker.com/story/show/144895441


graph/typeDefs.graphql
```diff
@@ -339,6 +339,9 @@ type Comment {
 
   # describes how the comment can be edited
   editing: EditInfo
+
+  # Indicates if it has a parent
+  hasParent: Boolean
```
alinhle, line 344: @wyattjoh Would you say it makes more sense to expose `parent_id: String` instead? I think there were other occasions were we just needed the `parent_id` but used the `parent` endpoint which fetches the whole comment from mongodb. Then we also know `hasParent` as `hasParent = comment.parent_id !== null`.

client/coral-admin/src/components/ReplyBadge.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import {Badge} from 'coral-ui';
+import t from 'coral-framework/services/i18n';
+
+const ReplyBadge = ({hasParent}) => hasParent ?
```
alinhle: Imo the logic for checking `hasParent` should not lie within `ReplyBadge`. It should not receive a `hasParent` property because what does that mean for the `ReplyBadge`? What is the parent of a reply badge? That's logic that should reside at the parent component that renders the `ReplyBadge`.

client/coral-admin/src/components/UserDetailComment.js
```diff
@@ -56,7 +57,11 @@ class UserDetailComment extends React.Component {
               ? <span>&nbsp;<span className={styles.editedMarker}>({t('comment.edited')})</span></span>
               : null
             }
-            <CommentType type={commentType} className={styles.commentType}/>
+
+            <div className={styles.badgeBar}>
+              <ReplyBadge hasParent={comment.hasParent} />
```
alinhle: Should be `{comment.hasParent && <ReplyBadge />}`. See above.

client/coral-admin/src/routes/Moderation/components/Comment.js
```diff
@@ -107,6 +108,7 @@ class Comment extends React.Component {
                 </ActionsMenu>
               }
               <div className={styles.adminCommentInfoBar}>
+                <ReplyBadge hasParent={comment.hasParent} />
```
alinhle: Should be `{comment.hasParent && <ReplyBadge />}`. See above.

alinhle (changes requested):
See above

client/coral-admin/src/components/UserDetailComment.js
alinhle: Or a `<CommentBadgeBar comment={comment} />` that contains the above logic.

client/coral-admin/src/routes/Moderation/components/Comment.js
alinhle: Or a `<CommentBadgeBar comment={comment} />` that contains the above logic.

okbel:
Changes added! @cvle

graph/typeDefs.graphql
wyattjoh: We want to encourage people to batch requests together. It's my opinion that instead, we do not provide the parent id at the top level.

It would encourage the use of the public api in a desired way, and force the user to think about batching because they can't get the ID otherwise.

alinhle (approved):

PR #870, opened 2017-08-21
Optimizations and User Sort

wyattjoh:
## What does this PR do?

- Adds support for tracking of action mods and new replies by emitting to a new server side event emitter.
- Added tests surrounding emitter usage.
- Renamed sort fields of CHRONO -> ASC/DSC.
- Removed unused and outdated E2E code.
- Removed unused dependancies.
- Moved Front End dependancies from devDependancies -> dependancies.
- Optimized comment count edges.
- Action based queries involving comments no longer grab all the things from the comments collection, instead utilized a smart, cached, indexed field on the comments to directly query them.
- Replaced docker image with `8-alpine` from `8` (due to new 🎊 full JS bcrypt library), reducing resulting image size by ~65% or cutting  off 600MB. **Yes, you heard that right, 65% smaller**.
- Added new `sortBy` to comment/reply edges which allows comments to be sorted in more flexible ways.
- Added more to plugin reaction HOC to support new `groupBy` functionality.
- Support the new sorting params in stream
- Add a `SortOption` presentational Component, a `withSortOption` HOC providing the props, and a `createSortOption` factory method combining the previous two to `plugin-api`
- Added new slots `viewingOptionsSort` and `viewingOptionsFilter`. Removed slot `viewingOptions`.
- Adapted `talk-plugin-offtopic`
- Added plugins `talk-plugin-sort-most-liked`, `talk-plugin-sort-most-replied`, `talk-plugin-sort-newest`, `talk-plugin-sort-oldest`
- Added support for `childFactory` and `component` props to the `Slot` component.
- Changed `sort` to `sortOrder`.

*Still in progress, expect more to come here...*

## Migration

In order for these non-atomic updates to be maintained, we require that after this feature is made available, you run:

```bash
./bin/cli verify db --fix
```

At least once to update the cached values on the models that depend on them.


.vscode/settings.json
```diff
@@ -0,0 +1,8 @@
+{
```
alinhle: Should this VS specific code be committed?

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8
+FROM node:8-alpine
```
alinhle, line 1: alpine :-) <3

graph/typeDefs.graphql
```diff
@@ -241,10 +244,14 @@ input CommentsQuery {
   limit: Int = 10
 
   # Skip results from the last created_at timestamp.
-  cursor: Date
+  cursor: Cursor
 
-  # Sort the results by created_at.
-  sort: SORT_ORDER = REVERSE_CHRONOLOGICAL
+  # Sort the results by from largest first.
+  sort: SORT_ORDER = DESC
```
alinhle: Hm I intuitively thought that this should be called `sortOrder` instead.

kgardnr (approved):

PR #874, opened 2017-08-23
Dependency Management

alinhle:
#### This PR fixes the following issues due to global dependencies:
- A lot of initialisation happens during import time.
- Problematic circular dependencies.
- Writing isolated tests is difficult or impossible in some cases
- Difficult to reason about macro code behavior
- Encourages direct coupling

#### The PR removes global dependencies on the following services:
- Apollo client
- Redux Store
- GraphQL Registry
- Pym
- Rest API
- Plugins
- Notification
- Storage (localStorage)
- History (Router history)

#### Remaining global dependencies
- Access to the URL
- i18n (pending refactor)
- perm (pending refactor)

#### Steps taken
- Removed framework reducers as they are currently convoluted with the embed stream and are not reusable.
- Refactored plugins code into a service
- Refactored rest calls into a service
- Refactored Apollo initialisation to decouple from the Redux Store vice versa
- Refactored graphql registry into a non singleton service
- Refactored localStorage calls into a service
- Refactored router history into a service
- Refactored notification code into a service that works in admin and embed
- Added `getErrorMessages` util that extracts and translates error message as returned from graphql into an array of strings, which can be passed to the notification service.
- Introduced a `bootstrap` service that creates and wires the dependencies by calling `createContext`
- Introduced `TalkProvider` that passes the context to the React components
- The context is also passed to redux actions using `redux-thunk` and its `withExtraArgument` method.
- The context currently contains the following services: client, pym, plugins, eventEmitter, rest, graphqlRegistry, notification, storage, history
- React DevTools was causing warnings in our plugins api. This PR contains a fix.


client/coral-framework/services/plugins.js
```diff
@@ -0,0 +1,179 @@
...
+  plugins.forEach((plugin) => {
+    const slots = plugin.module.slots;
+    slots && Object.keys(slots).forEach((slot) => {
+      slots[slot].forEach((component) => {
+
+        // Attach plugin name to the component
+        component.talkPluginName = plugin.name;
+
+        // Attach uuid to the component
+        component.talkUuid = uuid();
```
wyattjoh, line 68: For abbreviations, you should use `component.talkUUID` instead to keep access consistent.

alinhle: @wyattjoh I'd love to do that, because that is my own personal preference, but at least on the frontend side we have not been following this convention at all. Like `URL` in `baseUrl` and `buildUrl` or `URI` in `redirectUri`. That made me think that our convention was different.

If this is our convention we should make another PR and go through our code and make sure we follow it.

kgardnr (approved):
I'm good to merge this if @okbel and @wyattjoh agree!

PR #889, opened 2017-08-28
My Comment Reactions

okbel:
## What does this PR do?
MY COMMENT REACTIONS/REPLIES | As a commenter, I want to see in My Profile how many likes/reactions and replies my comments have received so that I can know if I want to revisit the comment section

https://www.pivotaltracker.com/story/show/150595787

### DIFF
--- a/client/coral-framework/utils/index.js
+++ b/client/coral-framework/utils/index.js
@@ -184,3 +184,11 @@ export function getShallowChanges(a, b) {
   return union(Object.keys(a), Object.keys(b))
     .filter((key) => a[key] !== b[key]);
 }
+
+export function getTotalReactionsCount(actionSummaries) {
+  return actionSummaries
+    .filter((s) => s.__typename !== 'FlagActionSummary')
+    .reduce((total, summary) => {
+      return total + summary.count;
+    }, 0);
+}
--- a/client/coral-settings/containers/ProfileContainer.js
+++ b/client/coral-settings/containers/ProfileContainer.js
@@ -108,6 +108,11 @@ const CommentFragment = gql`
     nodes {
       id
       body
+      replyCount
+	    action_summaries {
+        count
+        __typename
+      }
       asset {
         id
         title
--- a/client/coral-settings/containers/ProfileContainer.js
+++ b/client/coral-settings/containers/ProfileContainer.js
@@ -109,7 +109,7 @@ const CommentFragment = gql`
       id
       body
       replyCount
-	    action_summaries {
+      action_summaries {
         count
         __typename
       }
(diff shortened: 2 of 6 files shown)


kgardnr:
Hey @okbel the only thing is to singularize "reactions" and "replies" when there's only 1: 
my_reactions_replies

alinhle:
Also it should be hidden when the count is 0.

client/coral-framework/utils/index.js
```diff
@@ -184,3 +184,11 @@ export function getShallowChanges(a, b) {
   return union(Object.keys(a), Object.keys(b))
     .filter((key) => a[key] !== b[key]);
 }
+
+export function getTotalReactionsCount(actionSummaries) {
+  return actionSummaries
+    .filter((s) => s.__typename !== 'FlagActionSummary')
```
alinhle, line 190: This is not a reliable method to determine `reactions`. There is not only `FlagActionSummary` but also `DontAgreeActionSummary` and on top of that not all plugin action summaries are sure to be reactions. What's your stance on this @wyattjoh?

kgardnr:
That's a good idea @cvle, definitely reduces clutter to the eye as well - @IAmSamHankins has the 0 counts showing, but we can chat with her later.

alinhle:
I think it was a requirement from @losowsky but maybe not? :-D Anyway I think it's a good idea.

alinhle:
I'd also add more space between _reactions_ and _replies_ @IAmSamHankins to have that more visually separated.

okbel:
We decided that if the count was 0 it should still be shown. We can re-check this. @IAmSamHankins  thoughts?

client/coral-framework/utils/index.js
okbel: Seems fragile for me too as well. But I can't see other filtering option based on this structure. *Interested in what Wyatt has to say about this*

theysamtheirs:
@okbel, yes we should still show the counts even if zero. to address "clutter" issue, let's have anything displaying 0 with a color of #9E9E9E (lighter for zero, so as to be less distracting and also draw more attention when there are replies or reactions.)

okbel:
@IAmSamHankins Ready to be checked! :)

okbel:
Styles and translations added. // @kgardnr

kgardnr:
Awesome, thanks @okbel!

@IAmSamHankins is this good to go then?

reactions_replies_take_two

theysamtheirs:
LGTM!

kgardnr:
Cool will just wait for @wyattjoh's feedback before merging.

client/coral-framework/utils/index.js
wyattjoh: I think we could use a GraphQL union type? But I'm a bit unsure..

wyattjoh: For now we can leave this as is. This might be a bigger task to classify all reaction types properly.

wyattjoh (approved):

PR #894, opened 2017-08-29
Added graph API for assets

wyattjoh:
## What does this PR do?

- Provides all the route methods for assets as graph mutations
- Provides more query params to the `assets` edge

Related: https://www.pivotaltracker.com/story/show/149459989

Automated tests are already implemented to verify this functionality, this PR was not meant to include the client side implementation.


graph/typeDefs.graphql
```diff
@@ -714,7 +742,7 @@ type RootQuery {
   comment(id: ID!): Comment
 
   # All assets. Requires the `ADMIN` role.
-  assets(query: AssetsQuery): [Asset]
+  assets(query: AssetsQuery): AssetConnection
```
alinhle, line 824: I think the resolver has not been modified to return a `AssetConnection` type yet.

graph/typeDefs.graphql
```diff
@@ -884,6 +912,54 @@ input RejectUsernameInput {
   message: String!
 }
 
+# Configurable settings that can be overridden for the Asset.
+input AssetSettingsInput {
```
alinhle: Is it possible to only change a subset of Asset Settings?

alinhle: What happens if I only provide a subset of the fields

graph/typeDefs.graphql
```diff
@@ -884,6 +912,54 @@ input RejectUsernameInput {
...
+  # questionBoxContent is the content of the Question Box.
+  questionBoxContent: String
+
+  # questionBoxIcon is the icon for the Question Box.
+  questionBoxIcon: String
+}
+
+# UpdateAssetStatusInput contains the input to change the status of a comment as
+# it relates to being open/closed for commenting.
+input UpdateAssetStatusInput {
```
alinhle, line 1017: See above

graph/typeDefs.graphql
```diff
@@ -1044,6 +1120,13 @@ type RootMutation {
   # Removes a tag.
   removeTag(tag: ModifyTagInput!): ModifyTagResponse!
 
+  # Updates settings on a given asset.
+  updateAssetSettings(id: ID!, input: AssetSettingsInput!): UpdateAssetSettingsResponse
```
alinhle: Shouldn't that be `UpdateAssetSettingsResponse!`?

graph/typeDefs.graphql
```diff
@@ -1044,6 +1120,13 @@ type RootMutation {
...
+
+  # Updates the status of an asset allowing you to close/reopen an asset for
+  # commenting.
+  updateAssetStatus(id: ID!, input: UpdateAssetStatusInput!): UpdateAssetStatusResponse
```
alinhle: See above

graph/typeDefs.graphql
```diff
@@ -1044,6 +1120,13 @@ type RootMutation {
   # Removes a tag.
   removeTag(tag: ModifyTagInput!): ModifyTagResponse!
 
+  # Updates settings on a given asset.
```
alinhle, line 1293: This mutation is restricted >:-D ?

graph/typeDefs.graphql
wyattjoh: I've updated the comment to reflect the behaviour. You should update all fields when posting, you cannot specify a subset. This is because for some fields, the `null` value is used.

graph/typeDefs.graphql
wyattjoh: I've updated the comment to reflect the behaviour. You should update all fields when posting, you cannot specify a subset. This is because for some fields, the `null` value is used.

services/assets.js
```diff
@@ -108,19 +108,57 @@ module.exports = class AssetsService {
...
+      } else {
+        assets.merge({
+          created_at: {
+            $gt: cursor,
+          },
+        });
+      }
+    }
+
+    return assets.limit(limit);
```
alinhle: The result  should be sorted by `createdAt` depending on `sortOrder` otherwise the pagination is wrong.

alinhle (approved):
Looks good!

PR #896, opened 2017-08-29
Fix endless interval

alinhle:
## What does this PR do?
- Fix bug when you quickly toggle multiple refresh of the login popup than after closing the popup it will lead to endless login checks.


kgardnr (approved):
Woot! :100:

PR #897, opened 2017-08-29
Sort enhancements

alinhle:
## What does this PR do?
- Always refetch when changing `sortOrder` or `sortBy` to get most current state.
- Only exclude newly added top-level comments in sorts other than _newest first_
- Move Spinner down to `TabContent` to make the UX experience nicer.


kgardnr (approved):
🎉

PR #899, opened 2017-08-29
Added graph API for settings

wyattjoh:
## What does this PR do?

- Increased fields exposed on the `settings` edge to match REST api.
- Added `updateSettings` mutation.
- Added `updateWordlist` mutation.

Related: https://www.pivotaltracker.com/story/show/150488299

Automated tests are already implemented to verify this functionality, this PR was not meant to include the client side implementation.


graph/resolvers/settings.js
```diff
@@ -1,3 +1,21 @@
...
+// eyes.
+const PROTECTED_SETTINGS = [
+  'premodLinksEnable',
+  'autoCloseStream',
+  'wordlist',
+  'domains',
+];
+
+// decorate the fields on the settings resolver with a permission check.
+decorateWithPermissionCheck(Settings, VIEW_PROTECTED_SETTINGS, ...PROTECTED_SETTINGS);
```
alinhle: Nice^^ 

My suggestion is to allow different permissions per field so that it is more flexible: 

```js
const protect = {
  premodLinksEnable: [VIEW_PROTECTED_SETTINGS]
  autoCloseStream: ...
};

decorateWithPermissionCheck(Settings, protect);
```

graph/typeDefs.graphql
```diff
@@ -977,6 +1035,91 @@ type EditCommentResponse implements Response {
   errors: [UserError!]
 }
 
+# UpdateSettingsInput is the input used to input the global site settings. This
+# will override the existing settings, so all fields must be included.
+input UpdateSettingsInput {
```
alinhle, line 1061: Same here as the other PR, can I submit just a subset of the input and only modify the subset keeping the rest the same?

graph/typeDefs.graphql
```diff
@@ -977,6 +1035,91 @@ type EditCommentResponse implements Response {
...
+# UpdateSettingsResponse contains any errors that were rendered as a result
+# of the mutation.
+type UpdateSettingsResponse implements Response {
+
+  # An array of errors relating to the mutation that occurred.
+  errors: [UserError!]
+}
+
+# UpdateWordlistInput is the list of words that composes the Wordlist.
+input UpdateWordlistInput {
```
alinhle, line 1127: This means both `banned` and `suspect` must be provided right?

graph/typeDefs.graphql
```diff
@@ -1044,6 +1187,12 @@ type RootMutation {
   # Removes a tag.
   removeTag(tag: ModifyTagInput!): ModifyTagResponse!
 
+  # updateSettings will update the global settings.
```
alinhle, line 1211: This mutation is restricted >:-D

graph/typeDefs.graphql
```diff
@@ -1044,6 +1187,12 @@ type RootMutation {
...
+  updateSettings(input: UpdateSettingsInput!): UpdateSettingsResponse!
+
+  # updateWordlist will update the given Wordlist.
```
alinhle: This mutation is restricted >:-D

graph/typeDefs.graphql
wyattjoh: That's correct.

graph/resolvers/settings.js
wyattjoh: Good idea, will implement.

graph/typeDefs.graphql
wyattjoh: Same as the other one, all fields are required because we can't distinguish in-between a null/undefined answer.

alinhle (approved):
👍

PR #900, opened 2017-08-29
Fixed errors when launching not-defined commands

wyattjoh:
## What does this PR do?

- When a cli command is incorrectly entered, the application will not crash with a reference error.
- Closest match commands will be suggested when the correct command wasn't found.

Related: https://www.pivotaltracker.com/story/show/150519132

## How do I test this PR?

1. Try `./bin/cli servee`
2. See it suggest `cli serve`, and not crash!

### DIFF
--- a/bin/cli
+++ b/bin/cli
@@ -4,8 +4,9 @@
  * Module dependencies.
  */
 
-// const util = require('./util');
 const program = require('./commander');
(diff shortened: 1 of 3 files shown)


kgardnr (approved):
Nice!

PR #903, opened 2017-08-30
Paginate flagged users and more

alinhle:
## What does this PR do?
- Implement `UserConnection` on the graph
- Allow property `statuses` for filtering in a user query.
- Refactor `FlaggedUsers` related code
- Refactor Community styles
- Add transition effect
- Implement graphql mutation updators.
- Always refetch when switching to flagged Users.

## How to test this PR?
- Have more than 10! flagged users to see pagination ^.^
- See transition effect when removing a user
- See refetch taking place whenever switching to flagged user view.


graph/typeDefs.graphql
```diff
@@ -730,8 +751,8 @@ type RootQuery {
   # role.
   me: User
 
-  # Users returned based on a query.
-  users(query: UsersQuery): [User]
+  # Users returned based on a query. Requires the `ADMIN` role.
```
wyattjoh, line 754: Shouldn't need to indicate role here.

alinhle: Well yeah but you need special privilege to access it, not sure how to describe that. (Just copied what we used before)

alinhle: Requires the `SEARCH_OTHER_USERS` permission?

wyattjoh: I would suggest we remove all the role requirements from the comments to be honest until we can get custom annotations.

alinhle: Custom annotations, is that a thing?

kgardnr (approved):

PR #904, opened 2017-08-30
Exclude Ignored users in Featured Comments Plugin

okbel:
## What does this PR do?
- Exclude Ignored users in Featured Comments Plugin

https://www.pivotaltracker.com/story/show/150605682

### DIFF
--- a/plugins/talk-plugin-featured-comments/client/containers/TabPane.js
+++ b/plugins/talk-plugin-featured-comments/client/containers/TabPane.js
@@ -102,6 +102,7 @@ const enhance = compose(
             tags: ["FEATURED"]
             sortOrder: $sortOrder
             sortBy: $sortBy
+            excludeIgnored: true
           }
           deep: true
         ) @skip(if: $hasComment) {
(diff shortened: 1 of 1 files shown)


plugins/talk-plugin-featured-comments/client/containers/TabPane.js
```diff
@@ -102,6 +102,7 @@ const enhance = compose(
             tags: ["FEATURED"]
             sortOrder: $sortOrder
             sortBy: $sortBy
+            excludeIgnored: true
```
alinhle: Should be `excludeIgnored: $excludeIgnored`, taking the variable from the query directly, which is set to false when user is not logged in otherwise true.

okbel: Got it. I tried that and it didn't work. I will check again. One sec.

okbel: Perfect 👍

okbel: Thanks for the clarification

alinhle (approved):
I didn't test, but looks good!

PR #907, opened 2017-08-31
Hide menu when clicking on a sort option

alinhle:
## What does this PR do?
- Hide menu when clicking on a sort option


kgardnr (approved):
Works perfectly!

PR #910, opened 2017-08-31
Make rendering `edit` pure

alinhle:
## What does this PR do?
- Make rendering `edit` pure
- _edit button_ should now disappear after expiration


kgardnr (approved):
:100:

PR #911, opened 2017-08-31
Create both a PreMod and a New Queue when stream is switched from one status to another

okbel:
## What does this PR do?
- Creates both a PreMod and a New Queue when stream is switched from one status to another.

## Edge Case
News Room enables premod every night, and they switch to POST every day. If there are comments with `PREMOD` status we should provide a queue for them. 

## How do I test this PR?
- Enable `PREMOD`
- Post a comment
- Disable `PREMOD`
- Go to Moderation (Both queues should appear and as soon as you reject or accept the PREMOD comment the `PREMOD` queue should disapear)


kgardnr:
Really nice work @okbel! As soon as @IAmSamHankins approves I'll merge it :)

client/coral-admin/src/routes/Moderation/containers/Moderation.js
```diff
@@ -190,7 +190,7 @@ class ModerationContainer extends Component {
   };
 
   render () {
-    const {root, root: {asset, settings}, data} = this.props;
+    const {root, root: {asset}, data, settings} = this.props;
```
alinhle: ^ Let's change it back an use `network-only` fetchPolicy as discussed in the Arch meeting

theysamtheirs:
LGTM!!!

kgardnr (approved):

PR #914, opened 2017-09-01
Top-level comments by ignored user are tombstoned, but replies should be as well

okbel:
## What does this PR do?
- Solves this issue https://www.pivotaltracker.com/story/show/150605584
- Requesting me from `Comment` updates the props and triggers a re render.

_Note: I didn't move `commentIsIgnored` fn because `AllCommentsPane` component is using it_


kgardnr (approved):
Woot yes!

kgardnr:
I'll wait until @cvle gives the ok before I merge :)

alinhle (dismissed):
We should remove all traces of `commentIsIgnored` because it is unpure and creates coupling between Stream Props with any other components that uses this function. It is the main cause of this bug. One way is to let the comment itself detect if it is ignored and render a tombstone instead.

kgardnr:
I'm going to merge this PR and create a new ticket with @cvle's suggestion above.

PR #921, opened 2017-09-05
Change Websocket URI

wyattjoh:
## What does this PR do?

- Adds support for a new runtime variable that is injected to the client to specify a different root url to use for subscriptions.

`process.env.TALK_WEBSOCKET_LIVE_URI` when set should point to the endpoint (path included) to the web socket server.

Example, the Talk API is being served out of `https://sitea.com`. Specifying `TALK_WEBSOCKET_LIVE_URI=wss://siteb.com/api/v1/live` will allow you to have the server use that instead.

 **Warning: if used without managing the auth state manually, auth cannot be persisted**

### DIFF
--- a/client/coral-framework/constants/config.js
+++ b/client/coral-framework/constants/config.js
@@ -1,3 +1,0 @@
-const configElement = document.querySelector('#data');
-
-export const CONFIG = JSON.parse(configElement ? configElement.textContent : '{}');
--- a/client/coral-framework/constants/config.js
+++ b/client/coral-framework/constants/config.js
@@ -1,0 +1,3 @@
+const configElement = document.querySelector('#data');
+
+export const CONFIG = JSON.parse(configElement ? configElement.textContent : '{}');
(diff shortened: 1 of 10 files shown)


client/coral-framework/constants/config.js
```diff
@@ -0,0 +1,3 @@
+const configElement = document.querySelector('#data');
+
+export const CONFIG = JSON.parse(configElement ? configElement.textContent : '{}');
```
alinhle: For now just move this logic inside of bootstrap, if we need this config somewhere else, we'll pass it through the context created by the bootstrap service.

alinhle (changes requested):

client/coral-framework/services/bootstrap.js
```diff
@@ -16,6 +16,25 @@ import {createStorage} from 'coral-framework/services/storage';
 import {createHistory} from 'coral-framework/services/history';
 
 /**
+ * getStaticConfiguration will return a singleton of the static configuration
+ * object provided via a JSON DOM element.
+ */
+const getStaticConfiguration = (() => {
+  let staticConfiguration = null;
```
alinhle, line 23: Usually this will not be called more than once, I wonder if it's worth using a singleton pattern here.

alinhle (approved):

client/coral-framework/services/bootstrap.js
wyattjoh: The idea was that the remainder of the files here seemed like they were returned as functions, and I didn't want to introduce any other side effects by loading something that really should never change every call.

PR #936, opened 2017-09-08
Init, Introspection service, pymStorage, and talk-plugin-remember-sort

alinhle:
## What does this PR do?
- Introduce `init` property for plugins that is called with the context and resolved before render.
- The `bootstrap` service now accepts a `preInit`, and `init` option.
- Use new bootstrap options to load config _before_ rendering the embed.
- Implement a `introspection` service with a `isValidEnumValue` method.
- Implement a `pymStorage` which uses the `localStorage` of the parent and is available on the context
- Implement a `talk-plugin-remember-sort` that saves the last user sorting choice into `pymStorage` and loads it before render.


kgardnr (approved):

plugins/talk-plugin-remember-sort/client/index.js
```diff
@@ -0,0 +1,40 @@
...
+    // in order to mitigate strict cross domain security settings.
+
+    let sort = JSON.parse(await pymStorage.getItem(STORAGE_PATH));
+    if (
+      sort &&
+      introspection.isValidEnumValue('SORT_ORDER', sort.sortOrder) &&
+      introspection.isValidEnumValue('SORT_COMMENTS_BY', sort.sortBy)
+    ) {
+      store.dispatch(setSort(sort));
+    }
```
wyattjoh, line 27: Maybe if the sort isn't valid we remove it from the store? Might improve future refresh performance?

alinhle: I played with that thought before, and went with leaving it as it is. I can add it though :-)

PR #940, opened 2017-09-09
improve docker compose getting started docs

boazsender:
I think that the env config vars used to be documented in the project's repo, but no longer are.

This PR adds link to new location for config vars, and includes the minimum needed env vars in docker-compose.yml example file.


kgardnr (approved):

kgardnr:
Thanks @boazsender!

PR #942, opened 2017-09-11
In-Stream Moderation Updates

okbel:
## What does this PR do?
- Adds modifications to In-Stream Moderation such as:
  - Naming conventions
  - Updating the cache
  - Adding a reducer for the Moderation Menu
  - Handling errors
  - UX improvements


plugins/talk-plugin-moderation-actions/client/containers/ApproveCommentAction.js
```diff
@@ -1,28 +1,16 @@
...
-        commentId: comment.id,
-        status: 'ACCEPTED'
-      });
-
-      if (!isNil(result.data.setCommentStatus)) {
-        throw result.data.setCommentStatus.errors;
-      }
-      
-    } catch (err) {
-      notify('error', getErrorMessages(err));
```
alinhle: The framework doesn't automatically fire notifications to notice the user about the error, so this line is still needed. Also `notify` is a redux action and need to be connected.

plugins/talk-plugin-moderation-actions/client/containers/RejectCommentAction.js
```diff
@@ -1,28 +1,16 @@
...
-        commentId: comment.id,
-        status: 'REJECTED'
-      });
-
-      if (!isNil(result.data.setCommentStatus)) {
-        throw result.data.setCommentStatus.errors;
-      }
-      
-    } catch (err) {
-      notify('error', getErrorMessages(err));
```
alinhle: The framework doesn't automatically fire notifications to notice the user about the error, so this line is still needed. Also `notify` is a redux action and need to be connected.

plugins/talk-plugin-moderation-actions/client/containers/ModerationActions.js
```diff
@@ -1,14 +1,72 @@
...
+        return false;
+      }
+    }
+    
+    // Prevent Slot from rerendering when no props has shallowly changed.
+    return changes.length !== 0;
+  }
+
+  toogleTooltip = () => {
+    if (this.props.showTooltipForComment === this.props.comment.id) {
```
alinhle: We should probably extract this logic into something like
```js
isTooltipVisible: (props = this.props) => props.showTooltipForComment === props.comment.id
```

plugins/talk-plugin-moderation-actions/client/actions.js
```diff
@@ -0,0 +1,10 @@
+import {OPEN_TOOLTIP, CLOSE_TOOLTIP} from './constants';
```
alinhle: According to wikipedia, a `tooltip `is sth that shows when hovering over it, so I don't think this is the correct term here. Maybe sth like _menu_.

plugins/talk-plugin-moderation-actions/client/constants.js
```diff
@@ -0,0 +1,4 @@
+const prefix = 'TALK_MODERATION_ACTIONS';
+
+export const OPEN_TOOLTIP = `${prefix}_OPEN_TOOLTIP`;
```
alinhle: According to wikipedia, a `tooltip `is sth that shows when hovering over it, so I don't think this is the correct term here. Maybe sth like _menu_.

plugins/talk-plugin-moderation-actions/client/containers/ModerationActions.js
alinhle: Following my other comments this should rather be sth like `isMenuVisibile`

alinhle (changes requested):
Great job so far! Besides the above comments, we also identified two changes:

image

1. The `approved` button should not have a hover styling, so that it does not suggest that it's clickable, as it does nothing when clicking on it.

2. After performing an action, the menu should hide.

okbel:
All the issues above solved, but one. When you feature a comment the Menu is not closing. I need a clean way to pass `hideMenu` to the Slot. Any thoughts?

okbel:
_Note: Since I can't just import the action from a plugin that the user might not even have._

I tried to pass that to the `<Slot>` . It doesn't look very clean.

alinhle:
Just connect the action using redux

plugins/talk-plugin-featured-comments/client/index.js
```diff
@@ -84,8 +85,28 @@ export default {
           }
 
           return updated;
-        },
-      }
+        }
+      },
+      update: (proxy) => {
+
+        if (name !== 'FEATURED') {
```
alinhle: Where does the variable `name` come from?

plugins/talk-plugin-featured-comments/client/index.js
```diff
@@ -84,8 +85,28 @@ export default {
...
+          return;
+        }
+
+        const fragmentId = `Comment_${variables.id}`;
+
+        const fragment = gql`
+          fragment Talk_ModerationActions_addTag on Comment {
```
alinhle: This should be called something with `FeaturedComments`

plugins/talk-plugin-featured-comments/client/index.js
```diff
@@ -84,8 +85,28 @@ export default {
           }
```
alinhle, line 74: Line 63-73 needs to be removed as it is replaced with the new `update` below.
https://github.com/coralproject/talk/pull/942/files#diff-8103bea6e6543f80c7f897579ce2dcbeL63

alinhle:
- Approved button still shows a pointer cursor when hovering it and clicking on it causes the menu hide and send a mutation to the server. Approved button shouldn't perform any action and it's styling shouldn't suggest that it's clickable.

image

- Another bug I found is when I feature a comment, the button doesn't change its text to `un-feature` unless I leave and rehover with the mouse. So the text remains `feature` and clicking it will _unfeature_ the comment.

image

okbel:
Thanks for reporting them! I will be working on them.

okbel:
All the issues have been addressed. Ready to be checked. @cvle

plugins/talk-plugin-moderation-actions/client/containers/RejectCommentAction.js
```diff
@@ -1,25 +1,24 @@
...
 import {getErrorMessages} from 'plugin-api/beta/client/utils';
-import {withSetCommentStatus} from 'plugin-api/beta/client/hocs';
 import {notify} from 'plugin-api/beta/client/actions/notification';
 import RejectCommentAction from '../components/RejectCommentAction';
-import isNil from 'lodash/isNil';
+import {connect, withSetCommentStatus} from 'plugin-api/beta/client/hocs';
 
 class RejectCommentActionContainer extends React.Component {
 
-  rejectComment = async () => {
```
alinhle: Need `async` otherwise `try` `catch` doesn't work.

plugins/talk-plugin-moderation-actions/client/containers/RejectCommentAction.js
```diff
@@ -1,25 +1,24 @@
...
-    const {setCommentStatus, comment} = this.props;
-
+  rejectComment = () => {
+    const {setCommentStatus, closeMenu, comment} = this.props;
+    
     try {
-      const result = await setCommentStatus({
```
alinhle: Need `await` otherwise `try` `catch` doesn't work.

plugins/talk-plugin-moderation-actions/client/containers/RejectCommentAction.js
```diff
@@ -1,25 +1,24 @@
...
-import isNil from 'lodash/isNil';
+import {connect, withSetCommentStatus} from 'plugin-api/beta/client/hocs';
 
 class RejectCommentActionContainer extends React.Component {
 
-  rejectComment = async () => {
-    const {setCommentStatus, comment} = this.props;
-
+  rejectComment = () => {
+    const {setCommentStatus, closeMenu, comment} = this.props;
```
alinhle: Need to get `notify` from props, otherwise its unbounded to the store.

plugins/talk-plugin-moderation-actions/client/containers/ApproveCommentAction.js
```diff
@@ -1,25 +1,24 @@
...
-import isNil from 'lodash/isNil';
+import {connect, withSetCommentStatus} from 'plugin-api/beta/client/hocs';
 
 class ApproveCommentActionContainer extends React.Component {
 
-  approveComment = async () => {
-    const {setCommentStatus, comment} = this.props;
-
+  approveComment = () => {
+    const {setCommentStatus, closeMenu, comment} = this.props;
```
alinhle: Need to get `notify` from props, otherwise its unbounded to the store.

plugins/talk-plugin-moderation-actions/client/containers/ApproveCommentAction.js
```diff
@@ -1,25 +1,24 @@
...
 import {getErrorMessages} from 'plugin-api/beta/client/utils';
-import {withSetCommentStatus} from 'plugin-api/beta/client/hocs';
 import {notify} from 'plugin-api/beta/client/actions/notification';
 import ApproveCommentAction from '../components/ApproveCommentAction';
-import isNil from 'lodash/isNil';
+import {connect, withSetCommentStatus} from 'plugin-api/beta/client/hocs';
 
 class ApproveCommentActionContainer extends React.Component {
 
-  approveComment = async () => {
```
alinhle: Need `async` otherwise `try` `catch` doesn't work.

plugins/talk-plugin-moderation-actions/client/containers/ApproveCommentAction.js
```diff
@@ -1,25 +1,24 @@
...
-    const {setCommentStatus, comment} = this.props;
-
+  approveComment = () => {
+    const {setCommentStatus, closeMenu, comment} = this.props;
+    
     try {
-      const result = await setCommentStatus({
```
alinhle: Need `await` otherwise `try` `catch` doesn't work.

alinhle (changes requested):

alinhle (approved):
Looks good!

PR #943, opened 2017-09-12
Added support for deep comment counts

wyattjoh:
## What does this PR do?

Adds support for deep comment counts to be returned. It does this by overriding the `replyCount` edge to provide the new loader to use instead.

**Warning: This feature introduces possibly a large performance impact for sites with many levels of nesting. Not recommend for any site exceeding 3 levels deep of commenting.**

Related: https://www.pivotaltracker.com/story/show/150848457

## How do I test this PR?

1. Add `talk-plugin-deep-reply-count` to the `plugins.json` file.
2. Reload server.
3. Notice that counts returned are deep!


kgardnr:
@wyattjoh I can't get this to work at all, when I try My Profile after enabling the plugin, rebuilding, and restarting the server, I get a spinner but no errors. I have it set to default threading levels.

cc @cvle in case you've tried testing it too

alinhle:
That's problematic. The field replyCount shouldn't be just replaced. The contract for `replyCount` is that it only has the number of direct replies. This is sth that the client may rely on.

alinhle:
Code looks good. Do we need to implement a client part too? Or separate ticket?

wyattjoh:
I think that any newsroom that wants to display the reply count can provide a plugin that does that and instead uses the deepReplyCount value @cvle.

alinhle (approved):

PR #944, opened 2017-09-12
Better wordlist service and highlighter

alinhle:
## What does this PR do?
- Implement our own custom regexp and linkify-it based highlighter
- Banned and supsected words/phrases are now only highlighted when matching the whole word/phrase
- Rewrite wordlist service on server using regexp solution which is more robust, simpler, supports unicode and emoticons, and on top of that it's 20 times faster

## How do I test this PR?
- Post comments with banned, suspected and link text.


wyattjoh:
I added the phrase "it is suspectful" to the suspect word list, and it was not highlighted on the frontend panel, but was successfully caught by the backend.

wyattjoh (commented):
I added the phrase "it is suspectful" to the suspect word list, and it was not highlighted on the frontend panel, but was successfully caught by the backend.

alinhle:
Ah ok, I thought the word list only has _words_ so I didn't add support for phrases. Will do!

wyattjoh (approved):
Looks good to me!

PR #949, opened 2017-09-14
Allow Moderator to have read access on assets and settings

alinhle:
## What does this PR do?
- Allow Moderator to have read access on assets and settings
https://www.pivotaltracker.com/story/show/150905653


wyattjoh (approved):
Looks good to me!

PR #952, opened 2017-09-14
In-Stream Ban

okbel:
## What does this PR do?
- Adds In-Stream Ban! 🎉


perms/mutationReducer.js
```diff
@@ -10,13 +10,16 @@ module.exports = (user, perm) => {
   case types.EDIT_COMMENT:
     return true;
   case types.ADD_COMMENT_TAG:
+    return check(user, ['ADMIN', 'MODERATOR', 'STAFF']);
```
alinhle: These 3 additional lines seems unnecessary as they fall through the same result. (This line and the other similar ones below).

plugins/talk-plugin-moderation-actions/client/containers/BanUserAction.js
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const mapDispatchToProps = (dispatch) =>
+  bindActionCreators({
+    openDialog
+  }, dispatch);
+
+const enhance = compose(
+  connect(null, mapDispatchToProps),
+  excludeIf(({root: {me}, comment}) => !me || me.id === comment.user.id),
```
alinhle: This only works when this component is being used inside a slot, which is not the case.

plugins/talk-plugin-moderation-actions/client/actions.js
```diff
@@ -8,3 +8,12 @@ export const openMenu = (id) => ({
 export const closeMenu = () => ({
   type: CLOSE_MENU,
 });
+
+export const openDialog = (comment) => ({
```
alinhle: I would avoid passing too much data. Moreover `comment` is always a fragment and the content may vary.
Changing this to sth like `commentId, authorId` makes the data requirement clear.

alinhle (changes requested):
Only small comments, this is well done! 
I let the UX testing to @IAmSamHankins

plugins/talk-plugin-moderation-actions/client/actions.js
okbel: Hm, while I was doing this I realized this change does not look good. I also need `commentStatus` and it's a lot to rename. The amount of data passed through `comment` is being handled by the `fragment` so I don't think this change is needed.

plugins/talk-plugin-moderation-actions/client/containers/BanUserAction.js
okbel: Oh, It makes sense. Is there something else I can use to have the same behavior? Just to stop the execution of the compose... :thinking:

alinhle: Just go back to the parent where you include this component, and add the logic there not to include it in render when it's the same user.

okbel: Yeah. I was wondering if there was something in the compose that we could add.

okbel: Will do!

plugins/talk-plugin-moderation-actions/client/actions.js
alinhle: By looking at this action in isolation it's not defined what data I need to pass to `comment`. Only if I have a good knowledge of the whole plugin, I'd know that it needs the comment id the author id and the comment status and it must follow graph schema. Under the condition that this action is not used anywhere else outside of the plugin (e.g. someone wants to trigger the ban dialog from a different plugin) than we could argue that it's fine. My recommendation is to have clear data requirements and only pass what is necessary (right now `tags` are also passed to redux but it is not used).

plugins/talk-plugin-moderation-actions/client/reducer.js
```diff
@@ -1,7 +1,9 @@
-import {OPEN_MENU, CLOSE_MENU} from './constants';
+import {OPEN_MENU, CLOSE_MENU, OPEN_DIALOG, CLOSE_DIALOG} from './constants';
 
 const initialState = {
   showMenuForComment: null,
+  showDialog: false,
```
alinhle: Maybe call it `showBanDialog`, as we might also have a `showSuspendDialog` later?

plugins/talk-plugin-moderation-actions/client/constants.js
```diff
@@ -2,3 +2,5 @@ const prefix = 'TALK_MODERATION_ACTIONS';
 
 export const OPEN_MENU = `${prefix}_OPEN_MENU`;
 export const CLOSE_MENU = `${prefix}_CLOSE_MENU`;
+export const OPEN_DIALOG = `${prefix}_OPEN_DIALOG`;
```
alinhle: Maybe _OPEN_BAN_DIALOG_ as we have a suspend dialog later?

theysamtheirs:
A couple issues:
- Modal scrolls view to what appears to be the middle of the stream (some kind of fixed location) when you try to ban a user.
- Interaction is correct (banning the user, and rejecting the comment from which the commenter is banned), but message in the modal is incorrect. Should match Admin side ban language.

### Correct Language (which includes the user name of who is currently being banned):
image

kgardnr (approved):

alinhle (approved):

PR #955, opened 2017-09-14
Added support for changing the static url

wyattjoh:
## What does this PR do?

Adds support for new static root configuration:

- `TALK_STATIC_URI` (_optional_) - Used to set the uri where the static assets should be served from. This is used when you want to upload the static assets through your build process to a service like Google Cloud Storage or Amazon S3 and you would then specify the CDN/Storage url. (Default `process.env.TALK_ROOT_URL`)
- `TALK_DISABLE_STATIC_SERVER` (_optional_) - When `TRUE`, it will not mount the static asset serving routes on the router. (Default `FALSE`)

Related: https://www.pivotaltracker.com/story/show/151066761

## How do I test this PR?

1. Setup a service like [ngrok](https://ngrok.com).
2. Set it's http url as the `TALK_STATIC_URI` environment variable.
3. Reload Talk, and observe that the logs for ngrok are showing the static files being loaded from that url.


alinhle:
image

I think the `default.css` should be served on the  `STATIC_URL`

kgardnr (approved):

PR #958, opened 2017-09-15
Detect and display stream errors

alinhle:
## What does this PR do?
- Show error message when the query yields no results.

https://www.pivotaltracker.com/story/show/151134916

## How to test this
- Use a non existing comment id for permalink view


wyattjoh (approved):
Fix works for me.

PR #963, opened 2017-09-18
Linting pass

alinhle:
## What does this PR do?
- Fix some linting errors
- Set to warn for
  -  `react/prop-types` (This will cause >700 linter warnings, we can work on it one by one)
- Set to default (= error) for
  - react/jsx-no-duplicate-props
  - react/jsx-no-target-blank
  - react/no-find-dom-node
  - react/no-unknown-property


wyattjoh (approved):
Linting fixes look good to me!

PR #964, opened 2017-09-18
replace hard-coded string on all comments tab with translation

ebirving:
## What does this PR do?
Replaces the hard-coded string, "All Comments", with a translation string on the All Comments Pane Tab
## How do I test this PR?
Open an embed stream. Confirm that the text in All Comments the tab matches `settings.all_comments` for your language.

### DIFF
--- a/client/coral-embed-stream/src/components/Stream.js
+++ b/client/coral-embed-stream/src/components/Stream.js
@@ -160,7 +160,7 @@ class Stream extends React.Component {
           loading={loading}
           appendTabs={
             <Tab tabId={'all'} key='all'>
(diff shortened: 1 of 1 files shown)


alinhle:
Thanks @ebirving, there is a linting issue with tabs that prevent this PR from being merged. 
I also plan to move the translation key from `settings.all_comments` to `stream.all_comments` in all translation files. If you want you can do this, otherwise I'll just make this change in a subsequent PR.
Thanks.

ebirving:
It probably makes most sense for you to do it -- thanks for the update! 😄

PR #971, opened 2017-09-19
Flags: Implement Labels, bugfixes and improvements

alinhle:
## What does this PR do?
- New slot `adminCommentLabels`
- Fix toxic comments not adding a flag and not appearing in reported queue
- Fix `all` comments queue and `userDetail` `all` tab not showing all comments.
- Fix mod queue `loadMore` feature not considering tags.
- Return actions for `CreateComment` Mutation
- Posting a toxic comment twice will show a different notification now and not the default premod notification
- Show toxic comment flag label
- Implement core labels:
  - reply label
  - premod label
  - history flag label
  - suspect flag label
  - user flag label

## How do I test this PR?

- Post a toxic comment (twice), see new notification text that states that your text might still violates guidelines etc.
- Check the existance of the above mentioned labels in the modqueue and user detail
- Check that the `all` moderation tab contains  `SYSTEM_WITHHELD` comments
- Test as a priviledged and as a normal user.


graph/loaders/comments.js
```diff
@@ -91,12 +91,19 @@ const getParentCountsByAssetID = (context, asset_ids) => {
 const getCommentCountByQuery = (context, {ids, statuses, asset_id, parent_id, author_id, tags, action_type}) => {
   let query = CommentModel.find();
 
-  if (ids) {
-    query = query.where({id: {$in: ids}});
+  if (
+    (context.user != null && context.user.can(SEARCH_NON_NULL_OR_ACCEPTED_COMMENTS)) ||
+    statuses && statuses.every((status) => ['NONE', 'ACCEPTED'].includes(status))
```
wyattjoh: If `statuses` is null/undefined, no comments will be returned. Is that desired?

graph/mutators/comment.js
```diff
@@ -393,13 +393,12 @@ const moderationPhases = [
...
 /**
- * This resolves a given comment's status to take into account moderator actions
- * are applied.
+ * This resolves a given comment's status and actions.
  * @param  {Object} context graphql context
  * @param  {String} body          body of the comment
  * @param  {String} [asset_id]    asset for the comment
  * @param  {Object} [wordlist={}] the results of the wordlist scan
- * @return {Promise}              resolves to the comment's status
+ * @return {Object}               resolves to the comment's status and actions
```
wyattjoh: Every `async` function returns a `Promise`

graph/mutators/comment.js
```diff
@@ -454,17 +456,20 @@ const createPublicComment = async (context, comment) => {
...
   // status to set the comment to.
-  let actions = await resolveCommentModeration(context, comment);
+  let {actions, status} = await resolveCommentModeration(context, comment);
+
+  // Assign status to comment.
+  comment.status = status;
 
   // Then we actually create the comment with the new status.
-  comment = await createComment(context, comment);
+  const result = await createComment(context, comment);
```
wyattjoh, line 465: Why can't we just assign this to the original comment variable? We don't use the old one here, and truly, it's realized the new comment here instead.

graph/resolvers/asset.js
```diff
@@ -23,6 +23,7 @@ const Asset = {
 
     // Include the asset id in the search.
     query.asset_id = id;
+    query.statuses = ['NONE', 'ACCEPTED'];
```
wyattjoh: Should not be applied at the resolver level, instead should apply at the schema level. This prevents any future query to support a `statuses` field, which currently, the graph schema says it does allow.

graph/resolvers/comment.js
```diff
@@ -26,6 +26,7 @@ const Comment = {
 
     query.asset_id = asset_id;
     query.parent_id = id;
+    query.statuses = ['NONE', 'ACCEPTED'];
```
wyattjoh, line 29: Same as above.

graph/resolvers/root_mutation.js
```diff
@@ -1,7 +1,13 @@
...
-    comment: await Comment.create(input),
-  }),
+  createComment: async (_, {input}, {mutators: {Comment}, loaders: {Actions}}) => {
+    const comment = await Comment.create(input);
+
+    // Retrieve flags that was assigned to comment.
+    const actions = await Actions.getByID.load(comment.id);
+    const flags = actions.filter(({action_type}) => action_type === 'FLAG');
+
+    return {comment, flags};
```
wyattjoh: Too specific, this should instead be `actions`, don't filter out the flags.

graph/typeDefs.graphql
```diff
@@ -875,6 +877,9 @@ type CreateCommentResponse implements Response {
   # The comment that was created.
   comment: Comment
 
+  # Flags that was assigned during creation of the comment.
+  flags: [FlagAction]
```
wyattjoh: See above, should return `actions: [Action]` instead.

wyattjoh (changes requested):

graph/loaders/comments.js
alinhle: It'll return all comments if permissions are there, look inside the if clause

graph/mutators/comment.js
alinhle: I think I added this because I used to have more changes here. 
But I think it is good to have a different variable here.
Because  `comment` in my mind was the `input` and
`result` is the created comment. Different concepts and as such I would assign them to different variables just to make it more clear.

graph/resolvers/comment.js
alinhle: Reply query currently does not allow specifying `statuses`.

graph/resolvers/asset.js
alinhle: Mixed this up with the `RepliesQuery` as that one does not allow specifying statuses.

graph/resolvers/root_mutation.js
alinhle: You said something about `filter` in the call, so I was thinking what possibly could I filter and ended up filtering out the flags 😅

alinhle: can revert it 👍

graph/typeDefs.graphql
alinhle: 👍

graph/loaders/comments.js
wyattjoh: For these if statements, if you can write a short comment above that describes how/why we're doing this. It'll make it super clear.

alinhle: 👍

wyattjoh (approved):

PR #985, opened 2017-09-22
Moderate Indicator

okbel:
## What does this PR do?
- Adds indicators for Community and Moderation on the Admin Menu
More info here: https://www.pivotaltracker.com/story/show/144641979
- Lifts the queries, and deletes old code. Adds proptypes, etc.

## How do I test this PR?
- Flag a username, the indicator should show for Community
- Send a comment to pre-mod, the indicator should show for Moderation


client/coral-admin/src/containers/Layout.js
```diff
@@ -72,19 +89,50 @@ class LayoutContainer extends Component {
...
+  toggleShortcutModal: PropTypes.func,
+  TALK_RECAPTCHA_PUBLIC: PropTypes.string,
+  checkLogin: PropTypes.func,
+  fetchConfig: PropTypes.func,
+  root: PropTypes.object.isRequired,
+  data: PropTypes.object.isRequired,
+};
+
+const withData = withQuery(gql`
+  query TalkAdmin_initialQuery {
```
alinhle: Second part should be component name: `TalkAdmin_Layout`

client/coral-admin/src/containers/Layout.js
```diff
@@ -72,19 +89,50 @@ class LayoutContainer extends Component {
...
+    ...${getDefinitionName(Header.fragments.root)}
+    ...${getDefinitionName(Community.fragments.root)}
+  }
+  ${Header.fragments.root}
+  ${Community.fragments.root}
+  `, {
+  options: {
+    fetchPolicy: 'network-only',
```
alinhle: This will not do anything, should remove the options.

client/coral-admin/src/routes/Community/containers/Community.js
```diff
@@ -90,5 +66,22 @@ export default compose(
   connect(mapStateToProps, mapDispatchToProps),
   withSetUserStatus,
   withRejectUsername,
-  withData,
+  withFragments({
```
alinhle: This should not be defined as fragments but as a query, because we only want to load the whole community data only when we are on the community page.

alinhle (changes requested):
See comments above

client/coral-admin/src/containers/Header.js
```diff
@@ -0,0 +1,16 @@
+import {gql} from 'react-apollo';
+import withFragments from 'coral-framework/hocs/withFragments';
+import Header from '../components/ui/Header';
+
+export default withFragments({
+  root: gql`
+    fragment TalkAdmin_Header on RootQuery {
```
alinhle: Actually this could be turned into a query instead of fragment.

okbel: I think this one is okay, because it expresses the needs from the Header component. And we will have more needs in the Layout component that we will need to spread. No need to make an extra query out of this one.

alinhle: What do you expect will need more?

okbel: We can take advantage of it. And preload stuff but now that I'm thinking that we only have that Header in the Layout, maybe it's not that needed as I thought.

alinhle: I can't think of any other cases, so that's why I think we can simplify it and just move the query directly to the header. But if you can think of some, go ahead :-)

client/coral-admin/src/containers/Header.js
```diff
@@ -0,0 +1,20 @@
...
+    })
+    reportedCount: commentCount(query: {
+      statuses: [NONE, PREMOD, SYSTEM_WITHHELD],
+      action_type: FLAG
+    })
+    flaggedUsernamesCount: userCount(query: {
+      action_type: FLAG,
+      statuses: [PENDING]
+    })
+  }
```
alinhle, line 19: You can turn on polling until we have live updates: http://dev.apollodata.com/react/receiving-updates.html#polling

alinhle (approved):
Looks good. I think we should turn on polling until we have live updates, other than that it's 👍

PR #988, opened 2017-09-25
Password reset fixes

wyattjoh:
## What does this PR do?

- Cleanups around email checking.
- Added domain validation against the root domain as well as whitelisted domains to support the admin site, previously the whitelisted domains were the only ones checked, not the root.
- Added the `parentUrl` to the login container so that the password reset redirect goes to the article page and not the root comment domain.

## How do I test this PR?

1. Run Talk off of `127.0.0.1:3000`
2. Add any other domain to the whitelist
3. Go to `http://127.0.0.1/admin` and go through the password reset flow.
4. Email!

### DIFF
--- a/services/domainlist.js
+++ b/services/domainlist.js
@@ -2,6 +2,8 @@ const debug = require('debug')('talk:services:domainlist');
 const _ = require('lodash');
 const SettingsService = require('./settings');
 
+const {ROOT_URL} = require('../config');
+
 /**
  * The root domainlist object.
  * @type {Object}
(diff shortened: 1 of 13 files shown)


services/domainlist.js
```diff
@@ -84,7 +82,7 @@ class Domainlist {
     let domain;
 
     // removes protocol and get domain
-    if (url.indexOf('://') > -1) {
+    if (url.indexOf('//') > -1) {
```
alinhle, line 85: Maybe using a library like https://www.npmjs.com/package/url-parse will work more reliably?

wyattjoh: Unfortunately, the `url-parse` and `url` packages don't do well with partial user input. My extensive tests: https://github.com/coralproject/talk/pull/988/files#diff-0b9df2d13c97387e2ccee23f13e6435eR29 have validated that it should always be parsed correctly.

alinhle (approved):

PR #997, opened 2017-09-30
Fix double preposition in string

vickychijwani:
Minor string fix for this error on the Configure tab:

```
The comment stream will close in in 1 week.
```

The second "in" comes from `timeago`.

Also updated the French and Portuguese translations with help from Google Translate, but someone who actually speaks those should maybe double-check :smile:


kgardnr (approved):

kgardnr:
Thank you @vickychijwani, great catch. Looking forward to future contributions!

PR #684, opened 2017-06-16
Action focus states

theysamtheirs:
## What does this PR do?

Fixes issue relating to PT BUG: "Liked" blue outline cuts off on left side on Mac Chrome

## How do I test this PR?

- click a button
- view the cat
- see the cat meow

PR #998, opened 2017-10-02
Reintroducing E2E testing with Puppeteer

okbel:
## What does this PR do?
- Reintroduces e2e with Pupeteer
- Adds missing classes and prop-types

## Summary and Concerns 
I’ve been using Puppeteer for a couple of days now and I ran into a couple of issues to set our e2e test environment. One of them ([Clicking element inside iFrame does not give any response · Issue #684 · GoogleChrome/puppeteer · GitHub](https://github.com/GoogleChrome/puppeteer/issues/684)) has been a blocker to develop tests for our `pym` handlers.  I’ve been able to come with a solution/workaround for it but I realised that there are a lot of features that interact with our `pym` service. 

Since Puppeteer does not provide a framework around testing we have to build the suite around it. The good thing about this is that we have more control over it. 

Another thing is that Puppeteer works only with Chrome, so we lose the API for testing cross-browser support that Selenium provides.

We can either explore more solutions for this, exploring other frameworks, other options or wait until they solve this bug.

If we move forward with this I plan on using more test architecture stuff like, page objects and better handling for selectors and errors. (Like screenshots for errors, etc.)

## How do I test this PR?
- run `yarn e2e`
- You should see: 
screen shot 2017-10-02 at 12 04 06 am

https://github.com/coralproject/talk/pull/188


test/e2e/spec/test.spec.js
```diff
@@ -0,0 +1,175 @@
+const {test} = require('../browser');
+const {murmur3} = require('murmurhash-js');
+const uuid = require('uuid');
+const {getEmbedStream} = require('../utils/frame');
+const {expect} = require('chai');
+
+describe('Embed Stream', () => {
+
+  it('creates a new asset', test(async (browser, {opts: {url}}) => {
+    const rHash = murmur3(uuid.v4());
```
wyattjoh, line 10: Why hashing the uuid instead of just using the id? Mostly asking because curious.

test/e2e/spec/test.spec.js
```diff
@@ -0,0 +1,175 @@
...
+    await page.waitForSelector($.Login.loginButton);
+    await page.click($.Login.loginButton);
+    await page.waitForNavigation({waitUntil: 'networkidle'});
+
+    // Going to the Embed Stream
+    await page.goto(data.embedStreamUrl, {waitUntil: 'networkidle', networkIdleTimeout: 3000});
+
+    // Checking logged-in Username
+    const userBoxEl = await page.waitForSelector($.Stream.authUserboxUsername);
+    expect(userBoxEl).to.not.equal(null);
```
wyattjoh: We could add a tiny bit more to this test to check to see if the username was what we expected.

test/e2e/spec/test.spec.js
```diff
@@ -0,0 +1,175 @@
...
+
+  it('signed in user sees comment history', test(async (browser, {$}) => {
+    const page = await browser.currentPage;
+    await page.click($.Stream.myProfileTab);
+    await page.waitForNavigation({waitUntil: 'networkidle'});
+    await page.waitForSelector($.Stream.tabContent);
+
+    // Checking if Comment History exists
+    const myCommentHistory = await page.$($.MyProfile.myCommentHistory);
+    expect(myCommentHistory).to.not.equal(null);
```
wyattjoh: Maybe extend this to check at least for one comment's text that we expect?

wyattjoh:
Also, super excited to see this PR!! Awesome work @okbel 😸

test/e2e/spec/test.spec.js
okbel: I think I did it because it triggered the same id in the second test I ran 🤔 I wanted to avoid collision at all cost. I think we can remove that now

okbel: Oh, I remember. It was because `uuid` generates uuids with dashes and for example username does not allow that. I couldn't find a way to overide that easily.

kgardnr:
@okbel @cvle, can we close this PR now?

PR #1002, opened 2017-10-03
Es6 modules + tree shaking

alinhle:
## What does this PR do?
- Switching frontend code base to full es6 modules 
- Removing deprecated commonjs module style export and imports
- Turn on Tree Shaking
  - Without in depth optimization we only got like 10kb savings of gzipped and minified data. I expected more, but probably will yield better results when we do more file size optimizations.
- Webpack can now analyse our dependency tree and yielded several warnings of importing undefined exports, which are solved in this PR as well.
- Move all `.babelrc` files into one file at the root dir.

## How do I test this PR?

-  Should not break any existing functionality.


wyattjoh (commented):
Why were all the .babelrc files removed? Were they not required?

alinhle:
Not essentially, the root `.babelrc` is valid for all subfolders including the plugins. I removed them because otherwise I needed to copy and paste the same changes in 20+ files. If we decide that they are required, we can make a preset like with the linters, as a separate story.

.babelrc
```diff
@@ -12,4 +11,4 @@
     "transform-react-jsx",
     "syntax-dynamic-import"
   ]
-}
\ No newline at end of file
+}
```
wyattjoh, line 14: Doesn't this mean that any file under the server JS will be processed with babel? I don't think this is a good solution if that's the case. I want to avoid server side transpiling if at all possible.

alinhle: There is no transpiling, because our cli does not use babel. We only have babel when running using webpack (and jest later). If the client reach out to server code, then that would be transpiled but as we currently dont share server code and client, it doesnt happen.

wyattjoh (approved):

PR #1006, opened 2017-10-03
Fixes Featured Tag Permissions

wyattjoh:
## What does this PR do?

- When a comment is created, there is a possibility of injecting a tag into the payload, this removes that permission error


okbel (approved):
:+1:

PR #1045, opened 2017-10-04
Flags locales

wyattjoh:
## What does this PR do?

- Flags now are all required to pre-define their types in the graph, or at the very least provide their translations in locales.

Ref: https://www.pivotaltracker.com/story/show/151179316

## How do I test this PR?

1. Create some flags
2. Check out the translated responses in the admin


graph/typeDefs.graphql
```diff
@@ -948,7 +972,7 @@ input CreateFlagInput {
   item_type: ACTION_ITEM_TYPE!
 
   # The reason for flagging the item.
-  reason: String!
+  reason: FLAG_REASON
```
alinhle, line 975: Required?

alinhle:
We also need to adapt toxic comment for its TOXIC_COMMENT flag.

graph/typeDefs.graphql
wyattjoh: Nope!

alinhle (changes requested):
image
The graphql refuses to load any system flags whose reasons are not defined in the `FLAG_REASON` type, e.g. trust, suspect and toxic comment flags

alinhle (dismissed):

wyattjoh:
Future PR will merge this branch in.

PR #1052, opened 2017-10-04
Add a max-width to flag-reason text box to prevent cutoff

### LINKED ISSUE #1036: Tried to resize the "Other" box when reporting a comment, if resized too far the right side of the window goes past the right side of the window. The box cuts off and cuts off the text.
report_other_box_resize_bug

lancekruegger:
Fixes #1036 

Without a max-width a resizable textarea will expand into infinity. This fix doesn't allow it to expand beyond it's parent container.  

There is a margin left of 20px on the textarea the 100% - 40px is to maintain consistent gutters on both sides of the textarea. 

End result: 
screen shot 2017-10-04 at 11 22 44 am

### DIFF
--- a/client/coral-embed-stream/style/default.css
+++ b/client/coral-embed-stream/style/default.css
@@ -270,6 +270,7 @@ button.comment__action-button[disabled],
   width: 75%;
   font-size: 16px;
   border: 1px solid #ccc;
+  max-width: calc(100% - 40px);
 }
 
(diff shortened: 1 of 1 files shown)


kgardnr:
Thank you @lancekruegger! We will review this.

okbel (approved):
Awesome! Thank you so much for contributing! @lancekruegger

PR #1057, opened 2017-10-05
Slack notifications plugin for new comments

### LINKED ISSUE #1016: [Plugin] Slackbot for Talk
AC:
- Slackbot that automatically posts new comments to a Slack channel
- This would be useful for small newsrooms and blogs

vickychijwani:
Closes #1016.

## What does this PR do?

Adds `talk-plugin-slack-notifications` which posts all new comments to the given [Slack Webhook](https://api.slack.com/incoming-webhooks).

## How do I test this PR?

1. Create a [new Slack app](https://api.slack.com/apps/new) and connect it to any one of your Slack workspaces for testing.
2. In the app configuration, enable "Incoming Webhooks".
3. In the Webhook settings, click "Add New Webhook to Workspace". Slack will open an OAuth dialog requesting access to your development workspace (the one we linked in step 1). Click Authorize.
4. Back on the Webhook settings page, you should now have a valid webhook URL. Try the sample curl request and make sure the Webhook is working - you should see posted messages in your Slack channel/DM.
5. Add the webhook to your .env file (`TALK_SLACK_WEBHOOK_URL=https://<your_webhook_url>`) and restart the server.
6. Post a new comment and it should appear in Slack.

I've tested these cases:

- Many kinds of inputs in a comment: punctuation, unicode symbols, emoji, large comments (4500 chars) - all of it reached Slack verbatim
- Plugin enabled but `TALK_SLACK_WEBHOOK_URL` not set - a descriptive error is thrown

## What remains to be done

- Set up an official Slack app for Talk, and update the error message in `config.js` to point to it. Additionally update docs if needed. (*deferred*)

## Notes

- Performance: The request to Slack takes ~1.5 sec, but I assume that's not a problem because of Node's non-blocking I/O? (NOTE: I'm in India and Slack doesn't have regional servers here apparently). I've set the default timeout to 5000 ms for this reason, but that can be changed.
- I tried 2 different designs and went with the latter one below - feedback is welcome.
- The notification could be made richer by perma-linking it to the comment, but I couldn't figure out how to access the asset URL in the `createComment` hook - probably need a loader or resolver, not sure which.
- The plugin could behave differently when the toxic comments plugin is enabled, e.g., so only comments above the toxic threshold would be posted to Slack, or toxic comments would be [colored](https://api.slack.com/docs/message-attachments#attachment_structure) red, etc.

talk-slack-title

talk-slack-footer

### DIFF
--- a/plugins/talk-plugin-slack-notifications/server/hooks.js
+++ b/plugins/talk-plugin-slack-notifications/server/hooks.js
@@ -19,24 +19,26 @@ module.exports = {
           }
         } = result;
         const username = context.user.username;
-        const response = await fetch(SLACK_WEBHOOK_URL, {
-          method: 'POST',
-          headers: {
-            'Content-Type': 'application/json',
-          },
-          timeout: SLACK_WEBHOOK_TIMEOUT,
-          body: JSON.stringify({
-            attachments: [{
-              text: text,
-              footer: `Comment by ${username}`,
-              ts: Math.floor(Date.parse(createdAt) / 1000),
-            }]
-          }),
+        process.nextTick(async () => {
+          const response = await fetch(SLACK_WEBHOOK_URL, {
+            method: 'POST',
+            headers: {
+              'Content-Type': 'application/json',
+            },
+            timeout: SLACK_WEBHOOK_TIMEOUT,
+            body: JSON.stringify({
+              attachments: [{
+                text: text,
+                footer: `Comment by ${username}`,
+                ts: Math.floor(Date.parse(createdAt) / 1000),
+              }]
+            }),
+          });
+          if (!response.ok) {
+            const responseText = await response.text();
+            console.trace(`Posting to Slack failed with HTTP code ${response.status} and body '${responseText}'`);
+          }
         });
-        if (!response.ok) {
-          const responseText = await response.text();
-          console.trace(`Posting to Slack failed with HTTP code ${response.status} and body '${responseText}'`);
-        }
         return result;
       },
     },
--- a/plugins/talk-plugin-slack-notifications/server/hooks.js
+++ b/plugins/talk-plugin-slack-notifications/server/hooks.js
@@ -1,0 +1,44 @@
+const fetch = require('node-fetch');
+const {SLACK_WEBHOOK_URL, SLACK_WEBHOOK_TIMEOUT} = require('./config');
+const debug = require('debug')('talk:plugin:slack-notifications');
+
+// We don't add the hooks during _test_ as the Slack API is not available.
+if (process.env.NODE_ENV === 'test') {
+  return null;
+}
+
+module.exports = {
+  RootMutation: {
+    createComment: {
+      async post(_, {input}, context, _info, result) {
+        debug(`Posting notification to Slack webhook: ${SLACK_WEBHOOK_URL}`);
+        const {
+          comment: {
+            body: text,
+            created_at: createdAt
+          }
+        } = result;
+        const username = context.user.username;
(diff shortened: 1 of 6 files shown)


plugins/talk-plugin-slack-notifications/server/hooks.js
```diff
@@ -0,0 +1,44 @@
...
+// We don't add the hooks during _test_ as the Slack API is not available.
+if (process.env.NODE_ENV === 'test') {
+  return null;
+}
+
+module.exports = {
+  RootMutation: {
+    createComment: {
+      async post(_, {input}, context, _info, result) {
+        debug(`Posting notification to Slack webhook: ${SLACK_WEBHOOK_URL}`);
```
wyattjoh, line 14: What I would suggest doing here is placing the call to Slack into the next tick, this will ensure that the operation will not block the comment creation process:

```js
process.nextTick(() => {
  // Your slack code goes here, don't return `result` from this function.
});

// Return it here instead.
return result;
```

wyattjoh (changes requested):
One small change and I think we're good to go! Great suggestion with the Talk Slack App. We'll hold off setting that up right now as it's a bit involved.

plugins/talk-plugin-slack-notifications/server/hooks.js
vickychijwani: That's a pretty good idea, thanks! :+1:

Just one question to cement my understanding of the timing of events here. If you look at my 2nd commit, the callback to process.nextTick is marked `async`, so I think the order of execution will be:

post-resolve hook starts => schedules async callback for next tick of event loop => returns `result`, comment is created => makes request to Slack => until response arrives, Node processes other events / sits idle => reads response (again non-blocking)

Is this correct?

wyattjoh: Almost correct!

post-resolve hook starts => schedules async callback for next tick of event loop => returns result, comment is created => makes request to Slack => schedules processing of slack response when received, do other things, slack response came back => log ok!

So when the fetch goes out, it's now an I/O bound event, because it's a http request, so the system will switch over to doing things while it's waiting for a response.

plugins/talk-plugin-slack-notifications/server/hooks.js
```diff
@@ -19,24 +19,26 @@ module.exports = {
...
-          },
-          timeout: SLACK_WEBHOOK_TIMEOUT,
-          body: JSON.stringify({
-            attachments: [{
-              text: text,
-              footer: `Comment by ${username}`,
-              ts: Math.floor(Date.parse(createdAt) / 1000),
-            }]
-          }),
+        process.nextTick(async () => {
```
vickychijwani: I thought it would be better to keep the field extraction from `result` and `context` outside the nextTick callback to avoid holding references to those objects for longer than the lifetime of the request - but I'm new to this stuff so comments are welcome :smile:

wyattjoh: You can certainly extract those outside to the function and access them from the closure. It's up to you!

plugins/talk-plugin-slack-notifications/server/hooks.js
vickychijwani: Makes sense. Probably uses kernel I/O events under the hood thanks to libuv (have briefly read about it in the past). :slightly_smiling_face:

wyattjoh (approved):

wyattjoh:
Thanks so much for your contribution @vickychijwani!

Issue #1046, opened 2017-10-04
If the Edit Timeframe Window is set to > 59 seconds, display time remaining in minutes instead of seconds


kgardnr:
Currently, the edit timeframe window is always displayed in seconds. We should be clever about this and display in minutes and seconds when it's more appropriate.

E.g. "300 seconds" should display as "5 minutes"

This would only be the case in displaying this on the Comment Stream to the commenter. We can keep the Configuration setting in seconds to be 100% precise.

Rishabh04-02:
I'd like to work on the issue.

kgardnr:
Hi @Rishabh04-02! That sounds great, welcome and thank you!

holmberd:
I'm sorry to butt in, but I ended up writing a PR for it, see https://github.com/coralproject/talk/pull/1058

Rishabh04-02:
@kgardnr Please review #1059

kgardnr:
Hi @Rishabh04-02 and @holmberd! Thank you both for your efforts! I'm going to have to go with Risha's PR, Dag, because he did claim the issue first. I do hope you both continue to contribute to Talk, we would love to have you both involved :)

holmberd:
Haha no worries. Surprisingly similar solution ;)

kgardnr:
Fixed in #1059

PR #1059, opened 2017-10-05
displaying time in X minutes Y seconds

Rishabh04-02:
## What does this PR do?
Show the time in X minutes Y seconds instead of X seconds, when editing a comment
## How do I test this PR?

- post a comment and edit it after a  #minute.

#1046

### DIFF
--- a/client/coral-embed-stream/src/components/CountdownSeconds.js
+++ b/client/coral-embed-stream/src/components/CountdownSeconds.js
@@ -45,16 +45,13 @@ export class CountdownSeconds extends React.Component {
     if (typeof classNameForMsRemaining === 'function') {
       classFromProp = classNameForMsRemaining(msRemaining);
     }
-    if (wholeMinRemaining > 0) {
-      return (
-        <span className={classFromProp}>
-          {`${wholeMinRemaining} ${minUnit} ${wholeSecRemaining} ${secUnit}`}
-        </span>
-      );
-    }
+    const text = wholeMinRemaining > 0
+      ? `${wholeMinRemaining} ${minUnit} ${wholeSecRemaining} ${secUnit}`
+      : `${wholeSecRemaining} ${secUnit}`;
+
     return (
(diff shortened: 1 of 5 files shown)


client/coral-embed-stream/src/components/CountdownSeconds.js
```diff
@@ -35,16 +35,26 @@ export class CountdownSeconds extends React.Component {
...
+    const minUnit = t(wholeMinRemaining !== 1 ? 'edit_comment.minutes_plural' : 'edit_comment.minute');
     let classFromProp;
     if (typeof classNameForMsRemaining === 'function') {
       classFromProp = classNameForMsRemaining(msRemaining);
     }
+    if (wholeMinRemaining > 0) {
+      return (
+        <span className={classFromProp}>
+          {`${wholeMinRemaining} ${minUnit} ${wholeSecRemaining} ${secUnit}`}
+        </span>
```
alinhle: Instead of creating a whole new components tree, I think it would be better to compose the _text_ beforehand and use it to render the tree, e.g. 

```js
const text = wholeMinRemaining > 0
  ? `${wholeMinRemaining} ${minUnit} ${wholeSecRemaining} ${secUnit}`
  : `${wholeSecRemaining} ${secUnit}`;

return (
  <span className={classFromProp}>
    {text}
  </span>
);

kgardnr:
@Rishabh04-02 can you take a look at Kiwi's comments above when you get a moment?

client/coral-embed-stream/src/components/CountdownSeconds.js
Rishabh04-02: Yaa, you're right. It'll be more convenient.
I'll change it.

Rishabh04-02:
@cvle Required changes done.

alinhle (approved):

alinhle:
Thanks @Rishabh04-02

PR #1070, opened 2017-10-06
Viewing Options, and bug fixing

okbel:
# Viewing Options and Bug Fixing

## What does this PR do?
- Adds Viewing Options to the Moderation View
- Fix Header Full-Width
- Fix Table View - Icon Drawer
- FIx Drawer
- Adding container to the Moderation View
	- Aligned Moderation Keys under container
	- Aligned View Options

The Dropdown Component is too unstable. I will introduce a new Dropdown component in the next PRs.


kgardnr:
This is so great @okbel!!!! 💯

client/coral-admin/src/components/ui/Drawer.js
```diff
@@ -6,8 +6,8 @@ import styles from './Drawer.css';
 import t from 'coral-framework/services/i18n';
 import {can} from 'coral-framework/services/perms';
 
-const CoralDrawer = ({handleLogout, auth}) => (
-  <Drawer className={styles.header}>
+const CoralDrawer = ({handleLogout, auth = {}}) => (
```
alinhle, line 9: Just a note that dont need to be addressed in this PR. Why is `Drawer` in the folder `ui`. I thought ui is only for "generic ui building blocks".

alinhle (commented):
image

I think the viewing options should never be _behind_ the comments but should trigger the other layout earlier, so that it does not happen.

okbel:
That screenshot looks from an unreacheable resolution. I will lower the breakpoint.

alinhle:
What's an unreachable resolution?

client/coral-admin/src/components/ui/Drawer.js
okbel: Yeah, that one should be definitelly out.

okbel:
Ready to be tested! Thank you :) @cvle

alinhle (approved):

PR #1072, opened 2017-10-06
fix issue 1037

kamos1:
## What does this PR do?
- This pr prevents users from entering multiple newline characters and submitting a blank comment. 

## How do I test this PR?
- Press "return" multiple times and the "Post" button should remain light gray (disabled)

### DIFF
--- a/client/talk-plugin-commentbox/CommentForm.js
+++ b/client/talk-plugin-commentbox/CommentForm.js
@@ -90,7 +90,7 @@ export class CommentForm extends React.Component {
     const {maxCharCount, submitEnabled, cancelButtonClassName, submitButtonClassName, charCountEnable, body, loadingState} = this.props;
 
     const cleanedBody = this.removeNewLine(body);
-    const length = cleanedBody.length;
+    const length = cleanedBody.trim().length;
     const isRespectingMaxCount = (length) => charCountEnable && maxCharCount && length > maxCharCount; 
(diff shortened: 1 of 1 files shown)


kgardnr:
Great! Thank you @kamos1!!

kgardnr (approved):

client/talk-plugin-commentbox/CommentForm.js
```diff
@@ -82,11 +82,16 @@ export class CommentForm extends React.Component {
...
+    return string.replace(/\n/g, '');
+  }
+
   render() {
     const {maxCharCount, submitEnabled, cancelButtonClassName, submitButtonClassName, charCountEnable, body, loadingState} = this.props;
 
-    const length = body.length;
-    const isRespectingMaxCount = (length) => charCountEnable && maxCharCount && length > maxCharCount;
+    const cleanedBody = this.removeNewLine(body);
+    const length = cleanedBody.length;
```
alinhle: Even better when we just use `trim` e.g. `const length = body.trim().length`.
That way we also prevent empty spaces, tabs and other whitespace characters.

alinhle (changes requested):

kamos1:
@cvle I added your suggestion, much cleaner.

alinhle (approved):
Awesome, thanks for your contribution!

Issue #1023, opened 2017-10-04
Don't allow readers to ignore staff members


kgardnr:
AC:
- If a user ignores a staff member, it gives a warning notification "Staff members cannot be ignored"
- The staff member is not ignored and not added to the user's ignore list

dagjaneiro:
Hi! I'd like to tackle this issue. 

Since I'm new to the project can you clarify if by warning notification you mean the same notification system used to display errors?

Something like this:

![image](https://user-images.githubusercontent.com/10207818/31309503-0220f34e-ab7f-11e7-8e2c-abcd8b57db20.png)

kgardnr:
Fixed in #1078

PR #1078, opened 2017-10-07
Cannot ignore staff members

dagjaneiro:
## What does this PR do?

It prevents users from ignoring staff members.

## How do I test this PR?

- try to ignore a comment from a staff member
- A message is displayed: "Cannot ignore Staff members."

Addresses #1023.


kgardnr:
Thank you @dagjaneiro! We will review this shortly.

plugins/talk-plugin-ignore-user/client/index.js
```diff
@@ -14,7 +15,11 @@ export default {
   mutations: {
     IgnoreUser: ({variables}) => ({
       updateQueries: {
-        CoralEmbedStream_Embed: (previousData) => {
+        CoralEmbedStream_Embed: (previousData, {mutationResult: {data: {ignoreUser}}}) => {
+          if (get(ignoreUser, 'errors.length')) {
```
alinhle: You can remove this part, our framework does not run this function when there was an error.

alinhle (changes requested):
See remarks

kgardnr:
Hi @dagjaneiro we have some feedback for you - once it's addressed, this will be merged. Thank you again!

plugins/talk-plugin-ignore-user/client/index.js
dagjaneiro: Hi @cvle, thanks for your feedback. 

Actually, this was a source of confusion for me. 
It seems the framework only throws a GraphQL error if the error is a generic one. 
If you use an API error as a response, it wraps the error in a node specific to the mutation and prevents GraphQL from throwing an exception. 

I believe [this code](https://github.com/coralproject/talk/blob/master/graph/errorHandler.js#L8) is responsible for the behavior I just described.

The payload I'm getting in the frontend is:
```
{
  "data": {
    "ignoreUser": {
      "errors": [{
        "translation_key": "CANNOT_IGNORE_STAFF",
        "__typename": "GenericUserError"
      }],
      "__typename": "IgnoreUserResponse"
    }
  }
}
```
With this payload apollo is returning a successful `APOLLO_MUTATION_RESULT` instead of the expected `APOLLO_MUTATION_ERROR` which would prevent `updateQueries` from running.

Am I missing something to make this work without having to handle the error case explicitly as I did above? Thanks for your help 🙏

alinhle: @dagjaneiro This is correct from the server side view. On the client the mutations runs through our framework that powers the plugin system and it detects the error specific to the mutation and doesnt run the updaters https://github.com/coralproject/talk/blob/master/client/coral-framework/hocs/withMutation.js#L111. It also fails the promise when it detects these types of errors here: https://github.com/coralproject/talk/blob/master/client/coral-framework/hocs/withMutation.js#L144

dagjaneiro: @cvle with your guidance I figured out what's happening. 

When the `withMutation` hoc is processing the `updateQueries`, the [config array](https://github.com/coralproject/talk/blob/master/client/coral-framework/hocs/withMutation.js#L99) has two objects with keys `CoralEmbedStream_Embed`:
- the [featured-comments plugin ignoreUser mutation](https://github.com/coralproject/talk/blob/master/plugins/talk-plugin-featured-comments/client/index.js#L27)
- the [ignore-user plugin ignoreUser mutation I changed](https://github.com/coralproject/talk/pull/1078/files#diff-df3b0f249d219aa83e4324f0ed4fd808R16)

Because of that, when the [reducer function](https://github.com/coralproject/talk/blob/master/client/coral-framework/hocs/withMutation.js#L105) processes my mutation, the key is already present in the result object, so [the callback associated with the mutation](https://github.com/coralproject/talk/blob/master/client/coral-framework/hocs/withMutation.js#L120) doesn't have the error handling bailout code and ends up running `updateQueries` when it shouldn't.

This looks like a bug. If you're ok with it I can address it in this PR, unless there's a reason for it to be this way?

alinhle: Thanks for your investigation @dagjaneiro. I will take a look too.

alinhle: @dagjaneiro You are right, that something is odd, I will dive deeper tomorrow.

alinhle: Thanks a ton!

dagjaneiro:
@cvle I've removed the error handling code from the `ignoreUser` mutation, once the issue with the `withMutation` HoC is resolved this should be good to merge.

alinhle:
@dagjaneiro thanks, pending PR is here https://github.com/coralproject/talk/pull/1100. After that we can merge yours.

alinhle (approved):

PR #1081, opened 2017-10-10
Added notifications when flagging failed

PepeFranco:
## What does this PR do?

It adds notifications to user when they tried to flag a comment/user and the action fails

## How do I test this PR?

Have the api that responds to flags not started and try to flag a comment/user

### DIFF
--- a/client/talk-plugin-flags/components/FlagButton.js
+++ b/client/talk-plugin-flags/components/FlagButton.js
@@ -8,6 +8,8 @@ import ClickOutside from 'coral-framework/components/ClickOutside';
 import cn from 'classnames';
 import styles from './styles.css';
 
+import {getErrorMessages} from 'coral-framework/utils';
+
 const name = 'talk-plugin-flags';
 
 export default class FlagButton extends Component {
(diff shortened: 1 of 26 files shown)


client/talk-plugin-flags/components/FlagButton.js
```diff
@@ -99,6 +99,7 @@ export default class FlagButton extends Component {
             }
           })
           .catch((err) => {
+            this.props.notify('error', t('error.COMMENT_FLAG_FAILED'));
```
alinhle: Can you change these lines to `this.props.notify('error', getErrorMessages(err));` and add an `import {getErrorMessages} from 'coral-framework/utils';` at the top?

alinhle (changes requested):

client/talk-plugin-flags/components/FlagButton.js
PepeFranco: Sure, will do

PepeFranco:
Just did the change as requested by @cvle

alinhle (approved):
@PepeFranco Perfekt!

PR #1082, opened 2017-10-10
Dropdown and Option Component

okbel:
## Summary
We had 3 versions of dropdowns running around with different styling. Now we have a single component created from scratch

## What does this PR do?
- Adds `Dropdown` and `Option` Component
- Replaces dropdowns from the Admin Side

## Usage and example
```js
import {Dropdown, Option} from 'coral-ui';

<Dropdown 
  value={row.status}
  label={capitalize(row.status)}
  placeholder={t('community.status')}
  onChange={(status) => onCommenterStatusChange(row.id, status)}>     
    <Option value={'ACTIVE'}>{t('community.active')}</Option>
    <Option value={'BANNED'}>{t('community.banned')}</Option>
</Dropdown>  
```
`value` : "value from the current selected option"
`label` : "label from the current selected option (applies when value is not suitable for displaying)"
`placeholder`: "when value if falsy it shows the placeholder i.e "Select Status... "
`onChange` "func. When a user clicks another option"


client/coral-ui/components/Dropdown.css
```diff
@@ -0,0 +1,50 @@
...
+  color: #2c2c2c;
+  border-radius: 2px;
+  background: white;
+  box-shadow: 0 2px 2px 0 rgba(0,0,0,.14), 0 3px 1px -2px rgba(0,0,0,.2), 0 1px 5px 0 rgba(0,0,0,.12);
+  width: 100%;
+  list-style: none;
+  padding: 0;
+  margin: 0;
+  transform: scale(0);
+  transition: transform .3s cubic-bezier(.4,0,.2,1),opacity .2s cubic-bezier(.4,0,.2,1),-webkit-transform .3s cubic-bezier(.4,0,.2,1);
```
cvle: It feels a bit slow, try changing it to 0.2s, 0.1s and .0.2s.

alinhle: It feels a bit slow, try changing it to 0.2s, 0.1s and .0.2s.

client/coral-ui/components/Dropdown.js
```diff
@@ -0,0 +1,116 @@
...
+          label: newProps.label,
+          value: newProps.value
+        }
+      });
+    }
+  }
+
+  compoentDidMount() {
+    document.addEventListener('click', this.handleClick, false);
+    document.addEventListener('touchend', this.handleClick, false);
```
alinhle: Why do you need these global event listeners?

client/coral-ui/components/Dropdown.js
```diff
@@ -0,0 +1,116 @@
...
+import cn from 'classnames';
+import ClickOutside from 'coral-framework/components/ClickOutside';
+
+class Dropdown extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      selected: {
+        label: props.label || '',
```
alinhle: why is it necessary to know the selected label?

client/coral-ui/components/Dropdown.js
```diff
@@ -0,0 +1,116 @@
...
+    });
+  }
+
+  renderLabel() {
+    if (this.props.label) {
+      return this.props.label;
+    } else if (this.props.value) {
+      return this.props.value;
+    } else {
+      return this.props.placeholder;
```
alinhle: What's the difference between `props.label` and `props.placeholder`?

client/coral-ui/components/Option.js
```diff
@@ -1,14 +1,18 @@
...
+const Option = ({className, children, onClick}) => (
+  <li className={cn(styles.option, className)} onClick={onClick}>
+    {children || ''}
+  </li>
+);
+
+Option.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.string,
+  onClick: PropTypes.func
```
alinhle: Missing proptype for `value`

client/coral-ui/components/Dropdown.js
alinhle: Also would it be possible to use the labels defined for the `Option`s to get the label?

alinhle (changes requested):
Some additional remarks:
  - The closing animation of the dropdown should be a lot faster.
  - Closing animation puts the dropdown behind other elements.
  - I'd love to see some accessibility enhancements like
being able to control using the keyboard and adding aria roles.

client/coral-ui/components/Dropdown.css
okbel: Why *should* be a lot faster? 😆  It's the one we have now. I will lover it down tho.

client/coral-ui/components/Dropdown.js
okbel: Read the description of the PR. There are the differences.

client/coral-ui/components/Dropdown.js
okbel: Because if there is no label it shows the value, and if there is no value it shows the placeholder. Read the description for more detailed info

client/coral-ui/components/Dropdown.js
okbel: > re: Also would it be possible to use the labels defined for the Options to get the label?

Yes, we could. I thought you wanted it to be semantic. That's why I kept going with the `children` to have something like 

```js
<Option value={'BANNED'}>{t('community.banned')}</Option>
```

Instead of:
```
<Option value={'BANNED'} label={t('community.banned')} />
```

client/coral-ui/components/Dropdown.js
alinhle: I don't find anywhere in the code that `this.state.label` is used. Can you point me to it?

client/coral-ui/components/Dropdown.js
alinhle: Why is the latter version not semantic?

client/coral-ui/components/Dropdown.css
alinhle: The previous one was faster. The animations should be fast enough to give you a feeling of instantanity. You should never have to wait for an animation to end even if its just some milliseconds. It's an important concept of material design :-)

client/coral-ui/components/Dropdown.js
```diff
@@ -0,0 +1,84 @@
...
+    super();
+
+    this.state = {
+      isOpen: false
+    };
+  }
+  
+  setValue = (value, label) => {
+    if (this.props.onChange) {
+      this.props.onChange(value, label);
```
alinhle: don't pass label, it should not be needed.

kgardnr (approved):

client/coral-ui/components/Dropdown.js
```diff
@@ -66,10 +66,13 @@ class Dropdown extends React.Component {
...
 
   renderLabel() {
-    if (this.props.label) {
-      return this.props.label;
-    } else if (this.props.value) {
-      return this.props.value;
+    const options = React.Children.toArray(this.props.children);
+    const option = options.find((option) => option.props.value === this.props.value);
+
+    if (option.label) {
```
alinhle: You need to check if an option was found.

```js
if (option) {
  return option.label ? option.label : option.value;
} else if (this.props.value) {
  return this.props.value;
} else {
  return this.props.placeholder;
}
```

client/coral-ui/components/Dropdown.js
```diff
@@ -69,10 +69,10 @@ class Dropdown extends React.Component {
     const options = React.Children.toArray(this.props.children);
     const option = options.find((option) => option.props.value === this.props.value);
 
-    if (option.label) {
-      return option.label;
-    } else if(option.value) {
-      return option.value;
+    if (option.props.label) {
```
alinhle: You need to check if an option was found.

```js
if (option) {
  return option.props.label ? option.props.label : option.props.value;
} else if (this.props.value) {
  return this.props.value;
} else {
  return this.props.placeholder;
}
```

okbel:
Ready! Thank you! @cvle

client/coral-ui/components/Dropdown.css
```diff
@@ -3,15 +3,27 @@
...
+
+  line-height: 1.4;
+  font-size: 13px;
 }
 
-.dropdown {
+.toggle {
+  padding: 10px 15px;
   cursor: pointer;
+  outline: none;
```
okbel: We shouldn't remove `outline` this one is for accesibility purpuses.

alinhle (approved):

PR #1086, opened 2017-10-11
Count leading and trailing whitespace towards char count

alinhle:
## What does this PR do?
- Leading and trailing whitespace was not counted towards the max char count.
- https://www.pivotaltracker.com/story/show/151864634

## How do I test this PR?

- Turn on max char count
- Create comment with leading and trailing whitespace
- Submit button should be disabled if char count exceeds max char count.


kgardnr (approved):
Works perfectly! 💯

PR #1099, opened 2017-10-12
Email Verification Resend

wyattjoh:
## What does this PR do?

Prompts the user to resend their email verification if it has not been received.

Related: https://www.pivotaltracker.com/story/show/151851092


services/limit.js
```diff
@@ -0,0 +1,47 @@
...
+  key(value) {
+    return `limit[${this.prefix}][${value}]`;
+  }
+
+  async get(value) {
+    const key = this.key(value);
+
+    return client().get(key);
+  }
+
```
alinhle, line 38: Some comments for this method would be awesome :-)

kgardnr:
Nice! This works well:

new_pass

My only question is I noticed 2 users being generated - shouldn't these be generated in the test db, not the dev db? @cvle @wyattjoh

kgardnr (changes requested):
Outstanding question

kgardnr (approved):

kgardnr:
I think the above is something with my local env, approving! 💯

PR #1116, opened 2017-10-19
No Comments Text

okbel:
## No Comments Text
As a reader, I want to see copy that tells me if there are no comments to currently display
- https://www.pivotaltracker.com/story/show/152041888


kgardnr (approved):
:100:!

PR #1138, opened 2017-10-30
disable report button when comment already reported

### LINKED ISSUE #1044: When I've already reported a comment, the Reported button shouldn't then be clickable
It's not clickable, but it looks like it should be; just need to change the mouseover behavior

blackcathacker:
## What does this PR do?

Disabled the report button if the comment has already been reported. Fixes #1044 

## How do I test this PR?

Report a comment and see that the report button pointer is now default and the button is disabled.

Before changes
image

After changes
image


kgardnr (approved):

kgardnr:
This is great, thanks @blackhathacker!!

Issue #1051, opened 2017-10-04
Auto-approve reported usernames of Admins, Moderators, and Staff members


kgardnr:
Story: As a moderator, I want usernames of admins, mods and staff members to be auto-approved, so I don't have to ever moderate them

AC:
- If someone flags a username of an admin, moderator, or staff member, the username is not shown as Reported, and the username does not to go to Flagged Usernames queue in Community

holmberd:
I wouldn't mind taking a look at this.

kgardnr:
That would be awesome @holmberd!

kgardnr:
@holmberd Did you still wanna look into this? If so, make sure you go through the upgrade instructions for 4.0.0 https://coralproject.github.io/talk/migration/4/ - let us know if you need anything!

kgardnr:
This is fixed!

PR #1142, opened 2017-10-30
Reported comments

PepeFranco:
## What does this PR do?
Solves issue #1051 

## How do I test this PR?

- Report a comment or username of a user with Admin, Moderator or Staff
- See that the report is not shown


kgardnr (dismissed):

graph/mutators/action.js
```diff
@@ -33,12 +33,18 @@ const createAction = async ({user = {}, pubsub, loaders: {Comments}}, {item_id,
...
+      await UsersService.setStatus(item_id, 'PENDING');
+    }
   }
 
   if (comment) {
-    pubsub.publish('commentFlagged', comment);
+    let user = await UsersService.findById(comment.author_id);
+    if(!user.roles){
+      pubsub.publish('commentFlagged', comment);
+    }
```
alinhle: These changes only removes the live updates for staff comments. Can you revert these changes on the comment part? We will create a new ticket to tackle this as it requires a different approach. The reported queue shows comments that are flagged and neither approved nor rejected. It's too expensive to query whether the user is a staff or not for each comment to populate the list, so our approach is to mark staff comments as approved as soon as they are posted (and edit should not change that).

graph/mutators/action.js
```diff
@@ -33,12 +33,18 @@ const createAction = async ({user = {}, pubsub, loaders: {Comments}}, {item_id,
 
   if (item_type === 'USERS' && action_type === 'FLAG') {
 
-    // Set the user as pending if it was a user flag.
-    await UsersService.setStatus(item_id, 'PENDING');
+    // Set the user as pending if it was a user flag and user has no Admin, Staff or Moderation roles
+    let user = await UsersService.findById(item_id);
+    if(!user.roles){
+      await UsersService.setStatus(item_id, 'PENDING');
+    }
```
alinhle, line 40: 👍

alinhle (changes requested):
See my remarks. We'll merge the user flagging part, and create a new ticket for comment flags.

graph/mutators/action.js
alinhle: Thanks for your work.

PepeFranco: Sure I can do that and maybe take a look at that other ticket once it is created

alinhle (approved):
Thanks @PepeFranco!

Issue #1013, opened 2017-10-04
Show Staff Badge in Flag Details for Moderators


kgardnr:
STORY: As a moderator, I'd like to easily see if a "staff" member has flagged a comment, so that I know to "reject" that comment quickly without having to read it or think about it

AishaBlake:
I'd like to get involved but could use a little more guidance in getting my development environment up and running. @kgardnr could you help me out with that? And is there anything documenting what this flag should look like?

kgardnr:
Welcome @AishaBlake! That sounds great.

We don't have designs for where it should show on the Moderator side, I'll see if I can get our designer @IAmSamHankins to give some guidance on that.

In regards to getting your dev env set up - here's the best place to start: https://coralproject.github.io/talk/

A few tips:
1. You can either [use Docker if you're comfortable with that](https://coralproject.github.io/talk/installation-from-docker/), or you can [install from source and just run Mongo and Redis separately](https://coralproject.github.io/talk/installation-from-source/)
2. Here are the required Environment variables you'll need: https://coralproject.github.io/talk/configuration/ (you can either load these from a .env file locally OR you can leverage docker-compose.yml if you go the Docker route)
3. We have a group of amazing people working in a bunch of different timezones, so feel free to ask for help if you need it, we usually respond pretty quickly!
4. You can also join our Slack channel for 1:1 assistance, just send an email here and we can add you! coralcommunity@mozillafoundation.org

kgardnr:
I've made some edits above to clarify how the Staff badge should show:

Here are the other "Flag badges", e.g. markers that show Moderators why a comment is showing in Reported.

![reported comment markers](https://user-images.githubusercontent.com/1077300/31220109-4c6a271e-a9b7-11e7-8c99-9ddc3b0aab0c.png)

The User one shows it was flagged by a User. I think you could re-use this styling and just change the word User to "Staff", indicating it was flagged by a Staff member. Staff members are: Admins, Moderators, or Staff.

AishaBlake:
Thank you so much! That's exactly what I was looking for! I see now that there's a link in the main README, but that would perhaps be a good addition to `CONTRIBUTING.md` as well. Thanks also for following up on the design. I think that's everything I need to get started and I'll certainly reach out if I have questions or get stuck!

kgardnr:
Amazing @AishaBlake!

PR #1143, opened 2017-10-31
show staff instead of user badge when reported by staff

blackcathacker:
## What does this PR do?

Fixes issue #1013 . Displays 'Staff' in badge when Admin, Moderator, or Staff report a comment instead of just 'User' I know this issue was claimed by someone else, but...enjoying this project and looking for my hacktoberfest numbers. :) Feel free to close if I'm overreaching here.

## How do I test this PR?

report a comment as a staff member and see the comment in the reported queue says 'Staff' in the badge instead of 'User'.

image


kgardnr (dismissed):

client/coral-admin/src/components/CommentLabels.js
```diff
@@ -6,10 +6,19 @@ import FlagLabel from 'coral-ui/components/FlagLabel';
...
+
 function isUserFlagged(actions) {
   return actions.some((action) => action.__typename === 'FlagAction' && action.user);
 }
 
+function getUserFlaggedType(actions) {
+  return actions
+    .filter((action) => action.__typename === 'FlagAction' && action.user)
+    .map((action) => action.user.roles.some((role) => staffRoles.includes(role)))
+    .some((staff) => staff) ? 'Staff' : 'User';
```
alinhle: This chain of methods seems a bit complicated. Why not
```js
function getUserFlaggedType(actions) {
  return actions
    .some((action) =>
      action.__typename === 'FlagAction' &&
      action.user &&
      action.user.roles.some((role) => staffRoles.includes(role))
    ) ? 'Staff' : 'User';
}
```

alinhle (changes requested):
Looking good feature-wise. 
See my small remark below.

kgardnr:
Hey @blackcathacker just a small comment above and then we can merge it - thank you! 👍

client/coral-admin/src/components/CommentLabels.js
blackcathacker: That is cleaner. Updated.

alinhle (approved):
Awesome thanks @blackcathacker

PR #1144, opened 2017-10-31
add numbered shortcuts for mod queue

### LINKED ISSUE #1042: Keyboard shortcuts to move between Mod Queues
AC:
- Add keyboard shortcuts `1` `2` `3` `4` `5` etc to move between queues in order of queues.
- Would have to be careful with how to handle plugin-generated queues potentially?

blackcathacker:
## What does this PR do?

Fixes #1042. Adds numbered shortcuts for moderation queues. Pressing 1-5(or 6 if premod exists) will jump to that queue. This might be a bit of a WIP...not sure if this needs to handle the case of the queues changing while the moderation component is loaded.

## How do I test this PR?

Navigate to the moderation page and press 1-5.

image


kgardnr (dismissed):
Works great!

client/coral-admin/src/routes/Moderation/components/Moderation.js
```diff
@@ -34,25 +34,34 @@ class Moderation extends Component {
...
+    this.selectQueue(menuItems[nextQueueIndex].key);
+  }
+
+  selectQueue = (key) => {
+    const assetId = this.props.data.variables.asset_id;
+    this.props.router.push(this.props.getModPath(key, assetId));
+  }
+
+  getMenuItems = () => {
+    return Object.keys(this.props.queueConfig).map((queue) => ({
```
alinhle: This might as well be an `array` of keys no? `[key1, key2, ...]` instead of `[{key: key1}, {key: key2}, ...]`

locales/en.yml
```diff
@@ -296,6 +296,7 @@ en:
     show_shortcuts: "Show Shortcuts"
     singleview: "Toggle single comment edit view"
     thismenu: "Open this menu"
+    jumptoqueue: "Jump to specific queue"
```
alinhle: Use `jump_to_queue` to match the convention. (I know not all keys follow the convention yet..)

client/coral-admin/src/components/ModerationKeysModal.js
```diff
@@ -12,6 +12,7 @@ const shortcuts = [
       'k': 'modqueue.prev_comment',
       'ctrl+f': 'modqueue.toggle_search',
       't': 'modqueue.next_queue',
+      '1...5': 'modqueue.jumptoqueue',
```
alinhle: It would be super cool if the number `5` here would correspond to the exact number of queues. If you pass over a `queueCount` property from `Moderation`, and turn this from a static object to a function, it should work.

alinhle (commented):
Looks good, what do you think of my remarks? @blackcathacker

kgardnr:
Same for above @blackcathacker! 🎉

client/coral-admin/src/routes/Moderation/components/Moderation.js
blackcathacker: Yeah. Pulled that out directly from previous code, but I'll clean it up.

blackcathacker:
All of the comments make sense and I've updated based on the feedback.

alinhle (approved):
Sweet! Good job @blackcathacker

PR #1158, opened 2017-11-13
Allow CSS overrides manifest for custom styling

joshkadis:
## What does this PR do?

_As a developer implementing Talk, I'd like to be able to override CSS module import paths so I can have more control over Talk's look._

Using Talk's admin option to load a CSS file is great for minor tweaks, but bigger modifications to the styling of the UI require a lot of style overrides and `!important`. 

This PR allows a developer to create a `css-overrides.json` file in the Talk root directory. This file contains an array of resources, each with relative paths from the Talk root to the "old" and "new" resources.

If `css-overrides.json` exists, is valid JSON, and is not empty, Talk's Webpack config will use the [`NormalModuleReplacementPlugin`](https://webpack.js.org/plugins/normal-module-replacement-plugin/) to override the original resource with the new one. _Both the original and replacement resources must be a CSS files._

### Example

In `css-overrides.json`:

```
[
  {
    "oldPath": "client/coral-embed-stream/src/components/Comment.css",
    "newPath": "plugins/my-cool-plugin/styles/Comment.css"
  }
]
```

Now `Comment.js` will [import](https://github.com/coralproject/talk/blob/master/client/coral-embed-stream/src/components/Comment.js#L11) its `styles` object from `plugins/my-cool-plugin/styles` instead of the origin `client/coral-embed-stream/src/components`.

### Note about selectors

The developer is responsible for making sure that all CSS selectors required by the React component (e.g. `Comment.js`) exist in the new CSS file. 

I.e. if `Comment.js` refers to `<div className={styles.anExample} >`, it's on you to make sure that `.anExample{ ... }` exists in your new `Comment.css`. If not, it will fail silently and be rendered as `<div class="undefined">`.

## How do I test this PR?

1. Create a file `css-overrides.json` in the Talk root using the above JSON snippet.
1. Copy `client/coral-embed-stream/src/components/Comment.css` to `plugins/my-cool-plugin/styles/Comment.css`
1. In the new `Comment.css`, add the property `color: orange` to the selector `.content`
1. `yarn build`
1. The text of comments in the embed stream should be orange.
1. `git status` should not show any tracked changes.

Edge-case handling:

* Edit `css-overrides.json` so that it is no longer valid JSON
* Edit `oldPath` so that it ends with `Comment.js`
* Edit `newPath` so that it ends with `Comment.js`

Then rebuild. In all cases, the Webpack build should complete without error but the styles will be inherited from the original CSS file, `coral-embed-stream/src/components/Comment.css`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,4 +51,6 @@ plugins/*
 !plugins/talk-plugin-flag-details
 !plugins/talk-plugin-slack-notifications
 
+css-overrides.json
+
 **/node_modules/*
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,5 @@
 const path = require('path');
-const fs = require('fs');
+const fs = require('fs-extra');
 const CompressionPlugin = require('compression-webpack-plugin');
 const autoprefixer = require('autoprefixer');
 const precss = require('precss');
@@ -133,6 +133,22 @@ const config = {
     ]
   }
 };
+
+//==============================================================================
+// CSS overrides resolver
+//==============================================================================
+const cssOverrides = fs.readJsonSync(path.resolve(__dirname, 'css-overrides.json'), { throws: false });
+if (cssOverrides && cssOverrides.length) {
+  cssOverrides.forEach(({ regExpStr, newRelPath }) => {
+    // Must be a CSS file
+    if (regExpStr.endsWith('.css')) {
+      config.plugins.push(new webpack.NormalModuleReplacementPlugin(
+        new RegExp(`${regExpStr}$`),
+        path.resolve(__dirname, newRelPath)
+      ));
+    }
+  });
+}
 
 //==============================================================================
 // Production configuration overrides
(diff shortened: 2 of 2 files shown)


wyattjoh:
It is our desire in the future to utilize text extraction (via https://github.com/webpack-contrib/extract-text-webpack-plugin) instead of inline style tags as currently implemented by our webpack setup (via the https://webpack.js.org/loaders/style-loader/).

The advantage here is that CSS files can be maintained/tweaked outside the build process, and in the presence of a option, it would replace the loaded CSS file in place for the custom CSS url instead. From my perspective, this is a cleaner approach, as it allows developers to completely control the experience without introducing potentially breaking changes to the existing CSS system.

In the situation where you want to change the behavior of specific things, the custom CSS file can be included after the extracted text CSS, thereby removing the requirement for the constant `!important` statements to change existing behavior.

I'm hesitant at this point to introduce another configuration file to modify CSS rules, so at this point, I'm going to close this PR. If you are using the docker onbuild image (or deploying via source), you can still include a custom webpack configuration file that can add your override. Until we have implemented the style refactor (with extract text) I'd suggest that course of action.

wyattjoh:
We haven’t prioritized making talk completely design-modular (aka you can move everything around) - we wanna go in that direction, but for now we’re focused on maturing features that prevent trolling and harassment. It is our intention long term to separate the client side and server side code in a way that will allow newsrooms to combine components (React) to build a commenting system that leverages our shared ecosystem. Completely understand your motivations though.

PR #1166, opened 2017-11-16
Check status_history and fallback to 'NONE'

alinhle:
## What does this PR do?
- Check status_history exists and fallback to 'NONE'


alinhle:
Unrelated, oops.

PR #1175, opened 2017-11-21
User CLI Fix

wyattjoh:
## What does this PR do?

Removes attached actions, comments from users that are removed.

## How do I test this PR?

1. Create a user
2. Have them like/respect things, comment on things
3. Delete the user using the CLI
4. Notice that counts and comments are removed/updated.


kgardnr (approved):
Works perfectly!

PR #1177, opened 2017-11-22
Graph Context Connectors

wyattjoh:
## What does this PR do?

Provides the context with a connectors object that can be used to access internal API's from Talk that we expose through this context style interface.


kgardnr (approved):
Approved by @cvle 🎉

PR #1186, opened 2017-11-30
Performance enhancements

alinhle:
## What does this PR do?
- [x] Throttled and batched subscription handling
- [x] Preserve immutability and correctly detect when query data has changes (using custom optimzied version of `graphql-anywhere` (https://github.com/cvle/graphql-anywhere/tree/optimized)
- [x] Preoptimize graphql documents
- [x] Decouple apollo redux
- [x] After testing and review: Remove death star button, only optimize documents in production mode

## How do I test this PR?

- Check that everything is still working
- Open moderation panel
- Open stream in another window with another user and click on Death Star button which posts 100 comments in  10 seconds


kgardnr:
This is looking really great so far @cvle!

kgardnr (dismissed):
:100:! Great work Kiwi!!

package.json
```diff
@@ -105,7 +106,7 @@
     "fs-extra": "^4.0.1",
     "gql-merge": "^0.0.4",
     "graphql": "^0.9.1",
-    "graphql-anywhere": "^3.1.0",
+    "graphql-anywhere-optimized": "git+https://git@github.com/cvle/graphql-anywhere#optimized",
```
wyattjoh: Replace with version pin from npm.

wyattjoh (changes requested):

wyattjoh (dismissed):

wyattjoh (approved):

PR #1195, opened 2017-12-05
Use graphql-ast-tools

alinhle:
## What does this PR do?
- Move document optimization code to graphql-ast-tools 
- Use new package graphql-ast-tools


wyattjoh:
Shouldn't this be swapped out?

https://github.com/coralproject/talk/blob/9bce69a7cdb24b0b306c34808d220877318278e8/webpack.config.js#L123

wyattjoh:
Also should you remove the dependency on `apollo-utilities`?

alinhle:
Right, I'll remove `apollo-utilities`. `graphql-anywhere-optimized` is still required. This package now uses `graphql-ast-tools` internally and we still need it.

wyattjoh (approved):

PR #1199, opened 2017-12-06
Less Lazy Assets

wyattjoh:
## What does this PR do?

Improves the asset fetching algorithm to attempt a fetch before the upsert. This will result in less writes to the database as a result.

**Note to 3.8.1 users, this rewrites some of the indexes found in that version with sparse indexes** (to improve performance) **and must be removed prior to activating this on production.**


graph/loaders/comments.js
```diff
@@ -82,54 +82,52 @@ const getParentCountsByAssetID = (context, asset_ids) => {
...
+    (ctx.user == null || !ctx.user.can(SEARCH_NON_NULL_OR_ACCEPTED_COMMENTS))
   ) {
     return null;
   }
 
-  if (statuses && statuses.length > 0) {
-    query = query.where({status: {$in: statuses}});
-  }
-
-  if (ids) {
```
alinhle, line 107: This is not required anymore?

wyattjoh: We don't actually use that query anymore no. That was used before when we had to search for all the flags and then find all the comments from those flags with the given id's. The query paths in place now never use the `ids`.

graph/loaders/assets.js
```diff
@@ -54,58 +37,121 @@ const getAssetsByQuery = async (context, query) => {
...
+  // Try to validate that the url is valid. If the URL constructor throws an
+  // error, throw our internal ErrInvalidAssetURL instead. This will validate
+  // that the url contains a valid scheme.
+  try {
+    new URL(url);
+  } catch (err) {
+    throw ErrInvalidAssetURL;
   }
 
-  let asset = await AssetsService.findOrCreateByUrl(asset_url);
```
alinhle, line 71: Whats the reason behind not implementing the changes in `AssetsService.findOrCreateByUrl`. Is it still required?

graph/loaders/comments.js
```diff
@@ -82,54 +82,52 @@ const getParentCountsByAssetID = (context, asset_ids) => {
...
-    query = query.where({id: {$in: ids}});
-  }
+  const query = CommentModel.find();
 
   if (asset_id != null) {
-    query = query.where({asset_id});
+    query.merge({asset_id});
```
alinhle, line 106: Is this `merge` function documented somewhere? couldnt find it on the www.

alinhle:
What happens if the old indexes are not removed? It won't run?

alinhle (approved):
LGTM

PR #1206, opened 2017-12-13
Fix logic to generate asset_url even if the asset_id is provided

jtnelson:
an error is avoided

## What does this PR do?
Prior to this commit, the embed logic would only generate an `asset_url` if one wasn't provided and the `asset_id` also wasn't provided. This commit changes the logic so that the `asset_url` is generated if one is provided regardless of whether and `asset_id` is provided.

This is needed to support plugins that use fallback logic to create/load an asset by url if the asset id is invalid.


wyattjoh (approved):
Amazing catch! Looks great to me!

PR #1208, opened 2017-12-14
Mod Queue improvements

alinhle:
## What does this PR do?
- Integrate `react-virtualized`
- Infinite scrolling: When approaching end of the queue, new comments are automatically loaded.
- Limit notifications to comments that are close to current scroll position. 
- Comments now are never inserted in the middle of the queue, instead these are always appended to the end of the list.
- Introduce dangling comments: Instead of removing a comment mid-queue the comments are marked as dangling and they change their background color to indicate that the status of the comment has been modified by another moderator.
- Automatic queue cleanup: When switching queues dangling comments are removed and large queues are reduced to 100 comments.
- Better shortcut experience
  - Comments are marked as selected when clicked on (shortcut navigation starts from there)
  - When tabbing focus reaches a comment it's selected

## Goals
- Keep performance even with a large list
- Maintain performance even after longer use on heavy traffic
- Keep moderator distractions to a minimum (less sudden comment movements, less notifications)
- Better shortcut experience


kgardnr (approved):
This is amazing - I'll wait for reviews from @okbel and @wyattjoh before we merge!

okbel (approved):

PR #1209, opened 2017-12-14
i18n + Email Support

wyattjoh:
## What does this PR do?

- Support new server side plugin for translations:

> When a plugin provides a `translations` hook, they provide an absolute filename path to the translation yaml file that should be included, example: `translations: path.join(__dirname, 'client/translations.yml')`.

- Supports a new `TALK_EMAIL_SUBJECT_PREFIX` config that allows configuration of the email prefix. Defaults to `[Talk]`
- Supports sending emails in the default language selected with the `TALK_DEFAULT_LANG` config.

### DIFF
--- a/services/i18n.js
+++ b/services/i18n.js
@@ -41,7 +41,7 @@ const loadPluginTranslations = () => {
   }
 
   // Load the plugin translations.
-  plugins.forEach('server', 'translations').forEach(({plugin, translations: filename}) => {
+  plugins.get('server', 'translations').forEach(({plugin, translations: filename}) => {
     debug(`added plugin '${plugin.name}'`);
 
     const pack = yaml.parse(fs.readFileSync(filename, 'utf8'));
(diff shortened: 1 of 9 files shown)


services/i18n.js
```diff
@@ -1,57 +1,91 @@
...
+const loadPluginTranslations = () => {
+  if (loadedPluginTranslations) {
+    return;
+  }
+
+  // Load the plugin translations.
+  plugins.get('server', 'translations').forEach(({plugin, translations: filename}) => {
+    debug(`added plugin '${plugin.name}'`);
+
+    const pack = yaml.parse(fs.readFileSync(filename, 'utf8'));
```
alinhle, line 45: I think you need to deal with resolving the path relative to the plugin here.

wyattjoh: It's the expectation that the plugin will use something like ``` path.join(__dirname, 'client/translations.yml')``` to resolve it in file.

kgardnr (approved):

PR #1210, opened 2017-12-15
Redis Optimization Pass

wyattjoh:
## What does this PR do?

- removes call for `commentCount` when we already have `totalCommentCount`
- removes job processor cleaner on non-processing nodes
- fixed some template issues

### DIFF
--- a/client/coral-embed-stream/src/graphql/utils.js
+++ b/client/coral-embed-stream/src/graphql/utils.js
@@ -38,7 +38,7 @@ function applyToCommentsOrigin(root, callback) {
 function findAndInsertComment(parent, comment) {
   const isAsset = parent.__typename === 'Asset';
   const [connectionField, countField, action] = isAsset
-    ? ['comments', 'commentCount', '$unshift']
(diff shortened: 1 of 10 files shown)


alinhle (changes requested):
Need to remove https://github.com/coralproject/talk/blob/6850dd1a5359c00256461021748fb6fde4ee8254/client/coral-embed-stream/src/graphql/utils.js#L111 and https://github.com/coralproject/talk/blob/6850dd1a5359c00256461021748fb6fde4ee8254/client/coral-embed-stream/src/graphql/utils.js#L74

alinhle:
I realized unrelated to this PR that when I post a comment and then refresh, the number does not reflect the new count, seems like because its cached in redis, do we need to invalidate the cache somewhere?

alinhle (approved):

alinhle (changes requested):
Need to update paths to selenium-standalone due to the upgrade in `nightwatch.conf.js`

alinhle (approved):

PR #1214, opened 2017-12-18
Confirmation Dialog for Featuring Comments on the Admin Side

okbel:
## What does this PR do?
- Adds a Confirmation Dialog for Featuring Comments on the Admin Side

## How do I test this PR?
- Write a comment
- Go to the Admin Panel
- Go to Mod
- Feature a comment
- A dialog should pop up


plugins/talk-plugin-featured-comments/client/components/ModTag.js
```diff
@@ -51,10 +42,19 @@ export default class ModTag extends React.Component {
       </span>
     ) : (
       <span className={cn(styles.tag, {[styles.featured]: alreadyTagged})}
-        onClick={this.postTag} >
+        onClick={() => this.props.openFeaturedDialog(comment, asset)} >
```
alinhle: Use a bound method like `this.openFeaturedDialog `

plugin-api/beta/client/hocs/withTags.js
```diff
@@ -100,6 +100,7 @@ export default (tag, options = {}) => hoistStatics((WrappedComponent) => {
         postTag={this.postTag}
         deleteTag={this.deleteTag}
         config={config}
+        {...rest}
```
alinhle: Put `{...rest}` first so it doesn't overwrite the others.

plugins/talk-plugin-featured-comments/client/components/ModTag.js
okbel: Done. What's the reason behind it?

alinhle: It's better for performance to use a single instance of this function instead of creating a new one on each render.

kgardnr (approved):

PR #1215, opened 2017-12-18
*Fixes for sorted window handler

wyattjoh:
## What does this PR do?

Replaces some code to wrap more logic checks.


alinhle (approved):
If this works for you I'm fine. I'll take a closer look later at the overall integration code. Alternatively just disable e2e-ci for now.

PR #1230, opened 2017-12-20
Safari Browserstack changes

alinhle:
## What does this PR do?
- Use local IP instead of localhost to support Safari in Browserstack
- Safari is still not turned on for CI due to still existing bugs on BS


config.js
```diff
@@ -261,13 +263,13 @@ CONFIG.JWT_COOKIE_NAMES = uniq(CONFIG.JWT_COOKIE_NAMES.concat([CONFIG.JWT_COOKIE
 // testing environment. Every new mongo instance comes with a test database by
 // default, this is consistent with common testing and use case practices.
 if (process.env.NODE_ENV === 'test' && !CONFIG.MONGO_URL) {
-  CONFIG.MONGO_URL = 'mongodb://localhost/test';
```
wyattjoh: I'd leave the mongo url + redis url as localhost, as browserstack isn't affected by server side only params.

wyattjoh (dismissed):

config.js
alinhle: sure!

kgardnr (approved):
Wyatt approved 💯

PR #1237, opened 2017-12-22
CSS Customization and i18n fixes

wyattjoh:
## What does this PR do?

- Custom CSS is now loaded on all admin pages, giving admins the ability to change the styling on those pages.
- Fixes and improvements were made to the mailing service.
- Existing styles for secondary admin pages (password reset, confirm email) were updated.

## How do I test this PR?

1. Create a fresh Talk install.
2. Install Talk.
3. Create a new account (with email enabled).
4. Click confirm email dialog, and follow steps, notice the styling.
5. Change the custom css url to something like bootstrap.
6. Repeat 1-4, but notice the new styles :).

## Pending tasks

- [ ] translate admin page ui elements to es


routes/index.js
```diff
@@ -189,7 +189,7 @@ router.use('/', (err, req, res, next) => {
   if (err instanceof errors.APIError) {
     res.status(err.status);
     res.render('error', {
-      message: err.message,
+      message: res.locals.t(err.translation_key),
```
alinhle: ```res.locals.t(`error.${err.translation_key}`)``` ?

services/mailer.js
```diff
@@ -54,102 +54,108 @@ templates.render = (name, format = 'txt', context) => new Promise((resolve, reje
...
-}); // ends templates.render
+});
+
+const mailer = {};
+
+// enabled is true when the required configuration is available. When testing
+// is enabled, we will be simulating that emails are being sent, because in a
+// production system, emails should and would be sent.
+mailer.enabled = Boolean(
+  SMTP_HOST && SMTP_HOST.length > 0 &&
```
alinhle: Isn't it enough to just check `SMTP_HOST` instead of `SMTP_HOST && SMTP_HOST.length > 0`?

services/mailer.js
```diff
@@ -54,102 +54,108 @@ templates.render = (name, format = 'txt', context) => new Promise((resolve, reje
...
+ */
+mailer.send = async ({template, locals, to, subject}) => {
+  if (!mailer.enabled) {
+    throw new Error('email is not enabled because required configuration is not available');
+  }
 
-    if (!subject) {
-      return Promise.reject('sendSimple requires a subject for the email');
+  // Attach the template locals.
+  attachStaticLocals(locals);
```
alinhle: Instead of modifying what was passed as an argument I'd recommend copying the locals e.g.  `{...locals, t: i18n.t, staticLocals}`. It's best to reduce side effects and keep functions pure.

services/users.js
```diff
@@ -767,6 +767,36 @@ module.exports = class UsersService {
     }, tokenOptions);
   }
 
+  static async verifyEmailConfirmationToken(token) {
```
alinhle: Difference to `verifyEmailConfirmation`?

services/mailer.js
wyattjoh: I can look into using lodash.merge, we don't have object spread on the server :(

services/users.js
wyattjoh: `verifyEmailConfirmation` actually marks the email as verified, and uses `verifyEmailConfirmationToken` under the hood. Comments that I'll add now should make that clear :)

services/mailer.js
alinhle: This is the same as object spread: `Object.assign({}, locals, {t: i18n.t}, staticLocals)`

alinhle (approved):

PR #1255, opened 2018-01-08
Fix live status updates

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/154060536
- Live updates about user being banned, suspended or username rejected was not working anymore since apollo has been decoupled from the redux store.

## How do I test this PR?

- Ban, Suspend, Reject username and see it immediately active on the embed stream side


okbel (approved):

wyattjoh:
It seems that unbanning and unsuspending the user doesn't propagate, should we add events for this?

alinhle:
That'd be a new feature for another time :-)

PR #1265, opened 2018-01-10
Use lodash with es modules

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/153290233
- Optimize bundle size by using lodash es6 modules.


wyattjoh (approved):
Looks good to me!

PR #1267, opened 2018-01-10
Permission fix

wyattjoh:
## What does this PR do?

- Resolves permissions bug with server that permitted access to the community page to change roles

## How do I test this PR?

- Sign in as a moderator
- Try to set your account to an Admin
- You shouldn't be able to 😄


alinhle:
Shouldn't we allow the moderator to view the role?

image

kgardnr (approved):

PR #1269, opened 2018-01-11
Translations refinement for plugins

larrylizhao:
## What does this PR do?
Refine the translations in talk plugins for both simplified and traditional Chinese

## How do I test this PR?
- set TALK_DEFAULT_LANG to 'zh_TW' or 'zh_CN'
- run the talk instance with plugins
- see if everything is fine in Chinese


kgardnr (approved):

kgardnr:
Thank you @larrylizhao!

PR #1276, opened 2018-01-11
Missing Email Error

wyattjoh:
## What does this PR do?

- Does not error out when sending emails, instead prints a warning to the console (in the future, we can tell the frontend that the user didn't have an emaill, but often times, the sending of the email is secondary to the actual purpose, suspending a user doesn't _need_ an email but if they have it they'll get a good message.)

## How do I test this PR?

- Create a new Talk account with Facebook
- Login as an admin user (a different account)
- Suspend that user
- Notice the lack of error on the frontend and a warning in the console.


kgardnr:
This looks great to me! I'll wait for @okbel to give the ok before we merge!

okbel (approved):

PR #1288, opened 2018-01-16
Replies to hidden/deleted comments

wyattjoh:
## What does this PR do?

- When a reply is made against a comment that is not visible or has been deleted, the comment will be rejected.

## How do I test this PR?

1. Create a comment.
2. In another tab, reject that comment.
3. In the first tab, try replying to that comment.
4. See the error!

Tracking: https://www.pivotaltracker.com/story/show/153522862


kgardnr:
@wyattjoh @cvle we should improve this error:

replying_to_removed_comment

`error.COMMENT_PARENT_NOT_VISIBLE`

kgardnr:
@wyattjoh @cvle ah disregard, it is showing the proper message.

kgardnr (approved):

PR #1296, opened 2018-01-18
Auth Callback Rewrite

wyattjoh:
## What does this PR do?

- Rewrites the auth callback to properly sanitize the auth data and apply content security policies to that page


models/user.js
```diff
@@ -209,10 +209,6 @@ const UserSchema = new Schema(
         delete ret.__v;
         delete ret._id;
         delete ret.password;
-        delete ret.status.username.history;
-        delete ret.status.banned.history;
-        delete ret.status.suspension.history;
-        delete ret.tokens;
```
alinhle, line 215: Is keeping `tokens` a deliberate choice?

wyattjoh: It does not contain any critical data besides id's, and eventually might be a useful property.

alinhle (dismissed):

alinhle (approved):

PR #1301, opened 2018-01-22
Fixes for report dialog

alinhle:
## What does this PR do?
- When a user is flagged twice with the same reason show `You have already flagged this username.` instead of `Resource already exists`.
- Prevent dialog from progressing when no items where selected
- Previously when a comment was flagged or the username was flagged, the same button used to perform this action, will not execute any other mutation, even though the UI suggests that it does and successfully so. This patch contains changes to actually perform the mutations as the ui suggests.
- Fix bug where error messages where not translated anymore.
- https://www.pivotaltracker.com/story/show/154524763
- https://www.pivotaltracker.com/story/show/152472590

## How do I test this PR?

- Test 1
  - Flag a username  
  - Flag same username with the same reasons
  - It should yield `You have already flagged this username`

- Test 2
  - Click on continue on the report dialog without selecting any items.
  - It should not proceed

### DIFF
--- a/client/talk-plugin-flags/components/FlagButton.js
+++ b/client/talk-plugin-flags/components/FlagButton.js
@@ -9,7 +9,7 @@ import cn from 'classnames';
 import styles from './styles.css';
 import * as REASONS from '../helpers/flagReasons';
 
-import { getErrorMessages } from 'coral-framework/utils';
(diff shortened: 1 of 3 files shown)


kgardnr (approved):

DHorse2:
#ingnoreThisPost #1301 I will be deleted shortly.

PR #1303, opened 2018-01-22
Moderation refactor

wyattjoh:
## What does this PR do?

Replaces previous moderation array with a more modular approach and replaces `pre-git` with `pre-commit` to simplify the development workflow.


alinhle (approved):
It looks good to me

PR #1305, opened 2018-01-23
Auth Callback Regression

wyattjoh:
## What does this PR do?

Resolves a regression with the HTML entities encoding for the auth details when being serialized for HTML/JSON.

### DIFF
--- a/views/admin/password-reset.ejs
+++ b/views/admin/password-reset.ejs
@@ -33,6 +33,15 @@
     <script src="/public/javascripts/admin.js"></script>
     <script>
       $(function() {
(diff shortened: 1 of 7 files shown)


views/admin/password-reset.ejs
```diff
@@ -33,6 +33,15 @@
     <script src="/public/javascripts/admin.js"></script>
```
alinhle: You removed this file `public/javascripts/admin.js`

alinhle (changes requested):
look above

alinhle (approved):

PR #1309, opened 2018-01-24
Fix wrong text on "premod if contains links" setting

hellp:
## What does this PR do?

This fixes a wrong text for the "enable premod for comments with links" settings on the embed stream. The plain premod text was re-used which is incorrect.

## How do I test this PR?

- go to these settings
- see the correct text


kgardnr:
Nice catch @hellp! Thank you for fixing this!

kgardnr (approved):

PR #1311, opened 2018-01-25
Migration updates

wyattjoh:
## What does this PR do?

- Improves migrations to support larger scale migrations
- Added new flags to the migrations cli:

```
$ ./bin/cli migration run --help

  Usage: run [options]

  runs all pending migrations

  Options:

    -q, --query-batch-size <n>   change the size of queried documents that are batched at a time (default: 100)
    -u, --update-batch-size <n>  change the size of documents that are batched before the update is sent (default: 1000)
    -y, --yes                    will answer yes to all questions
    -h, --help                   output usage information
```

## How do I test this PR?

1. Create a 8GB collection of users + actions
2. Run the migration
3. Watch it take a long, long time, but not crash!


erikreyna (approved):
The changes Wyatt made have been tested against a copy of production data at The Washington Post.

Given the large dataset, the migration performed exceptionally well when compared to a previous test.

kgardnr (approved):

PR #1313, opened 2018-01-26
Port Activity dot polling to subscriptions

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/154365617
- https://www.pivotaltracker.com/story/show/154366467
- Switch activity dot polling to be based on subscriptions
- Fix bugs in modqueue sorting logic

## How do I test this PR?
- Activity dot should continue to work as before, regardless of
  - whether I'm currently in the modqueue 
  - whether I'm on another view (e.g. community)
- Try emptying the premod and reported queue with another user,
  while being in another view (e.g. community)


kgardnr:
@cvle the activity dot for Moderate seems to always stay when there is no Pre-mod queue, and once I've cleared things from the New and Reported queues. Are you seeing this too?

kgardnr (approved):

PR #1315, opened 2018-01-29
Better handle comment edits

alinhle:
## What does this PR do?
- Handle comment edits that does not rely on this bug: https://www.pivotaltracker.com/story/show/154722676

## How to test this PR
- Modqueue and activity dot should work as expected, especially when involving edits


client/coral-admin/src/routes/Moderation/graphql.js
```diff
@@ -124,12 +128,16 @@ function determineLatestChange(comment) {
 }
 
 function reconstructPreviousCommentState(comment) {
-  const history = comment.status_history;
+  const statusHistory = comment.status_history;
+  const bodyHistory = comment.body_history;
   const actions = comment.actions;
```
wyattjoh, line 133: Is there a reason why this style is used rather than destructuring?

```diff
- const statusHistory = comment.status_history;
- const bodyHistory = comment.body_history;
- const actions = comment.actions;
+ const {
+   status_history: statusHistory,
+   body_history: bodyHistory,
+   actions,
+ } = comment;
```

services/comments.js
```diff
@@ -37,14 +38,14 @@ module.exports = class CommentsService {
...
-                  created_at: new Date(),
+                  created_at,
                 },
               ]
             : [],
           body_history: [
             {
               body: input.body,
-              created_at: new Date(),
+              created_at,
```
wyattjoh, line 48: 👍

wyattjoh (approved):

PR #1319, opened 2018-01-29
User Deletion Cleanup

wyattjoh:
## What does this PR do?

- Removes old code related to db verifications from user deletion
- Introduces migration concepts to resolve updating counts


kgardnr (approved):
Awesome!

PR #1320, opened 2018-01-29
Stopstart

chrismika:
## How do I test this PR?

- click a button
- view the cat
- see the cat meow


kgardnr:
Hey @chrismika we're concerned about ya'll forking Talk and making changes to core - it's not the advised way to use Talk. Can you reach out to us via Slack when you get a moment?

PR #1322, opened 2018-01-30
Make Approve/Reject button styles more robust for longer labels

hellp:
## What does this PR do?

For other languages than English the 'Approve' text did not fit into the button. (We are currently working on a German localization of Talk.)

This commit makes the CSS a bit more flexible by removing some very pixel-specfic styles. (Also, imho, I think that aligning the buttons to the right side of the comment box increases usability when scrolling through a list of comments when moderating.)

## How do I test this PR?

Check the Approved/Rejected lanes in the admin interface. Buttons are aligned a bit more to the right. If you change the buttons' text to something longer than 'Approved' (e.g. the German 'Freigegeben') the button does not break anymore by having a ugly newline in it.

Before:
bildschirmfoto am 2018-01-30 um 14 35 36

After:
bildschirmfoto am 2018-01-30 um 14 36 18


kgardnr (approved):

kgardnr:
Thanks @hellp! This is great! Can't wait to see the German translation.

Issue #1241, opened 2017-12-29
Incorrect URL for "Ask Readers a Question" in the Configure tab


gakhov:
### Expected behavior

_Configure Comment Stream_ -> _Ask Readers a Question_ should call  `/${MOUNT_PATH}/client/0.js`

### Actual behavior

_Configure Comment Stream_ -> _Ask Readers a Question_ calls  `/client/0.js`

### Steps to reproduce behavior

1. Setup a CoralTalk Instance

```
'TALK_ROOT_URL=https://example.com/test/',
'TALK_ROOT_URL_MOUNT_PATH=TRUE',
'TALK_PORT=7777',
```
2. Initialize the instance (`/test/admin/install`)
3. Create a test page that is accessible under `https://example.com/test/` with content:

```
...
<div id='coralStreamEmbed'></div>
<script src="https://example.com/talk/embed.js" async onload="
  Coral.Talk.render(document.getElementById('coralStreamEmbed'), {
    talk: 'https://example.com/talk/'
  });
"></script>
...
```

4. Navigate to `https://example.com/test/` and login as superuser
5. Click the `Configure` tab
6. Under _Configure Comment Stream_ click on checkbox _Ask Readers a Question_
7. See that handler for the checkbox calls `https://example/client/0.js` that returns 404 because the actual script is under `https://example/client/test/0.js`

wyattjoh:
This is related to this line here:

https://github.com/coralproject/talk/blob/0af8a3115477a96ea58761f669276a71f2cb1d1e/webpack.config.js#L39

We might be able to patch it without requiring a rebuild however.

PR #1330, opened 2018-01-31
Public path

leeeandroo:
## What does this PR do?
Introduce a new environment variable `TALK_PUBLIC_PATH` to use on webpack config ans allow to set different static path. Useful when you're running the Talk not on root url. 

## How do I test this PR?

- Insert the variable `TALK_PUBLIC_PATH` on .env;
- Build and run Talk;
- Check if your public files are loading from your new path;

Also, It fix this issue: #1241


wyattjoh:
Sorry for not getting back to this issue @leeeandroo, and thanks for your contribution! I feel however that I don't want to introduce another environment variable into our build process, as it will complicate our already complicated configuration.

I've pushed up PR https://github.com/coralproject/talk/pull/1331 that should resolve this issue without requiring additional variables, or recompilation if the variable change.

If you can review that solution to see if it works for you (by checking out the `public-path` branch), we can have that merged into 4.1.0 soon.

PR #1333, opened 2018-02-01
Cleanly remove auth data from storage

alinhle:
## What does this PR do?
- Cleanly remove auth data from storage when they are not needed anymore

### DIFF
--- a/plugins/talk-plugin-auth/client/components/SignInContainer.js
+++ b/plugins/talk-plugin-auth/client/components/SignInContainer.js
@@ -38,8 +38,7 @@ class SignInContainer extends React.Component {
   }
 
   componentDidMount() {
-    window.addEventListener('storage', this.handleAuth);
-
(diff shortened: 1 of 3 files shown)


plugins/talk-plugin-auth/client/components/SignInContainer.js
```diff
@@ -60,6 +60,7 @@ class SignInContainer extends React.Component {
     if (e.key === 'auth') {
       const { err, data } = JSON.parse(e.newValue);
       authCallback(err, data);
+      localStorage.removeItem('auth');
```
wyattjoh: Because there is still a listener to the `storage` when this function mutates the storage, won't this cause a loop until the component is unmounted?

Can you check the `newValue` property of `e` to see if the value is a string? (See https://developer.mozilla.org/en-US/docs/Web/Events/storage)

wyattjoh (approved):

PR #1346, opened 2018-02-06
Adds logout for non-admin users

mattmontgomery:
## What does this PR do?

This adds support for logging out when a non-administrative user attempts to access the admin section.

## How do I test this PR?

- As a non-administrative user, visit the admin section
- Attempt to use the 'sign out' button. You should now be logged out.

### DIFF
--- a/client/coral-admin/src/containers/Layout.js
+++ b/client/coral-admin/src/containers/Layout.js
@@ -75,10 +75,7 @@ class LayoutContainer extends React.Component {
       );
     } else if (loggedIn) {
       return (
-        <Layout
-          handleLogout={logout}
-          {...this.props}
-        >
+        <Layout handleLogout={logout} {...this.props}>
           <p>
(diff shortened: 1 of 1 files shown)


kgardnr:
Hey @mattmontgomery, I tried testing this and I'm not seeing the Logout button on the Admin. 

Can you explain the use case for this feature?

mattmontgomery:
@kgardnr Sure. It looks like this is being masked by another bug, wherein the admin section (/admin/moderate) isn't showing the full height of the settings menu at the top-right when you get an access denied screen. It will scroll, however, and if you scroll down in that context, you can find the 'sign out' menu item, which is the item that was not working. That item only did not work when you receive an access denied screen.

The following screenshots show where that field exists.

talk_-_coral_admin

talk_-_coral_admin

wyattjoh (approved):
Works for me!

PR #1350, opened 2018-02-06
Notifications

wyattjoh:
### Events

Rewrite of the events service used in Talk. Events are now the source for the subscription system which simply use the `Broker` interface to send events to the internal PubSub as well now as the instance event emitter. This allows action by action resolution to plugins implemented on the Talk nodes which can allow for some interesting features.

### Connect API

Exposes the `graph/connectors.js` via the `connect` hook.

```js
module.exports = {
  connect(connectors) {
    // use `connectors`, contents of https://github.com/coralproject/talk/blob/b758dc91cb1f1969ecd895b6059306b318995b33/graph/connectors.js#L104
  }
}
```

See https://github.com/coralproject/talk/blob/90290cfa2de88e62f687e1ed0235ba6dfe4cde26/plugins/talk-plugin-notifications/server/connect.js for an example.

### Notifications

This also introduces the notifications plugins:

- talk-plugin-notifications
- talk-plugin-notifications-category-reply

That creates a NotificationManager that creates and manages events send from the event emitter that is linked to the Graph API PubSub system.

Adding the `talk-plugin-notifications` plugin will also enable the `notifications` plugin hook. Any plugin that registers before the `talk-plugin-notifications` plugin will get picked up by.

See https://github.com/coralproject/talk/blob/8b669a31c551a042f0f079d8cfc16825673eb8f0/plugins/talk-plugin-notifications-reply/index.js for an example.

### Email Templates

Email templates can now be registered using the connect API, see https://github.com/coralproject/talk/blob/8b669a31c551a042f0f079d8cfc16825673eb8f0/plugins/talk-plugin-notifications/server/connect.js#L12-L28


graph/resolvers/comment.js
```diff
@@ -29,15 +31,8 @@ const Comment = {
...
-  replyCount({ reply_count }) {
-    // A simple remap from the underlying database model to the graph model.
-    return reply_count;
-  },
-  actions({ id }, _, { user, loaders: { Actions } }) {
-    if (!user || !user.can('SEARCH_ACTIONS')) {
-      return null;
-    }
-
+  replyCount: property('reply_count'),
```
alinhle, line 34: Interesting interesting

serve.js
```diff
@@ -79,6 +80,9 @@ async function onListening() {
  * Start the app.
  */
 async function serve({ jobs = false, websockets = false } = {}) {
+  // Run the deffered plugins.
```
alinhle: Typo :-P

kgardnr (dismissed):

kgardnr (dismissed):

kgardnr (approved):

PR #1351, opened 2018-02-06
Provides username for subscription fields

mattmontgomery:
## What does this PR do?

Fixes an issue where usernames in events for moderated comments across multiple sessions show as `undefined`

image

## How do I test this PR?

- Open two instances of the administration interface, both with different logins 
- Open the same queue with both (e.g., rejected)
- Perform an action on one item, such as approve
- You should see the username appear with the rest of the message


wyattjoh (approved):
Works for me! Nice small change :)

PR #1353, opened 2018-02-08
Add missing translation keys

kgardnr:
## What does this PR do?
- Updates missing core translation keys for all current languages
- Updates missing plugin translation keys for all current languages

## How do I test this PR?

- Change between languages in local storage
- See that the page is being translated properly


kgardnr:
Hey @larrylizhao, we've been updating some translation keys that were missing and we have a few Chinese missing - both Simplified and Traditional. If you have some free time, we'd love your help filling in those gaps - thank you!

wyattjoh (approved):

larrylizhao:
Hi @kgardnr, sorry for the late reply.
I'd love to add the missing translations for both Simplified and Traditional Chinese when have time.
Will send pr and let you know then!

Issue #1035, opened 2017-10-04
When a user is suspended and tries to post, show them an informative error message


kgardnr:
AC:
- Currently says: "An unexpected error has occurred".
- Should say: "Your account is currently suspended. It will be reactivated in [TIME]. Please contact us if you have any questions."

fernandezpablo85:
I'd be happy to work on this if it's still available.

kgardnr:
@fernandezpablo85 yes it is! Welcome and thank you! :)

kgardnr:
Hey @fernandezpablo85 how's it going with this? Can we help or support you in any way?

fernandezpablo85:
Sorry, I had to travel for work and I'm having trouble finding free time to
setup the dev environment :(. If there's someone else that can tackle it
faster feel free to unassign me and give the issue to him/her.

kgardnr:
No worries @fernandezpablo85! Thanks for letting us know, but feel free to take your time and get to this when you can too :)

facuesr:
Hi @kgardnr , hope you are doing well!! Can I take a look at this?

kgardnr:
Sure that sounds great!

kgardnr:
Hey @facuesr did you still want to work on this by chance? If so, make sure you go through the upgrade instructions for 4.0.0 https://coralproject.github.io/talk/migration/4/ - let us know if you need anything!

PR #1354, opened 2018-02-12
Informative msg when the user is suspended and tries to post a message

facuesr:
## What does this PR do?
- Adds an informative message when the user is suspended and tries to post a message. Solves #1035 

## How do I test this PR?

- Log in with a test user. 
- Suspend the test user.
- Try to post a comment with the suspended user.


kgardnr:
Thank you @facuesr!

Fixes #1035

kgardnr (approved):

PR #1355, opened 2018-02-12
Lock install issue

leeeandroo:
## What does this PR do?
Fix wrong syntax on `serve.js` file. 
If you set `TALK_INSTALL_LOCK=TRUE` on environment and try to start, system you throw an error `'install lock active'` because the switch statement has a syntax error.

## How do I test this PR?
- Set `TALK_INSTALL_LOCK=TRUE`
- Try to start application `sudo yarn start`
- See the application running


wyattjoh (approved):
The inital proposed changes where the errors are `||` together, simply yields the first result (as it's truthy). The amended code correctly breaks (without throwing an error) with the environment variable present.

PR #1362, opened 2018-02-13
Added/updated missing Dutch translations

bartee:
## What does this PR do?
Added Dutch translations that were missing for both core and plugins.

## How do I test this PR?
- Change between languages in local storage
- See that the page is being translated properly


kgardnr (approved):

kgardnr:
@bartee awesome!! Thank you so much!

PR #1390, opened 2018-02-21
Yarn Workspaces

wyattjoh:
## What does this PR do?

Replaces custom module dependency management with [Yarn Workspaces](https://yarnpkg.com/en/docs/workspaces). This should reduce package size requirements and simplify the plugin flow.


alinhle (dismissed):
What a great match for us right? I would check if this is affecting our release scripts.

kgardnr (approved):

PR #1391, opened 2018-02-21
Rich Text Editor Plugin

okbel:
# Rich Text Editor Plugin
Support for Rich Text edition for Talk 🎉 - `talk-plugin-rich-text` and `talk-plugin-rich-text-pell`

This PR introduces two plugins: 
- `talk-plugin-rich-text`
- `talk-plugin-rich-text-pell`

`talk-plugin-rich-text` provides rich text support server-side and `talk-plugin-rich-text-pell` client-side.

`talk-plugin-rich-text-pell` uses Pell - [pell - jaredreich.com](https://jaredreich.com/pell) - Pell is the simplest and smallest WYSIWYG text editor for web, with no dependencies. Pell is around **1.11kB** (min+gzip) 

We need to **always** persist the original body in case we need it for future chores. This PR adds `htmlBody` to the `comment` schema  that returns the comment of the body with HTML.

#### Why we should store both `htmlBody` and `body`
- In case someone wants to toggle the plugin off/on.  We can offer support for both. Nothing breaks.
We show `body` or `htmlBody` field.

- No need for migration. 
If someone has the plugin on for a while, and then turns it of for a short period of time and then turns it back on. If you enable the plugin again you will provide rich text for old comments. They will remain untouched. 

#### Why html and not markdown?
- If we pick markdown we can decrease the size of the body field. But we will have to add another library on the client in order to render it. I'd prefer not adding another library to the client.
- If we pick html, we already have a renderer. The browser :)

#### Why should we store the original body?
- Because should **always** have the original. This is the real input from the user. 
- Comment Edit
- To make sure we moderate always based on the user input
- We might add features in the future, like search and we don't want that to be polluted with html or markdown stuff.

#### Any safety issue that we should know storing and rendering HTML?
We can add [DOMPurify](https://cure53.de/purify) in the server to make sure that the input is safe and the output that we return is also safe.

## Summary of changes
- Adds new slot `draftArea` 🎉 
- Provides a default text area component
- Adds a new RTE plugin that uses [Pell](https://jaredreich.com/pell)
- Stores original body
- Stores body as html
- Renders comment html if exists
- Comment editing
- Comment History
- Add DOMPurify
- Adds README.md
- Adds plugin to the docs

## TODO:
- Make the editor user friendly [Waiting for Design]


.vscode/settings.json
```diff
@@ -0,0 +1,2 @@
+{
```
alinhle: File accidentally commited

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
```diff
@@ -24,8 +24,8 @@ export default class DraftAreaContainer extends React.Component {
     return `${STORAGE_PATH}_${this.props.id}`;
   };
 
-  onChange = e => {
-    this.props.onChange && this.props.onChange(e.target.value);
+  onChange = (...args) => {
+    this.props.onChange && this.props.onChange(...args);
```
alinhle: Better be explicit `(body, data) =>`. That way it's immediately obvious what the arguments mean.

package.json
```diff
@@ -184,6 +184,7 @@
     "timeago.js": "^2.0.3",
     "timekeeper": "^1.0.0",
     "tlds": "^1.196.0",
+    "turndown": "^4.0.1",
```
alinhle: Not used

plugins/talk-plugin-rich-text-pell/client/index.js
```diff
@@ -0,0 +1,78 @@
...
+          const editCommentFragment = gql`
+            fragment Talk_EditComment on Comment {
+              body
+              richTextBody
+            }
+          `;
+
+          const fragmentId = `Comment_${id}`;
+
+          const data = proxy.readFragment({
```
alinhle: No need to read fragment when you don't need it, just write directly.

services/comments.js
```diff
@@ -110,6 +110,7 @@ module.exports = class CommentsService {
       $set: {
         body,
         status,
+        metadata,
```
alinhle: Here you are essentially removing all other metadata. See https://github.com/coralproject/talk/blob/master/services/settings.js#L59 on how to solve it.

services/comments.js
```diff
@@ -164,11 +165,14 @@ module.exports = class CommentsService {
       body,
       created_at,
     });
+
     editedComment.status_history.push({
       type: status,
       created_at,
     });
 
+    editedComment.metadata = metadata;
```
alinhle: Also here, you shouldn't overwrite other metadata.

plugins/talk-plugin-rich-text-pell/client/index.js
```diff
@@ -0,0 +1,78 @@
...
+          editComment: {
+            comment: {
+              richTextBody: edit.richTextBody,
+            },
+          },
+        },
+        update: proxy => {
+          const editCommentFragment = gql`
+            fragment Talk_EditComment on Comment {
+              body
```
alinhle: I think you are not using `body` anywhere.

client/coral-embed-stream/src/tabs/stream/components/EditableCommentContent.js
```diff
@@ -72,8 +74,14 @@ export class EditableCommentContent extends React.Component {
     }
   }
 
-  handleBodyChange = body => {
-    this.setState({ body });
+  handleBodyChange = (body, ...data) => {
```
alinhle: How about just `handleBodyChange = (body, data)`, we don't seem to utilize the array of data?

client/coral-embed-stream/src/tabs/stream/components/EditableCommentContent.js
```diff
@@ -72,8 +74,14 @@ export class EditableCommentContent extends React.Component {
...
+    this.setState(state => ({
+      body,
+      data: {
+        ...state.data,
+        ...data[0],
```
alinhle: Just `data`, see above.

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
```diff
@@ -88,8 +90,14 @@ class CommentBox extends React.Component {
       });
   };
 
-  handleBodyChange = body => {
-    this.setState({ body });
+  handleBodyChange = (body, ...data) => {
```
alinhle: Same here see above

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
```diff
@@ -88,8 +90,14 @@ class CommentBox extends React.Component {
...
+    this.setState(state => ({
+      body,
+      data: {
+        ...state.data,
+        ...data[0],
```
alinhle: Same here see above

client/coral-embed-stream/src/tabs/stream/components/DraftArea.js
```diff
@@ -35,22 +38,30 @@ export default class DraftArea extends React.Component {
...
     } = this.props;
 
+    const tASettings = {
+      value,
+      placeholder,
+      id,
+      onChange,
+      rows,
+      disabled,
+      comment,
```
alinhle: Don't pass comment here as a regular property, it should be passed as `queryData`, and a `DraftArea` Container with the slot fragments must be created and several other containers connecting up to the appropiate parent fragment container e.g. `Comment`.

alinhle (changes requested):
See comments

client/coral-embed-stream/src/tabs/stream/components/DraftArea.js
alinhle: Otherwise we have interdependent data dependencies of different components that do not have a parent-child relationship which we want to avoid. This also wouldn't work in relay because it has an accurate props filter.

kgardnr (dismissed):

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
```diff
@@ -71,4 +77,16 @@ DraftAreaContainer.propTypes = {
...
   rows: PropTypes.number,
   label: PropTypes.string.isRequired,
+  comment: PropTypes.object,
 };
+
+const slots = ['commentBox'];
+
+export default withFragments({
+  comment: gql`
+  fragment CoralAdmin_CommentDetails_comment on Comment {
```
alinhle: Need to adapt fragment name.

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
```diff
@@ -71,4 +77,16 @@ DraftAreaContainer.propTypes = {
   disabled: PropTypes.bool,
   rows: PropTypes.number,
   label: PropTypes.string.isRequired,
+  comment: PropTypes.object,
 };
+
+const slots = ['commentBox'];
+
+export default withFragments({
```
alinhle: The fragments still needs to be attached to related parent fragments leading up to a query.

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
alinhle: Something like `TalkEmbedStream_DraftArea_comment`^^

services/comments.js
```diff
@@ -110,6 +111,7 @@ module.exports = class CommentsService {
       $set: {
         body,
         status,
+        metadata: dotize(metadata),
```
alinhle: The resulting $set object should be:
```
$set {
  body: '',
  status: '',
  'metadata.dataA': 'valueA',
  'metadata.dataB': 'valueB',
}
```

Check out what `dotize` does. You might need to extend it's functionality.

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
```diff
@@ -201,8 +213,21 @@ CommentBox.propTypes = {
...
-export default connect(mapStateToProps, null)(CommentBox);
+export default compose(
+  connect(mapStateToProps, null),
+  withFragments({
+    comment: gql`
+    fragment TalkEmbedStream_CommentBox_comment on Comment {
+      __typename
+      ${getSlotFragmentSpreads(slots, 'comment')}
+      ...${getDefinitionName(DraftArea.fragments.comment)}
+    }
```
alinhle: Add after `}` another line: `${DraftArea.fragments.comment}`.

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
```diff
@@ -201,8 +213,21 @@ CommentBox.propTypes = {
   tags: PropTypes.array,
 };
 
+const slots = ['commentBox'];
```
alinhle: Don't add slots here

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
```diff
@@ -201,8 +213,21 @@ CommentBox.propTypes = {
...
 });
 
-export default connect(mapStateToProps, null)(CommentBox);
+export default compose(
+  connect(mapStateToProps, null),
+  withFragments({
+    comment: gql`
+    fragment TalkEmbedStream_CommentBox_comment on Comment {
+      __typename
+      ${getSlotFragmentSpreads(slots, 'comment')}
```
alinhle: Don't add slots here, it's included in draft area

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
```diff
@@ -71,4 +77,16 @@ DraftAreaContainer.propTypes = {
   disabled: PropTypes.bool,
   rows: PropTypes.number,
   label: PropTypes.string.isRequired,
+  comment: PropTypes.object,
 };
+
+const slots = ['commentBox'];
```
alinhle: I think we should rename this slot to `draftArea` because it confuses with our `CommentBox` container which is only used for `post` and `replies` but not for `edit`

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
```diff
@@ -1,17 +1,21 @@
...
+import DraftArea from './DraftArea';
 import Slot from 'coral-framework/components/Slot';
 import { connect } from 'react-redux';
 import { CommentForm } from '../components/CommentForm';
 import { notifyForNewCommentStatus } from '../helpers';
+import { compose } from 'react-apollo';
 
 // TODO: (kiwi) Need to adapt CSS classes post refactor to match the rest.
-export const name = 'talk-plugin-commentbox';
+export const name = 'talk-commentbox';
```
alinhle: I didn't rename it yet, because I scared to break some existing styling adaptions of newsrooms. Kim advised against changing the `css` names and do them all at once, e.g. when we make a major release.

client/coral-embed-stream/src/tabs/stream/containers/Comment.js
```diff
@@ -17,12 +19,13 @@ const slots = [
...
   'commentInfoBar',
   'commentActions',
-  'commentContent',
   'commentReactions',
   'commentAvatar',
   'commentAuthorName',
   'commentAuthorTags',
   'commentTimestamp',
+  'commentContent',
+  'commentBox',
```
alinhle: Don't add slot here, it's included in `DraftArea`

client/coral-embed-stream/src/tabs/stream/containers/Comment.js
```diff
@@ -94,7 +97,8 @@ const singleCommentFragment = gql`
       edited
       editableUntil
     }
-    ${getSlotFragmentSpreads(slots, 'comment')}
+    ${getSlotFragmentSpreads(slots, 'comment')}.
+    ...${getDefinitionName(CommentBox.fragments.comment)}
   }
```
alinhle: Add another line below the  `}`: `${CommentBox.fragments.comment}`

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
```diff
@@ -71,4 +77,16 @@ DraftAreaContainer.propTypes = {
...
+
+export default withFragments({
+  comment: gql`
+  fragment TalkEmbedStream_DraftArea_comment on Comment {
+    __typename
+    ${getSlotFragmentSpreads(slots, 'comment')}
+  }
```
alinhle: It would make sense to connect to `root` to, so the RTE plugin can e.g. access settings.

client/coral-embed-stream/src/tabs/stream/containers/DraftArea.js
alinhle: Need to also attach to `EditableComment` parent

alinhle (dismissed):

client/coral-embed-stream/src/tabs/stream/containers/CommentBox.js
kgardnr: Yep that's a good idea

alinhle: @kgardnr 'Cause it was your idea ;-)

plugins/talk-plugin-rich-text/server/config.js
```diff
@@ -0,0 +1,27 @@
...
+    className: 'talk-plugin-rich-text-link',
+    tagName: 'a',
+    target: {
+      url: '_blank',
+    },
+  },
+
+  // TODO: move to admin eventually
+  // Super strict rules to make sure users only submit the tags they are allowed
+  dompurify: { ALLOWED_TAGS: ['b', 'i', 'blockquote'] },
```
alinhle: New lines are currently just stripped out, I guess adding `br` should fix it?

alinhle:
Some display issues on the admin comments:
image
image

alinhle:
Long comments run out of the boundaries:
image

alinhle:
User Comment Detail does not include the RTE styling. we probably should add it.
image

kgardnr (dismissed):

wyattjoh (dismissed):
Made a small change, but looks good go to =D

wyattjoh (approved):

PR #1394, opened 2018-02-22
postMessage API

wyattjoh:
## What does this PR do?

Replaces the `localStorage` events edition of syncing the access tokens with the [Window.postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage) API.

This allows Talk to work flawlessly when 3rd party cookies are disabled/blocked by using cross origin safe communication means.

This also patches the storage service to use `sessionStorage` if `localStorage` is not accessible (like when 3rd party cookies are blocked).


client/coral-framework/services/postMessage.js
```diff
@@ -0,0 +1,86 @@
...
+import get from 'lodash/get';
+import remove from 'lodash/remove';
+
+/**
+ * createPostMessage returns a service that deals with cross
+ * window communication using the postMessage API.
+ * @param  {Object}  options
+ * @return {Object}  messenger service
+ */
+export function createPostMessage(origin, scope = 'client') {
```
wyattjoh: At what point do we replace a function that returns an object with a class constructor instead?

alinhle (approved):

Issue #1022, opened 2017-10-04
Quiet logs on production mode


kgardnr:
Example from The Washington Post:
`bundle.js:967 WARNING: heuristic fragment matching going on!`

wyattjoh:
To be more specific, this is related to logging from the client code.

nikoferro:
Hey, if this one is not taken, i would like to contribute to the project :) 
Just to double check, does this means suppressing all logging on the client side (on production mode), or are there any logs that should still appear after the change and just hide warnings and errors?

wyattjoh:
From my read of this warning, it's coming from an external package.

My best guess at a direction to take is to remove all `console.log` statements at build in production mode. Feel free to take a stab at if you like @nikoferro =D

nikoferro:
@wyattjoh sorry for the late reply, been disconnected the last few days. I ll take a shot at this one!

PR #1395, opened 2018-02-22
issue #1022 - quiet logs on production mode

nikoferro:
## What does this PR do?

removes console warnings and logs on production mode (issue #1022)

## How do I test this PR?

- build the app on production mode
- load client served by this build
- you shouldn't see any warnings except this one (This page does not include a canonical link tag. Talk has inferred this asset_url from the window object. Query params have been stripped, which may cause a single thread to be present across multiple pages.)


wyattjoh:
Looks great! Biggest concern is stripping this tag:

https://github.com/coralproject/talk/blob/011a6e646532ad6f813360520f905c60f67f382b/client/coral-embed/src/index.js#L19-L21

Which is critical for newsrooms while their installing Talk. Do you think there's a way to retain this one line?

nikoferro:
@wyattjoh From what i ve read in the docs, i *think* there is. Gonna take a look at whats possible and update the PR if there is a good solution for this

wyattjoh (approved):
Looks great!

PR #1402, opened 2018-02-23
slotSize Slot

wyattjoh:
## What does this PR do?

In some slots, it only makes sense to load a specific number of plugins for a slot (like 1). While this can be addressed with not enabling duplicate plugins that interface with the same slot, it's also possible to assist this on the client by limiting the visible slots by a single one if needed.

## How do I test this PR?

1. Create a plugin that implements the `commentContent` slot (a slot that has `slotSize={1}` set)
2. Enable it and the `talk-plugin-comment-content`
3. Notice that the order in the `plugins.json` dictates which will fill the slot.

### DIFF
--- a/client/coral-framework/components/Slot.js
+++ b/client/coral-framework/components/Slot.js
@@ -38,7 +38,7 @@ class Slot extends React.Component {
       'inline',
       'className',
       'reduxState',
-      'singleSlot',
+      'slotSize',
       'defaultComponent_',
       'queryData',
       'childFactory',
@@ -47,7 +47,7 @@ class Slot extends React.Component {
   }
 
   getChildren(props = this.props) {
-    const { singleSlot = false } = props;
+    const { slotSize = 0 } = props;
     const { plugins } = this.context;
 
     return plugins.getSlotElements(
@@ -55,7 +55,7 @@ class Slot extends React.Component {
       props.reduxState,
       this.getSlotProps(props),
       props.queryData,
-      { singleSlot }
+      { slotSize }
     );
   }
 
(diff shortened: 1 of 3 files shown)


client/coral-framework/components/Slot.js
```diff
@@ -32,27 +33,29 @@ class Slot extends React.Component {
...
+      'inline',
+      'className',
+      'reduxState',
+      'singleSlot',
+      'defaultComponent_',
+      'queryData',
+      'childFactory',
+      'component',
+    ]);
   }
```
wyattjoh, line 47: I see this hiding from the linter 😉

client/coral-framework/components/Slot.js
```diff
@@ -32,27 +33,29 @@ class Slot extends React.Component {
...
-      ...rest
-    } = props;
-    return rest;
+    return omit(props, [
+      'fill',
+      'inline',
+      'className',
+      'reduxState',
+      'slotSize',
+      'defaultComponent_',
```
alinhle: you copy and pasted a typo :-))

client/coral-framework/components/Slot.js
alinhle: Yeah I initially avoided `omit` apparently this triggers the inclusion of a big chunk of lodash, but even without `omit` we already do, so that's fine now.

alinhle (approved):

alinhle (changes requested):
Fix typo, rest looks good!

alinhle (approved):

alinhle (changes requested):
As discussed, outputting a warning when Slot has more components than `slotSize` allows, helps identifying misconfigurations

alinhle (dismissed):

alinhle (approved):

PR #1403, opened 2018-02-25
Ability to extend Configuration Settings

tomasdev:
## What does this PR do?

By adding a `metadata` object to the `Setting` model, we can leverage all current Plugins hooks to let plugin authors have their own configurations for plugins in the administration dashboard (you know, that place with the other `<ConfigureCard>`s)

## How do I test this PR?

Plugin «Must Read» coming soon, which has a configurable option to set a different Words Per Minute value will be coming up shortly, and should solve https://twitter.com/coralproject/status/902966149943615488

But I do need in the core in order to create the plugin (or well, I will release it with a hard-coded median WPM)
:-)

**Update 02/25/2018:** Sample plugin at https://github.com/tomasdev/talk-recipes/tree/master/plugins/admin-setting (it just adds a checkbox to enable/disable the plugin itself)


okbel:
Nice 😍! I can see the value in having the `metadata` field available. Let's see what @wyattjoh thinks

wyattjoh:
This looks like a decent enough simple addition 😄! It's important to note though that while using the metadata field with mongoose here makes sense, and extending it with the graph as is done in other parts of the application does work, it comes with a caveat.

If the plugin is later disabled, the cached value for the old metadata value will be kept until the settings are saved again. This may not be a big issue, but it's worth noting anyways. We don't have a plugin-register or similar hook to dump caches yet.

wyattjoh (dismissed):

tomasdev:
@wyattjoh on the same topic, how do plugins on user or comment metadata currently make sure their namespace is not overwritten? (i.e., Plugin 1 uses metadata.variableName and Plugin 2 uses the same variableName for theirs)

wyattjoh:
Currently, it's up to the plugin maintainers to namespace their access themselves, as it's often that plugins will interact with their own fields as well as core + plugin fields as well.

okbel (dismissed):

wyattjoh:
If you could run `yarn lint:js --fix` @tomasdev, we can get this one merged =D

tomasdev:
Done! thank you.

wyattjoh (approved):

PR #1407, opened 2018-02-28
CircleCI 2.0

wyattjoh:
## What does this PR do?

- Adds support for CircleCI 2.0.
- Adds Firefox support for local testing of PR's 🔥 🦊 
- Re-added Safari to integration suite during releases 🍎 
- Added parallel support for CI jobs 🚄 🚋 
- Made CI really, really fast

## How do I test this PR?

- Beauty of this PR, it'll test itself!


.circleci/config.yml
```diff
@@ -0,0 +1,265 @@
...
+        command: ./bin/cli setup --defaults
+    - run:
+        name: Run the integration tests
+        command: bash .circleci/e2e.sh
+    - store_test_results:
+        path: /tmp/circleci-test-results
+
+version: 2
+jobs:
+  # npm_dependencies will install the dependancies used by all other steps.
```
alinhle: %s/dependancies/dependencies/g
😉

.circleci/config.yml
```diff
@@ -0,0 +1,265 @@
...
+      - test_integration_safari:
+          <<: *filter_deploy
+          requires:
+            - build_assets
+      - deploy:
+          <<: *filter_deploy
+          requires:
+            - lint
+            - test_unit
+            - test_integration_safari
```
alinhle: This used to fail unless BS fixed their issues, does it work?

nightwatch.conf.js
```diff
@@ -49,5 +49,12 @@ module.exports = {
         },
       },
     },
+    'firefox-headless': {
+      desiredCapabilities: {
+        chromeOptions: {
+          args: ['--headless', '--disable-gpu', 'window-size=1600,1200'],
+        },
```
alinhle: Can you try adding this instead? 
```
'moz:firefoxOptions': {
   args: ['-headless']
 }
```

Obviously `chromeOptions` shouldn't have any effect here..

alinhle (changes requested):

.circleci/config.yml
wyattjoh: We can remove the requirement, but keep the test there if we want?

alinhle: Yes!

alinhle (approved):

PR #1410, opened 2018-02-28
Staff Reply Notifications

wyattjoh:
## What does this PR do?

Adds a notification type for when a staff member replies to your comment, you can get a notification!

## How do I test this PR?

1. Enable the plugin (server + client)
2. Have your settings enabled for "A staff member replies to my comment"
3. Post a comment.
4. Using a **different** staff account, reply to the comment.
5. Check your email!

This option will supersede the "All replies" notification option, so if both are enabled, and a staff member replies, you will only get the staff reply notification.

_Note: This uses the new `  notifications.supersedesCategories` plugin hook which defines which notifications this type of notification will supersede for this particular event. The category supersede filter **will not work across different events**._

### DIFF
--- a/plugins/talk-plugin-notifications-category-staff/client/.eslintrc.json
+++ b/plugins/talk-plugin-notifications-category-staff/client/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": "@coralproject/eslint-config-talk/client"
+}
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ plugins/*
 !plugins/talk-plugin-moderation-actions
 !plugins/talk-plugin-notifications
 !plugins/talk-plugin-notifications-category-reply
+!plugins/talk-plugin-notifications-category-staff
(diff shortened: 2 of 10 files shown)


plugins/talk-plugin-notifications-category-staff/client/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
alinhle, line 1: I guess this file isn't required anymore? ;-)

alinhle (dismissed):

plugins/talk-plugin-notifications-category-staff/client/.eslintrc.json
wyattjoh: This one is, there's no way to reference the client eslint yet with patterns.

alinhle: True, my bad, got confused

kgardnr (approved):

PR #1427, opened 2018-03-09
checks for and mitigates username collisions with new external users

kitwestneat:
## What does this PR do?
Adds a username collision test for new external users. Currently if there is already a user of the format `<Firstname>_<Lastname>`, no one with the same name will be able to sign up using an external SSO like Facebook or Google. This patch checks for collisions, and if found, adds a random number to the end of the username.

## How do I test this PR?
- Create a user with a username in the format `<Firstname>_<Lastname>`
- Attempt to create a second user using an external SSO provider with the same firstname / lastname


kitwestneat:
fixed a couple comma and semicolon issues

wyattjoh (approved):
I took your construct and tweaked it a bit to reduce the number of database calls, thanks for the contribution!

kitwestneat:
Hmm, it only reduces the number of database calls if there are 2 or more collisions right? That seems like a really unlikely event to add so much complexity for, plus the new database call is significantly heavier than the original one. Is the latency to mongodb really so bad that it's worth it?

kitwestneat:
oh you already merged it in, ok!

wyattjoh:
Typically, the highest place of latency is when we have to reach to Mongo to perform a query. The `find($in)` query is against an indexed field, so it should be very fast as it just needs to do a index lookup. I want it to at most cost us 2 calls to the database, as the previous version had to make `n` calls for `n` username tests, which would have cost much more in terms of latency.

PR #1428, opened 2018-03-09
Disable safari and ie integration test on CI

alinhle:
## What does this PR do?
- Disable safari and ie integration tests. Even though a successful build is currently not required, but we need the psychologically important GREEN success button ✅ 😄


wyattjoh (approved):
With 😢 I approve.

PR #1440, opened 2018-03-12
User Delete

wyattjoh:
## What does this PR do?

Moved user deletion into graph in preparation for GDPR tickets.


kgardnr (approved):

graph/typeDefs.graphql
```diff
@@ -1523,6 +1529,9 @@ type RootMutation {
 
   # forceScrapeAsset will force scrape the Asset with the given ID.
   forceScrapeAsset(id: ID!): ForceScrapeAssetResponse
+
+  # delUser will delete the user with the specified id.
+  delUser(id: ID!): DelUserResponse
```
alinhle, line 1534: I prefer `deleteUser` and `DeleteUserResponse` instead of shortening it. Wdyt?

wyattjoh: `delete` is a reserved word in javascript :(

PR #1441, opened 2018-03-12
Restore the userCreated event

### LINKED ISSUE #1426: Event service no longer exists, not sure how to trigger on events now
We have a plugin that relies on the USER_NEW event, but it looks like events are completely gone in master. Is there a replacement planned, or any alternate method to trigger when a new user is created?

wyattjoh:
## What does this PR do?

- Re-introduces an event when a user is created (see below): fixes #1426 
- Modifies the `createLocalUser` and `findOrCreateExternalUser` functions to accept a context object as the first parameter that it can use to fire the event.
- Introduces a graph subscription for users that are created.

An example of utilizing the new event is as follows:

```js
module.exports = {
  connect(connectors) {
    const { graph: { subscriptions: { getBroker } } } = connectors;
    
    // Get the handle for the broker.
    const broker = getBroker();

    // The passed in method will fire for every user that is created.
    broker.on('userCreated', user => {
      // Do something with the user...
    });
  },
};
```

You can also listen with a graph subscription to `userCreated`.

**Note: External plugins utilizing the Users service's user create methods must adjust their call signatures to pass in the context.**


alinhle (approved):
LGTM

PR #1442, opened 2018-03-13
Use asset.url as a fallback when asset.title is missing

ollyd:
## What does this PR do?
In the admin console, it populates the 'Story' field of a comment with the `asset.url`, if an `asset.title` is unavailable. Currently, in admin, when there is no `asset.title` that field remains blank.
This is already the current behaviour in `talk/client/coral-embed-stream/src/tabs/profile/components/Comment.js`.

## How do I test this PR?

- post a comment
- open admin console
- check that the comment displays the url of the story that was posted from


kgardnr:
Thank you @ollyd!

wyattjoh (approved):
Looks great to me! Thanks for the contribution!

PR #1446, opened 2018-03-14
Improve Slot and GraphQL framework

alinhle:
### Simplified API for `Slot`, `IfSlotIsEmpty`, `IfSlotIsNotEmpty` 
- Deprecate passing _unknown props_ to Slot children, see `passthrough`.
- No longer require the `data` prop, will print deprecation message.
- `Slot`: changed `slotSize` to `size`.
- Deprecate _queryData_, see `passthrough`.
- New property `passthrough` to put all the props you want to pass over to the Slot children without the need to distinguish query data and _regular_ props. 

Old API

```js
<Slot
  fill="adminCommentContent"
  data={data}
  className={className}
  clearHeightCache={clearHeightCache}
  queryData={queryData}
  slotSize={1}
  defaultComponent={CommentFormatter}
/>

<IfSlotIsEmpty
  slot="adminCommentContent"
  data={data}
  clearHeightCache={clearHeightCache}
  queryData={queryData}
/>
```

New API

```js
const slotPassthrough = {
  clearHeightCache,
  root,
  asset,
  comment,
};

<Slot
  fill="adminCommentContent"
  className={className}
  defaultComponent={CommentFormatter}
  size={1}
  passthrough={slotPassthrough}
/>

<IfSlotIsEmpty
  slot="adminCommentContent"
  passthrough={slotPassthrough}
/>
```
### New beta HOC `withGraphQLExtension`

There is a new HOC also exposed in `plugin-api/beta/hoc` that will allow components to register GraphQLExtensions for the framework. _IMPORTANT:_ The extensions are only picked up when the component is used in a slot. 

```js
// MyComponent.js
withGraphQLExtension({
  mutations: {
    UpdateNotificationSettings: () => ({
      update: proxy => {...}
    })
  },
  fragments: {...},
  query: {...},
})(MyComponent);

// my-plugin/client/index.js
export default {
  mySlot: [MyComponent],
}
```

### New beta HOCs `withRefetch`, `withFetchMore`, `withSubscribeToMore`, `withVariables`
These HOCs will give you access to the functions previously accessed through `data.refetch`, `data.fetchMore`, `data.subscribeToMore` and `data.variables`.

In case you used the `data` prop that was passed to the Slot Components, you need to adapt your code to get what you need from using these new HOCs that provide you with the props `fetchMore`, `refetch`, etc.

```js
// Pass `refetch` prop to `MyComponent`.
withRefetch(MyComponent);
```

### New internal HOC `withSlotElements`
Refactored logic that was previously spread in various components like `Slot`, `IfSlotIsEmpty`, `IfSlotIsNotEmpty` and `ExtendableTabPanel` into a new hoc `withSlotElements`.

```js
/**
 * Exports a HOC that provides a property `slotElements`.
 * @param  {Object}         [options]                      configuration
 * @param  {string|array}   [options.slot]                 Slot name or Array of Slot Names
 * @param  {element|array}  [options.defaultComponent]     Default Components or Array of such
 * @param  {number|array}   [options.size]                 Slot size or an Array of slot size
 * @param  {string}         [options.passthroughPropName]  The property to find the passthrough prop
 * @param  {string}         [options.propName]             New property name, defaults to `slotElements`
 *
 * @return {func}           Returns a HOC that per default provides the property `slotElements` with an
 *                          Array of Slot Elements or in case of multiple slots, an Array of Slot Element
 *                          Arrays.
 *
 * Example:
 * withSlotElements({
 *  slot: 'awesomeSlot',
 *  size: 1,
 })(MyComponent);
 */
```

### Moved `talk-plugin-profile-settings` to core
With the accomplished refactoring above, `talk-plugin-profile-settings` has been reimplemented back to core.

### Refactored `talk-plugin-notifications` and related plugins to take advantage of `withGraphQLExtension`

Checkout the new simplified code:

```js
import translations from './translations.yml';
import { t } from 'plugin-api/beta/client/services';
import { createSettingsToggle } from 'talk-plugin-notifications/client/api/factories';

const SettingsToggle = createSettingsToggle('onReply', () =>
  t('talk-plugin-notifications-category-reply.toggle_description')
);

export default {
  slots: {
    notificationSettings: [SettingsToggle],
  },
  translations,
};
```

### Other Stuff
- Refactored plugins to not use `data` prop and if necessary use the new HOCs `withVariables`, `withRefetch`, and alike.
- Core has been also adapted to not pass `data` prop and use the new Slot API
- Fixed numerous bugs where the Slot API was not properly used


wyattjoh (approved):

kgardnr (approved):

okbel (approved):
This solves most of the issues I've encountered while developing plugins. Huge improvement! Great work Kiwi🎉

PR #1447, opened 2018-03-14
Logout Fix

wyattjoh:
## What does this PR do?

Fixes a bug when a user tries to logout without having an authenticated request.

## How do I test this PR?

Run the following on `master` and on this branch:

```
curl --request DELETE \
  --url http://localhost:3000/api/v1/auth
```

On this branch, you'll see a not authorized error, but the server won't crash 😉

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ client/coral-framework/graphql/introspection.json
 *.swp
 *.DS_STORE
 .prettierrc.json
+.vscode
 
 coverage/
 test/e2e/reports/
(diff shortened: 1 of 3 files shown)


kgardnr (approved):

clintabrown:
Thanks for looking at this so quickly @wyattjoh @kgardnr!

TBeijen:
Can confirm that this fixes #1472 

As stated in the op, in it's current form it's a bit chatty (having the ApiError 401 in logs) and also shows errors in the browser console (see below). But both server and client code don't crash so it does the job.

Client errors raised after fix:
```
VM270:1 DELETE http://127.0.0.1:5000/api/v1/auth 401 (Unauthorized)
(anonymous) @ VM270:1

bundle.7b6d1099a8f457a095cc.js:1 Uncaught (in promise) Error: You are not authorized to perform this action.
    at bundle.7b6d1099a8f457a095cc.js:1
    at <anonymous>
```

PR #1452, opened 2018-03-15
Context Performance

wyattjoh:
## What does this PR do?

Proxied access to expensive context objects to reduce unnecessary calls. On local machine tests, it reduced the average request down by about 40ms.


graph/context.js
```diff
@@ -42,6 +42,32 @@ const decorateContextPlugins = (context, contextPlugins) => {
...
+/**
+ * Some pieces of the Context are quite complex to setup, using multiple merges
+ * and other lodash functions. This proxies that access such that it is only
+ * loaded if it is used. Helpful for a query that only uses a loader, and not a
+ * mutator.
+ *
+ * @param {Object} ctx the graph proxy
+ * @param {Function} loader the loadable component that should be proxied
+ */
+const proxyContextLoader = (ctx, loader) =>
```
alinhle: Let's call it `createLazyContextLoader` or `lazyLoad` or similar. It took me a bit to figure out what it does.

alinhle (changes requested):

alinhle (approved):

PR #1460, opened 2018-03-21
Comment Drafting and Editing Refactor + Coral RTE

alinhle:
## What does this PR do?
- Include a default `client` side implementation with `Coral RTE`, which is just a simple RTE based on the `contenteditable` and `execCommand` api with support for formatting `bold`, `italic`, and `blockquote`.
- Simplified `draftArea` API, now using `input` and `onInputChange`
- `preSubmit` hooks now accepts a `return value` as it's new input.
- Refactored hook support into `withHooks`
- Refactored and added hooks support to `EditableCommentContent`
- Deprecated previous `comment input tags` api. Now simply using `input` and `onInputChange` works.
- Better crossbrowser support

## How do I test this PR?

- Offtopic feature should remain intact
- Post, Reply and edit comments using the RTE


kgardnr (dismissed):

kgardnr:
This looks great @cvle and @okbel! I'll leave it to you both to merge it when it's ready.

kgardnr (approved):

PR #1464, opened 2018-03-21
Docs for the Plugins API

okbel:
- Adds docs for Plugins API


kgardnr:
@okbel @cvle @wyattjoh I moved the Client and Slots docs under the API section. I'll do a bit of formatting cleanup on it tomorrow and then we should be able to at least merge a WIP. If you want to take a look and provide any more feedback that would be great!

kgardnr (dismissed):

kgardnr (approved):

Issue #1250, opened 2018-01-08
Add "responding to X" (user) when comment nesting is flattened or always add this info.


DHorse2:
### Expected behavior
When I am reading a long thread of comments I need to know which commenter is being responded to. The conversation in the threads completely lose context.
### Actual behavior
Users on the TI site are now adding the specific person they are responding to at the top of their comment. When they remember.
### Steps to reproduce behavior
You can observe this on most any article.

**Context:**
[Commenter Features](https://coralproject.github.io/talk/commenter-features/)

> Threading
Talk supports by default 3 levels of threading, meaning each top-level comment has a depth of 3 replies; replies beyond that are not nested below the 3rd level. You can adjust this using the TALK_THREADING_LEVEL configuration variable. We don’t recommend deep threading because it can cause issues with styling, especially on mobile.

DHorse2:
@ehalferty your enhancement is perfectly acceptable and addresses the largest complaint in the community. Please allow me to provide some technical feedback on the issue directly related to user experience.

In short this solution:
If CommentNesting > MaxNesting Then Dispaly RespondingTo

This addresses user confusion over who a reply is addressed to when nesting is no longer a visual clue. However, this can be viewed in broader terms of when does the user become confused and loose track of the mental stack of comment replies (nested comment authors?) These following other metrics are correlated to the broader issue:

- 1. The number of comments seen since viewing the parent comment.
- 2. The elapsed time passed.
- 3. The total number of words (or lines) separating the two comments. (Amount read?)
- 4. The number of words typed or number of replies the user made in between.
- 5. The number of off-site links followed.

Some of them are interesting (causes of confusion/display triggers) although not really practical to implement, the elapsed time one is fairly straight forward. Over several months I saw confusion mentioned a number of times. It may have been mentioned 10 times re Coral so it is clearly important.

As a commenter, evaluator and fact checker, I personally find it impossible to keep track of deeply nested comment threads without the presence of RespondingTo. On numerous occasions I have returned to comments section after more than one hour elapsed.

UI Note: Newspapers want control over the style and what's displayed for numerous reasons. RespondingTo, Link copy, Share and other features can optionally be part of a hover or ellipse button. As long as it is easily available to the user their experience is not compromised in an important way.

I'll close this by noting the following about this change. It is structured, predictable and addresses the problem (for the user) and uses values already present in the application. Sounds good.

Regards, Dave Horsman

DHorse2:
**ignore this note** **(off topic)**
[...]

PR #1471, opened 2018-03-22
When comment reply is flattened, prefill comment with @<parent username>.

halferty:
## What does this PR do?

When the max threading level is hit, prefill replies with "@<parent username>" so users know which commenter the reply was meant for.

This is a potential improvement for https://github.com/coralproject/talk/issues/1250

What do you all think? Is this a reasonable solution to the problem? Or should we hardcode the "actual parent" information in some way? Or should this new behavior be controlled by a config option? Should this go through the roadmap or something? I just wanted to throw this PR up here to get your thoughts. :)

## How do I test this PR?

- Set TALK_THREADING_LEVEL to a low number, or leave it blank to use the default value of 3.
- Run the server.
- Post some replies to yourself to create a deeply-nested thread.
- Note that once the max threading level is hit, your posts will begin to be pre-filled with @<your username>.
- Note that if you reply to a different user, the posts are correctly pre-filled with @<their username>.

### DIFF
--- a/client/coral-embed-stream/src/tabs/stream/components/Comment.js
+++ b/client/coral-embed-stream/src/tabs/stream/components/Comment.js
@@ -373,6 +373,9 @@ export default class Comment extends React.Component {
       charCountEnable,
       root,
     } = this.props;
+
+    const reachedMaxThreadingLevel = depth >= THREADING_LEVEL;
+
     return (
       <ReplyBox
         root={root}
@@ -381,7 +384,10 @@ export default class Comment extends React.Component {
         charCountEnable={charCountEnable}
         maxCharCount={maxCharCount}
         setActiveReplyBox={setActiveReplyBox}
-        parentId={depth < THREADING_LEVEL ? comment.id : parentId}
+        parentId={reachedMaxThreadingLevel ? parentId : comment.id}
+        parentAuthorName={
+          reachedMaxThreadingLevel ? comment.user.username : undefined
+        }
         notify={notify}
         postComment={postComment}
         currentUser={currentUser}
--- a/client/coral-embed-stream/src/tabs/stream/components/ReplyBox.js
+++ b/client/coral-embed-stream/src/tabs/stream/components/ReplyBox.js
@@ -24,6 +24,7 @@ class ReplyBox extends React.Component {
       currentUser,
       notify,
       parentId,
+      parentAuthorName,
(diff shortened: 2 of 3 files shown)


wyattjoh:
Interesting idea! We were exploring ways to address this issue.

One of the proposed solutions was actually to remove the reply button on the deepest level, because the representation of the replies is directly tied to the max reply depth setting, it helps keep the conversation focused on the article/asset that's being discussed.

We'll bring this up with our team to review as a potential solution!

DHorse2:
@ehalferty your enhancement is perfectly acceptable and addresses the largest complaint in the community.  Please allow me to provide some technical feedback on the issue directly related to user experience.

In short this solution:
If CommentNesting > MaxNesting Then Dispaly RespondingTo

This addresses user confusion over who a reply is addressed to when nesting is no longer a visual clue. However, this can be viewed in broader terms of when does the user become confused and loose track of the mental stack of comment replies (nested comment authors?)  These following other metrics are correlated to the broader issue:

- The number of comments seen since viewing the parent comment.
- The elapsed time passed.
- The total number of words (or lines) separating the two comments. (Amount read?)
- The number of words typed or number of replies the user made in between.
- The number of off-site links followed.

Some of them are interesting (causes of confusion/display triggers) although not really practical to implement, the elapsed time one is fairly straight forward. Over several months I saw confusion mentioned a number of times.  It may have been mentioned 10 times re Coral so it is clearly important.

**As a commenter, evaluator and fact checker,** I personally find it impossible to keep track of deeply nested comment threads without the presence of RespondingTo.  On numerous occasions I have returned to comments section after more than one hour elapsed.

UI Note: Newspapers want control over the style and what's displayed for numerous reasons. RespondingTo, Link copy, Share and other features can optionally be part of a hover or ellipse button. As long as it is easily available to the user their experience is not compromised in an important way.

 I'll close this by noting the following about this change.  It is structured, predictable and addresses the problem (for the user) and uses values already present in the application. Sounds good.

Regards, Dave Horsman

wyattjoh:
Hi @ehalferty! We had a chat with the team, and we feel like we wanted to take a deeper dive to resolve this at a more fundamental level as it relates to our existing threading system. We don't want to introduce a solution that would not be compatible with future releases, so we've decided to close this for now. We'll update this PR when we have more to share!

PR #1473, opened 2018-03-22
Email Verification Copy

wyattjoh:
## What does this PR do?

When a email address is already verified, the copy will reflect that correctly.

## How do I test this PR?

1. Create a new account.
2. Click the email confirmation link.
3. Click the email confirmation link again.
4. Notice the correct copy!


kgardnr (approved):
💯

PR #1474, opened 2018-03-22
Bug fixes 

okbel:
## What does this PR do?
Fixes:
- https://www.pivotaltracker.com/story/show/156203156
- https://www.pivotaltracker.com/story/show/156203171


kgardnr (dismissed):

views/admin.ejs
```diff
@@ -4,17 +4,8 @@
...
-        height: 100%;
-        margin: 0;
-        background-color: #FAFAFA;
-        font-family: 'Roboto', sans-serif;
-      }
-
-      #root > div {
-        height: 100%;
-      }
+      html, body, #root, #root > div { min-height: 100%; }
```
alinhle, line 7: Scrolling issues on several pages with these changes. It took me a lot of trial and error to find the correct css controlling `height`. Try to find a solution that is very targeted, solves your issue but leaves the rest intact.

alinhle (dismissed):

views/admin.ejs
alinhle: The reason for the difficulties was to properly support `react-virtualized`.

alinhle: For instance you can just increase the height of the paragraph on the particular views that are causing problems.

client/coral-admin/src/components/Layout.css
```diff
@@ -1,4 +1,6 @@
 .layout {
   margin: 0 auto;
   background-color: #FAFAFA;
-}
+  height: inherit;
+  min-height: 100vh;
```
alinhle: As layout does not start at height: 0, this always means height > 100% and we'll always have a scrollbar.

views/admin.ejs
alinhle: Better yet, Just give `Layout` just the right amount of min-height to cover the popup, but not more.

client/coral-admin/src/components/Layout.css
alinhle: Give it sth like `300px`, this should be all you need to fix  https://www.pivotaltracker.com/n/projects/1863625/stories/156203156

okbel: I don't think this is a solution. If someone adds another link this could potentially be a bug again.

views/admin.ejs
okbel: Could you be more especific on how `min-height` could introduce errors? I see the queues perfectly working. Maybe I'm missing something. (the `height` is still `inherit` as you marked it on the first place)

kgardnr (approved):

PR #1477, opened 2018-03-23
Specify Subject Prefix config

mkonikov:
## What does this PR do?
Simply specifies in the notification plugin docs where the email subject prefix is set.

## How do I test this PR?
This PR only affects documentation.


wyattjoh:
We already specify the subject in our advanced configuration section:

https://docs.coralproject.net/talk/advanced-configuration/#TALK-EMAIL-SUBJECT-PREFIX

I'm hesitant to duplicate configuration docs, as the main docs search revealed it when I searched for "email subject".

mkonikov:
Maybe reference it somewhere/somehow?
It's not clear that an ENV var controls the subject without digging through the code. It doesn't even apply evenly to all notifications - prefix only appears before some. I assumed that it was fully controlled buy the `translations.yml` file.

wyattjoh:
Fair point, will address!

wyattjoh:
I've pushed a new PR that addresses this https://github.com/coralproject/talk/pull/1483. Let me know if the wording there makes sense @mkonikov!

PR #1478, opened 2018-03-23
Clarifying importance of order in plugins.json

mkonikov:
## What does this PR do?
Outlines the importance of ordering the plugins correctly when trying to override copy from another plugin. 

Before knowing this it presented an issue for me as I was trying to override the copy of the notifications plugin yet it did not work because I was using the `translations` hook inside my custom notification plugin which also needs to be listed before the core `talk-plugin-notifications` in order for the templates to register correctly.

## How do I test this PR?
This PR affects documentation only.


wyattjoh (approved):
Great comment!

PR #1484, opened 2018-03-26
Terms and conditions

okbel:
## What does this PR do?
- Adds the ability to dispatch `enableSubmitSignUpForm` and `disableSubmitSignUpForm` for `talk-plugin-auth`
- Adds `talk-plugin-health-report` that adds one last field to the Sign Up Form

More here: https://www.pivotaltracker.com/n/projects/1863625

Note: The translation service does not support receiving elements I was going to add an update to it but it's not worth it since we are going to upgrade our translation service. And more changes to it 

`talk-plugin-health-report` will be removed from this PR as soon as the review is done :)


wyattjoh:
`talk-plugin-health-report` should be extracted to another repository @okbel, as it's not a part of the core installation.

okbel:
👍 I'll leave it for testing and then I'll remove that folder. The PR is needed though, it has updates to the `talk-plugin-auth`

okbel:
@cvle ready to be checked! Thanks for the `hasBlocker` pattern 😄

plugins/talk-plugin-auth-checkbox/client/components/TermsAndConditionsField.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { Checkbox } from 'plugin-api/beta/client/components/ui';
+import { t } from 'plugin-api/beta/client/services';
+import styles from './TermsAndConditionsField.css';
+import cn from 'classnames';
+
+const TermsLink = () => (
+  <a
+    className={styles.link}
+    href={t('talk-plugin-health-report.terms-link')}
```
alinhle: Is the name `talk-plugin-health-report` here correct while the plugin name is `talk-plugin-auth-checkbox`?

plugins/talk-plugin-auth-checkbox/client/index.js
```diff
@@ -0,0 +1,9 @@
+import TermsAndConditionsField from './components/TermsAndConditionsField';
+import translations from './translations.yml';
+
+export default {
+  slots: {
+    'talkPluginAuth-formField': [TermsAndConditionsField],
```
alinhle: It feels a bit odd using the dash in combination with camelCase. What do you think of `talkPluginAuth.formField` instead?

alinhle (changes requested):
I removed `docs/source/_data/introspection.json` from the PR which seems to have been accidently commited and added it to `.gitignore`. I also changed the plugins service to produce a descriptive key which I replaced throughout the codebase where we relied on the talkPluginName instead. This makes sure no matter how many components a plugins adds to a slot, it still works. 

I adapted your blockers API to the above changed and simplified it a little bit. Besides the small review comments, it looks good!

plugins/talk-plugin-auth-checkbox/client/index.js
okbel: I like it :D

alinhle (dismissed):

kgardnr (approved):

PR #1485, opened 2018-03-26
Assets loader: fix typo getAssetsByID

dsteinel:
## What does this PR do?
This is about fixing a typo in a const. 
`genAssetsByID` it seems quite obvious that this should be named `getAssetsByID`

### DIFF
--- a/graph/loaders/assets.js
+++ b/graph/loaders/assets.js
@@ -2,7 +2,7 @@ const DataLoader = require('dataloader');
 const { URL } = require('url');
 const { singleJoinBy, SingletonResolver } = require('./util');
 
-const genAssetsByID = ({ connectors: { models: { Asset } } }, ids) =>
+const getAssetsByID = ({ connectors: { models: { Asset } } }, ids) =>
   Asset.find({
     id: {
       $in: ids,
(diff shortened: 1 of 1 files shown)


wyattjoh:
This was actually chosen intentionally! The concept was related to the fact that the function was designed for the https://github.com/facebook/dataloader, and was a "generator" style function for matching the signature of:

```
(ctx: graph.Context, queryIDs: Array<string>) -> Promise<Array<any>>
```

You can see the same pattern of naming here:

https://github.com/coralproject/talk/blob/0b624927b5e44a0aeedff90ee010c0765800bd07/graph/loaders/users.js#L212

And here:

https://github.com/coralproject/talk/blob/0b624927b5e44a0aeedff90ee010c0765800bd07/graph/loaders/actions.js#L177-L182

Though we seemed to have missed it here:

https://github.com/coralproject/talk/blob/0b624927b5e44a0aeedff90ee010c0765800bd07/graph/loaders/comments.js#L450

It should probably be updated/justified with comments, but we definitely appreciate the fined tooth comb of contributors verifying the codebase! Would love to hear any differing opinions on the naming issue 😄

dsteinel:
@wyattjoh ahh I see ... ok like this it makes sense! Thanks for replying on this.

wyattjoh:
No problem @dsteinel, and thanks for checking the code out!

PR #1488, opened 2018-03-26
Notification fixes

wyattjoh:
## What does this PR do?

- When a comment reply is not visible, it will not trigger a
notification email.
- When a comment is approved/accepted, we'll check to see if the comment
was visible before, if not, we'll delegate to the new reply notification
handler to possibly send a notification email.

## How do I test this PR?

1. Create a user account that has "notifications on replies" turned on.
2. Create a comment.
3. Create a second user account, and reply to that comment with a banned word, notice the lack of email notification?
4. Approve that rejected comment from the admin, notice the new email in your inbox ;)

### DIFF
--- a/plugins/talk-plugin-notifications-category-reply/index.js
+++ b/plugins/talk-plugin-notifications-category-reply/index.js
@@ -2,6 +2,12 @@ const { get } = require('lodash');
 const path = require('path');
 
 const handle = async (ctx, comment) => {
+  // Check to see if this reply is visible.
+  if (!comment.visible) {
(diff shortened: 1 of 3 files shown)


alinhle (dismissed):
Code looks good, I had no chance to test it though

okbel (approved):

mkonikov:
Does this also affect notifications from users you ignored? (e.g. you wouldn’t get a notification about a reply from a user you ignored)

wyattjoh:
Great question @mkonikov, I think it does not! I'll file a bug for that. Great catch! 🎉

PR #1493, opened 2018-03-28
[GDPR] Download My Comments

wyattjoh:
## What does this PR do?

Allows the user to request a download of their comment data in a zip file.


mkonikov:
Will this work the same as notifications where the mailer job can be cancelled and email can be sent out by a separate email provider using the `connect` hook?

wyattjoh:
Hey @mkonikov. Any emails sent by Talk uses our mailer service. If you've gone ahead and made changes to implement your own job processor for the mailer as described in #1439, then it will intercept the email sent from this service.

I'm also in the process of adding another graph mutation that can be called with an administrative token that will provide you with the direct download link to the zip file containing the account zip file. You could then implement this functionality into your own user management section so that you can include the download contents in your own download archive prepared for your users.

kgardnr (approved):

PR #1498, opened 2018-03-30
Admin Settings: Save and discard changes / Routes

okbel:
## Admin Settings: Save and discard changes / Routes

- Add an alert if someone changes a configuration, doesn't save, and leaves/closes the tab
- Adds routes for Settings Sections
https://www.pivotaltracker.com/story/show/156312019


wyattjoh (commented):
Question, the nature of the dialog seems to indicate that when I try to navigate away, and I click "save", it should just do so and then resume navigating. Is that not possible? Also, this implementation does not warn the user if I navigate away from the page, as in, we don't stop the browser from leaving before answering if they want to discard changes.

okbel:
Great! Changes added. Ready to be checked!

wyattjoh (approved):
Looks great to me! Love the changes.

PR #1507, opened 2018-04-04
Bug: User Comment History - After loading more comments, rejecting/accepting a comment reloads the user history

okbel:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/155836892

We addressed the updates to the cache in a previous PR. This deletes unnecessary fetchs to the server.


wyattjoh:
Not sure if related.. The user reject rate no longer updates when I click the accept/reject buttons for that panel, when previously it did.

okbel:
Interesting. Well.. yeah. It's related since the refetches also refeched that. I will investigate further.

okbel:
Now it updates the rejectRate based on the user actions but if another user rejects or approves it will not reflect the changes. This is generated on the FE exclusively and is only applied to the client cache and it's not reliable.

kgardnr (approved):

PR #1509, opened 2018-04-05
Safari External Auth

wyattjoh:
## What does this PR do?

Persist the token as handed by external auth into Talk.

### DIFF
--- a/client/coral-framework/reducers/auth.js
+++ b/client/coral-framework/reducers/auth.js
@@ -5,6 +5,7 @@ const initialState = {
   checkedInitialLogin: false,
   initialLoginError: null,
   user: null,
+  token: null,
 };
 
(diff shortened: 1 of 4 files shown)


client/coral-framework/reducers/auth.js
```diff
@@ -14,12 +15,18 @@ const purge = user => {
 
 export default function auth(state = initialState, action) {
   switch (action.type) {
+    case actions.SET_AUTH_TOKEN:
+      return {
+        ...state,
+        token: action.token || null,
```
okbel, line 21: Normally logic is not allowed in the reducer. This should be break down into 2 actions or send the value when it's dispatched - This is standard, but I'm fine either way.

okbel (approved):

PR #1512, opened 2018-04-09
Disable autocomplete on username signup field

okbel:
- Disable autocomplete on username signup field
https://www.pivotaltracker.com/story/show/156287975


plugins/talk-plugin-auth/client/login/components/SignUp.js
```diff
@@ -85,6 +86,7 @@ class SignUp extends React.Component {
                 showErrors={!!usernameError}
                 errorMsg={usernameError}
                 onChange={this.handleUsernameChange}
+                autocomplete="off"
```
wyattjoh, line 89: I'd also add:

```js
autocapitalize="none"
```

For the username field.

wyattjoh (changes requested):
Small change! Otherwise, wonderfully small update 😄

plugins/talk-plugin-auth/client/login/components/SignUp.js
okbel: 👍

wyattjoh (approved):

PR #1517, opened 2018-04-10
Refactored Errors

wyattjoh:
## What does this PR do?

Previously, we took the golang approach to managing errors, that is to say, we created error objects and passed them around our server and compared them directly. This method, unfortunately, does not work well with the promise style and exception throwing nature of NodeJS/Javascript as the stack is extremely important and is retained when properly creating error objects.

This PR serves to rectify this by replacing direct error objects with error classes that are thrown by the affected code, and replacing old equal comparisons with `instanceof` checks.

## How do I test this PR?

Wait for the integration + unit tests to pass.


okbel (approved):
👍

PR #1523, opened 2018-04-12
Reject username

okbel:
## Reject Username
- Adds Reject Username
- Adds a Reject Username Modal

This performs flagging and then rejects the username

## What does this PR do?
- https://www.pivotaltracker.com/story/show/155238146

## How to test this PR?
- Create an Admin
- Create a test user
- Comment with the test user
- Log as admin, and go to the mod queue
- Click the test user
- Reject the username clicking "Actions" next to the name
- A modal should open, Reject the username with a reason
- Change the username from the test user
- Approve or reject from the admin
- Test multiple flows


okbel:
@wyattjoh The server is returning while trying to reject an username:
```js
{"data":{"rejectUsername":{"errors":[{"translation_key":"NOT_FOUND","__typename":"GenericUserError"}],"__typename":"SetUsernameStatusResponse"}}}
```

Note: It does not follow the normal flow: flag -> reject. This is just rejecting from the user detail.

wyattjoh:
That should have been resolved in my last push @okbel 😄

okbel:
@wyattjoh this is ready to be tested. One concern is that when I repeat the flow the BE returns  "Resource already exists"

kgardnr:
This looks great @okbel!! 

The only bug I'm seeing is you get Unexpected Error, if the user changes their username, and you reject their second username.

To replicate:
1. Reject user
2. Login as that user, change your username
3. Login as admin, see new username is Reported Username queue - click Reject - you'll see the error
(If you Approve, it works as expected)

alinhle:
image

When I reject a username the reported queue becomes `-3`

alinhle:
image

I also see an error when you reject a username the second time with the same reason

EDIT:
@okbel I saw you already mentioned it.
@wyattjoh What's your input?

alinhle:
image

Also I can verify the issue that Kim reported

client/coral-framework/utils/user.js
```diff
@@ -35,6 +37,41 @@ export const isBanned = user => {
...
+
+export const getActiveStatuses = user => {
+  const statusMap = {
+    suspended: isSuspended,
+    banned: isBanned,
+    usernameRejected: isUsernameRejected,
+    usernameChanged: isUsernameChanged,
+  };
+
+  return toPairs(mapValues(statusMap, fn => fn(user))).filter(x => x[1]);
```
alinhle: Should return a flat array instead of pairs. The pair has no additional information compared to a flat array and as such has no function.

client/coral-framework/utils/user.js
```diff
@@ -35,6 +37,41 @@ export const isBanned = user => {
...
+  return get(user, 'state.status.username.status') === 'CHANGED';
+};
+
+/**
+ * getActiveStatuses
+ * returns an array of active status(es)
+ * i.e if suspension is active, it returns suspension
+ */
+
+export const getActiveStatuses = user => {
```
alinhle: We need to be careful introducing terms into the codebase especially when adding to the framework. What does an `active status` mean? What qualifies to be an `active status`? Is `username unset` an active status? is  `username set` an active status?

client/coral-admin/src/components/RejectUsernameDialog.js
```diff
@@ -0,0 +1,138 @@
...
+    const { reason, message } = this.state;
+    return (
+      <Dialog
+        className={cn(styles.dialog, 'talk-admin-reject-username-dialog')}
+        id="rejectUsernameDialog"
+        onCancel={onCancel}
+        open={open}
+      >
+        <div className={styles.close}>
+          <button
```
alinhle: We have a `BareButton` so you don't have to reset the styles

client/coral-admin/src/components/RejectUsernameDialog.js
```diff
@@ -0,0 +1,138 @@
...
+import styles from './SuspendUserDialog.css';
+import cn from 'classnames';
+import { RadioGroup, Radio } from 'react-mdl';
+import Button from 'coral-ui/components/Button';
+import { username as flagReason } from 'coral-framework/graphql/flagReasons';
+import t from 'coral-framework/services/i18n';
+
+const initialState = { reason: flagReason.offensive, message: '' };
+
+class RejectUsernameDialog extends React.Component {
```
alinhle, line 13: Too bad the forms stuff are not yet merged

alinhle (dismissed):

client/coral-framework/utils/user.js
alinhle: Or maybe there is a better term to describe your intention?

client/coral-admin/src/components/RejectUsernameDialog.js
okbel: This could be an easy update :')

client/coral-admin/src/components/RejectUsernameDialog.js
okbel: 👍

client/coral-framework/utils/user.js
okbel: If the user is suspended and banned: activeStatuses are: ban and suspension (not username for example) - We can come up with a better name.

kgardnr:
To dos:
- [x] Hide the Resource already exists error
- [x] Fix Reported Username count
- [x] Kiwi to have convo with Belen about other items above
- [x] Fix Unexpected error resolved

kgardnr (approved):

PR #1526, opened 2018-04-12
don't show profile tab if user is logged out

kitwestneat:
## What does this PR do?
This small patch removes the tab bar for anonymous users. There is no functionality available when the user is logged out, so removing the tab bar reduces clutter and potential confusion.

## How do I test this PR?
After applying the patch, observe that the tab bar is gone when logged out, but appears when logged in.


losowsky:
Thanks so much! We're taking a look at it this week.

wyattjoh:
Made a slight alteration to support when the tab bar is extended via a plugin (like the featured comments). To simplify the experience, because the tab bar exists high in the UI, maintaining the button (for even one tab) provides a visual cue that there should be buttons there to navigate.

We will be exploring hiding the tab bar in the future when we get around to our UI refresh, but thanks for the work!

wyattjoh (approved):

PR #1528, opened 2018-04-13
Check if token is in localStorage before logout

cristiandean:
## What does this PR do?

We are using an external auth platform, so we need to logout user on talk after the user is logged out on my auth system. The problem is that there is no way to, externally, check if the user is already logged on Talk and if we call TalkEmbed.logout() and the user is not authenticated, it raises an error on the client and server, printing a lot of error messages on the talk log.

So we thought that doesn't make sense calling the backend and doing all logout steps if there is no token in local storage. We made the change in this PR.

I would like to now your opinion about this change, and also how could we write an automated test for this.

## How do I test this PR?

- open debugger
- call TalkEmbed.logout()
- call TalkEmbed.logout() again
- should NOT raise error on server and client

**Before PR:**
screen shot 2018-04-13 at 14 47 35

screen shot 2018-04-13 at 15 27 15

**After PR:**
screen shot 2018-04-13 at 14 48 16


cristiandean:
Hello, any news about PR? Or how could we write an automated test for this.

Thank you.

kgardnr:
Hi @cristiandean! Apologies for the delay, we are heads down on GDPR support for our 4.4.0 release. We won't be able to give this our full attention until next week sometime. But thank you for contributing this and we will give feedback as soon as we can!

And yes we do think adding an automated test is always a good idea :)

wyattjoh:
Hey @cristiandean!

Sorry about the delay here. I've made a few changes that I think resolve some of the bugs you experienced.

The server response that there was a 401 as a result of not having auth credentials when logout is called is, in my opinion, a valid response. The only issue with the change you proposed is that the client may/may not have a cookie instead of a token on the request (that is `HttpOnly`), so the rest call should always be made to ensure that the cookie is handled.

In terms of handing the change of the auth state, you can listen for auth changes via the `coral-auth-changed` event from the `events` bus:

https://github.com/coralproject/talk/blob/93bda87ad061a2dc5eb8dc5b65a579a20efb76f7/views/article.ejs#L34-L44

Which should be emitted every time the auth state changes.

Hopefully that gives you some insight! Let me know if this fix that I applied makes sense to you.

wyattjoh (approved):

PR #1535, opened 2018-04-18
Fix Dutch locale for Moment.JS

daan944:
NL locale for Moment.JS was never imported, so some date/times were displayed in Portuguese (I guess since that's the last imported value).

To test: Set locale to Dutch, view a crud list admin panel and see Dutch as datetime. And: Open a comment thread, click a username and see the date in Dutch.


client/coral-framework/services/i18n.js
```diff
@@ -18,21 +19,21 @@ import daTA from 'timeago.js/locales/da';
 import deTA from 'timeago.js/locales/de';
 import esTA from 'timeago.js/locales/es';
 import frTA from 'timeago.js/locales/fr';
+import nl from 'timeago.js/locales/nl';
```
kgardnr: Should this be `nlTA`?

daan944: I only sorted those entries. For consistency it might be better to rename it, yes.

kgardnr: Ok great can you change it and we will merge it? Thanks @daan944!

daan944: Yes! I've just commited the change.

kgardnr (approved):

Issue #1540, opened 2018-04-24
Feature Idea: (temporarily) disable commenting site-wide (read-only switch)


hellp:
- As an admin I want a switch in the settings interface that disables commenting on the whole site, i.e. existing comments will remain readable but writing new comments (also replies) would not be allowed.
- Additionally I want to provide a short text that will appear instead of the comment input field. (E.g. "Commenting has been temporarily disabled.")

Use cases:

- technical issues in the automated moderation back-end (when using premod)
- all of the moderation team is sick
- other "worst-case" scenarios that temporarily may affect a whole site (shitstorms etc.)

From the user's POV this would be very similar to the per-asset close commenting feature Talk already has.

---

We are currently trying to implement this as a plugin, but it appears to be not possible to do it with plugin-only code. Any interest to have this in core?

kgardnr:
Same here @hellp! Will get back to you on this.

losowsky:
Thanks @hellp! We think this would be a good thing to add to core. We'd suggest adding another text box, like the "closed comments message" for after the time limit expires on keeping comments open, for writing the message.

Let us know what you need from us!

PR #1544, opened 2018-04-26
New setting: Disable commenting site-wide

hellp:
## What does this PR do?

Implements feature #1540 

## How do I test this PR?

- Go to Configure > Stream Settings
- Enable the new setting *Deactivate commenting site-wide*; optionally provide a message
- Load the comment stream on any asset page
- Result:
  - existing comments are still there, but the input field has been replaced with the message that commenting is currently disabled
  - if you try to submit a comment from a (non-updated) UI input, the back-end will throw an `COMMENTING_DISABLED` error with a proper message

## TODO

- [x] Write a test case
- [x] Find a better name for the setting. Not happy with `globalSwitchoffEnable` for a switch that actually *disables* commenting. How about `commentingDisabled` or `commentingOff`?


hellp:
If you have a good name for this setting, please tell me :)
Also if I should rebase or squash commits.

wyattjoh:
Hey @hellp!

I'd recommend changing:

- `globalSwitchoffEnable` to `disableCommenting`
- `globalSwitchoffMessage` to `disableCommentingMessage`

And if you can, resolve the merge conflicts so I can give it a test!

hellp:
- renamed
- rebased
- The Reply buttons on comments are now also hidden when commenting is disabled.

kgardnr (approved):

kgardnr:
This looks good to be from a product perspective @hellp! I'll let @wyattjoh do a code review and then we should be good to merge.

Thanks again for contributing this!

PR #1550, opened 2018-04-28
nightwatchjs.org doesn't support HTTPS

theel0ja:
## What does this PR do?
changes https to http on nightwatchjs.org link, because it doesn't support https.


kgardnr (approved):

kgardnr:
Thank you @theel0ja!

PR #1552, opened 2018-04-30
Make TALK_DEFAULT_LANG higher priority than locale and navigator lang…

larrylizhao:
## What does this PR do?
This pr makes the TALK_DEFAULT_LANG to have a higher priority than local storage and navigator language when choosing the locale.

## How do I test this PR?
1.  ADD TALK_DEFAULT_LANG=zh_CN in .env file
2. yarn && yarn build && yarn start
3. Open http://127.0.0.1:3000/admin/install in browser that language preference is in en, the page will still show Chinese as TALK_DEFAULT_LANG indicated. 

(Browser language was the main determiner of locale. It may disable the TALK_DEFAULT_LANG which is used by developer to indicate their preferred language for the product and causing English commenting box shown on a Chinese product)


kgardnr (approved):

wyattjoh:
Currently, the feature is working as intended, as we hoped that a user who browsers to the product would then be presented with the correct language.

The proposed PR as it is currently written, will never allow the language to be detected by the browser, and will never deviate from the default language. It is planned to make improvements to our i18n system, we can update here when we get around to drafting that particular feature.

larrylizhao:
I see, I'll try to enable CN in TALK by mapping the language code to locale code in pr #1561 .

But my concern with using navigator language prior to TALK_DEFAULT_LANG is that many users may never change the default language for their browsers, they are set when the browser is installed. 
So English comment box will be shown for Chinese websites in a browser which default language is English.
screen shot 2018-05-02 at 15 01 12

wyattjoh:
We intend to release a patch after we release our GDPR features that will allow the force setting of the language to disable auto-detection.

PR #1561, opened 2018-05-02
Map browser language code to locale code in TALK

larrylizhao:
## What does this PR do?
The window.navigator.language returns language code in different format with locale code in TALK.
Adding function to map the language code to something TALK can use.
e.g: 'zh-HK' or 'zh-TW'  is mapped to 'zh_TW' in TALK

## How do I test this PR?

1. yarn && yarn build && yarn start
2. Open http://127.0.0.1:3000/admin/install in browser that language preference is in en, the comment box will show in English.
3. Change the browser language preference to Chinese. Clear localStorage and refresh the page, comment box will show in Chinese.


larrylizhao:
Test failed for test_integration_chrome_local after merge with master on GitHub, not sure the cause.

kgardnr:
Hey @larrylizhao, I'm not able to replicate this bug you're fixing.

When I submit this to my console on Talk:

`localStorage.setItem('locale', 'zh-TW')`

I do not see the page update into Chinese. However, when I do:

`localStorage.setItem('locale', 'zh_TW')`

this does indeed work.

Am I missing something?

kgardnr:
For reference @larrylizhao: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language

We can certainly add support for `zh_HK` though. Let me know your thoughts on this!

larrylizhao:
Hi @kgardnr , thanks for the reply.
Indeed, 'zh-TW' can not work with TALK since the available locale in TALK for Chinese are 'zh_TW' and 'zh_CN' (https://github.com/coralproject/talk/blob/master/client/coral-framework/services/i18n.js#L48-L49).

Current problem is that `window.navigator.language` returns 'zh-TW' ,'zh-CN' and 'zh-HK' which cannot be consumed by TALK's locale system. So I wrote a mapping function to transform what `window.navigator.language` returns to something TALK can use.

As for the zh_HK, I think zh_TW and zh_CN in locale is sufficient to handle both traditional  and simplified Chinese. But we can add it when it's necessary.

kgardnr:
But I'm seeing in the documentation that `window.navigator.language` returns with underscore. When you set your browser to another language, is it not working?

mkonikov:
I'm seeing it listed with a dash in the [docs](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language#Value) - `Examples of valid language codes include "en", "en-US", "fr", "fr-FR", "es-ES", etc.`

larrylizhao:
Hi @kgardnr , When I try `window.navigator.languages` in console in Chrome, the browser returns language code as follows: https://drive.google.com/open?id=17LYaEwUCa_8z2HUkkCOW9Q3vnXC7zZwI . Safari also returns language code with dash.
So when in Chrome the browser language code cannot work for TALK locale.

wyattjoh:
Hey @larrylizhao!

In investigating the issue you raised, we realized that the way we were parsing and negotiating the language packs was pretty messed up!

We swapped out our custom code for the negotiation algorithm implemented in [fluent](https://projectfluent.org/). I've opened a new PR that specifically does this #1641, feel free to review if this resolve the issue.

If it does, we can close this PR in favor of #1641.

larrylizhao:
Hi @wyattjoh , the PR 1641 looks great.
I've tested it locally, it works fine for 'zh-CN' and 'zh-TW'.
Thanks.

PR #1584, opened 2018-05-07
Tracing

wyattjoh:
## What does this PR do?

Include trace ID in each downstream request.

## How do I test this PR?

Load the stream, look at the `/api/v1/graph/ql` request in the network inspector, notice the `X-Talk-Trace-ID` header that matches the `traceID` in the logs corresponding to the request.


kgardnr (approved):

kgardnr (changes requested):
I'm not seeing the header in Chrome or FF

kgardnr (approved):

PR #1589, opened 2018-05-08
Fixes for notifications

wyattjoh:
## What does this PR do?

Fixes a bug where notifications can be affected for users without ignored users lists.

## How do I test this PR?

1. Create 3 users, A, B, and C.
2. Create comments with User A, B, and C.
3. Enable "On Reply" notifications for Users A, B, and C.
4. Have User A ignore B.
5. Have User B write reply to User A.
6. Notice no notification is sent.
7. Have User B write a reply to User C.
8. Notice a notification is sent.


alinhle:
Code looks good, I'll test it also.

alinhle (approved):

PR #1593, opened 2018-05-08
Pass postDontAgree to replies

alinhle:
## What does this PR do?
- Fix _Don't agree with a reply yields an error_

## How do I test this PR?

- Don't agree with any reply


kgardnr (approved):

wyattjoh (approved):
Works for me!

PR #1599, opened 2018-05-09
CSS and Template Fixes + Improvements

wyattjoh:
## What does this PR do?

- [Fixed CSS ordering bug](https://www.pivotaltracker.com/story/show/157434681)
- Improved developer experience

## How do I test this PR?

1. Add custom CSS rules to overrride a property in the embed stream, taking care to **not** use `!important`.
2. Refresh the embed stream to see the new design with the change applied.


client/coral-admin/src/components/App.css
```diff
@@ -0,0 +1,9 @@
+html, body, #root, #root > div {
```
alinhle: `#root` will be transformed by _CSS modules_. To keep all global you need to wrap everything `:global { .. }`

alinhle (changes requested):

client/coral-admin/src/components/App.css
wyattjoh: But `#root` is defined outside the react context is it not?

alinhle: That's why you need to do this :-)

alinhle (approved):

PR #1603, opened 2018-05-10
Corrections and clarity

losowsky:
Fixed the incorrect thresholds and made the description clearer

### DIFF
--- a/docs/source/03-07-product-guide-trust.md
+++ b/docs/source/03-07-product-guide-trust.md
@@ -16,7 +16,7 @@ Here are the default thresholds:
 ```text
 -1 and lower: Unreliable
 0 to +1: Neutral
-+1 and higher: Reliable (we don't do anything with this label right now)
++2 and higher: Reliable (we don't do anything with this label right now)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/source/03-07-product-guide-trust.md
```diff
@@ -3,47 +3,53 @@ title: Trust
...
 
 Here are the default thresholds:
 
 ```text
--2 and lower: Unreliable
--1 to +2: Neutral
-+3 and higher: Reliable
+-1 and lower: Unreliable
+0 to +1: Neutral
++1 and higher: Reliable (we don't do anything with this label right now)
```
wyattjoh: This should actually be +2:

https://github.com/coralproject/talk/blob/4413e5616c8655b30f08144e78606033dd4a62a5/config.js#L245

wyattjoh (dismissed):
I think there's only one change.

docs/source/03-07-product-guide-trust.md
losowsky: oh, good point.

wyattjoh (approved):
Looks good to me!

PR #1604, opened 2018-05-10
Email Attach Fixes

wyattjoh:
## What does this PR do?

- Fixes bug in `excludeIf` invocation to include `inProgress` flag to allow wizard to complete
- Wrapped `Popup` `onunload` inspection with `try {} catch {}` to protect against security exemption errors throwing off the logic
- Prevent _Attach Email Dialog_ from showing when there is already the _Set Username Dialog_

## How do I test this PR?

1. Create a new account with an external auth provider (google/facebook/etc)
2. Follow the first page of the attach email flow, click "add email address"
3. Notice you see the confirmation page correctly instead of only the username dialog


kgardnr (dismissed):

plugins/talk-plugin-local-auth/client/components/AddEmailAddressDialog.js
```diff
@@ -113,6 +123,8 @@ class AddEmailAddressDialog extends React.Component {
         email: emailAddress,
         password: confirmPassword,
       });
+
+      // TODO: translate
```
alinhle: Let's do it :-)

plugins/talk-plugin-local-auth/translations.yml
```diff
@@ -1,4 +1,12 @@
 en:
+  email:
+    email_change_original:
+      subject: Email change
+      body: Your email address has been changed from {0} to {1}. If you did not initiate this change, please contact {2}. # TODO: update translation
```
alinhle: What needs to be updated?

plugins/talk-plugin-local-auth/client/constants.js
```diff
@@ -0,0 +1,4 @@
+const prefix = 'TALK_LOCAL_AUTH';
+
+export const STARTED_ATTACH = `${prefix}_STARTED_ATTACH`;
```
alinhle: Let's call it `START_ATTACH` instead

alinhle (dismissed):
See small remarks

plugins/talk-plugin-local-auth/translations.yml
wyattjoh: Previously, all the translations lived in two files, which didn't work for the client code as that code didn't live in the right place, this fixes that :)

alinhle: I mean it says `#TODO: update translation` at the end, my question is what is it referring to. Oh I just saw you were just copying it over 🤔

wyattjoh: I think that content wasn't vetted by product/design so I wanted to flag it to get reviewed :)

alinhle (approved):

PR #1607, opened 2018-05-10
External Signin for Admin

wyattjoh:
## What does this PR do?

- Adds a slot to support different administrative login strategies.
- Fixes bug with `postMessage` service.

## How do I test this PR?

1. Enable google-auth
2. Visit admin and notice the "Sign in with Google" button


kgardnr:
Whoa @wyattjoh this is amazing!!

alinhle:
I changed the code to not close the popup when there is an error, it made debugging really hard. Now you can see the error and get more details in the console:

image

alinhle (approved):

kgardnr:
Nice addition @cvle!!

PR #1610, opened 2018-05-11
bugfix: throw error when email already exist on changing

leeeandroo:
## What does this PR do?
Fix `talk-plugin-local-auth` mutator to catch error and show in the frontend. 

## How do I test this PR?

- Login
- Go to My Profile
- Click on Edit Profile
- Change your e-mail to a already used one and see the error
- Change your e-mail to a new one and see the success message and receive email

### DIFF
--- a/plugins/talk-plugin-local-auth/server/mutators.js
+++ b/plugins/talk-plugin-local-auth/server/mutators.js
@@ -35,24 +35,22 @@ async function updateUserEmailAddress(ctx, email, confirmPassword) {
   email = email.toLowerCase().trim();
 
   // Update the Users email address.
-  try {
-    await User.update(
-      {
-        id: user.id,
-        profiles: { $elemMatch: { provider: 'local' } },
-      },
-      {
-        $set: { 'profiles.$.id': email },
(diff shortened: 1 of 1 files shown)


plugins/talk-plugin-local-auth/server/mutators.js
```diff
@@ -35,24 +35,22 @@ async function updateUserEmailAddress(ctx, email, confirmPassword) {
...
+  await User.update(
+    {
+      id: user.id,
+      profiles: { $elemMatch: { provider: 'local' } },
+    },
+    {
+      $set: { 'profiles.$.id': email },
+      $unset: { 'profiles.$.metadata.confirmed_at': 1 },
+    }
+  ).catch(err => {
```
wyattjoh, line 47: As this is in an async function, it should (and from my tests, currently does) already throw the `ErrEmailTaken` error.

wyattjoh (changes requested):

plugins/talk-plugin-local-auth/server/mutators.js
leeeandroo: You are right. I've tested it on 4.4.0, checked on master and It works fine. Thanks

PR #1614, opened 2018-05-14
Add new slot 'commentTombstone'

hellp:
## What does this PR do?

Adds a new slot 'commentTombstone' with which plugins can override the default tombstone for deleted comments.

## How do I test this PR?

- create a plugin that uses the new slot and see your custom component being used instead of Talk's standard `CommentTombstone`.
- if no plugin uses this slot, nothing changes

## Rationale

We plan to use this for a plugin that let's us have deleted comments with annotations, i.e. "Was deleted by the user" or "Was deleted for legal reasons". We'd store this `remove_reason` in the deleted comment's metadata and show it via our custom component that overrides `CommentTombstone` using this slot.


okbel:
Makes perfect sense to me. @cvle, @wyattjoh ?

alinhle (approved):
LGTM

PR #1617, opened 2018-05-14
removed the exclusion of deleted for the comment count on admin panel

angelo-hub:
## What does this PR do?
This PR fixes a bug where comment count on a multi-field query takes too long to return performantly.

## How do I test this PR?

- deploy the application
- open the admin panel
- with a full database set it should load without issue


wyattjoh:
Thanks for the contribution @AngefloMusic! We're actually planning on removing counts entirely from this next release so we can afford more time to optimize the counting: https://www.pivotaltracker.com/story/show/157525698

PR #1622, opened 2018-05-15
Set Akismet is_test to false

icco:
## What does this PR do?

This sets akismet api calls to be normal and not test calls. From https://akismet.com/development/api/#comment-check, is_test should be false if these are real queries and not test queries.

## How do I test this PR?

 - Configure the akismet plugin with a pro account
 - Make comments on the talk instance
 - Verify comments show up in dashboard


wyattjoh (approved):

icco:
Thanks for the quick turnaround!

wyattjoh:
Thanks for the quick find!

PR #1625, opened 2018-05-16
Displaying multiple external provider IDs

okbel:
## Displaying multiple external provider IDs
- https://www.pivotaltracker.com/story/show/157605877
screen shot 2018-05-16 at 12 05 45 pm

Let me know your thoughts, @wyattjoh! Thanks!


kgardnr:
Awesome @okbel!!

client/coral-admin/src/routes/Community/components/People.js
```diff
@@ -130,7 +130,14 @@ class People extends React.Component {
                             {user.username}
                           </button>
                           <span className={styles.email}>
-                            {user.profiles.map(({ id }) => id)}
+                            {user.email
+                              ? user.email
+                              : user.profiles.map(({ id }, i) => {
```
alinhle: Is this the same as `user.profiles.map(p => p.id).join(', ')` which would be simpler?

alinhle (changes requested):

client/coral-admin/src/routes/Community/components/People.js
wyattjoh: Typically, the `id` from the provider are quite long, hence why it probably shouldn't be displayed here, but is displayed on the `UserDetail` component.

alinhle (approved):

PR #1630, opened 2018-05-17
Karma Score

okbel:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/157481506

todo: it needs to retrieve the real karma value from the BE


okbel:
@wyattjoh Done! Cache updating and styling. There is only one little thing that seems inconsistent.
screen shot 2018-05-18 at 2 07 52 pm


Here the score is `-1` and it's returning null. It should be `false`.

client/coral-admin/src/components/KarmaTooltip.css
```diff
@@ -0,0 +1,80 @@
...
+  display: inline-block;
+  margin: 2px 4px 0;
+}
+
+.icon {
+  font-size: 16px;
+  color: #0D5B8F;
+  -ms-user-select:none;
+  -moz-user-select: none;
+  -webkit-user-select: none;
```
alinhle: These vendor prefixes should be already handled by our css preprocessor.

client/coral-admin/src/components/KarmaTooltip.css
```diff
@@ -0,0 +1,80 @@
...
+  cursor: pointer;
+}
+
+.menu {
+  background-color: white;
+  border: solid 1px #999;
+  border-radius: 3px;
+  padding: 10px;
+  position: absolute;
+  -webkit-box-shadow: 0 2px 2px 0 rgba(0,0,0,0.14), 0 1px 5px 0 rgba(0,0,0,0.12), 0 3px 1px -2px rgba(0,0,0,0.2);
```
alinhle: See above

client/coral-admin/src/graphql/index.js
```diff
@@ -24,6 +24,16 @@ const userRoleFragment = gql`
   }
 `;
 
+const toBoolean = value => {
```
alinhle: Misleading name, as the returned value is not only a boolean. Maybe better `isReliable` and a comment what null value means?

alinhle (changes requested):
See above. Also @wyattjoh please check trust logic for me.

client/coral-admin/src/components/KarmaTooltip.css
okbel: 👍

client/coral-admin/src/components/KarmaTooltip.css
okbel: 👍

client/coral-admin/src/graphql/index.js
okbel: Well tecnically is a Tristate (boolean), `toTristate` seems reasonable?

okbel: or ternary, or trivalent https://en.wikipedia.org/wiki/Three-valued_logic 🤔

client/coral-admin/src/components/KarmaTooltip.css
kgardnr: I'm going to create a separate ticket to address the css vendor prefixes, because we are using them in other places. So we can standardize that as part of the UI refactor.

client/coral-admin/src/components/KarmaTooltip.js
```diff
@@ -0,0 +1,81 @@
...
+            className={cn(styles.icon, 'talk-admin-karma-tooltip-icon')}
+          >
+            <Icon name="info" />
+          </span>
+
+          {menuVisible && (
+            <div className={cn(styles.menu, 'talk-admin-karma-tooltip-menu')}>
+              <strong>{t('user_detail.user_karma_score')}</strong>
+              <ul>
+                {/* <li>
```
alinhle: Reason this is commented out?

graph/typeDefs.graphql
```diff
@@ -865,6 +896,9 @@ type Settings {
 
   # domains will return a given list of domains.
   domains: Domains
+
+  # karma contains the currently set thresholds for triggering Trust beheviour.
+  karma: KarmaThresholds
```
alinhle: Maybe it'd be more descriptive to use `settings.karma.thresholds.flag.reliable`.
Right now accessing this we use `settings.karma.flag.reliable`, it wasn't immediately clear what this meant until I saw the schema.

client/coral-admin/src/components/KarmaTooltip.js
```diff
@@ -0,0 +1,81 @@
...
+import ClickOutside from 'coral-framework/components/ClickOutside';
+import t from 'coral-framework/services/i18n';
+
+const initialState = { menuVisible: false };
+
+class KarmaTooltip extends React.Component {
+  static propTypes = {
+    settings: PropTypes.shape({
+      reliable: PropTypes.number.isRequired,
+      unreliable: PropTypes.number.isRequired,
```
alinhle: Also here it wasn't clear for me what those props represents, maybe `thresholds: { reliable, unnreliable }` is more descriptive or even better `reliableThreshold`, `unreliableThreshold` to avoid using nested props.

alinhle (changes requested):

client/coral-admin/src/components/KarmaTooltip.js
wyattjoh: For now, we wanted to hide the labels for reliable + neutral because we don't do anything with it. When we do, it may make sense to display them. Commenting them out here leaves it there so we can restore it easily going forward.

client/coral-admin/src/components/KarmaTooltip.js
wyattjoh: I'll change it to `thresholds` to better reflect the meaning.

graph/typeDefs.graphql
wyattjoh: We don't have any information to display relevant to the karma besides the thresholds, and as this is already the `settings` for the `karma`, I would assume it was implied that it would be relating to the settings for the the karma on flags.

alinhle: I'd like to make it easier to understand the code, without needing to lookup the graphql schema. Sth like `settings.karma.flagThresholds.reliable` would just be crystal clear what we are accessing here.

client/coral-admin/src/components/KarmaTooltip.js
alinhle: Ok, I was just thinking if we should add comments there whenever we comment parts of code out so it just don't look like it's forgotten :-P, e.g. "wyatt: kept here for future use"

graph/typeDefs.graphql
wyattjoh: The issue with that solution is that we already have a graph endpoint: `user.reliable.flagger` and `user.reliable.flaggerKarma` which represents properties of their reliability, and their Karma. We could perhaps on the client, alias the field?

alinhle (approved):

kgardnr (approved):

PR #1632, opened 2018-05-18
Only display paragraph if delete requested

mkonikov:
## What does this PR do?
Removes an extra empty paragraph when user has not requested for a deletion. 

## How do I test this PR?
Load the settings panel in your profile and see the extra paragraph is no longer there.
Request account deletion and see additional paragraph with info then display.


wyattjoh (approved):
Looks good to me!

plugins/talk-plugin-profile-data/client/components/DeleteMyAccount.js
```diff
@@ -63,14 +63,13 @@ class DeleteMyAccount extends React.Component {
...
         </p>
-        <p className="talk-plugin-auth--delete-my-account-description">
-          {scheduledDeletionDate &&
-            t(
-              'delete_request.already_submitted_request_description',
-              moment(scheduledDeletionDate).format('MMM Do YYYY, h:mm:ss a')
-            )}
-        </p>
         {scheduledDeletionDate ? (
+          <p className="talk-plugin-auth--delete-my-account-description">
```
wyattjoh: I wasn't able to compile the code, as these two elements need to be wrapped in an enclosing tag (like a `div`) or break each block into their own conditional checks instead.

wyattjoh (changes requested):

plugins/talk-plugin-profile-data/client/components/DeleteMyAccount.js
mkonikov: Updated.

wyattjoh (approved):
Perfect, look great!

PR #1634, opened 2018-05-18
Show overflow text on download comments button when disabled

mkonikov:
## What does this PR do?
When comment history has been recently requested and the button is therefore deleted, button text is cutoff on smaller devices. This updates the CSS to display the overflow text.  

## How do I test this PR?
- Request comment history
- Visit Settings panel on smaller device and see Download button now displays all text


alinhle (approved):
The `overflow` property seems to be unnecessary so I removed it and additionally added a TODO comment to better deal with mobile view. Thanks for catching this @mkonikov!

Issue #1626, opened 2018-05-16
Bug: Akismet plugin failing


icco:
#### Do you want to request a **feature** or report a **bug**?

Bug!

#### Intended outcome:

We were trying to implement the talk akismet plugin. We should see reports of comments sent to Akismet.

#### Actual outcome:

So in prod, running 4.4.0, we weren't seeing any comments in the akismet dashboard. I reached out to the Akismet team, and they said they weren't seeing any requests land for our domain or API key. I filed (and you released :D ) #1622 after further discussion with the Akismet team said that:

> Yes, indeed. Using the `is_test` parameter set to `true`, or `1`, would do exactly what you described. The intent of that parameter is for things such as unit tests, so that you can get consistent results for such tests, by essentially turning off all of the machine learning aspects of our API, logging, and so on.

So, I went to dev and upgraded our Talk to 4.4.1, which also has DEBUG=* set in it's environment. That's when I saw the log line:

```
2018-05-15T19:58:17.994Z talk:plugin:akismet no ip on request
```

This seemed weird to me, so I forked your plugin and discovered that the `req` variable from https://github.com/coralproject/talk/blob/v4.4.1/plugins/talk-plugin-akismet/server/hooks.js#L31 seems to never have the following properties:

 - `Get()`
 - `ip`
 - `headers`

I tried lots of different ways to get the data we needed out of `req` but was unsuccessful. One example diff I tried was 

```diff
diff --git a/plugins/flm-akismet/server/hooks.js b/plugins/flm-akismet/server/hooks.js
index 025789a..5c0bd21 100644
--- a/plugins/flm-akismet/server/hooks.js
+++ b/plugins/flm-akismet/server/hooks.js
@@ -37,12 +37,11 @@ module.exports = {
 
         let spam = false;
         try {
-          debug(req)
-          const user_ip = get(req, 'ip', '');
+          const user_ip = get(req, 'ip', '') || req.headers['x-forwarded-for'];
 
           // Get some headers from the request.
-          const user_agent = req.get('User-Agent') || '';
-          const referrer = req.get('Referrer') || '';
+          const user_agent = req.headers['user-agent'];
+          const referrer = req.headers['referer'];
 
           // Get the Asset that the comment is being made against.
           const asset = await loaders.Assets.getByID.load(input.asset_id);
@@ -52,7 +51,7 @@ module.exports = {
           }
 
           // Send off the comment to Akismet to check to see what they say.
-          spam = await client.checkSpam({
[...]

PR #1638, opened 2018-05-22
Akismet Bug - Req not properly passed

okbel:
## Akismet Bug - Req not properly passed

Before:
screen shot 2018-05-22 at 12 42 12 pm

Now:
screen shot 2018-05-22 at 12 39 31 pm

I left the code explicit because we might need to refactor the `Context` object. In case we encounter more issues related to this bug. I leave this to wyatt to further investigation.

It's working. 👍


okbel:
Explanation: 

Keys from `ctx.parent.parent`
```[ '_readableState',
  'readable',
  'domain',
  '_events',
  '_eventsCount',
  '_maxListeners',
  'socket',
  'connection',
  'httpVersionMajor',
  'httpVersionMinor',
  'httpVersion',
  'complete',
  'headers',
  'rawHeaders',
  'trailers',
  'rawTrailers',
  'upgrade',
  'url',
  'method',
  'statusCode',
  'statusMessage',
  'client',
  '_consuming',
  '_dumped',
  'next',
  'baseUrl',
  'originalUrl',
  '_parsedUrl',
  'params',
  'query',
  'res',
  'id',
  'secret',
  'cookies',
  'signedCookies',
  'body',
  '_body',
  'length',
  'read',
  '_passport',
  'jwt',
  'user',
  'context' ] 
```
Keys from `ctx.parent`
```[ 'id',
  'log',
  'user',
  'connectors',
  'loaders',
  'mutators',
  'plugins',
  'pubsub',
  'parent' ]
```

kgardnr:
Awesome! This works!!

kgardnr:
Fixes https://github.com/coralproject/talk/issues/1626

icco:
Sweet! Fixes #1626 for me. Thanks!

icco:
Oh actually, one other thing, I didn't have a chance to test this, but I'm 95% sure https://github.com/coralproject/talk/blob/71b1259db695dc7d8a286d82fc4cea1ac558aecf/plugins/talk-plugin-akismet/server/hooks.js#L56 should be `const referrer = req.get('Referer');`. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer for spelling weirdness.

wyattjoh:
I'll check that @icco, good catch!

wyattjoh:
It seems that Express knows about this, and specifically solves it 😄 

https://github.com/expressjs/express/blob/b97faff6e2aa4d34d79485fe4331cb0eec13ad57/lib/request.js#L43-L44

wyattjoh (approved):

PR #1640, opened 2018-05-22
Remove wordbreak

mkonikov:
## What does this PR do?
Removes `word-break` CSS rule which makes text to appear as a block and hard to read/moderate. See screenshot below

Coral Talk admin panel

I welcome your thoughts as I presume there was a reason this rule was added initially.

TIA!


wyattjoh:
The main issues here is that without wordbreak, if a commenter:

```
leavesasuperlongstringthenthereisnowayitwilleverterminatemakingitimpossibletoactuallywork
```

We'll investigate the issue raised, and hopefully come up with a solution that works 😄

alinhle:
@mkonikov Can you try if  `word-break: break-word` fixes the issue? It should do what we initially intended to, see https://developer.mozilla.org/en-US/docs/Web/CSS/word-break

mkonikov:
Yes @cvle that works just fine. I took a quick look at caniuse.com and it mentioned something about it being unofficial but I see now I misunderstood. Thanks!

mkonikov:
I've updated the PR changing `break-all` to `break-word`.

alinhle:
@mkonikov No actually you were right! I checked and for example firefox do not support `break-word`. However it seems `overflow-wrap: break-word;` provides a similar desired behavior if not even better and is supported across the bord. Do you mind adding that change? https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap https://caniuse.com/#feat=wordwrap

mkonikov:
Updated - @cvle

alinhle (approved):
Awesome thank you!!

mkonikov:
For the record it doesn't actually seem to wrap @cvle :/

PR #1641, opened 2018-05-22
Locale Negotiation

wyattjoh:
## What does this PR do?

- Replaces our home-grown language negotiation with [fluent-langneg](https://www.npmjs.com/package/fluent-langneg).
- Simplifies substitution replacement using the correct `.replace` prototype instead of creating `n` regex expressions.

Special thanks to @larrylizhao for discovering the bug!

## How do I test this PR?

Change your browser to several different (also unsupported) languages to test to see if the browser displays the correct language pack.


alinhle (approved):
LGTM

PR #1651, opened 2018-05-24
New env variable can be set in order to force a language

jerodb:
## What does this PR do?
In order to avoid auto-detecting the language from the navigator's language, a new variable (TALK_FORCE_LANG) can be set in the .env file with the value of the language you want Talk to be loaded in all cases.
A similar pull request #1552 was made by @larrylizhao , but he proposed to change the behavior of the TALK_DEFAULT_LANG variable. In this new request the TALK_DEFAULT_LANG variable works as intended.
If TALK_FORCE_LANG is not set then the navigator's language will be used.

## How do I test this PR?
- add TALK_FORCE_LANG equal to the language you want to force ( ex. TALK_FORCE_LANG=es ) to the .env file
- yarn build
- run talk and check if the language is the one forced despite the value of the navigator's language.

### DIFF
--- a/client/coral-framework/services/i18n.js
+++ b/client/coral-framework/services/i18n.js
@@ -59,14 +59,14 @@ export const supportedLocales = Object.keys(translations);
 let LOCALE;
 let TIMEAGO_INSTANCE;
 
-const preferedLanguage = process.env.TALK_FORCE_LANG ? process.env.TALK_FORCE_LANG : navigator.languages;
+const preferredLanguage = process.env.TALK_FORCE_LANG ? [process.env.TALK_FORCE_LANG] : navigator.languages;
 
 // detectLanguage will try to get the locale from storage if available,
 // otherwise will try to get it from the navigator, otherwise, it will fallback
 // to the default language.
 const detectLanguage = () =>
   first(
-    negotiateLanguages(preferedLanguage, supportedLocales, {
+    negotiateLanguages(preferredLanguage, supportedLocales, {
       defaultLocale,
       strategy: 'lookup',
     })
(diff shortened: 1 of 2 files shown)


wyattjoh:
Thanks for this PR @jerodb! I saw your initial one, and included support for the server side as well, see #1650. I'm going to close this in favor of the other PR that was already merged.

jerodb:
Hi @wyattjoh . Sorry but I don't see how #1650 replaces this PR. What I'm proposing is to force a language despite the browser's language.
What I understand from #1650 is that can override the supported languages and in some specific case replace the TALK_DEFAULT_LANG. But the application language is not forced/changed, is still the browser's.
I might be confuse dough. 
Thanks.

wyattjoh:
Good question!

When you set `TALK_WHITELISTED_LANGUAGES`, those are the only languages that will be reported as supported by Talk. If you want to force the language to a specific language, you only need to specify one language:

```
TALK_WHITELISTED_LANGUAGES=es
```

And rebuild. That will force the browser to attempt to negotiate the language, but will only resolve to those listed in the `TALK_WHITELISTED_LANGUAGES` parameter. You can see:

https://github.com/coralproject/talk/blob/1c0bfdb89eca308000e82932c79ea67318a476c9/client/coral-framework/services/i18n.js#L75-L88

That if the `supportedLanguages` is specified (from `TALK_WHITELISTED_LANGUAGES`), that it can only resolve to a language from the whitelist.

Let me know if that makes sense!

jerodb:
yes, it makes sense. Thank you very much Wyatt!

jerodb:
This works perfect Wyatt. Thanks again.

PR #1653, opened 2018-05-25
Add missing translations (en/de) for plugins ignore-user and local-auth.

hellp:
## What does this PR do?

Adds some missing translations and translation markers.


hellp:
Added some more mini changes. Hope this is still okay for one PR.

wyattjoh (approved):

PR #1657, opened 2018-05-25
move placeholder right behind contentEditable

erikreyna:
## What does this PR do?

fixes a small bug that prevents users from clicking into the comment box, when clicking above the placeholder text

## How do I test this PR?

- click above the placeholder, if the input gets activated, success!

Attaching example Greg Barber gave me showing the problem. http://recordit.co/tiSA5x7rOk


plugins/talk-plugin-rich-text/client/components/rte/RTE.css
```diff
@@ -15,6 +15,7 @@
   position: absolute;
   margin: 12px 0 0 12px;
   color: #bbb;
+  z-index: -1;
```
alinhle: Instead of moving the  `placeholder` behind, I think the better solution would be to disable _pointer events_ and _user selection_ using

```css
user-select: none;
pointer-events: none;
```

Would this solve your issue?

plugins/talk-plugin-rich-text/client/components/rte/RTE.css
```diff
@@ -1,5 +1,5 @@
 .contentEditable {
-  background: #fff;
+  background: transparent;
```
alinhle: The above suggestion would render this change unnecessary. Also this change would mean that the background color of the parent container will shine through the RTE which might not be wanted (e.g. if it's not white)

alinhle (dismissed):

plugins/talk-plugin-rich-text/client/components/rte/RTE.css
erikreyna: Hey @cvle! Thanks for reviewing. I can confirm that this does actually do the trick. It's a much better solution. Let me make the changes.

plugins/talk-plugin-rich-text/client/components/rte/RTE.css
erikreyna: Agreed it's unnecessary since the above change does actually solve the issue.

erikreyna (commented):
made changes requested by @cvle

kgardnr (approved):

wyattjoh (approved):

PR #1660, opened 2018-05-28
add missing dependency

hjanuschka:
https://github.com/coralproject/talk/commit/95072be97e27945e071e62c9658b1d81520ef204 introduced `nunjucks`  which is not beeing installed via `yarn` as it is missing in the packages.json


wyattjoh:
Nunjucks was actually included transitively via the `consolidate` package, but I don't see a reason to not include it explicitly given we're importing it directly.

wyattjoh (approved):

PR #1664, opened 2018-05-30
remove unused method

leeeandroo:
## What does this PR do?
Fix ban user action, sending message value to method and send email to the user.

## How do I test this PR?

- access admin
- go to Comunnity
- Ban an user
- Check the email with the message


wyattjoh (approved):
Fixed one small thing I found, good to go though!

PR #1673, opened 2018-06-04
Use Markdown for the "asset is closed" message.

hellp:
## What does this PR do?

Makes the last remaining TextArea in the stream settings also a Markdown, and adapts the rendering on the stream. Now all long text inputs on that settings page are Markdowns (stream intro, asset closed , comments deactivated).


kgardnr:
This looks great @hellp! I'll let @okbel or @cvle give a quick review! Thanks!

okbel:
I like this change. What do you think @cvle?

kgardnr (approved):

PR #1674, opened 2018-06-04
Add scraper proxy support

cristiandean:
For security reasons several companies only allow exteno access from the backend via proxy. This PR adds this support.


wyattjoh (approved):
Looks good to me!

PR #1678, opened 2018-06-04
User Detail Drawer Fixes

wyattjoh:
## What does this PR do?

- Hides the local user profile "id", which is just the email address (which is already show as "email" on the drawer)
- Displays the user's ID

## How do I test this PR?

1. Visit the `Community -> People` tab
2. Click on a Social (Google/Facebook) User
3. Checkout the User drawer, and notice the lack of `Local ID`, and instead, just the `ID` and the `Facebook ID` and possibly the `Email`


kgardnr (approved):
LGTM! Ready for @okbel's review! 🎉

PR #1686, opened 2018-06-11
Add translations in some missing pages and fix others

cristiandean:
Some pages like the admin login was missing translations, so I added it. I also fixed some translations in pt_BR.


kgardnr:
Awesome, thanks @cristiandean!

kgardnr (approved):

PR #1692, opened 2018-06-14
Use neutral icon for aggregated reactions count

hellp:
## What does this PR do?

Replaces the _thumb_up_ icon on Profile > My comments with the _thumbs_up_down_ icon.

**Rationale:** the aggregated reactions count may not only reflect positive reactions. For instance in our project we use the reactions framework for a "Dislike" button, so the "thumb up" is irritating when most of the reactions are actually dislikes.

As an alternative to the _thumbs_up_down_ one could consider the _whathot_ icon from the Material icon set. If you think of many reactions as a kind of "hotness".

screenshot-20180614-012a20


kgardnr:
Hey @hellp, we really think you should reconsider using negative reactions. This has been shown to increase toxicity in communities.

In regards to the icon, I'll speak to the team and see what their thoughts are.

hellp:
@kgardnr True, this is a double-edged sword. We discussed this and settled with this:

- add Dislike reaction
- but **do not** allow to "sort by most disliked", i.e. we don't want users to see these toxic comments ranked and in bulk
- on the other hand a click on „Dislike!“ may discourage users from writing a toxic (especially very short ones that are not useful) comment at all, if this vote was enough to express their frustration.

kgardnr (approved):

kgardnr:
Hey @hellp -- I merged this too early -- could we actually with the fire icon? Could you create a new PR for that?? Thank you so much!!

PR #1697, opened 2018-06-18
Fixed blog links for new site

losowsky:


docs/source/05-03-when-youve-installed-talk.md
```diff
@@ -44,13 +44,13 @@ Your plans might include:
 
 ![[IMAGE] A screenshot of the pre-moderate option in the Configure tab](http://blog.coralproject.net/wp-content/uploads/2018/03/premod.png)
```
wyattjoh: Missed an image here.

wyattjoh (changes requested):

docs/source/05-03-when-youve-installed-talk.md
losowsky: I don't think so - the blog images are redirecting fine, for some reason it's only the articles that don't. Does the image not show up for you?

wyattjoh: It does redirect fine, I just expect that we could swap out the old url for the new one, less redirects means faster loads :)

wyattjoh (approved):

PR #1701, opened 2018-06-20
Update COC

kgardnr:


package.json
```diff
@@ -58,6 +58,7 @@
   "dependencies": {
     "@coralproject/gql-merge": "^0.1.0",
     "@coralproject/graphql-anywhere-optimized": "^0.1.0",
+    "@coralproject/talk-plugin-auth-checkbox": "^1.0.0",
```
wyattjoh: Why was the `auth-checkbox` added to the `package.json`? It shouldn't need to be there, as users wanting to add it can add it to their `plugins.json` themselves.

wyattjoh (changes requested):

package.json
kgardnr: Oh gotcha ok!

kgardnr: Fixed, please take a look!

wyattjoh (approved):

PR #1708, opened 2018-06-26
Set versions to 4.5.0 instead of latest

kgardnr:
## What does this PR do?

Let's have our docs say 4.5.0 instead of latest so people who copy code don't accidentally pull a non-stable version of Talk.


wyattjoh (approved):
Perfect!

PR #1709, opened 2018-06-26
Merge client into next

alinhle:
## What does this PR do?
- Merge client into next


patches/relay-compiler/bin/relay-compiler
```diff
@@ -0,0 +1,10330 @@
+#!/usr/bin/env node
```
wyattjoh: Is this a compiled file?

src/core/server/app/middleware/logging.ts
```diff
@@ -37,7 +37,7 @@ export const access: RequestHandler = (req, res, next) => {
   next();
 };
 
-export const error: ErrorRequestHandler = (err, req, res, next) => {
+export const error: ErrorRequestHandler = (err, _req, _res, next) => {
```
wyattjoh: What's the benefit of this rule for unused params if it just means that to fix bugs with airity, we have to prefix all our params with `_`?

tsconfig.json
```diff
@@ -2,24 +2,40 @@
   "compilerOptions": {
     "target": "es5",
     "module": "commonjs",
-    "esModuleInterop": true,
-    "noImplicitAny": true,
+    "jsx": "react",
```
wyattjoh: Shouldn't this live in the `src/core/client/tsconfig.json` file?

patches/relay-compiler/bin/relay-compiler
alinhle: Yes

tsconfig.json
alinhle: This was only added for `jest`. The only alternative is to run different instances of `jest` for server and for the client. We can have a discussion around this.

alinhle: Ah I probably can make it work, without this.

src/core/server/app/middleware/logging.ts
alinhle: I think we can turn it off, also not sure about its benefits.

tsconfig.json
alinhle: Looks like I don't need this

package.json
```diff
@@ -4,21 +4,39 @@
...
+    "lint:client": "tslint --project ./src/core/client/tsconfig.json",
+    "docz:watch": "docz dev"
   },
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
+    "@types/relay-runtime": "github:coralproject/patched#types/relay-runtime",
     "apollo-server-express": "^1.3.6",
     "bunyan": "^1.8.12",
+    "chalk": "^2.4.1",
```
wyattjoh: Shouldn't this be in `devDependencies`?

package.json
```diff
@@ -4,21 +4,39 @@
...
+    "lint-fix": "tslint --fix --project ./tsconfig.json && tslint --fix --project ./src/core/client/tsconfig.json",
+    "lint": "npm-run-all --parallel lint:*",
+    "lint:server": "tslint --project ./tsconfig.json",
+    "lint:client": "tslint --project ./src/core/client/tsconfig.json",
+    "docz:watch": "docz dev"
   },
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
+    "@types/relay-runtime": "github:coralproject/patched#types/relay-runtime",
```
wyattjoh: Shouldn't this be in `devDependencies`?

alinhle: Ah yeah that slipped through!

wyattjoh (approved):

PR #1713, opened 2018-06-29
lint js files, cleanup scripts

alinhle:
## What does this PR do?
- Lint and fix js files
- Setup VSCode to lint and fix js files using tslint
- Ignore `node_modules` in `tslint`
- Don't open browser in `start.js`
- Remove checks for required files in `start.js` and `build.js`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ yarn.lock
 coverage
 
 .idea/
+.vs
 .docz
 *.swp
 *.DS_STORE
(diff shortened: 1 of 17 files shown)


.gitignore
```diff
@@ -7,6 +7,7 @@ yarn.lock
 coverage
 
 .idea/
+.vs
```
wyattjoh, line 10: What is this exclude for?

alinhle: My vscode creates this file that should not be commited. See https://stackoverflow.com/questions/31878901/should-i-add-the-visual-studio-2015-vs-folder-to-source-control

wyattjoh (approved):

PR #1727, opened 2018-07-03
[next] Use JSDocs comments

alinhle:
## What does this PR do?
- Use JSDocs style comments, so typescript will pick them up.


okbel:
Awesome. Is there a way that we can enforce this in our linters or somewhere?

okbel (approved):

PR #1730, opened 2018-07-04
[next] Timestamp - Reader can see relative time of when that comment was posted

okbel:
### RelativeTime - Reader can see relative time of when that comment was posted
- Adds RelativeTime component
- Adds Timestamp component
- Adds "createdAt" field to the comment schema/resolvers
- Translations
- Show meaningful css names in dev
- Turn on antialiasing for font rendering
- Tests and test support for MatchMedia uses


.vscode/settings.json
```diff
@@ -16,5 +16,6 @@
   },
   "tslint.autoFixOnSave": true,
   "tslint.jsEnable": true,
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "git.ignoreLimitWarning": true
```
alinhle: Can you turn it to false and see if you still have the issue?
Just making sure we really need this, it seems to make sense to keep the default.

src/core/client/framework/lib/bootstrap/TalkContext.tsx
```diff
@@ -18,7 +23,7 @@ const { Provider, Consumer } = React.createContext<TalkContext>({} as any);
  */
 export const TalkContextConsumer = Consumer;
 
-/**
+/**s
```
alinhle: Here is a lonely `s`

src/core/client/stream/components/Comment.spec.tsx
```diff
@@ -8,6 +8,7 @@ it("renders username and body", () => {
       username: "Marvin",
     },
     body: "Woof",
+    createdAt: new Date().toISOString(),
```
alinhle: This won't work well with jest snapshot, adding a fixed time would be better.

src/core/client/stream/components/Comment.spec.tsx
```diff
@@ -20,6 +21,7 @@ it("renders with gutterBottom", () => {
...
```
alinhle: See above

src/core/client/framework/lib/bootstrap/createContext.tsx
```diff
@@ -16,6 +19,19 @@ interface CreateContextArguments {
   init?: ((context: TalkContext) => void | Promise<void>);
 }
```
alinhle, line 21: We should add a comment here.

src/core/client/ui/components/Timestamp/Timestamp.spec.tsx
```diff
@@ -0,0 +1,19 @@
...
+  expect(tree).toMatchSnapshot();
+});
+
+it("renders future date", () => {
+  const props = {
+    date: new Date("December 17, 2108 03:24:00").toISOString(),
+  };
+  const tree = create(<Timestamp {...props} />).toJSON();
+  expect(tree).toMatchSnapshot();
+});
```
alinhle: As suggested before, a test to check if consuming from the context is working correctly, would be helpful.

src/core/client/ui/components/Timestamp/Timestamp.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { withStyles } from "talk-ui/hocs";
+import { PropTypesOf } from "talk-ui/types";
+
+import * as styles from "./Timestamp.css";
+
+interface InnerProps {
+  date: string;
+  live?: boolean;
+  classes: typeof styles;
+  formatter?: Formatter;
```
alinhle: This is never used, shouldn't it be passed to `TimeAgo`?

src/locales/de/framework.ftl
```diff
@@ -3,3 +3,37 @@
 ### among different targets.
 
 framework-validation-required = Dies ist ein Pflichtpfeld.
+
+framework-timeago =
```
alinhle: These German Translations are in Spanish?

src/locales/en-US/framework.ftl
```diff
@@ -3,3 +3,38 @@
 ### among different targets.
 
 framework-validation-required = This field is required.
+
+framework-timeago =
+  { $value ->
```
alinhle: The `Spanish` translations is using `en 1 hora` but the English translation `1 hour from now`, I think it's better when we stay consistent and also translate to `in 1 hour`.

src/core/client/ui/components/Timestamp/Timestamp.tsx
```diff
@@ -0,0 +1,37 @@
...
+import TimeAgo, { Formatter } from "react-timeago";
+
+import { UIContext } from "talk-ui/components";
+import { withStyles } from "talk-ui/hocs";
+import { PropTypesOf } from "talk-ui/types";
+
+import * as styles from "./Timestamp.css";
+
+interface InnerProps {
+  date: string;
```
alinhle: We should add support for `className` here.

src/core/client/ui/components/Timestamp/Timestamp.css
```diff
@@ -0,0 +1,4 @@
+.root {
+  composes: button from "talk-ui/shared/typography.css";
```
alinhle: We need to add typography styling from Sam or if already available from the Stylesheet

src/core/client/stream/components/Comment.tsx
```diff
@@ -24,6 +25,7 @@ const Comment: StatelessComponent<CommentProps> = props => {
       <Typography className={styles.author} gutterBottom>
         {props.author && props.author.username}
       </Typography>
+      <Timestamp date={props.createdAt} />
```
alinhle: We need to figure out a way to have `responsiveness` here.

alinhle (dismissed):

alinhle:
Also the large introspection files are still in the commit history, can you take another look?

src/core/client/stream/components/Comment.tsx
alinhle: I'm having a reusable component in mind, that we can always rely on.

src/locales/de/framework.ftl
okbel: These are for you haha. I'm really rusty with german.

src/core/client/stream/components/Comment.tsx
alinhle: I'll work with you on this, I'm gonna test if one of my components work that I wrote in one of my earlier projects.

src/locales/en-US/framework.ftl
okbel: 💅 : ok

alinhle: I tried that the last time actually and I first didn't know how to do it, but now I do, we can talk about it on Monday too :-)

src/core/client/ui/components/Timestamp/Timestamp.tsx
okbel: 👍

src/locales/en-US/framework.ftl
alinhle: Actually in our current Talk we also changed from `1 hour from now` to `in 1 hour`, I don't know any reason who would use `1 hour from now`.

src/locales/de/framework.ftl
alinhle: Haha sure

src/core/client/stream/components/Comment.tsx
okbel: 👍

src/core/client/ui/components/Timestamp/Timestamp.css
okbel: 👍

src/core/client/ui/components/Timestamp/Timestamp.tsx
okbel: This is old. Removed!

package.json
```diff
@@ -45,6 +45,7 @@
     "mongodb": "^3.0.10",
     "passport": "^0.4.0",
     "performance-now": "^2.1.0",
+    "react-timeago": "^4.1.9",
```
wyattjoh: Should be in devDep instead of dep

src/core/client/framework/lib/bootstrap/TalkContext.tsx
```diff
@@ -1,14 +1,19 @@
...
 
 export interface TalkContext {
   // relayEnvironment for our relay framework.
   relayEnvironment: Environment;
 
   // localMessages for our i18n framework.
   localeMessages: MessageContext[];
+
+  // formatter for timeago.
+  timeagoFormatter: Formatter | null;
```
wyattjoh: Why not use:

```typescript
timeagoFormatter?: Formatter;
```

instead?

wyattjoh (changes requested):

package.json
alinhle: 👍

src/core/client/framework/lib/bootstrap/TalkContext.tsx
alinhle: I didn't want this to be optional, but to explicitly set to null in special cases e.g. for testing. Normally this is required.

wyattjoh (approved):

src/core/client/framework/lib/bootstrap/TalkContext.tsx
alinhle: Thought about it again and made it optional :-P

PR #1740, opened 2018-07-09
Add suggested index to comments

icco:
## What does this PR do?

This adds an index to the comment table based off of suggestions from MongoDB Atlas running in production on our site.

screen shot 2018-07-09 at 11 31 40 am

## How do I test this PR?

This index, as I understand it, should apply automatically when running talk.


wyattjoh (approved):
Everyone loves indexes :)

PR #1741, opened 2018-07-09
[next] Permalink

okbel:
### Permalink
- Adds Permalink functionality. A user can share a permalink for a comment.

### Side and core tasks
- Added: Popover Component
- Implementing: ClickOutside and ToggleShow Component https://github.com/coralproject/talk/pull/1751 && https://github.com/coralproject/talk/pull/1757
- Implements `forwardRef`s
- Adds missing imports for UI Components
- Adds functionality to the graph to query by a single comment `comment(id: ID!)`

### Tasks done:
https://www.pivotaltracker.com/story/show/159177760
https://www.pivotaltracker.com/story/show/159177728
https://www.pivotaltracker.com/story/show/158266699
https://www.pivotaltracker.com/story/show/159211298
https://www.pivotaltracker.com/story/show/158925820

## Issues and future discussions
- ~~We need a way to `compose` multiple utilities while dealing with components, i.e `withLocalStateContainer` - `withFragmentContainer`~~
- As is right now the `url` from the `asset` returns the url from the tenant, not the exact url. Comments to replace that in the future have been added.
 
## Updates
- Improved accessibility
- Improved styling
- Improved usability
- Integrated with embed
- Added Tests
- Refactored Stream (_App_ and _Queries_)
- Upgraded typescript to ^3.0.0
- Use local state mutation instead of a reload


src/core/client/stream/components/Comment.tsx
```diff
@@ -25,6 +23,20 @@ const Comment: StatelessComponent<CommentProps> = props => {
         {props.author && props.author.username}
       </Typography>
       <Typography>{props.body}</Typography>
+      <div className={cn("talk-comment-footer")}>
```
okbel: We need to talk about which pattern are we going to use for classes @cvle

package.json
```diff
@@ -24,6 +24,8 @@
   "author": "",
   "license": "Apache-2.0",
   "dependencies": {
+    "@types/react-copy-to-clipboard": "^4.2.5",
+    "@types/react-tooltip": "^3.3.5",
```
wyattjoh: Should be moved to deps.

okbel: 👍 on it

src/core/client/stream/components/Permalink/PermalinkView.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <div className={styles.root}>
+        <Logo />
+        <Flex direction="column" className={styles.comment}>
+          <CommentContainer data={props.comment} />
+          {assetURL && (
+            <Button
+              variant="filled"
+              color="primary"
+              onClick={() => {
+                window.location.href = assetURL;
```
alinhle: If we use a mutation instead and reset the `commentID`, we can avoid reload to whole page.

src/core/client/stream/components/Permalink/PermalinkView.tsx
```diff
@@ -0,0 +1,73 @@
...
+      <Logo />
+      <Typography className={styles.commentNotFound}>
+        Comment not found
+      </Typography>
+      {assetURL && (
+        <Button
+          variant="filled"
+          color="primary"
+          onClick={() => {
+            window.location.href = assetURL;
```
alinhle: See above

alinhle:
> We need a way to compose multiple utilities while dealing with components, i.e withLocalStateContainer - withFragmentContainer

You can do that, just chain it. `withLocalStateContainer(...)(withFragmentContainer()(MyComp))`

src/core/client/stream/components/Comment/Comment.css
```diff
@@ -0,0 +1,3 @@
+.footer {
+  margin: 6px 0;
```
alinhle: If you put a `margin-bottom` here then it'll get in the way of layout components like flex doing their work. Avoid putting margins on the borders, but its ok to layout the margins inside, here: the top margin. Also use `spacingUnits` here it looks like from the design that it's 1.5* spacing unit so, so this should do for this line:
`margin-top: calc(1.5 * var(--spacing-unit));`

src/core/client/stream/components/Permalink/PermalinkPopover.css
```diff
@@ -0,0 +1,6 @@
+.textField {
+  margin-right: 5px;
```
alinhle: Our `Flex` component can add the standard margins, we can use the Flex component instead

src/core/client/stream/components/Permalink/PermalinkPopover.tsx
```diff
@@ -0,0 +1,63 @@
...
+      copied: !state.copied,
+    }));
+  };
+
+  public render() {
+    const { permalinkUrl, toggleVisibility } = this.props;
+    const { copied } = this.state;
+    return (
+      <ClickOutside onClickOutside={toggleVisibility}>
+        <Flex>
```
alinhle: If you do `<Flex innerGutter="half">` instead, you'll get the margins for free without adding custom css classes :-)

src/core/client/stream/components/Permalink/PermalinkView.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+export interface InnerProps {
+  comment: {} | null;
+}
+
+const PermalinkView: StatelessComponent<InnerProps> = props => {
+  let assetURL: string = "";
+  const query = qs.parse(location.search);
+
+  if (!query.assetID && !query.commentID) {
```
alinhle: You can get these from the local state, there is no issue chaining withLocalStateContainer and withFragmentContainer :-)

src/core/client/stream/components/Permalink/PermalinkView.tsx
```diff
@@ -0,0 +1,73 @@
...
+          {assetURL && (
+            <Button
+              variant="filled"
+              color="primary"
+              onClick={() => {
+                window.location.href = assetURL;
+              }}
+            >
+              Back to the Stream
+            </Button>
```
alinhle: Hmm I think design is a little different, see: 

image

The button appears to be above and with a different styling and text

src/core/client/stream/components/Permalink/PermalinkView.css
```diff
@@ -0,0 +1,10 @@
+.root {
+  width: 100%;
+  max-width: 400px;
+  margin: 20px auto;
```
alinhle: Most of the time we'll use spacing units for margins :-)

src/core/client/stream/components/Permalink/PermalinkView.css
```diff
@@ -0,0 +1,10 @@
...
+}
+
+.comment,
+.commentNotFound {
+  margin: 20px auto;
```
alinhle: See above

src/core/client/stream/containers/CommentContainer.spec.tsx
```diff
@@ -8,6 +8,7 @@ import { CommentContainer } from "./CommentContainer";
 it("renders username and body", () => {
   const props: PropTypesOf<typeof CommentContainer> = {
     data: {
+      id: "comment-ID",
```
alinhle: For consistency we could use `comment-id`

src/core/client/stream/containers/PermalinkViewContainer.tsx
```diff
@@ -0,0 +1,28 @@
...
+const enhanced = withFragmentContainer<{ data: Data }>({
+  data: graphql`
+    fragment PermalinkViewContainerQuery on Query
+      @argumentDefinitions(commentID: { type: "ID!" }) {
+      comment(id: $commentID) {
+        ...CommentContainer
+      }
+    }
+  `,
+})(PermalinkViewContainer);
```
alinhle: You can just chain another `withLocalStateContainer` here to get the other data you need.

src/core/client/stream/containers/__snapshots__/CommentContainer.spec.tsx.snap
```diff
@@ -9,5 +9,6 @@ exports[`renders username and body 1`] = `
   }
   body="Woof"
   createdAt="1995-12-17T03:24:00.000Z"
+  id="comment-ID"
```
alinhle: For consistency we could use `comment-id` here

src/core/client/stream/local/initLocalState.ts
```diff
@@ -21,10 +21,23 @@ export default async function initLocalState(environment: Environment) {
...
+      localRecord.setValue(
+        `${location.origin}/?assetID=${query.assetID}`,
+        "assetURL"
+      );
+    }
+
+    if (query.commentID) {
+      localRecord.setValue(query.commentID, "commentID");
+    }
+
```
alinhle, line 40: greaaat

src/core/client/stream/queries/AppQuery.tsx
```diff
@@ -49,6 +77,8 @@ const enhanced = withLocalStateContainer<Local>(
   graphql`
     fragment AppQueryLocal on Local {
       assetID
+      commentID
+      assetURL
     }
   `
 )(AppQuery);
```
alinhle: We'll refactor this when we have the iframe.

src/core/client/ui/components/ClickOutside/index.ts
```diff
@@ -0,0 +1 @@
+export { default, ClickOutside } from "./ClickOutside";
```
alinhle: Don't export the `ClickOutside` class here (keep the `default`), this is a pattern reserved for testing in codebases with higher order components like ours, but not as a API for consumption.

See: 
- https://stackoverflow.com/questions/48515157/how-to-write-proper-jest-tests-with-hoc
- https://medium.com/@7ynk3r/react-testing-done-right-24fdb4ef43d8
- (there are much more resources somewhere in the big wide web)

Just export the `default` export here:

`export { default } from "./ClickOutside";`

src/core/client/ui/components/ClickOutside/ClickOutside.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { findDOMNode } from "react-dom";
 
-interface Props {
+export interface ClickOutsideProps {
   onClickOutside: () => void;
   children: React.ReactNode;
 }
 
-class ClickOutside extends React.Component<Props> {
+export class ClickOutside extends React.Component<ClickOutsideProps> {
```
alinhle: `default` export here is enough, so we don't need to export this here. This is reserved for testing when having higher order components.

See:
- https://stackoverflow.com/questions/48515157/how-to-write-proper-jest-tests-with-hoc
- https://medium.com/@7ynk3r/react-testing-done-right-24fdb4ef43d8
- (there are much more resources somewhere in the big wide web)

src/core/client/ui/components/Popover/Popover.css
```diff
@@ -0,0 +1,15 @@
+.root {
+  background: #ffffff;
+  border: 1px solid #c9cacb;
+  box-sizing: border-box;
+  box-shadow: 1px 0px 4px rgba(0, 0, 0, 0.25);
+  border-radius: 1px;
+  color: #222;
+  display: flex;
+  padding: 6px 10px;
```
alinhle: Use our magic spacing units here :-) it's probably:
`padding: calc(0.5 * var(--spacing-unit)) var(--spacing-unit);`

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+                <div
+                  style={props.style}
+                  className={cn(styles.root, className)}
+                  ref={props.ref}
+                >
+                  {isValidElement(body)
+                    ? body
+                    : body({
+                        toggleVisibility: this.toggleVisibility,
+                        forwardRef: props.ref,
```
alinhle: You can only use `props.ref` once, it has been already used in line `120`. You probably don't need to pass this at all.

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+  | "right-end"
+  | "bottom-end"
+  | "bottom"
+  | "bottom-start"
+  | "left-end"
+  | "left"
+  | "left-start";
+
+interface InnerProps {
+  body: (props: RenderProps) => any | React.ReactElement<any>;
```
alinhle: `body` needs different `RenderProps` from `children`, it shouldn't forward the ref because its already used in line `120`

alinhle: This should resolve to `=> React.ReactNode | React.ReactElement<any>`

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+  children: (props: RenderProps) => any;
```
alinhle: Use `React.ReactNode` instead of `any`

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+}
+
+interface State {
+  visible: false;
+}
+
+interface Props {
+  ref: any;
+  style: CSSProperties;
+}
```
alinhle: Is this interface really necessary? Doesnt `react-popper` come with types?

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+      className,
+      placement,
+    } = this.props;
+
+    const { visible } = this.state;
+
+    return (
+      <Manager>
+        <Reference>
+          {(props: Props) =>
```
alinhle: Is this type really necessary? Doesnt `react-popper` come with types?
Can't you just destructure it? Shouldnt typescript already know the types?

e.g. `{({ref}) =>`

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+            })
+          }
+        </Reference>
+        <Popper
+          placement={placement}
+          modifiers={{ preventOverflow: { enabled: false } }}
+          eventsEnabled
+          positionFixed={false}
+        >
+          {(props: Props) =>
```
alinhle: Same as above, `react-popper` should already have types, you should be able just to destructure or use it.

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+  | "bottom-start"
+  | "left-end"
+  | "left"
+  | "left-start";
+
+interface InnerProps {
+  body: (props: RenderProps) => any | React.ReactElement<any>;
+  children: (props: RenderProps) => any;
+  description?: string;
+  id?: string;
```
alinhle: Make this required field, so we enforce accessibility

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+  | "bottom"
+  | "bottom-start"
+  | "left-end"
+  | "left"
+  | "left-start";
+
+interface InnerProps {
+  body: (props: RenderProps) => any | React.ReactElement<any>;
+  children: (props: RenderProps) => any;
+  description?: string;
```
alinhle: Make this required field, so we enforce accessibility

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+                aria-labelledby={`${id}-ariainfo`}
+                aria-hidden={!visible}
+              >
+                <AriaInfo id={`${id}-ariainfo`}>{description}</AriaInfo>
+                <div
+                  style={props.style}
+                  className={cn(styles.root, className)}
+                  ref={props.ref}
+                >
+                  {isValidElement(body)
```
alinhle: I think `isValidElement` doesn't cover all the valid types for allowed react childrens and is very limiting. Maybe check for `typeof body === 'function'` to check for render  props instead?

src/core/client/ui/components/TextField/TextField.css
```diff
@@ -0,0 +1,11 @@
+.root {
+  font-family: "Source Sans Pro";
+  font-weight: 400;
+  font-size: 16px;
+  letter-spacing: calc(0.57em / 16);
+  background: #ffffff;
+  border: 1px solid #979797;
+  box-sizing: border-box;
+  border-radius: 1px;
+  padding: 5px 20px;
```
alinhle: We should move typography to well typography.css, but you can just add a TODO instead and we'll add it once we have our first real form :-)

src/core/client/ui/components/TextField/TextField.d.ts
```diff
@@ -0,0 +1 @@
+export const root: string;
```
alinhle: This file looks commited by accident

src/core/client/ui/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,14 @@
+import cn from "classnames";
+import React, { InputHTMLAttributes, StatelessComponent } from "react";
+
+import * as styles from "./TextField.css";
+
+interface InnerProps extends InputHTMLAttributes<HTMLInputElement> {
+  classes?: typeof styles;
```
alinhle: Not used, you can remove this for now.

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+  | "right"
+  | "right-end"
+  | "bottom-end"
+  | "bottom"
+  | "bottom-start"
+  | "left-end"
+  | "left"
+  | "left-start";
+
+interface InnerProps {
```
alinhle: Maybe call it PopoverProps, we used InnerProps because of HOCs, we don't have one here :-)

src/core/client/ui/components/Popover/Popover.tsx
```diff
@@ -0,0 +1,138 @@
...
+  };
+
+  public close = () => {
+    this.setState((state: State) => ({
+      visible: false,
+    }));
+  };
+
+  public handleEsc = (e: KeyboardEvent) => {
+    if (e.key === "Escape") {
```
alinhle, line 73: Loooove it

src/core/client/ui/components/Popover/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Popover";
```
alinhle: I realized this pattern doesn't always work and isn't future proof (e.g. when testing suddenly requires other exports and we don't to export them here), let's remove this line and explicitly export what we want.

src/core/client/ui/components/TextField/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./TextField";
```
alinhle, line 1: I realized this pattern doesn't always work and isn't future proof (e.g. when testing suddenly requires other exports and we don't to export them here), let's remove this line and explicitly export what we want.

src/core/client/ui/components/TextField/TextField.tsx
```diff
@@ -0,0 +1,14 @@
+import cn from "classnames";
+import React, { InputHTMLAttributes, StatelessComponent } from "react";
+
+import * as styles from "./TextField.css";
+
+interface InnerProps extends InputHTMLAttributes<HTMLInputElement> {
```
alinhle: Let's call it TextFieldProps, we don't use HOCs for this yet

src/locales/es/stream.ftl
```diff
@@ -0,0 +1,3 @@
+share = Compartir
+copy = Copiar
+copied = Copiado
```
alinhle: Just remove these, we don't need to translate now either, we can later use pontoon

src/core/client/stream/components/Permalink/Permalink.tsx
```diff
@@ -0,0 +1,38 @@
+import { Localized } from "fluent-react/compat";
+import React from "react";
+import { Button, ButtonIcon, Popover } from "talk-ui/components";
+import PermalinkPopover from "./PermalinkPopover";
+
+interface InnerProps {
```
alinhle: Call it `PermalinkProps`? We usually used `InnerProps` when we use HOCs but we don't here.

src/core/client/stream/components/Permalink/Permalink.tsx
```diff
@@ -0,0 +1,38 @@
...
+class Permalink extends React.Component<InnerProps> {
+  public render() {
+    const { commentID, assetURL } = this.props;
+    return (
+      <Popover
+        placement="top"
+        body={({ toggleVisibility, forwardRef }) => (
+          <PermalinkPopover
+            permalinkUrl={`${assetURL}&commentID=${commentID}`}
+            forwardRef={forwardRef}
```
alinhle: It looks like the ref is already consumed inside the `Popover`, so this won't be needed, see review of the `Popover` component.

src/core/client/stream/components/Permalink/PermalinkPopover.tsx
```diff
@@ -0,0 +1,63 @@
+import { Localized } from "fluent-react/compat";
+import React, { CSSProperties } from "react";
+import CopyToClipboard from "react-copy-to-clipboard";
+import { RefHandler } from "react-popper";
+import { Button, ClickOutside, Flex, TextField } from "talk-ui/components";
+import * as styles from "./PermalinkPopover.css";
+
+interface InnerProps {
```
alinhle: Call it `PermalinkPopoverProps`? We usually used `InnerProps` when we use HOCs but we don't here.

src/core/client/stream/components/Permalink/PermalinkView.tsx
```diff
@@ -0,0 +1,73 @@
+import qs from "query-string";
+import * as React from "react";
+import { StatelessComponent } from "react";
+
+import Logo from "talk-stream/components/Logo";
+import { Button, Flex, Typography } from "talk-ui/components";
+import CommentContainer from "../../containers/CommentContainer";
+import * as styles from "./PermalinkView.css";
+
+export interface InnerProps {
```
alinhle: Call it `PermalinkViewProps`? We usually used `InnerProps` when we use HOCs but we don't here.

src/core/client/stream/containers/CommentContainer.tsx
```diff
@@ -2,16 +2,19 @@ import React, { StatelessComponent } from "react";
...
 import withFragmentContainer from "talk-framework/lib/relay/withFragmentContainer";
-import { Omit, PropTypesOf } from "talk-framework/types";
+import { PropTypesOf } from "talk-framework/types";
 import { CommentContainer as Data } from "talk-stream/__generated__/CommentContainer.graphql";
 
-import Comment, { CommentProps } from "../components/Comment";
+import Comment from "../components/Comment";
 
-type InnerProps = { data: Data } & Omit<CommentProps, keyof Data>;
+export interface InnerProps {
```
alinhle: Do we need the `InnerProps` to be exported here?

alinhle (dismissed):
Just lots of tiny things, and 2 not big ones: 
- Use mutation to reset the `commentID`
- Use local state instead of `location`
This looks awesooome @okbel.

src/core/client/stream/containers/CommentContainer.spec.tsx
alinhle: it's not really important though :-P

src/core/client/stream/containers/__snapshots__/CommentContainer.spec.tsx.snap
alinhle: it's not really important though :-P

src/core/client/stream/queries/AppQuery.tsx
alinhle: I mean this file

src/core/client/stream/components/Permalink/PermalinkView.tsx
alinhle: See https://github.com/coralproject/talk/blob/next/src/core/client/stream/mutations/SetNetworkStatusMutation.ts

src/core/client/ui/components/Popover/Popover.tsx
okbel: But it passes `toggleVisibility` 🤔 so we do need render props here.

src/core/client/ui/components/Popover/Popover.tsx
okbel: Let's have a talk over it next time :) I can change it for now!

src/core/client/ui/components/Popover/Popover.tsx
okbel: <3

src/core/client/ui/components/Popover/index.ts
okbel: Ok, removed. Then let's normalize that across our files bc I think I copied that over 💃

src/core/client/ui/components/TextField/index.ts
okbel: Same as above: Ok, removed. Then let's normalize that across our files bc I think I copied that over 💃

alinhle: Yes definitely

wyattjoh:
- [x] Resolve typescript issue

alinhle (approved):

PR #1743, opened 2018-07-10
[next] Passport Integration

wyattjoh:
## What does this PR do?

- Adds initial passport integration
- Adds server graph types based off of the graphql schema
- Added `@auth` custom graph directive

## Waiting on

- [x] https://github.com/auth0/node-jwks-rsa/issues/47


wyattjoh:
Looks like the test failures are due to the lack of url in the specs @cvle, do you think you could take a look into that?

src/core/client/tsconfig.json
```diff
@@ -12,7 +12,8 @@
       "talk-stream/*": ["./stream/*"],
       "talk-framework/*": ["./framework/*"],
       "talk-ui/*": ["./ui/*"],
-      "talk-common/*": ["../common/*"]
+      "talk-common/*": ["../common/*"],
+      "talk-locales/*": ["../../locales/*"]
```
alinhle: Merge conflict, this line isn't needed

package.json
```diff
@@ -3,28 +3,30 @@
...
-    "start": "node dist/index.js",
-    "test": "node scripts/test.js --env=jsdom",
-    "build": "npm-run-all compile --parallel build:*",
     "build:client": "node ./scripts/build.js",
     "build:server": "tsc -p ./src/tsconfig.json",
-    "watch": "NODE_ENV=development ts-node ./scripts/watcher/bin/watcher.ts ./config/watcher.ts",
-    "compile": "npm-run-all --parallel compile:*",
+    "build": "npm-run-all compile --parallel build:*",
     "compile:css-types": "tcm src/core/client/",
+    "compile:graphql": "node ./scripts/types.js",
```
alinhle: Maybe `compile:schema` is more on point?

scripts/types.js
```diff
@@ -0,0 +1,96 @@
+const { Linter, Configuration } = require("tslint");
```
alinhle: This file should have a more descriptive name, `generateSchemaTypes` or sth else

scripts/types.js
```diff
@@ -0,0 +1,96 @@
...
+    "..",
+    "src",
+    "core",
+    "server",
+    "graph",
+    name,
+    "schema",
+    "__generated__",
+    "types.ts"
+  );
```
alinhle: Why not `path.join(__dirname, "../src/core/server/graph/..")`?

src/core/server/app/handlers/auth/local.ts
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const SignupBodySchema = Joi.object().keys({
+  username: Joi.string().trim(),
+  password: Joi.string().trim(),
+  email: Joi.string().trim(),
+});
+
+// Extends the default signup body schema with the displayName to allow it to be
+// sent.
```
alinhle: `to bet set`, `sent` is misleading

src/core/server/models/user.ts
```diff
@@ -120,27 +115,79 @@ export async function create(db: Db, tenantID: string, input: CreateUserInput) {
...
+          : GQLUSER_USERNAME_STATUS.UNSET,
         history: [],
       },
     },
     created_at: now,
   };
 
+  if (input.password) {
+    // Hash the user's password with bcrypt.
+    input.password = await bcrypt.hash(input.password, 10);
```
alinhle: I recommend instead of modifying the original `input` causing an side-effect to use a new object or variable instead (side-effect-free / pure)

src/core/server/app/middleware/passport/oidc.ts
```diff
@@ -0,0 +1,336 @@
...
+    this.db = db;
+  }
+
+  private async verify(
+    tenant: Tenant,
+    token: OIDCIDToken,
+    done: VerifyCallback
+  ) {
+    try {
+      const user = await findOrCreateOIDCUser(this.db, tenant, token);
```
alinhle: I assume `verify` to be `side-effect` free, however this is creating a user in case it doesn't find.

src/core/server/app/middleware/passport/oidc.ts
```diff
@@ -0,0 +1,336 @@
...
+      }
+    );
+  };
+
+  /**
+   * keyFunc will provide the secret based on the given jwkw client.
+   *
+   * @param client the jwks client for the specific request being made
+   */
+  private keyFunc = (client: jwks.JwksClient): jwt.KeyFunction => (
```
alinhle: Maybe  `getSigningKey` is a better name?

src/core/server/app/middleware/passport/oidc.ts
```diff
@@ -0,0 +1,336 @@
...
+        scope: OIDC_SCOPE,
+        session: false,
+      });
+    } catch (err) {
+      return this.error(err);
+    }
+  }
+}
+
+function getEnabledIntegration(tenant: Tenant) {
```
alinhle: Maybe better `getEnabledIntegrationOrThrow` ?

src/core/server/app/middleware/passport/oidc.ts
```diff
@@ -0,0 +1,336 @@
...
+  if (!integration.enabled) {
+    // TODO: return a better error.
+    throw new Error("integration not enabled");
+  }
+
+  return integration;
+}
+
+export function createOIDCStrategy({ db }: OIDCStrategyOptions) {
+  return new OIDCStrategy({ db });
```
alinhle: I was just wondering if there is a way to make this better testable.

alinhle (changes requested):

src/core/server/app/middleware/passport/index.ts
```diff
@@ -1,13 +1,115 @@
...
+/**
+ * wrapAuthz will wrap a authenticators authenticate method with one that
+ * will return a valid login token for a valid login by a compatible strategy.
+ *
+ * @param authenticator the base authenticator instance
+ * @param signingConfig used to sign the tokens that are issued.
+ * @param name the name of the authenticator to use
+ * @param options any options to be passed to the authenticate call
+ */
+export const wrapAuthz = (
```
alinhle: I think Authz means Authorizations but this looks like it's about authentication, right?

alinhle (approved):

PR #1745, opened 2018-07-12
[next] Responsive Support

alinhle:
## What does this PR do?
- Implement `MatchMedia`, `Flex` Component
- Use CSS variables which support dynamically changing values inside `media queries`
- Keep Sass style variables for defining media queries e.g. `@media (min-width: $breakpoints-xs) {`
- Use a default spacing unit of `5` in small screens and `10` otherwise. 
- Remove `precss` preset in favor of our own selection of plugins that we actually use.

## How do I test this PR?

- `npm run docz:watch` to check out the new components.
- `npm run watch` and decrease / increase screen size to see dynamic spacing unit in practise.


okbel (approved):

src/core/client/stream/components/__snapshots__/Stream.spec.tsx.snap
```diff
@@ -1,7 +1,9 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`renders correctly 1`] = `
-<div>
+<div
+  className="Stream-root"
```
wyattjoh, line 5: Should classnames be uppercased? Is this something we can address via the css transformer?

wyattjoh (approved):

src/core/client/stream/components/__snapshots__/Stream.spec.tsx.snap
alinhle: It’s fine to have camelcased classnames as long as it’s consistent. That aside it doesn’t really play much of a role in these snapshots as they are just for testing. It just helps you see where the css class names came from ${ComponentName}-${className} and is different from how it ends up in the browser which currently is just a hash

PR #1748, opened 2018-07-13
[next] Watcher 2.0

alinhle:
## What does this PR do?
- More reliable process killing under `linux`, `mac` and `windows`. Inspired by `nodemon` kill functionality.
- Implement `sane` watcher backend, which is the same as the `jest` watcher backend.
- Automatically detect `watchman` and turn on watchman support in `sane`.
- Use `sane` watcher as default watcher.
- Improved and fixed watcher and executor cleanup and initialisation
- Support watcher `sets` and `defaultSet`
- Changed `docz` default port to `3030` so it doesn't conflict with the server.
- Upgrade deps
- `watcher` bin signature is now: `watcher --config <configFile> [sets or watches] ...`

## How do I test this PR?

- `npm run watch`
- `npm run watch -- server`
- `npm run watch -- client`
- `npm run watch -- docz`


config/webpackDevServer.config.js
```diff
@@ -9,6 +9,7 @@ const paths = require("./paths");
 const protocol = process.env.HTTPS === "true" ? "https" : "http";
 const host = process.env.HOST || "0.0.0.0";
 const serverPort = process.env.PORT || 3000;
```
wyattjoh, line 11: At some point, we need to harmonize what the client uses to read config with what the server is using.

scripts/watcher/LongRunningExecutor.ts
```diff
@@ -51,18 +50,43 @@ export default class LongRunningExecutor implements Executor {
     });
   }
 
-  private restart(): void {
+  private async restart(): Promise<void> {
```
wyattjoh: Why define the return type here while it's can be calculated?

scripts/watcher/pstree.remy.d.ts
```diff
@@ -0,0 +1,6 @@
+declare module "pstree.remy" {
```
wyattjoh: Why not move this into a `types` folder under `scripts/watcher`?

config/webpackDevServer.config.js
alinhle: Yes we should, step by step :-)

scripts/watcher/pstree.remy.d.ts
alinhle: Sure

scripts/watcher/LongRunningExecutor.ts
alinhle: Sure

wyattjoh (approved):
Looks good to me!

wyattjoh (dismissed):
seems to not stop restarting??

wyattjoh (approved):

PR #1756, opened 2018-07-20
[next] Implement AriaInfo

alinhle:
## What does this PR do?
- A component that shows text only to screen readers.

screen shot 2018-07-20 at 5 15 33 pm

### DIFF
--- a/src/core/client/ui/components/AriaInfo/AriaInfo.css
+++ b/src/core/client/ui/components/AriaInfo/AriaInfo.css
@@ -1,0 +1,10 @@
+.root {
+  position: absolute !important;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border: 0;
+}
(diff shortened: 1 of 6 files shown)


src/core/client/ui/components/AriaInfo/AriaInfo.css
```diff
@@ -0,0 +1,10 @@
+.root {
```
wyattjoh, line 1: Is there any document or standards doc that you can link here with justification on how this is applicable to ARIA? I tried to find some myself, but can't really find it.

wyattjoh (changes requested):

src/core/client/ui/components/AriaInfo/AriaInfo.css
alinhle: It's inspired by a proven library https://foundation.zurb.com/sites/docs/visibility.html. Most other examples are similiar, see:

-  https://www.coolfields.co.uk/2016/05/text-for-screen-readers-only-updated/
- https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/
- https://make.wordpress.org/accessibility/handbook/markup/the-css-class-screen-reader-text/

alinhle: However there does not seem to be a golden standard, even though most solutions look similar.

wyattjoh (approved):

PR #1758, opened 2018-07-26
[next] Settings/Tenant

wyattjoh:
## What does this PR do?

Adds support for the tenant cache.


src/core/server/services/tenant/cache.ts
```diff
@@ -1,31 +1,61 @@
...
+    this.mongo = mongo;
+
+    // Create a new client application ID. This prevents duplicated messages
+    // generated by this application from being handled as external messages
+    // as we should have already processed it.
+    this.clientApplicationID = uuid.v4();
 
     // Prepare the list of all tenant's maintained by this instance.
-    this.tenants = new DataLoader(ids => retrieveManyTenants(db, ids));
+    this.tenantsByID = new DataLoader(ids => {
```
alinhle: Change this to  a _property intializer_?

```ts
tenantsByID = new DataLoader(ids => {
  logger.debug({ ids: ids.length }, "now loading tenants");
  return retrieveManyTenants(this.mongo, ids);
});
```

src/core/server/services/tenant/cache.ts
```diff
@@ -1,31 +1,61 @@
...
+      logger.debug({ ids: ids.length }, "now loading tenants");
+      return retrieveManyTenants(mongo, ids);
+    });
+    this.tenantsByDomain = new DataLoader(domains => {
```
alinhle: Same as above?

alinhle (approved):
Small comments, but rest looks good

src/core/server/services/tenant/cache.ts
wyattjoh: Can you elaborate on this one?

PR #1759, opened 2018-07-26
[next] Comments and Moderation

wyattjoh:
## What does this PR do?

Adds support for moderation pipelines.


src/core/server/models/settings.ts
```diff
@@ -0,0 +1,246 @@
...
+   */
+  oidc?: OIDCAuthIntegration;
+
+  /**
+   * sso is the external auth integration for the Google auth.
+   */
+  google?: GoogleAuthIntegration;
+
+  /**
+   * sso is the external auth integration for the Facebook auth.
```
alinhle, line 124: sso=facebook also the others above too :-)

alinhle (approved):

PR #1760, opened 2018-07-26
[next] Templates

wyattjoh:
## What does this PR do?

Added initial template support.


alinhle (approved):
I added a commit to fix the css issues, also removed `cross-env` from our deps. Rest looks good!

PR #1762, opened 2018-07-27
[next] Add support for embed

alinhle:
## What does this PR do?
- Refactor webpack build scripts into `src/core/build`
- Harmonize server and client configuration into `src/core/common/config.ts` 
- Added `npm run tscheck` that runs type checking 
- Added `embed` build configuration
- Implement `embed`
- Added `eventEmitter` to `TalkContext`
- Support detecting clicks outside of iframe and other v4 embed features


scripts/start.ts
```diff
@@ -13,25 +27,8 @@ process.on("unhandledRejection", err => {
...
-  prepareProxy,
-  prepareUrls,
-} = require("react-dev-utils/WebpackDevServerUtils");
-const paths = require("../config/paths");
-const config = require("../config/webpack.config.dev");
-const createDevServerConfig = require("../config/webpackDevServer.config");
-
-const PORT = parseInt(process.env.DEV_SERVER_PORT, 10) || 8080;
-const HOST = process.env.HOST || "0.0.0.0";
+const PORT = 8080;
```
wyattjoh: I'd recommend instead replacing this with something like `config.get("port") + 1` or something similar.

scripts/start.ts
```diff
@@ -51,41 +48,42 @@ if (process.env.HOST) {
...
       if (err) {
         return console.log(err);
       }
       console.log(chalk.cyan("Starting the development server...\n"));
     });
 
-    ["SIGINT", "SIGTERM"].forEach(function(sig) {
-      process.on(sig, function() {
+    ["SIGINT", "SIGTERM"].forEach((sig: any) => {
+      process.on(sig, () => {
```
wyattjoh: Should instead be `process.once`, no need to double call for SIGINT/SIGTERM especially if they both exit.

scripts/start.ts
```diff
@@ -51,41 +48,42 @@ if (process.env.HOST) {
...
         devServer.close();
         process.exit();
       });
     });
   })
-  .catch(err => {
+  .catch((err: Error) => {
     if (err && err.message) {
```
wyattjoh: If the error is being checked for truthyness, why does the type only say `Error`? Shouldn't it be `Error | null | undefined`?

src/core/client/embed/index.ts
```diff
@@ -0,0 +1,30 @@
...
+import qs from "query-string";
+
+import createStreamInterface from "./Stream";
+
+export interface Config {
+  assetID?: string;
+  assetURL?: string;
+  rootURL?: string;
+  id?: string;
+  events?: (eventEmitter: EventEmitter2) => null;
```
wyattjoh: Should this instead be:

```diff
- events?: (eventEmitter: EventEmitter2) => null;
+ events?: (eventEmitter: EventEmitter2) => void;
```

tsconfig.json
```diff
@@ -13,10 +13,11 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "noErrorTruncation": true,
-    "lib": ["es6", "esnext.asynciterable"]
+    "lib": ["dom", "es6", "esnext.asynciterable"]
   },
   "include": [
     "./src/**/.*.js",
+    "./src/types/**/*.d.ts",
```
wyattjoh: How were the `*.d.ts` files included before?

tsconfig.json
```diff
@@ -13,10 +13,11 @@
     "noImplicitAny": true,
     "strictNullChecks": true,
     "noErrorTruncation": true,
-    "lib": ["es6", "esnext.asynciterable"]
+    "lib": ["dom", "es6", "esnext.asynciterable"]
```
wyattjoh, line 16: `lib` is already overridden in both the client and server side tsconfig files, maybe we don't need it here too?

wyattjoh (changes requested):
Practical test seems great! Works lovely, look forward to having it automatically use the development URL instead of still requiring the `assetID=` query param.

scripts/start.ts
alinhle: True!

src/core/client/embed/index.ts
alinhle: Oh yeah 👍

tsconfig.json
alinhle: They weren't as it wasn't required, I ported some scripts to ts and they require types

tsconfig.json
alinhle: This `tsconfig` currently influences ts files outside of src `scripts` and `config` so it is required. I added the `dom` because typescript requires all types in `@types` and `src/types/*.d.ts` and some types needs that even though we don't really use it. Another way would be to use a whitelist. I'd put more thought in it later, when we have e2e tests back, which probably would live in a `test` root level folder which would require additional tsconfig tweaking.

scripts/start.ts
alinhle: I wanted to keep 8080 as a default as this is a often used port for http development.
Also I didn't want to put dev configuration , so I might readd `const PORT = parseInt(process.env.DEV_SERVER_PORT, 10) || 8080;` (not sure why I removed it xD)

scripts/start.ts
alinhle: Will investigate

wyattjoh (approved):

PR #1775, opened 2018-08-01
Use scheduledDeletionDelayHours variable on hoc

leeeandroo:
## What does this PR do?
Introduce use of scheduledDeletionDelayHours on scheduling deletion date

## How do I test this PR?

- change scheduledDeletionDelayHours on `config.js`
- login
- request delete account
- check if the new schedule value is right


wyattjoh (approved):
Great catch!

PR #1784, opened 2018-08-07
[next] Use anchor for better accessibility and restore history

alinhle:
## What does this PR do?
- Use anchor tag for better accessibility
- Make the back button work

## How do I test this PR?

- `Show All Comments` button should be a A tag and points to the correct URL
- Browser history should remain intact and works when going back and forward


okbel (approved):
👍

PR #1794, opened 2018-08-09
[next] Flex improvements

alinhle:
## What does this PR do?
- Put ref and root className to first div
- Support double itemGutters

## How do I test this PR?

- `npm run watch -- docz`


okbel (approved):
👍

PR #1795, opened 2018-08-09
[next] Comment Editing

wyattjoh:
## What does this PR do?

Adds the `editComment` mutation and `editing` metadata.

## How do I test this PR?

1. `createComment` mutation
2. `editComment` mutation
3. `comment` query to view `editing` metadata


src/core/server/graph/tenant/mutators/comment.ts
```diff
@@ -1,11 +1,14 @@
...
+  GQLEditCommentInput,
+} from "talk-server/graph/tenant/schema/__generated__/types";
 import { Comment } from "talk-server/models/comment";
-import { create } from "talk-server/services/comments";
+import { create, edit } from "talk-server/services/comments";
 
 export default (ctx: TenantContext) => ({
-  create: (input: GQLCreateCommentInput): Promise<Comment> => {
-    return create(
+  create: (input: GQLCreateCommentInput): Promise<Comment | null> =>
```
alinhle: Why would it return `null`?

src/core/server/graph/tenant/mutators/comment.ts
```diff
@@ -16,6 +19,17 @@ export default (ctx: TenantContext) => ({
         parent_id: input.parentID,
       },
       ctx.req
-    );
-  },
+    ),
+  edit: (input: GQLEditCommentInput): Promise<Comment | null> =>
```
alinhle: Same question, why would it return null?

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -470,13 +482,18 @@ type Comment {
...
   replies(
     first: Int = 10
     orderBy: COMMENT_SORT = CREATED_AT_DESC
     after: Cursor
   ): CommentsConnection
+
+  """
+  editing returns details about the edit status of a Comment.
+  """
+  editing: EditInfo
```
alinhle: What do you think of adding it directly to `Comment` instead of using a separate type `EditInfo`? Makes it slightly easier to consume on the client. Just a thought.

alinhle: Besides that, shouldn't this be `EditInfo!`?

alinhle (changes requested):

src/core/server/graph/tenant/schema/schema.graphql
wyattjoh: I can change it to `EditInfo!`, but without repeating the same logic twice, shouldn't it be the same easiness level to access `comment.editInfo.?` than accessing `comment.?`?

alinhle: I guess it's fine. Optimistic updates are a bit more code first getting the comment, then getting the linked record and then setting the value. The more nested it is, the more cumbersome it is with the vanilla relay api, one level should be ok.

wyattjoh: Sounds good! Just don't take a look at the Karma API :P

alinhle (approved):

PR #1800, opened 2018-08-12
Enhance security agains malicious GraphQL queries

lytvynenko:
## What does this PR do?
 This PR introduces two extra security measures: 
 - query depth limiter 
 - query page size limiter. 

Both are intended to prevent DDoS attacks via malicious GraphQL query ([Medium article](https://blog.apollographql.com/securing-your-graphql-api-from-malicious-queries-16130a324a6b)). 
  
Query depth limiter utilises [graphql-depth-limit](https://github.com/stems/graphql-depth-limit) npm module and introduces a new environment variable `TALK_QUERY_DEPTH_LIMIT` (default  value 15).  

Query page size limiter implemented as custom GraphQL scalar type `QueryLimit` which must be used instead of `Int` in every GraphQL query with `$limit` variable. `graphql-js` version was bumped to support default values for custom scalars (Issue [#461](https://github.com/graphql/graphql-js/issues/461))


wyattjoh:
We've explored this feature in the past, but ran into issues preventing us from enabling them.

We provide the ability to adjust the maximum query depth on the client with the [TALK_THREADING_LEVEL](https://docs.coralproject.net/talk/advanced-configuration/#talk-threading-level) variable, which would make this feature quite complicated.

Unfortunately as well, many newsrooms using the GraphQL API to perform administrative actions, this type of limiters would be.. limiting.

For Talk 5.0, we will be enabling query whitelisting, pretty much eliminating DOS style attacks this way, and allowing any query from users presenting an administrative token.

Thank you so much for the contribution, but we'll likely want to delay this until 5.0.

PR #1808, opened 2018-08-20
[next] Production Usage

wyattjoh:
## What does this PR do?

Sets up Talk to be deployed in production by:

- Providing a `Dockerfile`
- Compressing static assets
- Adding cache control headers


Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ENV NODE_ENV production
+
+# Bundle app source
+COPY . /usr/src/app
+
+# Install app dependencies and build static assets.
+RUN npm install && \
+  npm run compile && \
+  npm run build && \
+  npm prune --production
```
alinhle: dev dependencies were not installed for me, but this would work:
```dockerfile
# Install app dependencies and build static assets.
RUN npm install --only=production && \
  npm install --only=dev && \
  npm run compile && \
  npm run build && \
  npm prune --production
```

alinhle (changes requested):

Dockerfile
alinhle: Fixing this and it works great!

alinhle (approved):

PR #1810, opened 2018-08-21
fix language detection in IE11 and Edge

### LINKED ISSUE #1801: ie11/edge ignores default_lang
bug

#### Intended outcome:

coral should appear in german, which has been defined with `TALK_DEFAULT_LANG=de`

bildschirmfoto 2018-08-14 um 10 54 11

#### Actual outcome:

coral appears in english, but only in **IE11 and Edge**

bildschirmfoto 2018-08-14 um 10 53 52

it happens only with **IE11 and Edge**, in other browsers coral appears in german.

#### Version and environment

talk v4.5.1
**IE11 and Edge**

hjanuschka:
## What does this PR do?
Fixes the language detection in IE11 and Edge Browsers

## How do I test this PR?
open the `/admin`  in an IE11 using DE as locale.

fixes: https://github.com/coralproject/talk/issues/1801


hjanuschka:
i am pretty confident, that the failing integration tests are not due to this PR 😨

wyattjoh (approved):
Good catch! For some reason, the documentation https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/languages indicates that IE/Edge supports this property.

hjanuschka:
@wyattjoh thx for merging it, the documentation is indeed a bit weird. IE11 has a `*` - and in the details it says ` Closest available (non-standard) properties are userLanguage and browserLanguage.`.
i think marking IE as green is just wrong, since you folks are from mozilla too, any idea where i can submit a PR for that doc. error?

wyattjoh:
You might be able to submit an issue here? https://github.com/mdn/browser-compat-data

PR #1817, opened 2018-08-23
Subscription Optimizations

wyattjoh:
## What does this PR do?

Optimizes user extraction during subscription operations.


alinhle (approved):
Code looks good, double check to be sure it really batches and does not cache :-)

PR #1819, opened 2018-08-24
support canonical links with query string separator;

gontranz:
## What does this PR do?

Allows `view_context` links from moderation to be represented accurately in the query string when canonical link includes a `?` separator or query string.

## How do I test this PR?

- click a button
- view the cat
- see the cat meow


client/coral-admin/src/routes/Moderation/components/Comment.js
```diff
@@ -168,7 +168,11 @@ class Comment extends React.Component {
                 <div className={styles.commentContentFooter}>
                   <a
                     className={styles.external}
-                    href={`${comment.asset.url}?commentId=${comment.id}`}
+                    href={
+                      comment.asset.url.includes('?')
```
wyattjoh, line 172: At some point in the code, I'm sure we added a url module. I'd instead recommend that you replace the code associated with:

```js
import { URL } from "url";

const assetURL = new URL(comment.asset.url);
assetURL.searchParams.set("commentId", comment.id);
const commentURL = assetURL.href;
```

wyattjoh (changes requested):

paulwellnerbou:
Fixes #1851.

client/coral-admin/src/routes/Moderation/components/Comment.js
paulwellnerbou: @wyattjoh, I am trying to get this working, as we need this fix.
Unfortunately this code does not work in my approach, maybe I am doing something wrong. I get errors:

```
bundle.f891298bb75164af31a0.js:1 Uncaught TypeError: v.URL is not a constructor
```

https://github.com/paulwellnerbou/talk/blob/ff81b10ecb791c24a5fadf137cb7f9e43c1c0958/client/coral-admin/src/routes/Moderation/components/Comment.js#L53-L57

I found out that this code was already implemented:

https://github.com/coralproject/talk/blob/b9d88251e6cd45da61f1e46ef43ed12a6d47000a/graph/resolvers/comment.js#L122-L124

Is there a way to reuse this anyhow?
(I am not experienced with React but I'd like to dive in, so with a little hint, I would be happy to fix this)

wyattjoh:
Closing in favor of #1860

Issue #1615, opened 2018-05-14
[Question] Disable websocket support


cristiandean:
Our website has huge daily access and keeping websocket enabled is becoming too network intensive. We would like to be able to disable websocket for at least a first version in production.

As for now we can't do that. What do you think about having an option to completelly disable websocket on client and server? We can send a pull request with this feature if you think that this behaviour would be desired.

Thanks

wyattjoh:
Without websockets, a refresh would be required to load any new content. It's something we can certainly explore for sure though.

cristiandean:
Yes, we are going to lose freshness in favor of network stability for a first moment. A pull request for that is valid?

wyattjoh:
I'm worried that a PR might get too big if it were to disable websockets. Our code is quite tightly coupled to the fact that subscriptions "just work". A PR would we welcome, but would likely consume quite a bit of time to write + review (just a fair warning ;))

cristiandean:
Hello, I'm working on this PR and I came across a question.
I would like to use this setting to disable the websocket in the application client, more precisely in this section:

```
window.TalkEmbed = Coral.Talk.render(document.getElementById('coralStreamEmbed'), {
    talk: '{{ BASE_URL }}',
    asset_url: '{{ asset_url }}',
    asset_id: '{{ asset_id }}',
    websocket: {{ websocket_enable }},
```

My question is, what is the best way to get this setting? More precisely in the file **client/coral-framework/services/client.js**.

kgardnr:
I'm closing this as we won't be including this in v4.x, but we will consider it for v5.

PR #1833, opened 2018-08-30
Optional websocket support

cristiandean:
## What does this PR do?
Allows that websocket become optional maintaining the behavior as possible. As discussed in this issue #1615  we are going to lose freshness in favor of network stability. We would like know your oppinion about this Pull Request.

## How do I test this PR?
Set **TALK_WEBSOCKET_CLIENT_DISABLE=TRUE**

### DIFF
--- a/client/coral-admin/src/routes/Community/containers/FlaggedAccounts.js
+++ b/client/coral-admin/src/routes/Community/containers/FlaggedAccounts.js
@@ -126,9 +126,9 @@ class FlaggedAccountsContainer extends Component {
       },
     ];
 
-    this.subscriptions = parameters.map(param =>
-      this.props.data.subscribeToMore(param)
-    );
+    this.subscriptions = parameters
+      .map(param => this.props.data.subscribeToMore(param))
+      .filter(param => param);
   }
 
   unsubscribe() {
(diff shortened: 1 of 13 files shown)


hjanuschka:
i love this! ❤️  thx for starting it, and putting in your time.
how about a client side refresh using REST in an interval?


getting rid of websockets, i guess would eliminate alot of hw resources at our end.

kgardnr:
Hey @cristiandean and @hjanuschka - we've taken a look at this and it's not a full implementation of disabling websockets - since websockets are so core to Talk 4.x, this might be difficult to do fully.

We're looking at alternatives for 5.x, and will keep you posted.

I would suggest testing this patch in a production environment and seeing if it solves the issues you're facing. We cannot merge this into core as-is, without more extensive testing in a live environment. We'd be interested to see what comes of your testing. Please report back!

hjanuschka:
since our infra. is now able to handle the amount of websockets, and we are probably soon moving the websockets to akamai anyway - we do not really need it.

i would like to test it anyway, but we'd need a sort of `setInterval` alternative to get a bit of freshness atleast.


oh and keep WS in admin dashboard.

wyattjoh:
Sounds great @hjanuschka! Feel free to patch your version to play with websocket support in Talk 4, we plan on focusing first class support of websocket toggling for the embed stream in Talk 5.

PR #1839, opened 2018-08-31
Change proxy-agent to support http/https through http/https/socks/pac…

thor27:
When using SCRAPER_PROXY_URL and the page with comments isn't HTTPS the asset title would be saved as "400 The plain HTTP request was sent to HTTPS port".

The solution proposed is to change the code to use the library "proxy-agent"[1]. That way we support more proxy types and also works with HTTP and HTTPS connections.

[1] - https://www.npmjs.com/package/proxy-agent


wyattjoh (approved):
Looks good to me!

PR #1840, opened 2018-09-03
fix long urls breaking moderation box

### LINKED ISSUE #1835: Long urls break comment cards in Moderator view

Bug

#### Intended outcome:
Long urls should wrap within the comment cards in Admin > Mod Queues.

#### Actual outcome:

long_urls

Create a new comment with a long url. You can use this: https://longurlmaker.com/

#### Version and environment
4.6.1 on Coral Staging

hjanuschka:
fixes: https://github.com/coralproject/talk/issues/1835

bildschirmfoto 2018-09-03 um 08 20 18

### DIFF
--- a/client/coral-admin/src/routes/Moderation/components/Comment.css
+++ b/client/coral-admin/src/routes/Moderation/components/Comment.css
@@ -61,6 +61,7 @@
   line-height: 1.5;
   font-weight: 300;
   margin-bottom: 8px;
+  word-break: break-all;
 }
 
 .body {
--- a/client/coral-admin/src/routes/Moderation/components/Comment.css
+++ b/client/coral-admin/src/routes/Moderation/components/Comment.css
@@ -61,7 +61,7 @@
   line-height: 1.5;
   font-weight: 300;
   margin-bottom: 8px;
-  word-break: break-all;
+  word-break: break-word;
 }
 
 .body {
--- a/client/coral-admin/src/routes/Moderation/components/Comment.css
+++ b/client/coral-admin/src/routes/Moderation/components/Comment.css
@@ -61,7 +61,7 @@
   line-height: 1.5;
   font-weight: 300;
   margin-bottom: 8px;
-  word-break: break-word;
+  overflow-wrap: break-word;
 }
 
 .body {


kgardnr:
Thank you for fixing this @hjanuschka!!

Tagging @clintabrown as he was the one who reported this.

We will review this asap!

client/coral-admin/src/routes/Moderation/components/Comment.css
```diff
@@ -61,6 +61,7 @@
   line-height: 1.5;
   font-weight: 300;
   margin-bottom: 8px;
+  word-break: break-all;
```
alinhle: Hi @hjanuschka, thanks for this PR.

`word-break: break-all;` breaks in the middle of words even if it doesn't have to. Would `word-wrap: break-word;` also do?

alinhle (changes requested):

hjanuschka:
feedback adressed 👍

client/coral-admin/src/routes/Moderation/components/Comment.css
```diff
@@ -61,7 +61,7 @@
   line-height: 1.5;
   font-weight: 300;
   margin-bottom: 8px;
-  word-break: break-all;
+  word-break: break-word;
```
cvle: Even better break is `overflow-wrap: break-word;` which only breaks word if absolute necessary.

See note in, which explains the contrast to `word-break`.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

client/coral-admin/src/routes/Moderation/components/Comment.css
```diff
@@ -61,6 +61,7 @@
   line-height: 1.5;
   font-weight: 300;
   margin-bottom: 8px;
+  word-break: break-word;
```
alinhle: Even better break is overflow-wrap: break-word; which only breaks word if absolutely necessary.

See note in, which explains the contrast to word-break.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap

alinhle (changes requested):

hjanuschka:
@cvle nice, didnt' know that one, thx, feedback addressed

alinhle (approved):

alinhle:
Thanks!

PR #1843, opened 2018-09-03
[next] Save Comment Draft + Pym Storage

alinhle:
## What does this PR do?
- Implement PymStorage
- Implement save comment draft
- Tests

## How do I test this PR?

- Write a comment
- Refresh


src/core/client/embed/decorators/withPymStorage.ts
```diff
@@ -0,0 +1,52 @@
+import { Decorator } from "./";
```
wyattjoh: Kind of a strange looking import???

src/core/client/embed/decorators/withPymStorage.ts
```diff
@@ -0,0 +1,52 @@
...
+
+const withPymStorage = (
+  storage: Storage,
+  type: "localStorage" | "sessionStorage",
+  prefix = "talkPymStorage:"
+): Decorator => pym => {
+  pym.onMessage(`pymStorage.${type}.request`, (msg: any) => {
+    const { id, method, parameters } = JSON.parse(msg);
```
wyattjoh, line 9: These are coming from the client.. Are we _sure_ that `JSON.parse` won't throw?

src/core/client/framework/lib/bootstrap/TalkContext.tsx
```diff
@@ -27,6 +28,12 @@ export interface TalkContext {
   /** Session storage */
   sessionStorage: Storage;
 
+  /** Session Storage over pym */
```
wyattjoh: Comment `s/Session/Local/`

src/core/client/framework/lib/bootstrap/createContext.tsx
```diff
@@ -118,6 +119,8 @@ export default async function createContext({
     postMessage: new PostMessageService(),
     localStorage: createLocalStorage(),
     sessionStorage: createSessionStorage(),
+    pymLocalStorage: pym && createPymStorage(pym, "localStorage"),
```
wyattjoh, line 122: Does this constructor `thing && thing` not possibly result in `false` versus undefined?

src/core/client/framework/lib/storage/PymStorage.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Creates a storage that put requests onto pym.
+ * This is the counterpart of `connectStorageToPym`.
+ * @param  {string} pym  pym
+ * @return {Object} storage
+ */
+export default function createPymStorage(
+  pym: Pym,
+  type: "localStorage" | "sessionStorage"
+): PymStorage {
```
wyattjoh: Isn't this best represented as another class that implements PymStorage?

src/core/client/stream/test/create.tsx
```diff
@@ -0,0 +1,59 @@
...
+  });
+
+  const context: TalkContext = {
+    relayEnvironment: environment,
+    localeBundles: [createFluentBundle()],
+    localStorage: createInMemoryStorage(),
+    sessionStorage: createInMemoryStorage(),
+    pymLocalStorage: createFakePymStorage(),
+    pymSessionStorage: createFakePymStorage(),
+    rest: new RestClient("http://localhost/api"),
```
wyattjoh, line 47: Don't you need to specify the port?

src/core/client/stream/test/postComment.spec.tsx
```diff
@@ -92,25 +66,26 @@ it("renders comment stream", async () => {
...
   expect(testRenderer.toJSON()).toMatchSnapshot();
 
   // Wait for loading.
   await timeout();
 
-  // Travel to the time where the "timeout" has executed.
-  timekeeper.travel(new Date("2018-07-06T18:24:01.002Z"));
-
   // Test after server response.
   expect(testRenderer.toJSON()).toMatchSnapshot();
```
wyattjoh, line 90: When timekeeper is reset above, time begins to tick, so this snapshot (which depends on time) is no longer guaranteed to be deterministic.

Can you explain the changes here?

wyattjoh (changes requested):

src/core/client/framework/lib/bootstrap/createContext.tsx
alinhle: It won't, also typescript would have complained :-)

src/core/client/framework/lib/storage/PymStorage.ts
alinhle: Copied that from v4. I could turn it into a class, but is it _really_ going to be _better_? 🤔

src/core/client/stream/test/create.tsx
alinhle: This is just the hull, it won't call to any server, because it'd be mocked during the tests.

src/core/client/stream/test/postComment.spec.tsx
alinhle: between line `108` and `114`, there is really no logic being executed, so this basically is a noop.

src/core/client/framework/lib/bootstrap/createContext.tsx
alinhle: `thing1 && thing2` if anything is falsey, it returns that particular falsey value e.g. `null` or `undefined` or `false` or `""`

src/core/client/embed/decorators/withPymStorage.ts
alinhle: Well it works 😅, yeah I can change that, it might encourages someone to do a circular import _(which currently isn't because it's just a type and this import is not actually being executed 😆)_

src/core/client/embed/decorators/withPymStorage.ts
alinhle: Well if some code in the stream sends some malformed input, then it should throw no?

src/core/client/stream/test/postComment.spec.tsx
alinhle: Except for optimistic responses – which is what the timekeeper covers – we don't have time dependent output. We avoid it whenever possible, to make testing simpler / possible.

src/core/client/stream/test/create.tsx
alinhle: with `sinon.mock`

src/core/client/stream/test/postComment.spec.tsx
wyattjoh: Note: snapshot data does not change in a time dependent way.

wyattjoh (approved):

PR #1846, opened 2018-09-05
External User Fix

wyattjoh:
## What does this PR do?

Returns the user when the user was not initially found, and optionally publishes to the graph context if it was non null.

### DIFF
--- a/services/users.js
+++ b/services/users.js
@@ -98,8 +98,12 @@ async function upsertUser(
   // Save the user in the database.
   await user.save();
 
-  // Emit that the user was created.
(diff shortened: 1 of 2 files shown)


services/users.js
```diff
@@ -68,6 +68,8 @@ async function upsertUser(
     },
   });
   if (user) {
+    user.wasUpserted = false;
+    user.$ignore('wasUpserted');
```
alinhle, line 72: Not the most pretty way, but alright ;-)

alinhle (approved):

PR #1850, opened 2018-09-06
[next] Implement Reply

alinhle:
## What does this PR do?
- Implement Reply functionality
- Adapt reply list styles to current design
- Focus on RTE when opening reply form
- Introduce `PromisifiedStorage` and simplify storage handling.
- Fix issues in the `removeFragmentRefs` function
- Move `uuidGenerator` into `TalkContext` for easier testing.
- Some fixes to allow Stream to run outside of an embed context e.g. for debugging
- Integration tests now only pass when all translations were provided.
- Disable submit button when form is empty
- Disable rte  when form is submitting
- Saves reply comment draft
- Wrap long words in comments

## How do I test this PR?

- Login 
- Post a comment reply


src/core/client/framework/helpers/getMe.ts
```diff
@@ -3,7 +3,9 @@ import { Environment, ROOT_ID } from "relay-runtime";
 export default function getMe(environment: Environment) {
   const source = environment.getStore().getSource();
   const root = source.get(ROOT_ID)!;
-  const meKey = Object.keys(root).find(s => s.startsWith("me("))!;
+  const meKey = Object.keys(root)
```
okbel, line 6: Nice! So I'm assuming this gets the `me` out of the Relay Store. Right? We can make a new section in our docs for our helpers

okbel (commented):
Works as expected! Amazing work. Continuing with the review

okbel:
fyi: Can't post a reply to a reply. I'm assuming this will be handled with **THREADING** https://www.pivotaltracker.com/story/show/158640079

src/core/client/framework/helpers/getMe.ts
alinhle: Yes! It's only useful when handling mutations. Otherwise you'd get me using a fragment.

alinhle:
Correct 👍

okbel (approved):

PR #1852, opened 2018-09-07
[next] Adding a Message component

okbel:
### What does this PR do?
- Adds Message Component, needed for Editing
- `ValidationMessage` relies on `Message` Component
screen shot 2018-09-07 at 10 32 42 am

### How do I test this PR?
- run docz
- check the Message Component
- check the values to match the styles, if it does we should correct the ones in `ValidationMessage`

Note: This component is similar to `ValidationMessage`, we could have a `<Message variant="validation" />` or `<Message variant="info" />`, etc.
I like them both separated because it looks semantic. But the API it's up for discussion.  :)

Note2: I know we talked about having `Icon` off this but this has specific styling that's why I think it should be inside the component


okbel:
API updated!

src/core/client/ui/components/Message/Message.tsx
```diff
@@ -0,0 +1,75 @@
...
+  */
+  fullWidth?: boolean;
+  /*
+  * Name of the icon, render if provided
+  */
+  icon?: string;
+  /*
+  * Name of color, "info" stays by default - common gray one
+  */
+  color?: "validation" | "info";
```
alinhle: Rather than having a new set of new enums that maps to our theme colors, it'd be simpler to just pass in the colors from our color palette of our theme `grey | primary | secondary | error`. The Button, Typography, TextField components are also refering to these colors.

src/core/client/ui/components/Message/Message.mdx
```diff
@@ -0,0 +1,27 @@
...
+  <HorizontalGutter>
+    <Message>This is a message</Message>
+    <Message fullWidth>Contrary to popular belief, Lorem Ipsum is not simply random text.</Message>
+  </HorizontalGutter>
+</Playground>
+
+## Usage with icon
+<Playground>
+  <HorizontalGutter>
+    <Message icon="alarm">Edit: 1 min 23 secs Remaining</Message>
```
alinhle: From my experience the API to pass in the icon (`<Message icon="alarm">`) has proven to be inferior. It's inflexible as it does not allow to customize which component to use and it does not allow to customize other props of the icon.

A different way would be to allow passing in the Element itself like `<Message icon={<Icon>alarm</Icon>}>`, which is an improvement over the above API. Some flaws include that layouting props must be handled by the `Message` component e.g. `<Message icon={<Icon>alarm</Icon>} alignIcon="right" />`.  Another downside is when using this pattern it can lead to hardly readable Element trees like `<X propA={<Y propB={<Z propC="u" />} propD="v" />} propE="z" />`.

Lastly there is this API that I personally prefer: `<Message><Icon>alarm</Icon> Edit: 1 min 23 secs Remaining</Message>`. It offers the flexibility to pass in any Icon Element, the Message component doesn't need to contain layouting props like when aligning the icon to the right: `<Message>Edit: 1 min 23 secs Remaining <Icon>alarm</Icon></Message>` and if using this pattern often it never leads to hardly readable element trees because it fits nicely into the `jsx` syntax.

The `Button` Component follows the third API pattern already, how do you feel about following this pattern also in this component?

alinhle (changes requested):

src/core/client/ui/components/Message/Message.tsx
okbel: Agreed

okbel: 👍

src/core/client/ui/components/Message/Message.css
```diff
@@ -4,11 +4,18 @@
   display: inline-flex;
   justify-content: flex-start;
   align-items: center;
-  padding: calc(0.5 * var(--spacing-unit)) var(--spacing-unit);
+  padding: calc(0.8 * var(--spacing-unit)) var(--spacing-unit);
```
alinhle: We usually use half steps of spacing unit, e.g. 0.5, 1.0, 1.5 whenever possible unless there is a specific reason not to.

src/core/client/ui/components/Message/Message.css
```diff
@@ -4,11 +4,18 @@
...
   box-sizing: border-box;
   border-radius: var(--round-corners);
-  border-left-width: calc(0.5 * var(--spacing-unit));
-  border-left-style: solid;
+  border-width: 1px;
+  border-style: solid;
+  border-left-width: calc(0.8 * var(--spacing-unit));
```
alinhle: Same as above

src/core/client/ui/components/Message/Message.css
```diff
@@ -23,5 +30,5 @@
 }
 
 .icon {
```
alinhle: I think this class is no longer used?

src/core/client/ui/components/ValidationMessage/ValidationMessage.tsx
```diff
@@ -16,40 +12,28 @@ export interface ValidationMessageProps {
...
+   * If set renders a full width message
    */
-  classes: typeof styles;
+  fullWidth?: boolean;
   /*
-  * If set renders a full width message
+  * Name of the icon, if not provided it will default to warning icon
   */
-  fullWidth?: boolean;
+  icon?: string;
```
alinhle: This is not used, we can remove it?

src/core/client/ui/components/ValidationMessage/ValidationMessage.tsx
```diff
@@ -16,40 +12,28 @@ export interface ValidationMessageProps {
...
-
-  const rootClassName = cn(
-    classes.root,
-    classes.colorError,
-    {
-      [classes.fullWidth]: fullWidth,
-    },
-    className
-  );
+  const { className, fullWidth, children, icon, ...rest } = props;
```
alinhle: It think `fullWidth` should be passed to `Message`

okbel:
@cvle Updated! Thanks for the review!

alinhle (dismissed):

alinhle (approved):

PR #1853, opened 2018-09-07
[next] Start a clean session when user logs in / out

alinhle:
## What does this PR do?
- `createContext` has been changed to `createManaged` which returns a `ManagedTalkContextProvider` that handles context changes.
- `clearSession` is now passed along the context, and is called whenever the user session changes.
  - A new relay environment will be created and passed into the context
  - Session storage will be cleared (only keys with our talk prefix)
- Adapt code that expected a static context, but now the context can change.

## Depends on
- https://github.com/coralproject/talk/pull/1850

## How do I test this PR?

- Login, Logout


src/core/client/auth/test/create.tsx
```diff
@@ -45,6 +46,8 @@ export default function create(params: CreateParams) {
     postMessage: new PostMessageService(),
     browserInfo: { ios: false },
     uuidGenerator: createUUIDGenerator(),
+    eventEmitter: new EventEmitter2(),
```
wyattjoh: Just note that the default event emitter has caps on the number of listeners.

src/core/client/framework/lib/bootstrap/createManaged.tsx
```diff
@@ -120,28 +196,38 @@ export default async function createContext({
 
   const localeBundles = await generateBundles(locales, localesData);
 
+  const localStorage =
+    (pym && inIframe && createPymStorage(pym, "localStorage")) ||
```
wyattjoh: Maybe possible to clean this up a bit? It's a bit hard to follow.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -809,12 +809,8 @@ type Query {
 
   """
   me is the current logged in User.
-
-  clientAuthRevision is an implementation detail that is only
-  used on the client to invalidate the cache.
-  TODO: This should move to a client side directive if this becomes possible.
   """
-  me(clientAuthRevision: Int): User
```
wyattjoh, line 817: thank goodness

src/core/client/framework/lib/bootstrap/createManaged.tsx
alinhle: Sure

src/core/client/auth/test/create.tsx
alinhle: Interesting, yeah not directly using this yet, but we will eventually.

src/core/server/graph/tenant/schema/schema.graphql
alinhle: I know ;-)

wyattjoh (approved):

PR #1877, opened 2018-09-17
[next] Embed plus

alinhle:
## What does this PR do?
- Implement new StreamEmbed API
```ts
const TalkStreamEmbed = Coral.Talk.createStreamEmbed({ id: 'coralStreamEmbed' });   
TalkStreamEmbed.render();
```
- Emit event `showPermalink` when a permalink request is detected
- Added `article.html` that renders a long article
- Added `articleButton.html` that renders a long article and hides the embed behind a button
- Don't produce embed htmls in production
- Resolve assetURL from canonical link ref or defer to current url like in v4 and pass it to the embed stream, which in turn use it to form queries
- Fix permalink URL creation
- Remove optional params from pym url
- Automatically scroll to permalinked comment, in permalink view
- Implement `autoRender` based on `IntersectionObserver`.


src/core/client/embed/StreamEmbed.ts
```diff
@@ -0,0 +1,125 @@
...
+  private pymControlFactory: PymControlFactory;
+
+  constructor(
+    config: StreamEmbedConfig,
+    pymControlFactory = defaultPymControlFactory
+  ) {
+    this.config = config;
+    this.pymControlFactory = pymControlFactory;
+    if (config.commentID) {
+      setTimeout(() => config.eventEmitter.emit("showPermalink"), 0);
```
wyattjoh: Can you add a comment as to why this is a timeout of 0?

alinhle: Sure!

wyattjoh (approved):

PR #1897, opened 2018-09-20
[next] Improve cross browser support

alinhle:
## What does this PR do?
- Improves cross browser support

## How do I test this PR?
- Still builds and runs :-)


src/core/build/createWebpackConfig.ts
```diff
@@ -288,6 +288,25 @@ export default function createWebpackConfig({
                 },
               ],
             },
+            // Makes sure node_modules are transpiled the way we need them to be.
+            {
+              test: /\.js$/,
```
wyattjoh, line 293: Should we include those files coming from TS at all?

src/core/build/createWebpackConfig.ts
```diff
@@ -363,6 +382,22 @@ export default function createWebpackConfig({
...
+        // of CSS changes), or refresh the page (in case of JS changes). When you
+        // make a syntax error, this client will display a syntax error overlay.
+        // Note: instead of the default WebpackDevServer client, we use a custom one
+        // to bring better experience for Create React App users. You can replace
+        // the line below with these two lines if you prefer the stock client:
+        // require.resolve('webpack-dev-server/client') + '?/',
+        // require.resolve('webpack/hot/dev-server'),
+        require.resolve("react-dev-utils/webpackHotDevClient"),
+      ];
+
```
wyattjoh, line 400: How is this related to cross browser support?

alinhle: Just deduplicated some lines there ;-)

alinhle: Nothing has changed

src/core/build/createWebpackConfig.ts
alinhle: What do you mean?

wyattjoh: Well, does this loader ensure that files transpiled through the `.ts` loader are also ran through the babel transformer?

alinhle: It transpiles js files in `node_modules` to the javascript version we need for our supported browsers, because a couple of modules are not in pure es5.

wyattjoh: What would happen if I for example, used sticky regex inside our typescript files? Would it still transform it?

alinhle: The configuration for `tsx?` files is right above this :-)

wyattjoh (approved):

PR #1901, opened 2018-09-22
[next] Parent Threading

wyattjoh:
## What does this PR do?

Introduces a new concept for storing thread information for a given parent/child relationship to drastically improve deep parent retrieval.

## 🚨 Caveat 🚨

This does introduce a technical limitation however, as [MongoDB limits the size of documents to 16 MB](https://docs.mongodb.com/manual/reference/limits/#bson-documents), it roughly equates to a given comment having ~20K ID's referencing a given comment in the upwards or downwards direction.

If for example, a comment is 100 levels deep, it can have a maximum of (20K - 100) direct replies to it. Safeguards have been added to Talk to ensure that this database limitation is not reached, but based on real world comment installations, you are pretty much guaranteed to not hit this limit.


src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -624,11 +640,22 @@ type Comment {
   """
   status: COMMENT_STATUS!
 
+  """
+  parentCount is the number of direct parents for this Comment. Currently this
+  value is the same as depth.
```
alinhle, line 645: When will this ever be different?

src/core/server/models/comment.ts
```diff
@@ -273,6 +303,96 @@ export async function retrieveCommentRepliesConnection(
...
+  if (!comment.parent_id) {
+    return createConnection({
+      pageInfo: {
+        hasNextPage: false,
+        hasPreviousPage: false,
+      },
+    });
+  }
+
+  // TODO: (wyattjoh) maybe throw an error when the limit is zero?
```
alinhle, line 331: I'd throw if all it does is returning an empty connection.

src/core/server/models/comment.ts
```diff
@@ -273,6 +303,96 @@ export async function retrieveCommentRepliesConnection(
...
+  // Fetch the subset of the comment id's that we are going to query for.
+  const parentIDSubset = parentIDs.slice(skip + 1, skip + 1 + limit);
+
+  // Retrieve the parents via the subset list.
+  const parents = await retrieveManyComments(mongo, tenantID, parentIDSubset);
+
+  // Loop over the list to ensure that none of the entries is null (indicating
+  // that there was a misplaced parent). We can assert the type here because we
+  // will throw an error and abort if one of the comments are null.
+  const nodes = parents.filter(parentComment => {
```
alinhle: `forEach` would also do :-)

alinhle (changes requested):

src/core/server/models/comment.ts
wyattjoh: An empty connection is still technically valid, albeit not super useful.

src/core/server/graph/tenant/schema/schema.graphql
wyattjoh: If we decide to change the count based on deleted/rejected/tombstoned comments

alinhle (approved):

PR #1904, opened 2018-09-24
[next] Implement LocalReplyList for last level replies

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/160729795
- Reader can reply to the last level of comments and see only their replies

## How do I test this PR?

- Reply until deepest comment level (6 indentions)
- Deepest level should not contain server replies but only comments from the CreateComment mutation and on refresh the comments should disappear.
- These comments should not have a Reply Button


src/core/client/stream/mutations/CreateCommentMutation.ts
```diff
@@ -102,9 +153,13 @@ function commit(
...
     } as any, // TODO: (cvle) generated types should contain one for the optimistic response.
     optimisticUpdater: store => {
+      sharedUpdater(store, input);
       store.get(id)!.setValue(true, "pending");
     },
-    configs: getConfig(input),
+    updater: store => {
+      sharedUpdater(store, input);
+    },
+    // configs: getConfig(input),
```
wyattjoh: Can you remove this line if if isn't in use?

src/core/client/stream/tabs/comments/components/Indent.tsx
```diff
@@ -20,6 +20,7 @@ const Indent: StatelessComponent<IndentProps> = props => {
           [styles.level3]: props.level === 3,
           [styles.level4]: props.level === 4,
           [styles.level5]: props.level === 5,
+          [styles.level6]: props.level === 6,
```
wyattjoh: Would it not be better to write some kind of getter function instead of this??

src/core/client/stream/mutations/CreateCommentMutation.ts
alinhle: Oh sure, forgot that one

src/core/client/stream/tabs/comments/components/Indent.tsx
alinhle: I'm doing it this way, so that it is typechecked. Do you know a better way that does preserve the type checking?

alinhle: Ok I refactored it, should be better now :-)

wyattjoh (approved):

PR #1910, opened 2018-09-25
[next] First Iteration of Admin

okbel:
## What does this PR do?
- First iteration over the Admin Client
- Implements `found` `found-relay`
- Routing to handle multiple views  -   MODERATE, COMMUNITY, STORIES, CONFIGURE 

## How do I test this PR?
- Go to http://localhost:8080/admin/moderate


config/watcher.ts
```diff
@@ -22,6 +22,9 @@ const config: Config = {
         "core/client/auth/**/*.ts",
         "core/client/auth/**/*.tsx",
         "core/client/auth/**/*.graphql",
+        "core/client/admin/**/*.ts",
+        "core/client/admin/**/*.tsx",
+        "core/client/admin/**/*.graphql",
```
alinhle: Not sure how the `auth` lines got here, but `npm run compile:relay-stream` does not touch these files. We can delete lines 22-27.

src/core/server/app/router.ts
```diff
@@ -49,6 +50,9 @@ async function createTenantRouter(app: AppOptions, options: RouterOptions) {
   // Setup auth routes.
   router.use("/auth", createNewAuthRouter(app, options));
 
+  // Handle the admin handler.
+  router.get("/admin", adminHandler);
```
alinhle: I let @wyattjoh the server part :-)

alinhle (changes requested):
As briefly talked about. In the `admin` target we have full control over the location as we are not in an iframe nor in a popup and so instead of writing our own view router, we can rely on much more powerful libraries that helps us manage the various routes of the admin. One that comes with Relay modern integration would be https://github.com/relay-tools/found-relay

wyattjoh (approved):
Server side looks good to me, I cleaned up a few things!

src/types/farce.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "farce" {
```
alinhle, line 1: Let's add a todo comment here `// TODO: provide better typings.`

src/core/client/admin/components/App.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { StatelessComponent } from "react";
+import { HorizontalGutter } from "talk-ui/components";
+import styles from "./App.css";
+import Navigation from "./Navigation";
+
+const App: StatelessComponent = ({ children }) => (
+  <div className={styles.root}>
+    <HorizontalGutter>
```
alinhle: `HorizontalGutter` can be used as root and we would save one unnecessary `div`, e.g. `<HorizontalGutter className={styles.root}>`

src/core/client/admin/components/Community.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { StatelessComponent } from "react";
+import { HorizontalGutter, Typography } from "talk-ui/components";
+
+const Community: StatelessComponent = ({ children }) => (
+  <HorizontalGutter>
```
cvle: This container seems obsolete, this is repeated in other files too.

alinhle (changes requested):

alinhle (approved):

PR #1912, opened 2018-09-26
[next] SSO Refactor

wyattjoh:
## What does this PR do?

Refactors the SSO interface to allow seamless signin with SSO credentials without performing an exchange.


alinhle (approved):
Tests would be great to know if it really works, but implementation looks good! :-D

Issue #1927, opened 2018-09-28
"Copy" button on Share feature  is not copying the permalink


kgardnr:
#### How to reproduce the issue:

1. Go to an existing comment
2. Click on Share
3. Click "Copy" to copy the permalink
4. Try and paste the link somewhere and it will not paste, it's not being copied to the clipboard

moogacs:
@kgardnr , where can i got to existing comment ?

jqueryisamonad:
@kgardnr this issue's reproducible only on some browsers? All work fine on Chrome 69

kgardnr:
@hovoodd & @moogacs yeah this is so weird, I swear I was able to reproduce this before but now I can't. Might be a moot point then!

kgardnr:
Cannot reproduce this - ack!

mkonikov:
Hey @kgardnr - I encounter this consistently on iOS. Maybe we can reopen and investigate there?
Possibly related to other click event handler issues that previously were reported and fixed on iPad?

kgardnr:
Thanks @mkonikov! Is there any chance you could take a stab submitting a fix for this?

mkonikov:
I'll bring it up at our next grooming meeting and see if we can fit in an upcoming sprint. Thanks @kgardnr!

mkonikov:
With #2159 merged in, I think we can close.

PR #1953, opened 2018-09-30
Fixes permlink copying #1927

moogacs:
## What does this PR do?
fixing copying to permalink
## How do I test this PR?
click the button

### DIFF
--- a/plugins/talk-plugin-permalink/client/components/PermalinkButton.js
+++ b/plugins/talk-plugin-permalink/client/components/PermalinkButton.js
@@ -36,7 +36,9 @@ export default class PermalinkButton extends React.Component {
   };
 
   copyPermalink = () => {
+    this.permalinkInput.focus();
     this.permalinkInput.select();
+    this.permalinkInput.createTextRange();
     try {
       document.execCommand('copy');
       this.setState({


kgardnr:
Thanks for this @moogacs!

moogacs:
@kgardnr it's still open ? or you have merged and closed it ?

wyattjoh:
We have yet to review it @moogacs, I'll take a review shortly!

plugins/talk-plugin-permalink/client/components/PermalinkButton.js
```diff
@@ -36,7 +36,9 @@ export default class PermalinkButton extends React.Component {
   };
 
   copyPermalink = () => {
+    this.permalinkInput.focus();
     this.permalinkInput.select();
+    this.permalinkInput.createTextRange();
```
wyattjoh, line 41: It seems that this is not a valid method for input elements: https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement

wyattjoh (changes requested):
From tests conducted locally, the existing implementation works in Firefox, Chrome, and Safari.

What browsers does this version work on that the current implementation doesn't?

plugins/talk-plugin-permalink/client/components/PermalinkButton.js
moogacs: So you mean before my PR it was fine ? 
it suppose an open task !

kgardnr:
Ok so strangely I'm not able to replicate this issue somehow - I can't determine if it was fixed or not. Apologies for that @moogacs - I'm just going to close this - but let me know if you see it happen again!

PR #1957, opened 2018-10-02
[next] Install

wyattjoh:
## What does this PR do?

Adds support for installing Talk.

### New Routes

Adds the new `POST /api/tenant/install` route:

```bash
curl --request POST \
  --url http://localhost:3000/api/tenant/install \
  --header 'content-type: application/json' \
  --data '{
        "tenant": {
                "organizationName": "Awesome News",
                "organizationContactEmail": "contact@awesomenews.net",
                "organizationURL": "https://awesomenews.net/",
                "domains": [
                        "https://awesomenews.net/"
                ]
        },
        "user": {
                "username": "admin",
                "password": "admin",
                "email": "admin@awesomenews.net"
        }
}'
```

Which will respond with a `204` if successful.

## Test
0. Delete your database
1. Go to http://localhost:8080/install.html
2. Install Talk


okbel:
@cvle @wyattjoh This one is ready to be tested!

okbel:
lint failing under: `src/core/server/app/router.ts(8,31): error TS2307: Cannot find module 'talk-server/app/handlers/embed/stream'.` @wyattjoh

okbel:
Let's update this PR to create the settings for Reactions too. Since a clean install triggers: `Cannot return null for non-nullable field Settings.reaction.` @wyattjoh

okbel:
Ready @wyattjoh and @cvle ! Thanks!

src/core/client/install/steps/containers/PermittedDomainsContainer.tsx
```diff
@@ -0,0 +1,79 @@
...
+  private onGoToPreviousStep = () => {
+    if (this.props.onGoToPreviousStep) {
+      this.props.onGoToPreviousStep();
+    }
+  };
+  private onSubmit: PermittedDomainsForm["onSubmit"] = async (input, form) => {
+    try {
+      const domains = input.domains.split(",");
+      this.props.onSaveData({ domains });
+      this.props.install(shapeFinalData(this.props.data));
```
wyattjoh: Because the data is being set in the state, and the `onSaveData` simply calls `setState` on the outer container, `this.props.data` does not actually have the data including the `domains`, so when you observe the last button on finishing the form, the `domains` field is never set.

okbel: This is a race condition. I added an await since `setState` is not syncronous. Please check. Thanks!

alinhle (approved):

Issue #1920, opened 2018-09-28
Change text in Moderation View > Configure 


kgardnr:
Change "Include Comment Stream Description For Readers" in the configure menu to "Summary of your Code of Conduct"

Description text change to "This message will appear above every comments stream on your site. Click here to learn more about writing a good code."

Leodau:
I'll take this one please! :)

PR #1963, opened 2018-10-04
Update text and description in moderation configure view

twosmalltrees:
## What does this PR do?

Issue #1920 

## How do I test this PR?

- Navigate to Moderation View > Configure
- Check it all looks 👍


twosmalltrees:
One question - the description provided in the original issue (#1920) includes the text  "...Click here to learn more about writing a good code". Is the intent to provide a link here?

kgardnr:
Yes! @twosmalltrees - it should link here: https://guides.coralproject.net/create-a-code-of-conduct/

Thank you!

kgardnr (changes requested):
Needs link added - posted link in previous comment - sorry for leaving that out of the original description!

kgardnr:
Hey @twosmalltrees did my previous response make sense? Let me know if you have any questions! Thank you again 😸

twosmalltrees:
Hi @kgardnr - yep all made sense, sorry about the slow follow up! Have just pushed though changes with the link included. Let me know what you think 😄

kgardnr:
Thanks @twosmalltrees! When running the config page, for some reason the link doesn't seem to be clickable - are you seeing something different?

twosmalltrees:
Hmmm hadn't noticed that. What I'm seeing is that the link is clickable if the Card is enabled (ie the checkbox is selected), however not clickable when unselected. Looks like this is caused by a `pointer-events: none` that's being applied to disabled ConfigureCard components. I'm guessing this has been applied to prevent interaction with the dropdowns/inputs on other configuration cards when they're disabled. Is this the same behaviour you're seeing on your end?

Will have to take another look at this - I can see two approaches:

1. The link should be clickable at all times
2. The link should only be clickable when the card is enabled, but should be greyed out (so as to not appear so clickable) when the card is disabled.

kgardnr:
Ohhh I see - okay that makes sense! 

Let me defer to @cvle if he has any reasons why we would need to keep those links disabled if the card is disabled? Because that would be solution number 3 I would guess.

Thanks @twosmalltrees!

locales/de.yml
```diff
@@ -132,8 +132,8 @@ de:
     enable_questionbox: 'Stellen Sie den Lesern eine Frage'
     enable_questionbox_description: 'Diese Frage erscheint am Anfang des Kommentarbereichs. Regen Sie eine Diskussion an.'
     hours: Stunden
-    include_comment_stream: 'Einleitung zum Kommentarbereich'
-    include_comment_stream_desc: 'Verfassen Sie eine Einleitung, die über jedem Kommentarbereich erscheint. Nützlich z.B. für Community-Richtlinien.'
+    code_of_conduct_summary: 'Einleitung zum Kommentarbereich'
```
alinhle, line 135: As we don't have translations for these new translation keys yet, it's better to remove them from all the other languages rather than having an inaccurate translation, so that a missing translation can be spotted.

alinhle (dismissed):
Thanks @twosmalltrees! I did a review with a small remark, if you could address that, it'd awesome 👍 

@twosmalltrees the link being not clickable is a separate issue, that I'll address in a different PR later today. We can ignore it for this PR. Thanks for your work @twosmalltrees!

alinhle:
This fixes the pointer events issue: https://github.com/coralproject/talk/pull/2035

kgardnr (approved):

twosmalltrees:
Excellent thanks @kgardnr and @cvle 😄

PR #1965, opened 2018-10-04
[next] Relay GraphQL Batching

wyattjoh:
## What does this PR do?

Adds GraphQL batching support that is compatible with the client side relay transport: https://github.com/relay-tools/react-relay-network-layer/blob/master/src/express-middleware/graphqlBatchHTTPWrapper.js 

## How do I test this PR?

Submit a GraphQL query:

```json
[{"query": "{ me { id } }", "id": "1"}]
```

And see the response:

```json
[{"id": "1", "payload": {"data": {"me": null}}}]
```


alinhle:
@wyattjoh The user information seems to be lost in this PR, can you take a look?

wyattjoh:
Fixed @cvle!

src/core/server/app/middleware/graphqlBatch.ts
```diff
@@ -0,0 +1,97 @@
...
+function wrapResponse(req: Request, res: Response) {
+  // If the request is not an array, or has no elements, we should skip it.
+  if (!Array.isArray(req.body) || req.body.length === 0) {
+    return res;
+  }
+
+  // If the request is an array, but it does not have an ID field, then we
+  // should skip it.
+  const needsUpgrade = Boolean(typeof req.body[0].id !== "undefined");
+  if (!needsUpgrade) {
```
alinhle, line 15: Upgrade usually means upgrading the connection to a websocket, maybe a more generic term like `skip` will do here too?

alinhle: Not that this is really important xD

src/core/client/framework/lib/network/createNetwork.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  authMiddleware,
+  batchMiddleware,
+  cacheMiddleware,
+  RelayNetworkLayer,
+  retryMiddleware,
+  urlMiddleware,
+} from "react-relay-network-modern/es";
```
wyattjoh, line 8: Why the `/es`?

src/core/client/stream/tabs/profile/queries/ProfileQuery.tsx
```diff
@@ -64,6 +64,10 @@ const ProfileQuery: StatelessComponent<InnerProps> = ({
       assetID,
       assetURL,
     }}
+    cacheConfig={{
+      // TODO: enable caching after mutations are adapted.
+      force: true,
+    }}
```
wyattjoh, line 70: What's this about?

src/core/server/app/middleware/graphqlBatch.ts
wyattjoh: "needsUpgrade" to a relay batch connection :P the term "upgrade" isn't reserved here.

src/types/react-relay-network-modern.d.ts
```diff
@@ -0,0 +1,311 @@
+/* tslint:disable */
+
+declare module "react-relay-network-modern/es" {
```
wyattjoh, line 5: Can we consider submitting these to @types? With a `TODO` to replace them in the future?

src/core/client/framework/lib/network/createNetwork.ts
alinhle: `es` means it contains es modules and have not been compiled down to  `commonjs` => allows tree shaking to do its work.

src/types/react-relay-network-modern.d.ts
alinhle: Of course

src/core/client/stream/tabs/profile/queries/ProfileQuery.tsx
alinhle: If we adapt the `CreateCommentMutation` to also insert into the my profile list, we can activate caching here. We should however first integrate pagination for the my comments list first, before we can adapt them appropriately.

alinhle (approved):

PR #1969, opened 2018-10-04
Body Count

wyattjoh:
## What does this PR do?

Adjusts the behavior when a comment exceeds the maximum length, preventing it from being committed at all.

## How do I test this PR?

1. In the admin (/admin/configure/stream), ensure that the `Limit Comment Length` is unchecked, and saved.
2. In another browser, open the stream (/dev), and login as another user.
3. Back in the admin browser/tab, check the box for comment length, setting it to `5`
4. In the user tab, submit the comment `123456`, and see the error on the stream page indicating the comment length was exceeded.
5. In the admin tab, notice that the comment does not appear int he moderation screen.

### DIFF
--- a/errors.js
+++ b/errors.js
@@ -343,6 +343,20 @@ class ErrCommentTooShort extends TalkError {
   }
 }
 
+// ErrCommentTooLong is returned when the comment is too long.
+class ErrCommentTooLong extends TalkError {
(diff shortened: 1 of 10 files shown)


okbel (approved):
Perfect 👍

PR #1972, opened 2018-10-05
[next] Content Security Policy

wyattjoh:
## What does this PR do?

Adds support for the [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) header.


src/core/server/app/url.ts
```diff
@@ -10,3 +10,39 @@ export function reconstructURL(req: Request, path: string = "/"): string {
...
+  }
+
+  // If the referer header is defined, then use it.
+  if (req.headers.referer && req.headers.referer.length > 0) {
+    // If the header contains multiple values, return the first one.
+    if (Array.isArray(req.headers.referer)) {
+      return req.headers.referer[0];
+    }
+
+    return req.headers.referer;
```
alinhle: Here you return the whole referer instead of just the hostname

src/core/server/app/middleware/csp/tenant.ts
```diff
@@ -0,0 +1,67 @@
...
+  if (!parentsHostname) {
+    return "deny";
+  }
+
+  // As we can only return a single domain in the `allow-from` directive as per
+  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options
+  // We need to find the domain that is asking so we can respond with the right
+  // result, sort of like CORS!
+  const allowFrom = tenant.domains.find(
+    domain => new URL(domain).hostname === parentsHostname
```
alinhle: Unfortunately `new URL("localhost:8080").hostname` returns `""`. Shouldn't we compare also the port (and schema)?

alinhle (changes requested):
Great Job! I have only some comments. Maybe tests would help stabilize these bits here :-)

src/core/server/app/middleware/csp/tenant.ts
wyattjoh: In the future, I think the fact that we'll have the nice url fields, we'll not have to worry about this case, as all the url's will include the scheme.

alinhle (approved):

PR #1985, opened 2018-10-10
CORS

freder:
## What does this PR do?
enable CORS for whitelisted origins, by adding `cors-allowed-origins.js` to the project root dir.

example:

```
module.exports = [
  'http://localhost:3000',
  'http://127.0.0.1:3000',
];
```

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,4 +1,5 @@
 const express = require('express');
+const cors = require('cors');
 const nunjucks = require('nunjucks');
 const cons = require('consolidate');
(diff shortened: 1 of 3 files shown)


wyattjoh:
Thanks for the contribution @freder!

Unfortunately, we've decided not to support CORS for various security related reasons, and would instead suggest applying CORS headers (if you must) at the network layer (such as nginx/caddy/haproxy) instead of applying it globally across Talk.

PR #1987, opened 2018-10-11
[next] Permalink View

alinhle:
## What does this PR do?
- Implement Permalink View including support for unlimited threading and parent context

## How do I test this PR?

- post unlimited comments and follow the conversation :-)


src/core/client/stream/tabs/comments/components/ConversationThread.tsx
```diff
@@ -63,7 +66,7 @@ const ConversationThread: StatelessComponent<
                   Show hidden comments
                 </Button>
               </Localized>
-              <Counter>{props.remaining}</Counter>
+              {props.remaining > 1 && <Counter>{props.remaining}</Counter>}
```
wyattjoh, line 82: Is this something that should be included in the `Localized` wrapper?

alinhle: Been thinking about that too, besides putting things ltr vs rtl was the only thing I could think of, but that should be addressed in a broader context.

src/core/client/stream/mutations/SetCommentIDMutation.ts
```diff
@@ -19,6 +20,16 @@ export async function commit(
     const record = store.get(LOCAL_ID)!;
     record.setValue(input.id, "commentID");
     record.setValue("COMMENTS", "activeTab");
+
+    // Change iframe url, this is important
+    // because it is used to cleanly initialized
+    // a user session.
+    window.history.replaceState(
```
wyattjoh, line 27: Why are you using `replaceState` instead of `pushState`?

alinhle: Intentional, because `pushState` would change where the browser would go to when clicking `back`, but in this case we are in the iFrame and we don't want `back` to change anything inside of the iframe. The reason we are adapting the url is because, when a new user session is created, we freshly initialise our local state which looks into the url for e.g. the commentID.

In the `embed` we have the equivalent code that changes the history using `pushState`, in this case we want `back` to change the parent url.

wyattjoh (approved):

okbel (approved):
Amazing work 👍

PR #1992, opened 2018-10-12
[next] i18n short number support

alinhle:
## What does this PR do?
- Add custom builtin function `SHORT_NUMBER` to fluent
- Implementation based on Unicode Short Numbers: http://cldr.unicode.org/translation/number-patterns#TOC-Short-Numbers

## How do I test this PR?
- a. Post thousands of comments
- b. or just manipulate the number in `CommentCountTab.tsx` ;-)


babel.config.js
```diff
@@ -0,0 +1,17 @@
...
+module.exports = {
+  env: {
+    test: {
+      presets: [
+        ["@babel/env", { targets: "last 2 versions, ie 11", modules: false }],
+        "@babel/react",
+      ],
+    },
+  },
+};
```
wyattjoh, line 17: Is it a hard ask for you to move this file to a non-root location? Possibly inside `src/core/build`?

alinhle: Yes, there is literally no options to configure this, besides this is only needed for testing and I think shoudn't be in `src/core/build`

wyattjoh (approved):

PR #1994, opened 2018-10-15
[next] Static URI Support

wyattjoh:
## What does this PR do?

Adds support for a new `STATIC_URI` configuration option that can be used when you want to serve Talk along with a CDN capable of caching static assets.

## How do I test this PR?

1. Build the static assets via `npm run build`
2. Use a tool like ngrok to create a new url that isn't on localhost.
3. Copy the URL generated for ngrok to your `.env` file as `STATIC_URI=https://<ngrok url`.
4. View the Talk stream, notice that the assets are properly prefixed with the `{{ staticURI }}` template variable.


src/core/build/createWebpackConfig.ts
```diff
@@ -495,6 +500,9 @@ export default function createWebpackConfig({
           inject: "body",
           ...htmlWebpackConfig,
         }),
+        // Inject the pieces we need here to resolve all the now relative url's
+        // against the CDN if it's provided.
+        new CDNWebpackPlugin({ production: isProduction }),
```
alinhle: You could also have used: `...(isProduction ? [new CDNWebpackPlugin()] : [])` but you don't have to, if this looks ugly to you lol or how about a helper that is something like`...onlyInProduction( new CDNWebpackPlugin())`

It's more clear to me than `new CDNWebpackPlugin({ production: isProduction }),`

src/core/build/createWebpackConfig.ts
```diff
@@ -438,25 +439,29 @@ export default function createWebpackConfig({
...
           paths.appPolyfill,
+          paths.appPublicPath,
           ...devServerEntries,
           paths.appAuthIndex,
           // Remove deactivated entries.
         ],
         install: [
           // We ship polyfills by default
           paths.appPolyfill,
+          paths.appPublicPath,
```
alinhle: As talked in our meeting: Should only be applied in production

alinhle (changes requested):

alinhle (approved):

PR #2007, opened 2018-10-18
Base64 Secret Encoding

wyattjoh:
## What does this PR do?

Added base64 encoding option for symmetric secrets.

## How do I test this PR?

1. Set `TALK_JWT_SECRET=base64:dGVzdA==` as your secret.
2. Start the server, try and login and do a few actions.
3. Stop the server.
4. Change the `TALK_JWT_SECRET` to `test` as `TALK_JWT_SECRET=test`
5. Start the server, and try to do a few more actions, notice that the old token still works!

This is because the token that was issued before, is actually signed with the same secret, the first time however, encoded in base64!

### DIFF
--- a/services/jwt.js
+++ b/services/jwt.js
@@ -122,7 +122,7 @@ function SharedSecret({ kid = undefined, secret = null }, algorithm) {
 
   // If the secret is base64 encoded, then decode it!
   if (secret.startsWith('base64:')) {
-    secret = Buffer.from(secret.substring(6), 'base64').toString();
+    secret = Buffer.from(secret.substring(7), 'base64').toString();
   }
 
   return new Secret({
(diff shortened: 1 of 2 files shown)


services/jwt.js
```diff
@@ -120,6 +120,11 @@ function SharedSecret({ kid = undefined, secret = null }, algorithm) {
     throw new Error('Secret cannot have a zero length');
   }
 
+  // If the secret is base64 encoded, then decode it!
+  if (secret.startsWith('base64:')) {
+    secret = Buffer.from(secret.substring(6), 'base64').toString();
```
alinhle: image

Shouldn't it be `secret.substring(7)`?

alinhle (changes requested):

services/jwt.js
wyattjoh: ... yes

alinhle (approved):

PR #2011, opened 2018-10-19
Email Prefix Fix

### LINKED ISSUE #2009: Unable to remove email subject prefix
Bug

#### Intended outcome:
Set an empty value for `TALK_EMAIL_SUBJECT_PREFIX` so nothing is prepended to email subjects.

#### Actual outcome:
If the var has no value, the default prefix is used (`[Talk]`). If quotes are used to denote an empty string, the quotes end up in the subject.

Set the `TALK_EMAIL_SUBJECT_PREFIX` env var to each value and then trigger the application to send an email, like suspending a user:
* `TALK_EMAIL_SUBJECT_PREFIX=`
* `TALK_EMAIL_SUBJECT_PREFIX=''`
* `TALK_EMAIL_SUBJECT_PREFIX=""`

screen shot 2018-10-19 at 8 48 02 am

#### Version and environment
4.6.4, but we will be upgrading to 4.6.5 soon.

wyattjoh:
## What does this PR do?

Fixes #2009

## How do I test this PR?

1. Set `TALK_EMAIL_SUBJECT_PREFIX=""`
2. Do something that sends an email
3. See no prefix!

### DIFF
--- a/config.js
+++ b/config.js
@@ -30,7 +30,7 @@ const CONFIG = {
   ENABLE_TRACING: Boolean(process.env.APOLLO_ENGINE_KEY),
 
   // EMAIL_SUBJECT_PREFIX is the string before emails in the subject.
-  EMAIL_SUBJECT_PREFIX: process.env.TALK_EMAIL_SUBJECT_PREFIX || '[Talk]',
+  EMAIL_SUBJECT_PREFIX: process.env.TALK_EMAIL_SUBJECT_PREFIX,
 
   // DEFAULT_LANG is the default language used for server sent emails and
   // rendered text.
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -271,6 +271,10 @@ const CONFIG = {
 // CONFIG VALIDATION
 //==============================================================================
 
+if (typeof CONFIG.EMAIL_SUBJECT_PREFIX === 'undefined') {
```
alinhle, line 274: Have you thought about using `if (CONFIG.EMAIL_SUBJECT_PREFIX === undefined)` instead?

wyattjoh: There are some issues with using `var === undefined` in that if the variable is undeclared, it would throw an error on the compare. I'd rather stick to the `typeof` method to keep things consistent.

alinhle (approved):

PR #2016, opened 2018-10-20
Fixed stream status updating

### LINKED ISSUE #1952: Changing Stream Status in Admin > Stories should update the UI live so it's not confusing to the user

Bug

#### Intended outcome:
When you Open or Close stories in a filtered view in the Admin > Stories page, it's not clear that the status has indeed been updated, because the UI doesn't refresh to show this. Gif attached to show how this is currently working.

#### Actual outcome:
It looks like the status was not changed - even though it is updating the status correctly.

Filter Closed Stories in Admin > Stories list. Change Closed to Open. Notice the UI doesn't update. But if you search again for that Story, it has been updated correctly to Open status.

#### Version and environment
4.6.2 (reproduced also in 4.6.3)

opening_streams_strib

jqueryisamonad:
## What does this PR do?
Fixes #1952 

I can add optimistic update for redux store, if necessary.

## How do I test this PR?

- read issue description
- see the cat meow

### DIFF
--- a/client/coral-admin/src/routes/Stories/containers/Stories.js
+++ b/client/coral-admin/src/routes/Stories/containers/Stories.js
@@ -50,7 +50,7 @@ class StoriesContainer extends Component {
     const { updateAssetState } = this.props;
 
     try {
-      updateAssetState(id, closeStream ? Date.now() : null);
+      await updateAssetState(id, closeStream ? Date.now() : null);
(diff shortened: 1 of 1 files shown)


kgardnr:
Thank you @hovoodd! 🎉

kgardnr (approved):

PR #2035, opened 2018-10-25
Fix pointer events in configure card

alinhle:
## What does this PR do?
- Allow pointer events in configure card even when disabled

## How do I test this PR?

- Stream Settings Configure Cards should allow pointer events when disabled meanwhile input elements inside should be disabled too.


kgardnr (approved):
Thank you @cvle! 🤗

Issue #2027, opened 2018-10-24
[Feature] 'Select all user's comments' in admin 


aguilarm:
#### Do you want to request a **feature** or report a **bug**?

Bug? Incomplete feature perhaps.

#### Intended outcome:

By clicking 'select all' in Talk admin for a user's comments, admins would like to easily reject _all_ comments for a particular user (generally spammers, with hundreds of comments).

#### Actual outcome:

Bulk approving/rejecting comments (original issue https://github.com/coralproject/talk/issues/1027) only selects 10 comments at a time, or however many have been loaded on the admin sidebar. This makes removing a user's comments with anything more than 100 or so is an unnecessarily manual process.

#### How to reproduce the issue:

In admin, click select all on a users comments (original issue has screenshots).

#### Version and environment

4.6.2

Perhaps it should operate like gmail, where initially clicking 'select all' selects showing comments, and prompts to further select _all_ comments, like 

> All 10 comments on this page are selected. Select all 360 comments by $USERNAME

aguilarm:
Looked at what's happening now a bit.  It's iterating individual comments and triggering a setStatus on each one then waiting for all of those changes to resolve.

I think perhaps we could make a new mutation like `setCommentStatusForUser` or something that takes a query shape like `{ author_id, status }` and updates all comments by a particular user to match a new status. That query will use the existing index on author_id/created_at and looks to be totally fine at scale.

kgardnr:
Hey @aguilarm this is a cool idea, if you wanted to contribute something like this!

aguilarm:
@kgardnr Cool, I'll have a go at a PR!

aguilarm:
Turns out changing the status of a comment triggers some side effect operations including:
- adjusting reply counts for the parent comment, if it exists
- clearing comment count caches for associated assets so they can be refreshed
- adjusting user karma

While I can do the asset cache clear and karma all at once in separate post-update actions based on total changes, reply counts check the pre- and post-change visibility state of the affected comments.

There is no way (in mongo, that I know of) to atomically run a find and update on multiple documents at once (so I can figure out if the change should trigger an increment or decrement on the parent reply count) meaning I still have to iterate one at a time to make sure the parent is updated accurately.

That might take some time with large numbers of comments, so it's probably prudent to run it in chunks and display chunked progress to the user.  Perhaps update 50 comments each chunk and write feedback/progress to the UI up top somewhere?

PR #2038, opened 2018-10-26
'Select all user's comments' in admin #2027

aguilarm:
## What does this PR do?

Adds #2027 - gmail-like behavior for selecting _all_ of a user's comments in the user detail view of admin. 

## How do I test this PR?

- In admin, select a user
- Click 'Select All'
- A new button appears, 'Select all X comments for USER'
image
- Click new button, button switches to:
image
- Clicking accept/reject applies new status to _all_ of user's comments

I have one problem, hopefully with some input I can solve it:
- ~~I created a new set of comments to select in the gql query for the UserDetail container that only grabs comment id's and statuses. This then loads all of a user's comments to apply an update. I'm not sure however how to reset it to the first page after running the fetchMore a few times so right now this basically only works once. Perhaps my whole approach here is off.~~ Turns out it I missed default arguments for comment status.

And a couple TODOS:
- ~~I should probably make a fragment to share between the original query and load more query~~
- ~~Once 'Select all by user' is selected, Load More should be disabled until the selection is cleared. Further, once an action is clicked (approve/reject) when 'all by user' is selected, the list should be locked or replaced with a loading indicator, ideally with progress and a cancel button because this will take at least a few seconds.~~ Switched so the element switches to loading state while the batch is running.


kgardnr:
@IAmSamHankins here's a gif of this functionality for your approval!

bulk_gmail_esque

@cvle / @wyattjoh take a look and let us know if it's good code-wise. 

Thanks!

kgardnr (approved):

aguilarm:
I'm pretty sure I found why this was sometimes bugging out and only working once.

It's the CommentsQuery statuses default value being [NONE, ACCEPTED], so if you rejected comments - since my query didn't set a value for statuses - it was not selecting rejected comments to later approve.

Pushed a fix to set statuses to null explicitly and everything looks all set now.

I have it doing 5 comments at a time just for ease of testing with a smaller dataset, but may want to switch that to something like 50 or 100.

aguilarm:
I switched to a shared fragment, load one comment to start and adjusted the # per page to 50.

Still could probably be explicitly setting the status to either [NONE,ACCEPTED] (default), or [Rejected] so I'm not needlessly iterating the entire set every time but usually that's only iterating a handful of extra comments tops - moderation has generally been all-or-nothing when possible so far.

PR #2079, opened 2018-11-16
[next] UserBox with respect to auth settings

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/162022146
- If only SSO is activate, don't show the Login nor SignUp button.
- Only show logout when JTI is set in the token.
- Only show register button when `allowRegistration` is set in one of the integrations (except sso)

## How do I test this PR?

- Change settings (e.g. only enable SSO)
- Go to Stream

### DIFF
--- a/src/core/client/auth/local/local.graphql
+++ b/src/core/client/auth/local/local.graphql
@@ -6,6 +6,10 @@ enum View {
 }
 
 type Local {
+  authToken: String
+  authExp: Int
+  authJTI: String
+  loggedIn: Boolean!
   view: View!
 }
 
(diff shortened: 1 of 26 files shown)


src/core/client/auth/local/local.graphql
```diff
@@ -6,6 +6,10 @@ enum View {
 }
 
 type Local {
+  authToken: String
+  authExp: Int
+  authJTI: String
```
wyattjoh, line 11: Where does this get populated?

alinhle: Here: https://github.com/coralproject/talk/blob/next/src/core/client/framework/lib/relay/localState.ts

wyattjoh (approved):

PR #2084, opened 2018-11-19
[next] Prevent accidental lock out from admin or stream

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/162023050
- Prevent disabling all integrations from TalkAdmin using form validation
- Confirm disabling all integrations from Comment Stream before submitting

## How do I test this PR?

- Try disabling all integrations for Talk Admin
- Try disabling all integrations for Comment Stream


src/core/client/admin/routes/configure/components/Configure.tsx
```diff
@@ -19,9 +19,9 @@ const Configure: StatelessComponent<Props> = ({
   onChange,
   children,
 }) => (
-  <div id="configure-container">
+  <div data-test="configure-container">
```
wyattjoh, line 22: Why the change here? Are the `data-test` props stripped on production builds?

alinhle: Not yet, but yes possibly! I saw some projects using `data-test` and it made sense. Using `id` to find a particular component for a test was not reliable, consider this:
```ts
<Localized id="configure-auth-button">
  <Button id="configure-auth-button">Click me </Button>
</Localized>
```
We have this pattern quiet frequently and it made it impossible to find the right component only using `findProps({id})`. Introducing a separate prop like `data-test` allows me to precisely pick the component I want easily, with the bonus that a simple babel plugin can remove this prop entirely for production builds!

alinhle: Sounds promising: https://www.npmjs.com/package/babel-plugin-transform-remove-props

wyattjoh (approved):

PR #2101, opened 2018-11-28
[next] Auth Popup v2

alinhle:
## What does this PR do?
- https://www.pivotaltracker.com/story/show/161655175
- New design + Social Login


scripts/start.ts
```diff
@@ -52,45 +51,30 @@ if (process.env.HOST) {
...
-    const devServer = new WebpackDevServer(compiler, serverConfig);
-    // Launch WebpackDevServer.
-    devServer.listen(port, HOST, (err: Error) => {
-      if (err) {
-        return console.log(err);
-      }
-      console.log(chalk.cyan("Starting the development server...\n"));
-    });
+const protocol = "http";
+const appName = "Talk";
```
wyattjoh: As these are "thankfully" rarely to change, I'd suggest just in-lining these variables instead.

src/core/client/auth/local/initLocalState.ts
```diff
@@ -1,28 +1,33 @@
...
  * Initializes the local state, before we start the App.
  */
-export default async function initLocalState(environment: Environment) {
-  commitLocalUpdate(environment, s => {
-    const root = s.getRoot();
+export default async function initLocalState(
+  environment: Environment,
+  context: TalkContext
+) {
+  const authToken = window.location.hash
```
wyattjoh, line 8: Why not extract this to a separate function?

alinhle: Sure

scripts/start.ts
alinhle: Sure

wyattjoh (approved):

PR #2109, opened 2018-12-06
Support lazy rendering

alinhle:
## What does this PR do?
- Support lazy rendering using IntersectionObserver.

## How do I test this PR?
- Pass `lazy: true` to the render options e.g.
```js
Coral.Talk.render(document.getElementById('container'), {
    talk: 'https://my-talk-installation.com',
    lazy: true,
});
```
- You can enable lazy rendering by default using ENV variable `TALK_DEFAULT_LAZY_RENDER=TRUE`


client/coral-embed/src/index.js
```diff
@@ -2,6 +2,9 @@ import URLSearchParams from 'url-search-params';
 import Stream from './Stream';
 import StreamInterface from './StreamInterface';
 
+// Require the polyfill before requiring any other modules.
+import 'intersection-observer';
```
wyattjoh: Any way we can import this lazily (:)) only if the lazy option is enabled?

client/coral-embed/src/Stream.js
```diff
@@ -50,15 +51,46 @@ export default class Stream {
...
-      id: `${el.id}_iframe`,
-      name: `${el.id}_iframe`,
-    });
     this.snackBar = new Snackbar(snackBarStyles || {});
+    this.emitter = new EventEmitter({ wildcard: true });
+    // Attach to the events emitted by the pym parent.
+    if (events) {
+      events(this.emitter);
+    }
+    if (config.lazy || process.env.TALK_DEFAULT_LAZY_RENDER === 'true') {
```
wyattjoh: To keep this consistent with v4, can you replace `true` with `TRUE`?

wyattjoh (approved):

PR #2110, opened 2018-12-06
Lazy Load Fix

kgardnr:
Fixes a bug with the lazy load polyfill, version bump.


client/coral-embed/src/Stream.js
```diff
@@ -1,3 +1,5 @@
+/* global __webpack_public_path__ */ // eslint-disable-line no-unused-vars
```
alinhle, line 1: What's this line for?

wyattjoh: Eslint complains that it’s undefined, the `__${prop}__` is global.

alinhle (approved):

PR #2116, opened 2018-12-11
fix bug in cli-setup

immber:
## What does this PR do?
Fixes a bug that was found in cli-setup where setup was exiting even if previous settings did not exist

## How do I test this PR?

wait for CI to validate, some tests were commented out


immber (commented):
👍

wyattjoh (approved):

PR #2117, opened 2018-12-12
[next] Resolved Race Conditions

wyattjoh:
## What does this PR do?

Resolves the race conditions around creating assets and getting moderation counts.

## How do I test this PR?

Perform a graph query that requests the story twice (a new story URL), and notice the lack of error!


alinhle (approved):
Awesome, thanks wyatt!

PR #2125, opened 2018-12-14
[next] User Input Validation

wyattjoh:
## What does this PR do?

Adds validation for email, username, and password.

## How do I test this PR?

1. Try create a user with characters other than `[a-zA-Z0-9_]+`
2. Try to create a user with an email not matching `a@a.a`
3. Try to create a user with a password less than 8 characters


src/core/server/services/users/index.ts
```diff
@@ -11,14 +12,105 @@ import {
...
+function validateUsername(tenant: Tenant, username: string) {
+  // TODO: replace these static regex/length with database options in the Tenant eventually
+
+  if (!USERNAME_REGEX.test(username)) {
+    throw new Error("username contained illegal characters");
+  }
+
+  if (username.length > USERNAME_MAX_LENGTH) {
+    throw new Error("username exceeded maximum length");
+  }
```
alinhle, line 39: Should we also check for MIN_LENGTH?

src/core/server/services/users/index.ts
```diff
@@ -11,14 +12,105 @@ import {
...
+ * @param password the password to be tested
+ */
+function validatePassword(tenant: Tenant, password: string) {
+  // TODO: replace these static length with database options in the Tenant eventually
+  if (password.length < 8) {
+    throw new Error("password is too short");
+  }
+}
+
+const EMAIL_REGEX = new RegExp(/^\S+@\S+.\S+$/);
```
alinhle: The client currently uses a different REGEX: see `src/core/client/framework/lib/validation.tsx`. Should we maybe expose the MIN + MAX_LENGTH constants and the REGEXs in `talk-common` and reuse them in both? At least until we get to better error handling.

alinhle (changes requested):

alinhle (approved):

PR #2128, opened 2018-12-16
fix minor ui/css bug in moderation/comment component caused by long links

oafx:
Hi, 

Edit: this is related to issue #2129 

## What does this PR do?

We're experiencing an minor css bug in the moderation/comment component (noticed in versions 4.6.7 - 4.6.10). Long links aren't wrapped/broken which causes an overflow (see below). I figured the fastest way to fix this would be a quick PR 😄 

Example Screenshots:

before: 
image

after:
image

## How do I test this PR?

- run talk in dev mode
- navigate to /dev
- create a comment containing a long link without dashes
- navigate to /admin/moderate and search for the comment
- the line should be wrapped/broken now, buttons should be still visible

### DIFF
--- a/client/coral-admin/src/routes/Moderation/components/Comment.css
+++ b/client/coral-admin/src/routes/Moderation/components/Comment.css
@@ -62,6 +62,7 @@
   font-weight: 300;
   margin-bottom: 8px;
   overflow-wrap: break-word;
+  word-break:break-word;
 }
 
 .body {


oafx:
Hmm not sure why the `test_integration_chrome_local` fails.. it's just a small css change after all 🤔

kgardnr (approved):

kgardnr:
Thank you @oafx! Great catch!

PR #2136, opened 2018-12-20
[next] Concurrency

wyattjoh:
## What does this PR do?

Adds cluster support to next.

## How do I test this PR?

Set the `LOGGING_LEVEL=trace` and try playing with the `CONCURRENCY` environment variable unset, set to a number.


package.json
```diff
@@ -31,7 +31,7 @@
     "generate:schema": "node ./scripts/generateSchemaTypes.js",
     "docz": "docz",
     "start": "node dist/index.js",
-    "start:development": "ts-node --project ./src/tsconfig.json -r tsconfig-paths/register ./src/index.ts",
+    "start:development": "TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
```
alinhle: Do we need to set `CONCURRENCY=1` here?

alinhle (changes requested):

package.json
wyattjoh: During development, it's important in my opinion to test like production, so `CONCURRENCY=1` isn't required. The only reason it was added to the vscode `launch.json` is to support actual debugging.

alinhle (approved):

PR #2170, opened 2019-01-30
Update appropriate <section> class in Profile section

iign:
Download my comment history has `talk-plugin-ignore-user-section` class. Should be `talk-plugin-profile-data--download-my-comment-history` (couldn't figure out the proper naming convention used).

## What does this PR do?
Updates a class in the markup in the My Profile tab, to reflect the correct section name.

## How do I test this PR?

- Go to My Profile tab.
- Click on Settings tab inside My Profile.
- Inspect, via devtools, the markup wrapping Download My Comment History section.
- HTML class should not be `talk-plugin-ignore-user-section`


alinhle (approved):

alinhle:
Thank you!

PR #2175, opened 2019-02-01
Create Moderation actions notification plugin

mkonikov:
## What does this PR do?
Creates a notification plugin that when enabled allows users to receive notifications when a previously unpublished comment gets approved or rejected.

## How do I test this PR?

- Run Talk with env var `TALK_MODERATION_NOTIFICATION_TYPES` set 
- Sign up for notifications
- Approve a comment that was previously `PREMOD` or `SYSTEM_WITHELD`


plugins/talk-plugin-notifications-category-moderation-actions/index.js
```diff
@@ -0,0 +1,144 @@
...
+  const commentID = get(comment, 'id', null);
+  if (commentID === null) {
+    ctx.log.info('could not get comment id');
+    return;
+  }
+
+  // Check to see if this is a previously unpublished comment.
+  const commentHistory = get(comment, 'status_history', []);
+  // Check for last status before current one
+  const lastStatus = commentHistory[commentHistory.length - 2];
```
wyattjoh: This would error in the event that the status history is defined well.

Instead, perhaps walking backwards:

```suggestion
  let previouslyUnpublished = false;
  if (commentHistory.length >= 2) {
    for (let i = commentHistory.length - 2; i >= 0; i--) {
      const status = commentHistory[i];
      if (UNPUBLISHED_STATUS_TYPES.includes(status.type)) {
        previouslyUnpublished = true;
        break;
      }
    }
  }
```

It may need a bit of work!

mkonikov: The issue with the proposed change is that in the event a comment was moderated, approved, and then flagged and approved again the commenter would receive 2 notifications (one upon initial publish and another when comment is approved again after the flag).

mkonikov: @wyattjoh What do you mean by 'in the event that the status history is defined well'?

wyattjoh: That was a typo, my bad! I had meant:

> in the event that the status history is _not well defined_.

I'd just rather have defensive programming around accessing arrays, otherwise we could run into issues. Perhaps then this might work better?

```js
let previouslyUnpublished = false;
if (commentHistory.length >= 2) {
    const previousStatus = commentHistory[commentHistory.length - 2];
    if (UNPUBLISHED_STATUS_TYPES.includes(previousStatus.type)) {
        previouslyUnpublished = true;
    }
}
```

mkonikov: Fine by me.

FWIW: by checking if true with (`lastStatus && lastStatus.type`), basically accomplishing the same thing already.

alinhle:
> * Approve a comment that was previously `REJECTED` or `SYSTEM_WITHELD`
^ @mkonikov Did you mean `PREMOD` or `SYSTEM_WITHELD`?

mkonikov:
@cvle I think my thought at the time was comments that were rejected because of a banned word, but I don't think that's really a use case we need to worry about. So yes, will update to `PREMOD`

mkonikov:
@cvle Yes - my mistake.

wyattjoh (approved):

alinhle:
@mkonikov We've synced on the wording and changed `unpublished` to `pending`. Otherwise this PR is good to go! Are you good with the changes?

Related commit: https://github.com/coralproject/talk/pull/2175/commits/364b747ccfc0b3a71c8322c2961f09f332f104a4

mkonikov:
Fine by me. We could always overwrite in our yml files if necessary.

alinhle:
@mkonikov Thank you so much!

Issue #2171, opened 2019-01-31
Snackbar too short for longer messages


mkonikov:
We're finding for some of the Snackbar messaging, we'd like to extend the length of time the Snackbar appears on screen. 

Wondering the teams thoughts on adding a small close (X) button and extending it a bit longer.

kgardnr:
Hey @mkonikov! I don't think we'd have any problem if you wanted to do this in core - I think it's a good idea to extend the display time and adding an X.

mkonikov:
With https://github.com/coralproject/talk/pull/2182 merged in, will close now.

PR #2182, opened 2019-02-08
Snackbar improvements

mkonikov:
## What does this PR do?
Adds a close button and lengthen the time on screen, as per https://github.com/coralproject/talk/issues/2171. 

image

## How do I test this PR?

- Run Talk
- Enter a comment that would trigger a Snackbar notification (e.g. submit a one character comment)
- Voila!


mkonikov:
Regarding styling, we've had requests to slightly improve the look of it by adding radius across all 4 corners and bumping it up a bit from bottom of page. If those styles are not something you guys appreciate, can always apply those styles on our front end.

Styling the close button was a bit tedious as need the button to be a bit larger (font-size: 20px) than the standard text which is ~16px, so needed to move it a bit up with position relative and add bottom margin only to the text container. I welcome your suggestions if you have a better way to accomplish.

theysamtheirs (approved):

PR #2187, opened 2019-02-09
[CORL-264] Fixes bug with purify

wyattjoh:
## What does this PR do?

Fixes a bug with the purify configuration which whitelisted the wrong bold tag name.

## How do I test this PR?

Create a bolded comment!

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "generate:schema": "node ./scripts/generateSchemaTypes.js",
     "docz": "docz",
     "start": "NODE_ENV=production node dist/index.js",
-    "start:development": "NODE_ENV=development CONCURRENCY=${CONCURRENCY=2} TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
+    "start:development": "NODE_ENV=development CONCURRENCY=${CONCURRENCY:=2} TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
     "start:webpackDevServer": "ts-node ./scripts/start.ts",
     "lint": "npm-run-all --parallel lint:* tscheck:*",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
     "generate:schema": "node ./scripts/generateSchemaTypes.js",
     "docz": "docz",
     "start": "NODE_ENV=production node dist/index.js",
-    "start:development": "NODE_ENV=development CONCURRENCY=2 TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
+    "start:development": "NODE_ENV=development CONCURRENCY=${CONCURRENCY=2} TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
```
alinhle: Bash Syntax: `${CONCURRENCY=2}` assigns the value 2 to  `CONCURRENCY`, better to make this clear using `${CONCURRENCY:=2}` (it's synonymous), but as we don't actually need or want to assign anything we can use the default and don't assign using `${CONCURRENCY:-2}`.

alinhle (changes requested):

package.json
```diff
@@ -30,7 +30,7 @@
     "generate:schema": "node ./scripts/generateSchemaTypes.js",
     "docz": "docz",
     "start": "NODE_ENV=production node dist/index.js",
-    "start:development": "NODE_ENV=development CONCURRENCY=2 TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
+    "start:development": "NODE_ENV=development CONCURRENCY=${CONCURRENCY:=2} TS_NODE_PROJECT=./src/tsconfig.json node -r ts-node/register -r tsconfig-paths/register ./src/index.ts",
```
alinhle: Any thoughts on not using `${CONCURRENCY:-2}` :-)

alinhle (approved):

PR #2192, opened 2019-02-12
fix: rename devToolsExtension

StevenYuysy:
## What does this PR do?

> Note that starting from v2.7, window.devToolsExtension was renamed to window.__REDUX_DEVTOOLS_EXTENSION__ / window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__.

## How do I test this PR?

- click a button
- view the cat
- see the cat meow


alinhle (approved):
Thanks @StevenYuysy

PR #2204, opened 2019-02-26
[CORL 212] Implement CharCount Restriction feature

alinhle:
## What does this PR do?
- Implement CharCount Restriction feature on the Stream
- Removes "Powered by Coral" text [CORL-262]

## How do I test this PR?

- Set char count restriction and check on client (post, edit, reply)
- Update to a stricter char count restriction while client only knows about the outdated char count restriction. A server error would trigger the client to refresh the setting

### DIFF
--- a/src/core/client/stream/tabs/comments/containers/PostCommentFormContainer.tsx
+++ b/src/core/client/stream/tabs/comments/containers/PostCommentFormContainer.tsx
@@ -2,8 +2,14 @@ import React, { Component } from "react";
 
 import { withContext } from "talk-framework/lib/bootstrap";
 import { InvalidRequestError } from "talk-framework/lib/errors";
+import { graphql, withFragmentContainer } from "talk-framework/lib/relay";
 import { PromisifiedStorage } from "talk-framework/lib/storage";
 import { PropTypesOf } from "talk-framework/types";
+import { PostCommentFormContainer_settings as SettingsData } from "talk-stream/__generated__/PostCommentFormContainer_settings.graphql";
+import {
+  RefreshSettingsFetch,
+  withRefreshSettingsFetch,
+} from "talk-stream/fetches";
 
 import {
   CreateCommentMutation,
@@ -12,11 +18,14 @@ import {
 import PostCommentForm, {
   PostCommentFormProps,
 } from "../components/PostCommentForm";
+import { shouldTriggerSettingsRefresh } from "../helpers";
 
 interface Props {
   createComment: CreateCommentMutation;
+  refreshSettings: RefreshSettingsFetch;
   storyID: string;
   sessionStorage: PromisifiedStorage;
+  settings: SettingsData;
 }
 
 interface State {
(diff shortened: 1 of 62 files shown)


src/core/client/stream/tabs/comments/containers/PostCommentFormContainer.tsx
```diff
@@ -60,6 +69,9 @@ export class PostCommentFormContainer extends Component<Props, State> {
       form.reset({});
     } catch (error) {
       if (error instanceof InvalidRequestError) {
+        if (shouldTriggerSettingsRefresh(error.code)) {
+          await this.props.refreshSettings();
+        }
```
wyattjoh, line 74: This pattern appears more than once, is it possible to extract this?

src/core/client/stream/tabs/comments/containers/PostCommentFormContainer.tsx
```diff
@@ -85,13 +101,39 @@ export class PostCommentFormContainer extends Component<Props, State> {
         onSubmit={this.handleOnSubmit}
         onChange={this.handleOnChange}
         initialValues={this.state.initialValues}
+        min={
+          (this.props.settings.charCount.enabled &&
+            this.props.settings.charCount.min) ||
+          null
```
wyattjoh, line 107: How does this handle when `min = 0` which is falsey?

wyattjoh: Worst case, add a test maybe when `min === 0`?

src/core/client/stream/tabs/comments/helpers/getHTMLText.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * getHTMLText returns text representation of html.
+ * Includes a different implementation during test that works.
+ * @param html
+ */
+export default function getHTMLText(html: string) {
```
wyattjoh, line 6: To simplify this function, possibly replacing it with:

```ts
const getHTMLText = process.env.NODE_ENV === "test" ? () => {  /* */  } : () => {  /* */  };

export default getHTMLText;
```

src/core/client/stream/tabs/comments/containers/PostCommentFormContainer.tsx
alinhle: That's fine `0` and `null` is the same in this case.

src/core/client/stream/tabs/comments/containers/PostCommentFormContainer.tsx
alinhle: For a good abstraction it would require a Higher Order Component which adds other complexities, not sure if there is a win here.

src/core/client/stream/tabs/comments/helpers/getHTMLText.ts
alinhle: This is probably debatable, but I prefer my version for readability :-)

alinhle: I would be also worried if dead code elimination is smart enough to replace the unary operator?

wyattjoh (approved):

PR #2213, opened 2019-03-04
[CORL-130] Handle non-visible comment status

alinhle:
## What does this PR do?
- When a comment is posted, replied to, edited, and the status returned is non-visible, show a message to the user that it is in review.

## How do I test this PR?

- Post/Edit/Write reply to a comment that gets rejected or withheld like posting a comment with a banned word


src/core/client/stream/mutations/CreateCommentMutation.ts
```diff
@@ -127,10 +157,18 @@ function commit(
...
+      // Skip if comments are always moderated.
+      if (alwaysPremoderated) {
+        return;
+      }
       sharedUpdater(environment, store, input);
       store.get(id)!.setValue(true, "pending");
     },
     updater: store => {
+      // Skip if comments are always moderated.
+      if (alwaysPremoderated) {
```
wyattjoh: Can you clarify what it means when the update is skipped when premod is enabled?

When premod is enabled for the story, there may be a situation in the future when a comment is automatically accepted/rejected, and I would rather your updater handle this more gracefully.

src/core/client/stream/mutations/CreateCommentReplyMutation.ts
```diff
@@ -10,35 +11,44 @@ import { TalkContext } from "talk-framework/lib/bootstrap";
...
   store: RecordSourceSelectorProxy,
   input: CreateCommentReplyInput
 ) {
+  const commentEdge = store
+    .getRootField("createCommentReply")!
+    .getLinkedRecord("edge")!;
+  const status = commentEdge.getLinkedRecord("node")!.getValue("status");
+
+  // If comment is in review or has been rejected, we don't need to add it.
+  if (isInReview(status) || isRejected(status)) {
```
wyattjoh: Would it not be better here to rephrase this to say, `isVisable` instead? This way we can have more granular control over when a comment is visible instead of these special cases.

src/core/client/stream/mutations/CreateCommentReplyMutation.ts
```diff
@@ -159,10 +186,18 @@ function commit(
...
+      // Skip if comments are always moderated.
+      if (alwaysPremoderated) {
+        return;
+      }
       sharedUpdater(environment, store, input);
       store.get(id)!.setValue(true, "pending");
     },
     updater: store => {
+      // Skip if comments are always moderated.
+      if (alwaysPremoderated) {
```
wyattjoh: Same comment as above.

src/core/client/stream/mutations/CreateCommentMutation.ts
alinhle: That makes sense, I'll adapt it!

alinhle: Good catch @wyattjoh  😄

src/core/client/stream/mutations/CreateCommentReplyMutation.ts
alinhle: Makes sense 😄

wyattjoh (approved):

PR #2217, opened 2019-03-07
Upgrade Google Passport Package

### LINKED ISSUE #2215: Google+ API Shutdown
This issue is being created to track the issue in an upstream repository we're using related to the [Google+ API Shutdown](https://developers.google.com/+/api-shutdown): https://github.com/mstade/passport-google-oauth2/pull/15

Once this PR is merged and a new version of published, we will release a new bug release of Talk that will upgrade this particular package.

We don't expect it to affect existing users of Talk as Google has indicated that they have added mitigations to the existing behaviors implementation that should allow for a seamless transition.

wyattjoh:
## What does this PR do?

- Fixes #2215

### DIFF
--- a/plugins/talk-plugin-google-auth/package.json
+++ b/plugins/talk-plugin-google-auth/package.json
@@ -4,6 +4,6 @@
   "main": "index.js",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


immber (approved):
👍

PR #2219, opened 2019-03-08
[CORL-159, CORL-160] Stream Config Tab

alinhle:
## What does this PR do?
- Implements the Stream Config Tab
- Upgrades `found`, and  `React`.
- Introduce `expectAndFail` that will immediately fail when expectation is not met. Useful when running inside a try-catch block.

## How do I test this PR?

- Login as moderator or admin
- Do your thing


src/core/client/framework/lib/form/SubmitHookHandler.tsx
```diff
@@ -0,0 +1,85 @@
...
+  private addSubmitHook: AddSubmitHook = hook => {
+    this.submitHooks.push(hook);
+    return () => {
+      this.submitHooks = this.submitHooks.filter(h => h !== hook);
+    };
+  };
+
+  public render() {
+    return (
+      <>
```
wyattjoh, line 74: What does the extra `<>` add here?

src/core/client/framework/lib/form/helpers.tsx
```diff
@@ -44,6 +44,7 @@ export const formatPercentage = (v: any) => {
   return Math.round(v * 100).toString();
 };
 
+export const parseBool = (v: any) => !!v;
```
wyattjoh: Strong suggestion to replace with:

```suggestion
export const parseBool = (v: any) => Boolean(v);
```

src/core/server/models/story/index.ts
```diff
@@ -61,6 +64,16 @@ export interface Story extends TenantResource {
...
+
+  /**
+   * moderation determines whether or not this is a PRE or POST moderated story.
+   */
+  moderation?: GQLMODERATION_MODE;
+
+  /**
+   * premodLinksEnable will put all comments that contain links into premod.
+   */
+  premodLinksEnable?: boolean;
```
wyattjoh, line 83: These settings should be available inside `settings.moderation` on the Story.

src/core/client/framework/lib/form/SubmitHookHandler.tsx
alinhle: Nothing :-)

src/core/client/framework/lib/form/helpers.tsx
alinhle: image
Ok 😃

src/core/server/models/story/index.ts
alinhle: I see what you mean!

wyattjoh (approved):

PR #2220, opened 2019-03-11
External User Fix

wyattjoh:
## What does this PR do?

The advertised behavior of the `services/users.upsertExternalUser(ctx, id, provider, displayName)` incorrectly assigned the provider `id`. This resolves that issue by setting the user's ID to the one provided by the integration as well as falling back to the correct user lookup for when a user returns.


immber (approved):
👍

PR #2243, opened 2019-03-24
DOCS | Move docker-compose.yml out of README

mikemorris:
## What does this PR do?

- Adds a `docker-compose.yml` file with [required](https://docs.docker.com/compose/compose-file/#variable-substitution) `SIGNING_SECRET` environment variable.
- Adds `data/mongo` and `data/redis` directories with `.gitkeep` to skip manual `mkdir` step in setup.
- Removes the manual `docker-compose.yml` creation step and updates docs.

## How do I test this PR?

```
docker-compose up
```
Will return the following error message if the `SIGNING_SECRET` environment variable is unset or empty:
```
ERROR: Missing mandatory value for "environment" option in service "talk": SIGNING_SECRET
```


kgardnr:
Thank you @mikemorris!

wyattjoh:
It was an intentional decision to exclude a `docker-compose.yml` file from the project because we wanted to avoid anyone actually running Talk in production using it, instead preferring the use of managed database services that provide persistence and backups.

PR #2252, opened 2019-03-29
[CORL-317] Modal Component

alinhle:
## What does this PR do?
- Implement a `Backdrop` component that dims background.
- Implement a `NoScroll` component that helps turning of scrolling for `body`.
- Improve `TrapFocus` component
  - Autofocus after mount inside of `TrapFocus`
  - Restore focus to last `activeElement` when `unmounting`
- Implement a `CardCloseButton`
- Put it all together into a `Modal` Component <3 

## How do I test this PR?

- Run `npm run docz -- dev`
- Checkout `Backdrop`, `NoScroll`, and `Modal`


src/core/client/ui/components/Backdrop/Backdrop.tsx
```diff
@@ -0,0 +1,35 @@
...
+  ...rest
+}) => {
+  const rootClassName = cn(
+    classes.root,
+    {
+      [classes.active]: Boolean(active),
+    },
+    className
+  );
+  return <div className={rootClassName} {...rest} />;
```
wyattjoh, line 31: Is there a reason there is no `z-index` specification?

alinhle: Yes! This is a general purpose backdrop, you might want to use it for a modal, or maybe you just want to highlight a certain ui element by darken everything except the element. What I mean is, that you don't always apply an z-Index, and the z-Index will depend on where this Backdrop is used in and for what.

wyattjoh (approved):

Issue #2143, opened 2019-01-04
"Load More" in People tab does not apply filter to returned results


lriggle-strib:
#### Do you want to request a **feature** or report a **bug**?
Bug

#### Intended outcome:
After filtering users by role, clicking "Load More" should display more users of that role

#### Actual outcome:
After filtering users by role, clicking "Load More" does not display more users of that role, it displays users of any role.

#### How to reproduce the issue:
1. Filtered all people so only Administrators would show.
   - There are currently 14 administrators in the database.
   - Initial pageload displays 10 users
2. Click "Load More"
3. The additional 4 administrators should display. Instead 5 additional users of any role display.

<img width="1792" alt="screen shot 2019-01-04 at 12 03 31 pm" src="https://user-images.githubusercontent.com/7051012/50703357-6a96b800-1019-11e9-9e64-5b16256a21b2.png">

#### Version and environment
We're currently on 4.6.5. We had some issues upgrading to newer versions prior to our launches last month and haven't circled back around to figure out what went wrong.

immber:
This looks related to a similar bug ticket we have open for an issue with the Assets (Stories) page pagination. I've added a note to that bug ticket to investigate this at the same time. https://www.pivotaltracker.com/story/show/162740247

immber:
May be related to #2158, which is also reported as being and issue with Assets, but could be hitting the same issue of query missing limit.

kgardnr:
Fixed in https://github.com/coralproject/talk/pull/2174 @lriggle-strib!

lriggle-strib:
Thanks @kgardnr!

mkonikov:
Hey @immber -

I'm still finding this to be an issue in v4.8.0

immber:
@kgardnr You closed this as fixed in #2174, however that PR only addressed the issue on the Stories view, I believe this is still an issue on the People tab, so I'm re-opening. We might need to apply the same or similar fixes to Users as were done to Assets.

kgardnr:
Thanks @immber!

@mkonikov and/or @lriggle-strib would you mind submitting a fix for this potentially? Our team is heads down on v5 the next few weeks so it would really help us out if you could!

PR #2254, opened 2019-04-02
Fix for Load More on Community page

cpoonolly:
## What does this PR do?
Fixes https://github.com/coralproject/talk/issues/2143

## How do I test this PR?
- create a bunch of users
- set a bunch of them to admins (but leave a few non-admins so we can make sure load more works)
- set filter to admin
- click load more
- should only load non admins
- see the cat meow

Peek 2019-04-03 03-02


kgardnr:
Thank you @cpoonolly! 😻

alinhle (approved):
@cpoonolly thank you so much!

PR #2260, opened 2019-04-05
[next] Reaction Sort

wyattjoh:
## What does this PR do?

Adjusts the sort dropdown to use the reaction configuration name to permit customization.


alinhle:
Looks good wyatt! What do you think of my change? Instead of doing `Most {$reactionName}`, I added a separate field in the database called `sortLabel` to allow complete freedom in translating this string to better support a variety of languages.

alinhle (approved):

PR #2261, opened 2019-04-06
[next] Email

wyattjoh:
## What does this PR do?

- Adds support for emails that are localized with a potential of overriding the HTML templates per tenant via a custom [nunjucks loader](https://mozilla.github.io/nunjucks/api.html#loader).
  - Reset Password
  - Confirm Email
  - Ban
  - Suspend

- Add support for `Forgot Email` on the client
  - Add support for TalkError handling for REST calls
  - Add new `account` bundle

### Prerequisite
Update your tenant model in the database with the following email settings:
```json
  "email": {
    "enabled": true,
    "smtpURI": "smtp://localhost:2500",
    "fromAddress": "community@test.com"
  },
```
The above settings is configured to access the inbucket server that you can run using the following docker command

```sh
#!/bin/sh
docker run -d  -p 9000:9000 -p 2500:2500 -p 1100:1100  --name inbucket inbucket/inbucket 
```

### Test feature
- Go to comment stream and in the auth popup kickoff the forgot password process.
- Check `inbucket` at `http://localhost:9000` for the received email to click on the link supplied in the email.
- Change your password


src/core/server/app/handlers/api/auth/local/signup.ts
```diff
@@ -0,0 +1,113 @@
...
+import { validate } from "talk-server/app/request/body";
+import { RequestLimiter } from "talk-server/app/request/limiter";
+import { IntegrationDisabled } from "talk-server/errors";
+import { GQLUSER_ROLE } from "talk-server/graph/tenant/schema/__generated__/types";
+import { LocalProfile, User } from "talk-server/models/user";
+import { insert } from "talk-server/services/users";
+import { sendConfirmationEmail } from "talk-server/services/users/auth";
+import { RequestHandler } from "talk-server/types/express";
+
+export interface SignupBody {
```
alinhle, line 14: nitpik: maybe for consistency with the client call use `SignUpX` instead of `SignupX`?

src/core/server/queue/tasks/mailer/index.ts
```diff
@@ -175,20 +58,28 @@ export class MailerQueue {
...
-    });
+    const startTemplateGenerationTime = now();
+
+    let html: string;
+    try {
+      // Generate the HTML for the email template.
+      html = await this.content.generateHTML(tenant, template);
+    } catch (err) {
+      log.error({ err }, "could not generate the html");
+      // TODO: (wyattjoh) maybe throw an error here?
```
alinhle, line 69: Would it hurt?

src/core/server/queue/tasks/mailer/processor.ts
```diff
@@ -0,0 +1,269 @@
...
+  /**
+   * translateMessage will translate the message to the specified locale as well
+   * a juice the contents.
+   *
+   * @param templateName the name of the template to base the translations off of
+   * @param locale the locale to translate the email content into
+   * @param fromAddress the address that is sending the email (from the Tenant)
+   * @param data data used to send the message
+   */
+  async function translateMessage(
```
alinhle: Maybe move this out of the function and do a `createMessageTranslator(i18n) { return () => ...}`? It should be easier to read to have this big function in a function live outside.

src/core/server/app/request/limiter.ts
```diff
@@ -0,0 +1,87 @@
...
+      .multi()
+      .incr(key)
+      .expire(key, this.ttl)
+      .exec();
+
+    // if this is new or has no expiry
+    if (tries === 1 || expiry === -1) {
+      // then expire it after the timeout
+      this.client.expire(key, this.ttl);
+    }
```
alinhle, line 58: note to self: learn from _wyatt_ about _redis expiry stuff_ :-)

src/core/server/app/handlers/api/auth/local/signup.ts
wyattjoh: Only issue I see with this is the ramifications of unifying that naming.. Camelcase version makes sense, but for routes, it would have to become `/api/auth/local/sign-up`. Not 100% sold.

src/core/server/queue/tasks/mailer/index.ts
wyattjoh: The intended behaviour or reasoning around throwing the error is that we would want to have the user retry the operation. This isn't always the case though, and logging it may trigger the admin's to repair the issue, and then re-trigger the required behaviour after the fact. If you still agree re-throwing here would be worth it, it can be done.

src/core/server/app/handlers/api/auth/local/signup.ts
alinhle: Just the `interface` then maybe? I mean we also have `GraphQL` interfaces but still call the endpoint in the route `graphql` (maybe not the best example but you get the point)

src/core/server/queue/tasks/mailer/index.ts
alinhle: Ok no that's fine :-)

kgardnr (approved):
Tested and approved!

alinhle:
awesome <3

Issue #1780, opened 2018-08-03
Feature Idea: Customize styles in Admin panel


franciclo:
We would like to customize styles in the Admin panel, only css (logo, colors, spaces, typography).
The custom css only affects the iframe, something like that to the admin would be perfect.
We are thinking of overriding styles inside the Docker container, as a workaround, is there a simpler way?

wyattjoh:
Currently, there isn't a way to do so, but the current custom CSS feature could be pretty easily extended to add a new field for admin css stylesheets.

franciclo:
It was implemented in #1237 but then it was removed (don't know when precisely), but that was for using the same css file, having a separate css file as you mention could be better.

kgardnr:
@franciclo if you'd like to pick this one up, feel free, we'd love to add it.

kgardnr:
Featured added in #2279

PR #2279, opened 2019-04-23
apply custom CSS to admin panel

type-face:
## What does this PR do?
Applies styles from custom CSS file to the admin panel.
See https://github.com/coralproject/talk/issues/1780

## How do I test this PR?
1. View admin panel
2. Create .css file that overrides styles on an existing element on the admin panel
3. Update custom admin CSS url to reference .css file from step 2
4. Refresh admin panel to see updated styles


wyattjoh:
We actually initially had this feature, but decided against it as there were some conflicting rules that applied to the stream as well as the admin feature. I would be able to accept this if it included another separate custom CSS file for admin versus the stream.

type-face:
Yes, I saw that this existed before but wasn't sure of the reason for its removal. I'll revise to use a separate CSS file for the admin panel.

locales/en.yml
```diff
@@ -124,6 +124,8 @@ en:
     copy_and_paste: 'Copy and paste code below into your CMS to embed your comment box in your articles'
     custom_css_url: 'Custom CSS URL'
     custom_css_url_desc: 'URL of a CSS stylesheet that will override default Embed Stream styles. Can be internal or external.'
+    custom_admin_css_url: 'Custom Admin Panel CSS URL'
```
type-face, line 127: I'm sure how to indicate that other translations are needed.

wyattjoh: If a translation doesn't exist in the target language, English is used as the fallback. Typically, language authors just notice that change and resolve it, we're planning on some tooling in the future to help with this, but we have limited resources unfortunately.

type-face:
@kgardnr I noticed that the milestone for this was removed. I'm assuming that this is still just awaiting review, but if I need to take any other action to keep this moving, please let me know. Thanks!

wyattjoh:
That's correct @type-face, we had to release earlier to resolve a breaking bug.

When we have some cycles to review this PR it can be included in the next release.

wyattjoh (approved):

PR #2288, opened 2019-04-26
[next] Tests for new story filter and combo box

alinhle:
## What does this PR do?
- Implement `TransitionControl` to allow tests to have control over allowing a routing transition and accessing the history of transition requests.
- Tests to verify that stories are linked to moderation in the _Stories_ page
- Tests to verify the _Stories_ page use the query parameter from the url as the initial search filter.
- Tests to verify the functionality of the search box
- Basic tests to verify moderation works in "Single Story Mode"

### DIFF
--- a/src/core/client/framework/lib/bootstrap/TalkContext.tsx
+++ b/src/core/client/framework/lib/bootstrap/TalkContext.tsx
@@ -11,6 +11,7 @@ import { BrowserInfo } from "talk-framework/lib/browserInfo";
 import { PostMessageService } from "talk-framework/lib/postMessage";
 import { RestClient } from "talk-framework/lib/rest";
 import { PromisifiedStorage } from "talk-framework/lib/storage";
+import { TransitionControlData } from "talk-framework/testHelpers";
 import { UIContext } from "talk-ui/components";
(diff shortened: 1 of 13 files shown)


src/core/client/framework/lib/bootstrap/TalkContext.tsx
```diff
@@ -62,6 +63,9 @@ export interface TalkContext {
 
   /** Clear session data. */
   clearSession: () => Promise<void>;
+
+  /** Controls router transitions (for tests) */
+  transitionControl?: TransitionControlData;
```
wyattjoh, line 68: Is it possible to protect access against this during production with polymorphism?

```ts
type A<T extends boolean> = { a: string } & (T extends true
  ? { b: string }
  : {});

// no error
const a: A<false> = { a: "123" };

// error
const b: A<true> = { a: "123" };
```

alinhle: Possibly, but not straight forward. Couldn't find a straight forward way to pull this off on first glance. I would also say, as it is marked as optional, you can't accidentally access it without checking for its existance or forcing it with `!`, which probably would have made you think twice about what this is for and when it is available.

wyattjoh (approved):

PR #2291, opened 2019-04-29
Tags Permissions

wyattjoh:
## What does this PR do?

Fixes a bug that exposed tags without the correct permissions.

### DIFF
--- a/graph/resolvers/util.js
+++ b/graph/resolvers/util.js
@@ -64,13 +64,13 @@ const wrapCheck = (
 
 /**
  * checkPermissions checks that the current user has all the required
- * permissions.
+ * permissions. It will return true if that's the case.
  *
  * @param {Object} ctx graph context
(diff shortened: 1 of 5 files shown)


graph/resolvers/util.js
```diff
@@ -89,7 +89,7 @@ const wrapCheckPermissions = (
   wrapCheck(
     typeResolver,
     field,
-    (obj, args, ctx) => !checkPermissions(ctx, permissions),
```
alinhle, line 92: Great catch 😆

alinhle (approved):

alinhle (changes requested):
The PR says it fixes a bug, and yes the function was wrong but it was negated when used.. which makes the end result still correct, so was there really a bug?

alinhle (approved):

PR #2293, opened 2019-05-01
[CORL-224] Flag Details

wyattjoh:
## What does this PR do?

- Adds new flag details style API for retrieving all the flags left on a specific comment.
- Adds flag details to moderation comment card

## How to test
- Report a comment as offensive or spam
- Find it in the moderation queue
- Look at details

image


src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -197,6 +245,11 @@ type FlagActionCounts {
   flagged for.
   """
   reasons: FlagReasonActionCounts!
+
+  """
+  flags is the actual Flags that were left by the Users or the system.
+  """
+  flags(first: Int = 10, after: Cursor): FlagsConnection!
```
alinhle: Wouldn't it makes better sense to put this directly on the `Comment`?

wyattjoh: We can't predict how many flags will be attached to the comment, so embedding them there would be unwise.

kgardnr:
This looks great @cvle!

kgardnr (approved):

PR #2315, opened 2019-05-16
[CORL-339] Allow comment draft when not logged in

alinhle:
## What does this PR do?
- Allow drafting a comment before login
- Upgrade coralproject/rte which fixes some RTE issues (CORL-376 and CORL-377)
   - Undo sometimes produced an `undefined` string
   - Placeholder should be below Text Cursor

## How do I test this PR?

- Draft a comment
- Login
- See draft still in the editor


immber (approved):
👍

PR #2320, opened 2019-05-20
[CORL-380] Duplicate Email Error

alinhle:
## What does this PR do?
- Throw duplicate Email Error during sign up.

## How do I test this PR?
- Sign up with local auth and try reusing an existing email
- Should show an error about using an duplicate email.

### DIFF
--- a/src/core/server/models/user.ts
+++ b/src/core/server/models/user.ts
@@ -403,6 +403,10 @@ export async function insertUser(
     // Evaluate the error, if it is in regards to violating the unique index,
     // then return a duplicate Story error.
     if (err instanceof MongoError && err.code === 11000) {
+      // Check if duplicate index was about the email.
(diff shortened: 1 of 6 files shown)


kgardnr (approved):

src/core/server/models/user.ts
```diff
@@ -403,6 +403,10 @@ export async function insertUser(
     // Evaluate the error, if it is in regards to violating the unique index,
     // then return a duplicate Story error.
     if (err instanceof MongoError && err.code === 11000) {
+      // Check if duplicate index was about the email.
+      if (err.errmsg && err.errmsg.includes("tenantID_1_email_1")) {
```
wyattjoh, line 407: Might need a more concrete solution for this, as the index name isn't guaranteed.

PR #2323, opened 2019-05-23
[next] Add redis_options config

cristiandean:
This PR add Redis Option configuration support.

With it you can configure: keepAlive, connectTimeout, tls, redis sentinel, and others.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,6 +1,8 @@
 # job_environment will setup the environment for any job being executed.
 job_environment: &job_environment
   NODE_ENV: test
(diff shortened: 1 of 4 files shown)


cristiandean:
Hello, I've already installed and configured the Cicrcle CI locally and it runs successfully, but it's fail online :/

wyattjoh (approved):
Change looks good! Nice find. I would have preferred to have the options for redis added as query params to the `REDIS_URI`, but as there's some quite complicated options there, I'll leave it as this for now unless it becomes too complex to manage.

PR #2337, opened 2019-05-30
[CORl-278] Perspective Model

wyattjoh:
## What does this PR do?

- Enables the ability to switch the perspective model.
- Provide UI for toggling options


src/core/server/models/tenant.ts
```diff
@@ -170,6 +170,7 @@ export async function createTenant(
       perspective: {
         enabled: false,
         doNotStore: true,
+        model: "TOXICITY",
```
nick-funk: Remove

nick-funk (approved):

PR #2338, opened 2019-05-31
[next] Add pt-BR translation

cristiandean:
Hello, this PR add pt-BR translation support. However, I coudn't test because, for some reason, only the `en-US` folder is generated in: `/dist/core/server/locales/` .


alinhle:
Thank you so much for this @cristiandean. The translations you added are for the client side which are bundled into javascript bundles, for that reason they don't appear in `/dist/core/server/locales/`. There are translations for the server in this directory `/src/core/server/locales` too. Would you help us translate them to `pt-BR` as well?

alinhle:
We also don't currently expose changing the language through configuration on the ui yet. You'll need to change it directly in the mongo db database in `tenants` -> `locale`. v5 is still beta and we are working on improving this.

cristiandean:
@cvle 

>  Would you help us translate them to pt-BR as well?

Added.

> You'll need to change it directly in the mongo db database in tenants -> locale

I tried to do this and still didn't work :/

wyattjoh:
Once you've changed the value in the database, you'll likely need to restart the server. We have enabled tenant caching on the application layer, and re-fetching the tenant isn't completed every single load.

alinhle:
@cristiandean This looks awesome! I just realized we are not passing the locale settings from the server to the client yet. Something we will add pretty soon! 

As a workaround if you want to test it out locally with the current version you can change the client's locale directly in the webpack config in `src/core/build/createWebpackConfig` until we add proper support for switching the language!
image

Looking good:
image
image

alinhle (approved):

PR #2339, opened 2019-05-31
[next] Cookie Support

wyattjoh:
## What does this PR do?

In order to support AMP, cookie support is required to allow persistence.


alinhle (approved):

alinhle (changes requested):
Things to look at:
- [x] On my Safari 12.1.1, if and only if my embed is on a different domain, for some reason the cookie does not get removed on logout 🙄 

- [x] The server should handle expired or revoked cookies gracefully. It should try to remove it and carry on as if the user is not logged in

alinhle (approved):

PR #2349, opened 2019-06-07
[CORL-183] Invite Users

wyattjoh:
## What does this PR do?

Allow users to be invited via the admin panel by other administrators.


src/core/server/services/users/auth/invite.ts
```diff
@@ -0,0 +1,319 @@
...
+  const payloads: Array<{
+    email: string;
+    inviteURL: string;
+    invitedNow: Invite;
+  }> = [];
+  for (const email of uniq(emails)) {
+    // Check to see if the user with the specified email already has an account.
+    const userAlready = await retrieveUserWithEmail(mongo, tenant.id, email);
+    if (userAlready) {
+      return null;
```
nick-funk: Should this abort everything if just one is already found?

nick-funk (approved):

PR #2354, opened 2019-06-11
[CORL-404] Recent Comment History

wyattjoh:
## What does this PR do?

Replaces "Karma" with "Automatic Pre-Moderation".

## Required Migrations

```js
db.tenants.updateMany({}, {
  $set: {
    "recentCommentHistory": {
      "enabled": false,
      "timeFrame": 604800,
      "triggerRejectionRate": 0.3
    }
  }
});
```


nick-funk:
Found this while doing some testing. Looks like a possible divide by 0 error.

image

nick-funk (approved):

PR #2355, opened 2019-06-11
[CORL-409] Prevent users from ignoring staff members

nick-funk:
## What does this PR do?

This removes the ignore button on staff member's overlay popup.
This also validates ignore requests in the backing API to prevent users from ignoring staff members.

## How do I test this PR?

Front-End:
- Create a user with a staff role (Admin, Mod, or Staff).
- Comment on something in any test discussion.
- Create a regular user that is not a staff role.
- View the previous discussion comment, click on the staff user's name.
- The ignore button should not be visible on the staff user's overlay popup/card.

Back-end:
- Request an ignore for a user that is a staff member
- Should return a `USER_CANNOT_BE_IGNORED`

### DIFF
--- a/src/core/server/models/user/helpers.ts
+++ b/src/core/server/models/user/helpers.ts
@@ -1,11 +1,16 @@
+import { GQLUSER_ROLE } from "coral-server/graph/tenant/schema/__generated__/types";
 import { STAFF_ROLES } from "coral-server/models/user/constants";
 
 import { User } from ".";
 
-export function userIsStaff(user: User) {
-  if (STAFF_ROLES.includes(user.role)) {
+export function roleIsStaff(role: GQLUSER_ROLE) {
+  if (STAFF_ROLES.includes(role)) {
     return true;
   }
 
   return false;
+}
+
+export function userIsStaff(user: User) {
+  return roleIsStaff(user.role);
 }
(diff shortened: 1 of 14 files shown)


nick-funk:
Looking into the tests and adding some more now. Tests will likely fail, just pushing up for any preliminary feedback/transparency.

src/core/server/models/user/helpers.ts
```diff
@@ -0,0 +1,11 @@
+import { STAFF_ROLES } from "coral-server/models/user/constants";
+
+import { User } from ".";
+
+export async function userIsStaff(user: User) {
```
wyattjoh: Function does not use `await`, therefore you can remove the `async` term. Also, scoping is helpful here:

```suggestion
export function userIsStaff(user: Pick<User, "role">) {
```

src/core/server/services/users/index.ts
```diff
@@ -585,6 +587,11 @@ export async function ignore(
     throw new UserNotFoundError(userID);
   }
 
+  const userToBeIgnoredIsStaff = await userIsStaff(targetUser);
```
wyattjoh: Assuming we can remove `async`, remove the `await`, and probably just put it into the `if (...) {` clause as the value isn't used outside this check.

src/core/server/models/user/helpers.ts
nick-funk: Fixed this up, removed the async and also the await.

src/core/server/services/users/index.ts
nick-funk: You're right, I've removed the async and also the await.

wyattjoh (approved):

PR #2356, opened 2019-06-12
[next] Remove skew tolerance while parsing/checking expired token.

Chavao:
This PR removes skew tolerance while checking expired token.

The side effect of the skewTolerance set to 300 is a blank screen when refresh the admin page with the token expired.

With this PR, the login page is immediately called when token expires.

### DIFF
--- a/src/core/client/framework/lib/jwt.ts
+++ b/src/core/client/framework/lib/jwt.ts
@@ -24,7 +24,7 @@ export function parseJWT(token: string, skewTolerance = 300): JWT {
     header,
     payload,
     get expired() {
-      return Date.now() / 1000 - skewTolerance >= payload.exp;
+      return Date.now() / 1000 + skewTolerance >= payload.exp;
     },
   };
 }


cvle:
Hey @Chavao, thanks for pointing that out! This is a bug, it's supposed to be `+ skewTolerance`, would you change the `-` to `+` for us? This will help with smaller time skew issues on the client or server.

alinhle:
Hey @Chavao, thanks for pointing that out! This is a bug, it's supposed to be `+ skewTolerance`, would you change the `-` to `+` for us? This will help with smaller time skew issues on the client or server.

Chavao:
> Hey @Chavao, thanks for pointing that out! This is a bug, it's supposed to be `+ skewTolerance`, would you change the `-` to `+` for us? This will help with smaller time skew issues on the client or server.

Done!

alinhle (approved):

alinhle:
Awesome, thank you so much @Chavao

PR #2369, opened 2019-06-19
Favicon fix

leeeandroo:
## What does this PR do?
Load the partial favicon on base template.

## How do I test this PR?

- Open any Talk page
- Check if the favicon are loading correctly

### DIFF
--- a/views/admin.njk
+++ b/views/admin.njk
@@ -2,9 +2,6 @@
 
 {% block title %}Talk Admin{% endblock %}
 
-{# Favicon Configuration #}
-{% include "partials/favicon.njk" %}
-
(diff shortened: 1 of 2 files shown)


kgardnr:
Hey @leeeandroo! We made a decision to remove Coral branding outside the Admin. Thank you for submitting this though! We always appreciate your contributions 🏆

leeeandroo:
Hi @kgardnr , thanks for the feedback.

Right now, even on Admin it's not loading, right? I think the `include` should be placed inside a `block` (or something like this). 
Also, do you have plans to allow custom favicon on public pages?

Best,

PR #2371, opened 2019-06-19
[next] Fixes some issues with external STATIC_URI

Chavao:
This PR fixes admin when using STATIC_URI env variable.

When you set STATIC_URI variable, the `/admin/configure/general` page break because the `static_uri` value is set without slash at the and, so assets are loaded like `http://0.0.0.0assets/x/y/z.css` without the slash between the static uri and the asset path. The side effect of this issue, is a blank screen when load this page.

The second problem, is when the woff2 files are loaded inside css file, using absolute path. Suposing we have a `STATIC_URI=http://www.foobar.com/folder/folder/folder/`, when the font files are called with absolute path, it ignores `/folder/folder/folder/` from the static uri, resulting in `http://www.foobar.com/assets/media/css/foobar.woff2` while the correct is `http://www.foobar.com/folder/folder/folder/assets/media/css/foobar.woff2`.


wyattjoh:
Thanks for the PR @Chavao!

I've been trying to reproduce the error you described, but I can't seem to.

We already have code in our asset entry-point handlers that prefix all the resources with a `/`:

https://github.com/coralproject/talk/blob/e72b15c50514b2b03bd71c1417ec8069a1f1f2c8/src/core/server/app/helpers/entrypoints.ts#L75-L76

And because we already have all the font faces included as data blobs in our css files, we shouldn't be running into any issues relating to paths for fonts as they aren't loaded.

I'm going to close this issue for now, but please post with some examples of issues that you're running into if you experience them.

Chavao:
@wyattjoh @kgardnr I prefered to record two videos to show the problem we have facing with the STATIC_URI variable.

In the first video, as you can see, the fonts aren't loaded correctly and the CONFIGURE page breaks when I access: https://youtu.be/uX4kTsklBW0

And in the second video, with the corrections suggested in this PR, all the problems are solved: https://youtu.be/udYpqSSsPQ0
 
To simulate the static environment, I started a http server in the root of the project, so my STATIC_URI is `http://0.0.0.0:8000/dist/static/`. And I ran the project with the following command: `STATIC_URI="http://0.0.0.0:8000/dist/static/" PORT=80 npm run start`

Chavao:
@wyattjoh @kgardnr Any comments on the issues reported in the video?

wyattjoh:
Hey @Chavao, we were able to reproduce the issue you reported! Currently re-examining your solution.

wyattjoh:
I've provided a fix @Chavao that should address the issue in https://github.com/coralproject/talk/pull/2399 which doesn't use `url-loader` (this way, the fonts aren't embedded into the CSS file). Give that a shot and see if it fixes your issue.

Chavao:
Nice @wyattjoh! Thank you!

The fonts and `Configure > General` page are loading correctly.

PR #2374, opened 2019-06-20
Added missing German translations.

leeeandroo:
## What does this PR do?
Add missing locale keys and translate it to German.

## How do I test this PR?

- Set `TALK_WHITELISTED_LANGUAGES=de` on env
- Check Talk instance and see that It's in German now (e.g. always premoderate feature)


kgardnr (approved):

kgardnr:
This is awesome @leeeandroo! Danke schön! 🎉

PR #2385, opened 2019-07-02
[CORL-382] Allow moderators to ban a commenter from the comment stream 

nick-funk:
## What does this PR do?

- Adds a ban user drop down option in the comment stream's moderation drop down.
- Ban User has a loading state as it waits for a query response from the server when the drop down opens
- Fixes some optimistic update warnings
- `lookup` helper now supports resolving arrays with object references

## TODO
- [x] Refine loading state design
- [x] Reject comment when banning
- [x] Change copy to notify user that the comment will be rejected
- [x] Use default text to send email
- [x] Fix tests

## How do I test this PR?

- Login as a moderate capable user
- Open the comment stream
- Find a comment with a user you want to ban
- Click the moderation drop down via the top right caret on the comment
- Click the new Ban User option
- Confirm you want to ban them
- User will be banned and comment will be rejected


src/core/client/stream/tabs/Profile/ChangeEmail/UpdateEmailMutation.ts
```diff
@@ -35,6 +37,7 @@ const UpdateEmailMutation = createMutation(
         updateEmail: {
           clientMutationId: (clientMutationId++).toString(),
           user: {
+            id: getViewer(environment)!.id,
```
nick-funk, line 43: It would be nice if we had a comment why this can be assumed non-null/undefined. 😃

src/core/server/locales/en-US/email.ftl
```diff
@@ -11,7 +11,7 @@ email-subject-forgotPassword = Password Reset Request
 
 email-notification-template-ban =
   { $customMessage }<br /><br />
-  if you think this has been done in error, please contact our community team 
+  If you think this has been done in error, please contact our community team
```
nick-funk, line 14: Nice catch

nick-funk (commented):
Looks good, tested and functionality works wonderfully as well.
I can't approve as I apparently originally authored this PR, but well done on the additional changes @cvle!

alinhle (approved):

PR #2394, opened 2019-07-08
[next] Websocket Keep Alive

wyattjoh:
## What does this PR do?

Adds a new `WEBSOCKET_KEEP_ALIVE_TIMEOUT` variable for adjusting the keep alive message timeout to be sent across the websocket connection to prevent it from closing.

### DIFF
--- a/src/core/server/graph/tenant/subscriptions/server.ts
+++ b/src/core/server/graph/tenant/subscriptions/server.ts
@@ -223,6 +223,13 @@ export function createSubscriptionServer(
   schema: GraphQLSchema,
   options: Options
 ) {
+  const keepAlive = options.config.get("websocket_keep_alive_timeout");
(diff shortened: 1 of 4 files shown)


nick-funk (approved):

src/core/server/graph/tenant/subscriptions/server.ts
```diff
@@ -223,13 +223,21 @@ export function createSubscriptionServer(
   schema: GraphQLSchema,
   options: Options
 ) {
+  const keepAlive = options.config.get("websocket_keep_alive_timeout");
+  if (typeof keepAlive !== "number") {
```
nick-funk: `number` can be 0 or negative, may be worth a guard clause.

nick-funk (approved):

PR #2396, opened 2019-07-09
[CORL-401] Toxicity Score

wyattjoh:
## What does this PR do?

Displays a toxicity score along with comments that have been reviewed by the perspective API.

## How do I test this PR?

1. Enable the perspective API with an API key
2. Write a comment
3. See the comment in the moderation panel, see the details dropdown
4. Click the details dropdown and see the toxicity score

### DIFF
--- a/src/core/client/admin/routes/Moderate/ModerateCard/ToxicityLabel.tsx
+++ b/src/core/client/admin/routes/Moderate/ModerateCard/ToxicityLabel.tsx
@@ -1,0 +1,45 @@
+import { Localized } from "fluent-react/compat";
+import React, { FunctionComponent } from "react";
+
+import { Counter, Typography } from "coral-ui/components";
+
+import styles from "./ToxicityLabel.css";
+
+const ToxicityCounter: FunctionComponent<{ score: number }> = ({ score }) => (
+  <Counter>{Math.round(score * 100)}%</Counter>
+);
+
+const ToxicityLabel: FunctionComponent<{
+  score: number;
+  threshold: number;
+}> = ({ score, threshold }) => {
+  const counter = <ToxicityCounter score={score} />;
+
+  if (score > threshold) {
+    return (
+      <Localized id="moderate-toxicityLabel-likely" score={counter}>
+        <Typography
+          className={styles.root}
+          variant="bodyCopy"
+          color="errorDark"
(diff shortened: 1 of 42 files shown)


src/core/client/admin/routes/Moderate/ModerateCard/ToxicityLabel.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <Typography
+          className={styles.root}
+          variant="bodyCopy"
+          color="errorDark"
+        >
+          Likely
+        </Typography>
+      </Localized>
+    );
+  } else if (score <= 0.5) {
```
nick-funk, line 30: What if they set their toxicity score to 50? This 0.5 is hidden to the user, should it maybe be a ratio of what 0.5 is to the default 80 instead of a hard coded value?

i.e if user sets toxicity to 70 instead of 80.

50 is 'maybe' when using default of 80.
Compute x for 70, which is 43.75.  (50 * 70 / 80 = 43.75)

What do you think of this?

src/core/server/services/comments/pipeline/phases/toxic.ts
```diff
@@ -79,7 +79,7 @@ export const toxic: IntermediateModerationPhase = async ({
   }
 
   // TODO: (wyattjoh) replace hardcoded default with config.
-  const timeout = ms("500ms");
+  const timeout = ms("800ms");
```
nick-funk, line 82: Reason to up to 800ms? API getting slower on average?

wyattjoh: During my testing there was slowdowns. It is planned to replace this with a config variable, just hasn't been implemented yet.

src/core/client/admin/routes/Moderate/ModerateCard/ToxicityLabel.tsx
wyattjoh: This was going off of the design. I'll follow up to verify this.

wyattjoh: Design has confirmed that this is working as intended.

nick-funk (approved):

PR #2404, opened 2019-07-12
[CORL-444] Stories Tab Adjustments

wyattjoh:
## What does this PR do?

Adjusts the behaviour of the stories tab in the admin panel to use cursor based pagination instead of limit/skip.


immber (approved):
This is a significant improvement! 👍

PR #2408, opened 2019-07-16
[CORL-447] Stories Fix

wyattjoh:
## What does this PR do?

Addresses a fix to the stories tab that affected the display of the version and the open/close functionality.

## How do I test this PR?

1. Load the stories tab in the admin.
2. Close a story.
3. Open a story.
4. See no errors, and see the version number in the top left display load correctly.


immber (approved):
Tested branch and can confirm this corrects the bug introduced in 4.10.0!
👍 
Tried out other story behaviors and looks good:
* Open or Close is now option, not greyed out after changing status
* close/open from comment stream > is reflected in stories tab with refresh
* close/open from stories tab > is reflected in comment stream
* filters work with Open/Closed. Observed some slowness with 1.9 million open stories and 1 closed, but filters work, and query returns with no timeouts

PR #2416, opened 2019-07-24
[CORL-468] Whole Numbers

wyattjoh:
## What does this PR do?

Fixes a bug where decimal numbers are used when whole numbers are required.


immber (approved):
👍 
Verified that bug is fixed and token is correctly generated!

PR #2418, opened 2019-07-25
[next] Persist oidc meta fields from given attribute list.

Chavao:
This PR suggests to add some way to persist data from oidc response in the user profile.

Specific fields can be sent in the oidc response, for example: unique id (in the oidc provider), full name etc.

These fields, specified by an environment variable, will be saved in the user profile, in a field called `meta` as an object.

This implementation could be useful for another providers too, because of this, I used an environment variable specific to OIDC integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,6 +388,9 @@ the variables in a `.env` file in the root of the project in a simple
 - `WEBSOCKET_KEEP_ALIVE_TIMEOUT` - A duration in a parsable format (e.g. `30 seconds`
   , `1 minute`) that should be used to send keep alive messages through the
   websocket to keep the socket alive (Default `30 seconds`)
+- `OIDC_META_FIELDS` - A comma-separated list with field names for OIDC attribute that should be persisted in the user profile document.
+  Those fields represent additional attributes in JWT that are not present in OIDC specification (e.g. Unique id for oidc provider).
+  (Default `null`)
 
 ## License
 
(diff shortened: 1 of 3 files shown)


wyattjoh:
From my understanding of the [OpenID Connect Core 1.0](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) spec, the `sub` claim contained on the `id_token` should contain the unique identifier. I'm hesitant to store arbitrary data from a token in the database. What is your intended use case of the stored metadata?

Chavao:
> What is your intended use case of the stored metadata?

Some fields present in the JWT are not stored in the database, e.g. the full name.
This implementation allows customizing which information we will show in the front-end but keeps the core clean from these codes.

wyattjoh:
As the current iteration of Coral in v5 is not including plugins, I'm hesitant to save arbitrary fields to the metadata on a User when it isn't clear how the data will be surfaced.

Chavao:
Hey @wyattjoh​, ​in our installation, we have to show the full name of the user in the front-end, so because of this, I suggested this implementation to make generic to anyone to run with these specific fields.
At the moment you don't want to add this to the core code, how you suggest implementing this feature avoiding problem with the tests and updating the code?

kgardnr:
Hey @Chavao we don't currently and will not support showing more than the username in v5.

This is because it's been proven that forcing full names can be at the detriment of the commenters, including the fact that "Revealing personal information exposes people to greater levels of harassment and discrimination"; we wrote about this here: https://coralproject.net/blog/the-real-name-fallacy/

Chavao:
Yes @kgardnr, I know you won't add support for this.

This is why I'm asking if you have any tips on how implementing these features in the core having less pain when updating the codebase.

I the last update, we had to manually refactor the code, because of a big change.

PR #2423, opened 2019-07-25
User drawer details tweaks

nick-funk:
## What does this PR do?

Tweaks the details area of the user drawer for two items:

- Increase left/right padding on the status dropdown/popover control
- Adds alternative text/tooltips to the user details icons so that users can identify what these detail items are if they are confused by the icon

## How do I test this PR?

- Select a username in moderate queue or in community area
- See described changes above in the user drawer's top details area

### DIFF
--- a/src/core/client/admin/components/UserHistoryDrawer/UserHistoryDrawerQuery.tsx
+++ b/src/core/client/admin/components/UserHistoryDrawer/UserHistoryDrawerQuery.tsx
@@ -99,7 +99,7 @@ const UserHistoryDrawerQuery: FunctionComponent<Props> = ({
                   id="moderate-user-drawer-email"
                   attrs={{ title: true }}
                 >
-                  <Icon size="sm" className={styles.icon}>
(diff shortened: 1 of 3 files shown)


src/core/client/admin/components/UserHistoryDrawer/UserHistoryDrawerQuery.tsx
```diff
@@ -124,9 +134,14 @@ const UserHistoryDrawerQuery: FunctionComponent<Props> = ({
...
               </Flex>
               <Flex alignItems="center" className={styles.userDetail}>
-                <Icon size="sm" className={styles.icon}>
-                  people_outline
-                </Icon>
+                <Localized
+                  id="moderate-user-drawer-member-id"
+                  attrs={{ title: true }}
+                >
+                  <Icon size="sm" className={styles.icon}>
```
wyattjoh: Should include the `title` attribute here as a default localization fallback.

wyattjoh (approved):

PR #2426, opened 2019-07-26
[CORL-445] Change Password

wyattjoh:
## What does this PR do?

Adds a change password UI to the "My Profile" tab.

## Migration

```js
db.users.updateMany(
  {
    profiles: {
      $elemMatch: {
        type: "local",
      },
    },
  },
  {
    $set: {
      "profiles.$[profiles].passwordID": "temp-migration",
    },
  },
  {
    arrayFilters: [{ "profiles.type": "local" }],
  }
);
```

### Test change functionality

1. From the stream (as User A), navigate to the "My Profile -> Settings" and see the change password UI.
2. Input an incorrect password for "Old Password" and any new password into the "New Password" and submit to see that the "Old Password" field is in error.
3. Input the correct "Old Password" and resubmit, noticing the success UI and cleared form.
4. Click the "Forgot your password?" link and see the forgotten password UI popup with your account's email address pre-filled.

### Test hiding when not in use

1. From the stream (as User A), navigate to the "My Profile -> Settings" and see the change password UI.
2. From the admin (as User B), disable local auth, and enable SSO, save
3. As User A in the stream, reload the page and see there is no change password UI any more.


src/core/client/stream/tabs/Profile/Settings/IgnoreUserSettingsContainer.css
```diff
@@ -1,7 +1,3 @@
-.root {
-  max-width: 440px;
```
nick-funk, line 2: How does this behave if you up the internal browser default text size? Do we get excessive line-wrapping that makes the interface un-usable?

src/core/client/stream/tabs/Profile/Settings/SettingsContainer.css
```diff
@@ -0,0 +1,3 @@
+.root {
+  max-width: 440px;
```
nick-funk, line 2: Ditto about internal browser text size usability with large default font sizes. How does this behave?

src/core/common/errors.ts
```diff
@@ -136,6 +136,12 @@ export enum ERROR_CODES {
    */
   PASSWORD_TOO_SHORT = "PASSWORD_TOO_SHORT",
 
+  /**
+   * PASSWORD_INCORRECT is returned when a logged in operation that requires the
+   * password returns the wrong password.
+   */
+  PASSWORD_INCORRECT = "PASSWORD_INCORRECT",
```
nick-funk, line 143: I've found it is usually more secure if you're vague about what was incorrect when handling a login request.

Does specifying that the password specifically was incorrect make sense?

src/core/server/graph/tenant/resolvers/Mutation.ts
```diff
@@ -1,5 +1,7 @@
 import { GQLMutationTypeResolver } from "coral-server/graph/tenant/schema/__generated__/types";
 
+// TODO: (wyattjoh) add rate limiting to these edges
```
nick-funk, line 3: Do we have a maintenance ticket logged for this?

src/core/common/errors.ts
wyattjoh: This particular error is only thrown when a user is already logged in, and is being asked to provide their password for confirmation as is explicitly referenced in the enum description. This is why we have a different (more vague) error code for standard auth errors `INVALID_CREDENTIALS`.

nick-funk (approved):

PR #2415, opened 2019-07-23
[CORL-234] Show user's account history in the user drawer

nick-funk:
## What does this PR do?

Adds an account history tab to the user drawer.

## How do I test this PR?

- Head to the moderate section signed in as a moderate level user
- Click on a username on a moderate card or in the community section
- Click on the account history tab in the user drawer
- See history of bans

Notes:

- Suspension records will not exist as we can't create suspensions yet

PR #2431, opened 2019-07-29
[CORL-97] Add user status details popover to user drawer

tessalt:
## What does this PR do?
Adds a popover next to user status dropdown in user drawer that indicates further details about bans or suspensions. 

## How do I test this PR?
- log in as moderator
- select a banned or suspended user from communities tab
- click on "i" icon next to user status

Screen Shot 2019-07-29 at 5 06 49 PM
Screen Shot 2019-07-29 at 5 06 16 PM


src/locales/en-US/admin.ftl
```diff
@@ -607,3 +607,10 @@ invite-goToAdmin = Go to { -product-name } Admin
 invite-goToOrganization = Go to { $organizationName }
 invite-tokenNotFound =
   The specified link is invalid, check to see if it was copied correctly.
+
+userDetails-banned-on = <strong>Banned on</strong> { $timestamp }
+userDetails-banned-by = <strong>by</strong> { $username } 
+userDetails-suspension = Suspension 
+userDetails-suspended-by = <strong>by</strong> { $username }
```
nick-funk: Can we join together the two element `Suspension` and `by { $username }`?

The reason being that it's hard to translate two separate ordered items in a sentence. In other languages (say Japanese) the subject of the sentence `{ $username }` will come before the word "Suspension".

By creating two elements, we make it hard for them to reorder this translation and template it nicely.

tessalt: i had them separate because "suspension" is on a different line (it's more like the heading of the bubble). Could I wrap two separate `<Typography>` elements in the same `<Localization>` tag?

src/core/client/admin/components/UserHistoryDrawer/UserStatusDetailsContainer.tsx
```diff
@@ -0,0 +1,170 @@
...
+  if (!user.status.ban.active && !user.status.suspension.active) {
+    return null;
+  }
+  const activeBan = user.status.ban.history.find(item => item.active);
+  const activeSuspension = user.status.suspension.history.find(
+    item => item.active
+  );
+  const { locales } = useCoralContext();
+
+  function formatDate(date: Date): string {
```
wyattjoh: Suggest using https://github.com/coralproject/talk/pull/2415/files#diff-95ea4d35e57417f9324dcaf7ebfa3d19R123 instead

src/core/client/admin/components/UserHistoryDrawer/UserStatusDetailsContainer.tsx
```diff
@@ -0,0 +1,170 @@
...
+          active
+          history {
+            active
+            from {
+              start
+              finish
+            }
+            createdBy {
+              username
+              id
```
wyattjoh: Suggest removing `id`

src/locales/en-US/admin.ftl
nick-funk: I'm not sure if we can wrap two of them together like that. In that case I think this is probably fine?

nick-funk (approved):

PR #2432, opened 2019-07-30
[CORL-456] add empty state for comment history

tessalt:
## What does this PR do?
- adds an empty state to "my profile" comment history if user has no comments

## How do I test this PR?
- log in as user with no comments, click on "my profile" tab in stream

Used a material design icon, pending discussion about bringing in fontawesome icons 
Screen Shot 2019-07-30 at 10 38 56 AM

### DIFF
--- a/src/core/client/stream/tabs/Profile/CommentHistory/CommentHistory.tsx
+++ b/src/core/client/stream/tabs/Profile/CommentHistory/CommentHistory.tsx
@@ -10,9 +10,9 @@ import {
   Typography,
 } from "coral-ui/components";
 
-import styles from "./CommentHistory.css";
-
 import HistoryCommentContainer from "./HistoryCommentContainer";
+
+import styles from "./CommentHistory.css";
 
 interface CommentHistoryProps {
   story: PropTypesOf<typeof HistoryCommentContainer>["story"];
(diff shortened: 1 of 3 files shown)


src/core/client/stream/tabs/Profile/CommentHistory/CommentHistory.tsx
```diff
@@ -2,7 +2,15 @@ import { Localized } from "fluent-react/compat";
...
-import { Button, HorizontalGutter, Typography } from "coral-ui/components";
+import {
+  Button,
+  Flex,
+  HorizontalGutter,
+  Icon,
+  Typography,
+} from "coral-ui/components";
+
+import styles from "./CommentHistory.css";
```
nick-funk: Kiwi informed me you can run into styling issues if your local component styles import isn't the final import.

src/locales/en-US/stream.ftl
```diff
@@ -154,6 +154,8 @@ profile-historyComment-story = Story: {$title}
 profile-profileQuery-errorLoadingProfile = Error loading profile
 profile-profileQuery-storyNotFound = Story not found
 profile-commentHistory-loadMore = Load More
+profile-commentHistory-empty = You have not written any comments
+Profile-commentHistory-empty-subheading = A history of your comments will appear here
```
nick-funk: Not sure if it matters, but the P is capitalized here where the other's aren't.

src/core/client/stream/tabs/Profile/CommentHistory/CommentHistory.tsx
tessalt: good to know! wonder if there's a TS rule we can add.

tessalt:
@nick-funk fixed!

src/core/client/stream/tabs/Profile/CommentHistory/CommentHistory.tsx
nick-funk: Apparently it's not available for our linting yet, but I have hopes we can automate this at some point. 😄

nick-funk (approved):

PR #2436, opened 2019-07-30
[CORL-458] open user drawer when flagger username clicked

tessalt:
## What does this PR do?
- when you expand the "offensive" or "spam" markers and a `flagger` is defined, turn username of flagger into a button that opens the user drawer for flagger

## How do I test this PR?
- click on details disclosure in a reported comment in moderation card, click on username
Screen Shot 2019-07-30 at 4 34 03 PM


src/core/client/admin/components/ModerateCard/FlagDetailsContainer.tsx
```diff
@@ -17,13 +17,15 @@ import ToxicityLabel from "./ToxicityLabel";
 interface Props {
   comment: FlagDetailsContainer_comment;
   settings: FlagDetailsContainer_settings;
+  onUserNameClick: (id?: string) => void;
```
alinhle: Just for consistency, I think we used `Username` instead of  `UserName`

alinhle (changes requested):
Hey for some reason if I click on the authors username of the comment I'm getting a `User not found`. Can you look into that?
image

Also just had a small sync with Sam about the design. The username on the flag details should look like the authors username with the hovering and stuff instead of underlined: 
image
You can sync with @IAmSamHankins  on this for details if needed.

Otherwise this looks good!

alinhle (approved):

PR #2445, opened 2019-08-05
[CORL-149] Persisted Queries

alinhle:
## What does this PR do?
- Enabled persisted queries for the Client
- `npm run build` will build using persisted queries
- `npm run generate-persist` and `npm run generate:relay-persist` will also generate artefacts with persisted queries

## TODO
- [x] Server implementation


src/core/client/framework/lib/network/createManagedSubscriptionClient.ts
```diff
@@ -109,10 +109,14 @@ export default function createManagedSubscriptionClient(
...
       }
+      if (!operation.text && !operation.id) {
+        throw Error("Neither subscription query nor id was provided.");
+      }
       const subscription = subscriptionClient
         .request({
           operationName: operation.name,
-          query: operation.text!,
+          // Use id as query when not available e.g. when using persisted queries.
+          query: (operation.text || operation.id)!,
```
alinhle: Add a `pq:` prefix.

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "ES2018",
```
nick-funk, line 3: If I remember correctly, this is setting the transpile output target for TS transpiler.

Was this intentional? Jumping from ES5 (2009) to ES2018 seems like a big compatibility ask.

nick-funk (commented):
Submitting with comment, I've reviewed and all looks good, just have a question about the tsconfig target change.

tsconfig.json
wyattjoh: This is only on the server code, as we already depend on node >= 8, we have no issues here, the advantage is that we can now iterate over ES6 `Map`, before we couldn't ;)

nick-funk: Sweeeeeet!

nick-funk (approved):

PR #2447, opened 2019-08-06
[CORL-391] Fix jankiness when opening reply box of last comment in Stream 

alinhle:
## What does this PR do?
- Fix jankiness when opening reply box of last comment in Stream
- Introduce `useResizeObserver` hook
- Simplify auth popup resize behavior using the resize observer

## How do I test this PR?

- Go to stream
- Scroll to the bottom.
- Click on `Reply` of the last comment to open the reply form
- Should work smoothly without jumpy behavior
- Auth Popup automatic resizing should continue to work as well.


tessalt (approved):
cool!

PR #2452, opened 2019-08-07
[CORL-219] Change username

tessalt:
## What does this PR do?
- allow users to change their own username under *either* one of the following conditions:
1. they have not previously changed their username
2. their most recent username change is older than 14 days

- allow admins/moderators to view a history of username changes in commenter history panel 

## How do I test this PR?
- run migration: 

```
db.users.find().forEach((user) => {
  db.users.update({
    id: user.id
  }, {
    $push: {
      "status.username.history": {
        username: user.username,
        createdAt: user.createdAt,
        createdBy: user.id
      },
    },
  }) 
})
```

- log in as a new or existing user
- click "profile" tab
- click "edit" next to username
- complete new username form and submit to change username
- attempt to change username a second time


src/core/server/graph/tenant/mutators/Users.ts
```diff
@@ -118,8 +120,24 @@ export const Users = (ctx: TenantContext) => ({
...
+    ),
   updateUserUsername: async (input: GQLUpdateUserUsernameInput) =>
-    updateUsername(ctx.mongo, ctx.tenant, input.userID, input.username),
+    updateUserUsername(
+      ctx.mongo,
+      ctx.tenant,
+      input.userID,
+      input.username,
+      ctx.user!
+    ),
```
tessalt, line 140: oops this is probably broken bc the methods have the same name.

tessalt: er wait maybe it's not, but I had issues earlier with ts being mad if the method signatures weren't identical if the names were the same?

src/core/client/admin/components/UserHistoryDrawer/UserDrawerAccountHistory.tsx
```diff
@@ -153,10 +179,12 @@ const UserDrawerAccountHistory: FunctionComponent<Props> = ({ user }) => {
...
               <TableCell className={styles.action}>
-                {history.kind === "suspension" ? (
+                {history.kind === "suspension" && (
                   <SuspensionAction {...history.action} />
-                ) : (
-                  <BanAction {...history.action} />
+                )}
+                {history.kind === "ban" && <BanAction {...history.action} />}
+                {history.kind === "username" && (
+                  <UsernameChangeAction {...history.action} />
```
wyattjoh: Maybe you should create a function that lives outside this one that has a `switch (history.kind)` to keep things clean?

src/core/client/admin/components/UserHistoryDrawer/UserDrawerAccountHistory.tsx
```diff
@@ -117,6 +130,19 @@ const UserDrawerAccountHistory: FunctionComponent<Props> = ({ user }) => {
       });
     });
 
+    user.status.username.history.forEach((record, i) => {
+      history.push({
+        kind: "username",
+        action: {
+          username: record.username,
+          prevUsername:
+            i >= 1 ? user.status.username.history[i - 1].username : null,
```
wyattjoh, line 117: I'd add a comment here as to the logic being performed. After reading it, I know it's grabbing the previous username, but it would be helpful to have it documented here.

src/core/client/admin/components/UserHistoryDrawer/UsernameChangeAction.tsx
```diff
@@ -0,0 +1,38 @@
...
+          <span className={styles.tableLight}>New: </span>
+        </Localized>
+        {` ${username}`}
+      </div>
+      {prevUsername && (
+        <div>
+          <Localized id="moderate-user-drawer-username-change-old">
+            <span className={styles.tableLight}>Old: </span>
+          </Localized>
+          {` ${prevUsername}`}
```
wyattjoh: I'd prefer:

```tsx
          <Localized id="moderate-user-drawer-username-change-old">
            <span className={styles.tableLight}>Old:</span>
          </Localized>{" "}
          {prevUsername}
```

Instead as it reads better than a string template. Also note that trailing spaces cannot be included in translations, and should not be depended on, (hence me removing the ` ` from `<span className={styles.tableLight}>Old: </span>`).

src/core/client/stream/tabs/Profile/ChangeUsername/ChangeUsernameContainer.tsx
```diff
@@ -0,0 +1,327 @@
...
+  TextField,
+  Typography,
+} from "coral-ui/components";
+
+import { ValidationMessage } from "coral-framework/lib/form";
+
+import { ChangeUsernameContainer_viewer as ViewerData } from "coral-stream/__generated__/ChangeUsernameContainer_viewer.graphql";
+import UpdateUsernameMutation from "./UpdateUsernameMutation";
+
+import styles from "./ChangeUsername.css";
```
wyattjoh: Check the import ordering on this file.

Also, style files should always match the component/class/filename, in this case, `ChangeUsernameContainer.css` instead of `ChangeUsername.css`.

src/core/client/stream/tabs/Profile/ChangeUsername/ChangeUsernameContainer.tsx
```diff
@@ -0,0 +1,327 @@
...
+
+  const canChangeUsername = useMemo(() => {
+    const { username } = viewer.status;
+    if (username && username.history.length > 1) {
+      const lastUsernameEditAllowed = new Date();
+      const dateDiff = lastUsernameEditAllowed.getDate() - 14;
+      lastUsernameEditAllowed.setDate(dateDiff);
+      return username.history.find(history => {
+        return history.createdAt > lastUsernameEditAllowed;
+      });
```
wyattjoh: I think it would be cleaner here to instead just reference the last history item in the list (as history items are always pushed) instead of iterating through all the history items.

src/core/client/stream/tabs/Profile/ChangeUsername/ChangeUsernameContainer.tsx
```diff
@@ -0,0 +1,327 @@
...
+    return true;
+  }, [viewer]);
+
+  const canChangeUsernameDate = useMemo(() => {
+    const { username } = viewer.status;
+    if (username && username.history.length > 1) {
+      const date = new Date(
+        username.history[username.history.length - 1].createdAt
+      );
+      const diff = date.getDate() + 14;
```
wyattjoh: Special numbers should live in `coral-common/constants`.

src/core/client/stream/tabs/Profile/ChangeUsername/UpdateUsernameMutation.tsx
```diff
@@ -0,0 +1,62 @@
...
+        },
+      },
+      optimisticResponse: {
+        updateUsername: {
+          clientMutationId: (clientMutationId++).toString(),
+          user: {
+            username: input.username,
+            status: {
+              username: {
+                history: [
```
wyattjoh, line 49: A user could have previously had a username history, you should merge their existing username history here.

src/core/client/stream/test/fixtures.ts
```diff
@@ -129,6 +132,58 @@ export const baseUser = createFixture<GQLUser>({
...
+      username: {
+        history: [
+          {
+            username: "original",
+            createdAt: `${weekago.toISOString()}`,
+            createdBy: { id: "changed-user" },
+          },
+          {
+            username: "u_changed",
+            createdAt: `${yesterday.toISOString()}`,
```
wyattjoh: `toISOString()` already returns a string, no need for the template parameters here.

src/core/server/locales/en-US/email.ftl
```diff
@@ -24,6 +24,13 @@ email-notification-template-passwordChange =
 
 email-subject-passwordChange = Your password has been changed
 
+email-subject-updateUsername = Your username has been changed
+
+email-notification-template-updateUsername =
+  Hello { $username },<br/><br/>
+  Thank you for updating your {{ context.organizationName }} commenter account information. The changes you made are effective immediately. <br /><br />
```
wyattjoh: Looks like `{{ context.organizationName }}` was copied from the HTML template! It should be `{ $organizationName }`

src/core/server/locales/en-US/errors.ftl
```diff
@@ -54,3 +54,4 @@ error-rateLimitExceeded = Rate limit exceeded.
 error-inviteTokenExpired = Invite link has expired.
 error-inviteRequiresEmailAddresses = Please add an email address to send invitations.
 error-passwordIncorrect = Password provided was incorrect.
+error-usernameAlreadyUpdated = You may only change your username once every 14 days.
```
wyattjoh: 14 here should be templated.

src/core/common/errors.ts
```diff
@@ -124,6 +124,12 @@ export enum ERROR_CODES {
    */
   USERNAME_EXCEEDS_MAX_LENGTH = "USERNAME_EXCEEDS_MAX_LENGTH",
 
+  /**
+   * USERNAME_UPDATED_WITHIN_WINDOW is returned when the user attempts to associate
+   * a new username when they have previously changed their username less than 14 days previously.
```
wyattjoh: Remove reference to 14 in this comment here.

src/core/server/models/user/user.ts
```diff
@@ -716,40 +764,48 @@ export async function setUserUsername(
...
  *
  * @param mongo the database handle
  * @param tenantID the ID to the Tenant
  * @param id the ID of the User where we are setting the username on
  * @param username the username that we want to set
+ * @param createdBy the user making the change
  */
-export async function updateUserUsername(
+
+export async function updateUsername(
```
wyattjoh: The function name should remain `updateUserUsername` as with other model scoped functions.

src/core/server/queue/tasks/mailer/templates/update-username.html
```diff
@@ -0,0 +1,11 @@
...
+{{ context.username }},<br /><br />
+
+Thank you for updating your {{ context.organizationName }} commenter account
+information. The changes you made are effective immediately. If you did not make
+this change please reach out to
+<a
+  data-l10n-name="organizationContactEmail"
+  href="mailto:{{ context.organizationContactEmail }}"
+  >{{ context.organizationContactEmail }}</a
+>. {% endblock %}
```
wyattjoh: Check formatting here!

src/core/server/services/users/index.ts
```diff
@@ -360,23 +361,88 @@ export async function deactivateToken(
...
+ * updateUserUsername will update the current users.
  *
  * @param mongo mongo database to interact with
+ * @param mailer mailer queue instance
  * @param tenant Tenant where the User will be interacted with
  * @param userID the User's ID that we are updating
  * @param username the username that we are setting on the User
  */
-export async function updateUsername(
+export async function updateOwnUsername(
```
wyattjoh: We've used `updateUsername` to refer to changing ones own username, and `updateUserUsername` as the semantic in other places.

src/core/server/services/users/index.ts
```diff
@@ -360,23 +361,88 @@ export async function deactivateToken(
...
-  userID: string,
+  user: User,
   username: string
 ) {
   // Validate the username.
   validateUsername(username);
 
-  return updateUserUsername(mongo, tenant.id, userID, username);
+  const lastUsernameEditAllowed = new Date();
+  const dateDiff = lastUsernameEditAllowed.getDate() - 14;
```
wyattjoh: Special numbers should live in `coral-common/constants`.

src/core/server/services/users/index.ts
```diff
@@ -360,23 +361,88 @@ export async function deactivateToken(
...
+  lastUsernameEditAllowed.setDate(dateDiff);
+
+  if (user.status.username.history.length > 1) {
+    const recentUpdate = user.status.username
```
wyattjoh: Can rely on the natural ordering of the history rather than a sort.

src/core/server/services/users/index.ts
```diff
@@ -360,23 +361,88 @@ export async function deactivateToken(
...
+        name: "update-username",
+        context: {
+          username: user.username!,
+          organizationName: tenant.organization.name,
+          organizationURL: tenant.organization.url,
+          organizationContactEmail: tenant.organization.contactEmail,
+        },
+      },
+    });
+  }
```
wyattjoh, line 433: Should add a `else` that logs that the email wasn't sent because the user didn't have an email address.

src/locales/en-US/stream.ftl
```diff
@@ -206,6 +206,19 @@ comments-submitStatus-submittedAndWillBeReviewed =
...
+profile-changeUsername-success = Your username has been successfully updated
+profile-changeUsername-edit = Edit
+profile-changeUsername-heading = Edit your username
+profile-changeUsername-desc = Change the username that will appear on all of your past and future comments. Usernames can be changed once every 14 days.
+profile-changeUsername-current = Current username
+profile-changeUsername-newUsername-label = New username
+profile-changeUsername-confirmNewUsername-label = Confirm new username
+profile-changeUsername-cancel = Cancel
+profile-changeUsername-submit = Save
+profile-changeUsername-recentChange = Your username has been changed in the last 14 days. You may change your username again on { $nextUpdate }
```
wyattjoh: 14 should be templated here.

wyattjoh (changes requested):

src/core/server/graph/tenant/resolvers/UsernameHistory.ts
```diff
@@ -0,0 +1,16 @@
+import { GQLUsernameHistoryTypeResolver } from "coral-server/graph/tenant/schema/__generated__/types";
+import * as user from "coral-server/models/user";
+
+export const UsernameHistory: Required<
+  GQLUsernameHistoryTypeResolver<user.UsernameHistory>
+> = {
+  createdBy: ({ createdBy }, input, ctx) => {
+    if (createdBy) {
+      return ctx.loaders.Users.user.load(createdBy);
```
nick-funk, line 9: @wyattjoh, thoughts on ways we can work to clean up this kind of access?

In general, I find if we're being forced to do `something.set.otherSet.item.function(...)` it indicates that the architecture isn't abstracted/decoupled enough for us to easily access the functions/properties we want without digging in really deep.

I worry this will escalate and we'll end up with business logic retrieval that knows too much about the underlying data it's trying to provide when we could maybe abstract this into cleaner retrieval patterns.

src/core/server/locales/en-US/errors.ftl
nick-funk: I don't know if this is possible, but it would be cool if this timespan could be templated into this error message. That way we could update a constant somewhere and update all the messages if they decide it should be, say, only every 30 days.

src/core/server/graph/tenant/resolvers/UsernameHistory.ts
wyattjoh: To me, this reads cleanly. The models being retrieved are nicely written such that reading the chained access tells you what's loading. The last `.load(` even tells me it's a `DataLoader`, which gives me confidence that this call is efficient. We can discuss it in arch meeting today.

src/core/server/queue/tasks/mailer/templates/update-username.html
```diff
@@ -0,0 +1,11 @@
+{% extends "layouts/user-notification.html" %} {% block content %} Hello
```
nick-funk: @wyattjoh  showed me that you can use the Nunjucks extension with VSCode and it will stop this from being garbled every time you save.

Link to extension: https://github.com/ronnidc/vscode-nunjucks

src/core/client/admin/components/UserHistoryDrawer/UserDrawerAccountHistory.tsx
nick-funk: Alternatively, if you don't like a comment, a neatly named function clarifying what it's doing could be good too.

src/core/server/graph/tenant/resolvers/UsernameHistory.ts
nick-funk: That's fair, and it probably reads different to you than to me. We can for sure discuss further. 👍

nick-funk (commented):
Leaving this as a "Comment" review, as Wyatt has outstanding comments.

Wanted to say this all looks like great work, even with the comments made, functionality seems to be there and my testings shows it's pretty solid.

src/core/client/stream/tabs/Profile/ChangeUsername/UpdateUsernameMutation.tsx
tessalt: does that mean I should add the username history to the `UpdateUsernameInput` type in the schema? or create a different interface to use here

wyattjoh (approved):

PR #2457, opened 2019-08-09
Center sign in popovers over the browser window so they are easily visible

nick-funk:
## What does this PR do?

Calculates the popover positioning relative to the browser window so it appears centered nicely over the window when spawned.

## How do I test this PR?

- Go to stream
- Select to sign in or register
- Notice popover is perfectly centered over the browser window horizontally, and just slightly down vertically

### DIFF
--- a/src/core/client/stream/common/UserBox/UserBoxContainer.tsx
+++ b/src/core/client/stream/common/UserBox/UserBoxContainer.tsx
@@ -122,21 +122,11 @@ export class UserBoxContainer extends Component<Props> {
       return null;
     }
 
-    const winLeft =
-      window.screenLeft !== undefined ? window.screenLeft : window.screenX;
-    const winTop =
-      window.screenTop !== undefined ? window.screenTop : window.screenY;
-
-    const popupWidth = 350;
-    const popupLeft = winLeft + window.outerWidth / 2 - popupWidth / 2;
-    const popupTop = winTop + 100;
-
     return (
       <>
         <Popup
           href={this.authUrl}
           title="Coral Auth"
-          features={`menubar=0,resizable=0,width=350,height=450,top=${popupTop},left=${popupLeft}}`}
           open={open}
           focus={focus}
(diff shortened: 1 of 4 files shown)


src/core/client/stream/common/UserBox/UserBoxContainer.tsx
```diff
@@ -122,12 +122,21 @@ export class UserBoxContainer extends Component<Props> {
...
+    const popupLeft = winLeft + window.outerWidth / 2 - popupWidth / 2;
+    const popupTop = winTop + 100;
+
     return (
       <>
         <Popup
           href={this.authUrl}
           title="Coral Auth"
-          features="menubar=0,resizable=0,width=350,height=450,top=100,left=500"
+          features={`menubar=0,resizable=0,width=350,height=450,top=${popupTop},left=${popupLeft}}`}
```
wyattjoh: I'd suggest instead to make `features` accept an object with `Record<string, any>` where by default, the `top` and `left` are set as you've defined above for the `Popover` component.

nick-funk: Can do!

src/core/client/ui/components/Popup/Popup.ts
```diff
@@ -1,5 +1,13 @@
 import { Component } from "react";
 
+interface WindowFeatures {
+  resizable: number;
+  menuBar: number;
```
wyattjoh: The browser API has it all lowercase:

```suggestion
  menubar: number;
```

wyattjoh (changes requested):

src/core/client/ui/components/Popup/Popup.ts
nick-funk: Fixed this up, also defaulted the value like we did previously with new naming.

wyattjoh (approved):

PR #2461, opened 2019-08-12
[CORL-220] Change email address

tessalt:
## What does this PR do?
- allows user to change their own email address provided they correctly enter their password
- sets `emailVerified` to false and sends out a new email verification email
- gives user the option to re-send email verification if their email address is unverified.

## How do I test this PR?
- visit profile tab
- click "edit" next to email 
- fill out a new email address and enter password
- you should see a notice about email verification

Very similar to change username PR, with additional step on the backend of updating the user profile ID for local profiles.

### DIFF
--- a/src/core/client/stream/tabs/Profile/ChangeEmail/ChangeEmailContainer.tsx
+++ b/src/core/client/stream/tabs/Profile/ChangeEmail/ChangeEmailContainer.tsx
@@ -1,4 +1,4 @@
-import { FORM_ERROR, FormApi } from "final-form";
+import { FORM_ERROR, FormApi, FormState } from "final-form";
 import { Localized } from "fluent-react/compat";
 import React, { FunctionComponent, useCallback, useState } from "react";
 import { Field, Form } from "react-final-form";
@@ -95,6 +95,29 @@ const changeEmailContainer: FunctionComponent<Props> = ({ viewer }) => {
     },
     [updateEmail]
   );
+
+  const preventSubmit = (
+    formState: Pick<
+      FormState,
+      | "pristine"
+      | "hasSubmitErrors"
+      | "hasValidationErrors"
+      | "dirtySinceLastSubmit"
+    >
+  ) => {
+    const {
+      pristine,
+      hasValidationErrors,
+      hasSubmitErrors,
+      dirtySinceLastSubmit,
+    } = formState;
+    return (
+      pristine ||
+      hasValidationErrors ||
+      (hasSubmitErrors && !dirtySinceLastSubmit)
+    );
+  };
+
   return (
     <HorizontalGutter spacing={5} data-testid="profile-changeEmail">
       {!showEditForm && (
@@ -187,12 +210,9 @@ const changeEmailContainer: FunctionComponent<Props> = ({ viewer }) => {
               {({
                 handleSubmit,
                 submitError,
-                pristine,
                 invalid,
                 submitting,
-                hasValidationErrors,
-                hasSubmitErrors,
-                dirtySinceLastSubmit,
+                ...formProps
               }) => (
                 <form
                   onSubmit={handleSubmit}
@@ -268,24 +288,17 @@ const changeEmailContainer: FunctionComponent<Props> = ({ viewer }) => {
                       </Button>
                     </Localized>
                     <Localized id="profile-changeEmail-submit">
-                      {pristine ||
-                      hasValidationErrors ||
-                      (hasSubmitErrors && !dirtySinceLastSubmit) ? (
-                        <Button
-                          variant="outlined"
-                          type="submit"
-                          color="regular"
-                          disabled
-                        >
-                          <ButtonIcon>save</ButtonIcon>
-                          <span>Save</span>
-                        </Button>
-                      ) : (
-                        <Button variant="filled" type="submit" color="primary">
-                          <ButtonIcon>save</ButtonIcon>
-                          <span>Save</span>
-                        </Button>
-                      )}
+                      <Button
+                        variant={
+                          preventSubmit(formProps) ? "outlined" : "filled"
+                        }
+                        type="submit"
+                        color={preventSubmit(formProps) ? "regular" : "primary"}
+                        disabled={preventSubmit(formProps)}
+                      >
+                        <ButtonIcon>save</ButtonIcon>
+                        <span>Save</span>
+                      </Button>
                     </Localized>
                   </Flex>
                 </form>
--- a/src/core/client/stream/tabs/Profile/ChangeEmail/ChangeEmailContainer.tsx
+++ b/src/core/client/stream/tabs/Profile/ChangeEmail/ChangeEmailContainer.tsx
@@ -2,10 +2,12 @@ import { FORM_ERROR, FormApi } from "final-form";
 import { Localized } from "fluent-react/compat";
 import React, { FunctionComponent, useCallback, useState } from "react";
 import { Field, Form } from "react-final-form";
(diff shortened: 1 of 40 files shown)


src/core/client/stream/tabs/Profile/ChangeEmail/ChangeEmailContainer.tsx
```diff
@@ -0,0 +1,310 @@
...
+                      <Button type="button" onClick={toggleEditForm}>
+                        Cancel
+                      </Button>
+                    </Localized>
+                    <Localized id="profile-changeEmail-submit">
+                      {pristine ||
+                      hasValidationErrors ||
+                      (hasSubmitErrors && !dirtySinceLastSubmit) ? (
+                        <Button
+                          variant="outlined"
```
wyattjoh: I would also most suggest instead embedding the:

```pristine || hasValidationErrors || (hasSubmitErrors && !dirtySinceLastSubmit)```

Into a function that can be used to switch the variant/color on the button instead of switching the button itself.

File: src/core/server/models/user/user.ts [hunk 2]
```diff
@@ -606,6 +606,68 @@ export async function verifyUserPassword(
...
+export async function updateUserProfileEmail(
+  mongo: Db,
+  tenantID: string,
+  id: string,
+  emailAddress: string
+) {
+  const email = emailAddress.toLowerCase();
+
+  // Try to see if this local profile already exists on a User.
+  const existingUser = await retrieveUserWithProfile(mongo, tenantID, {
```
wyattjoh: This would be considered business logic that should instead live in the `services/users`.

src/core/server/models/user/user.ts
```diff
@@ -606,6 +606,68 @@ export async function verifyUserPassword(
...
+    type: "local",
+    id: email,
+  });
+  if (existingUser) {
+    throw new DuplicateEmailError(email);
+  }
+  // Update the user with the new email.
+  const result = await collection(mongo).findOneAndUpdate(
```
wyattjoh: You should wrap this in a `try/catch` that checks the error like:

https://github.com/coralproject/talk/blob/next/src/core/server/models/user/user.ts#L500

wyattjoh (changes requested):
Some feedback for you!

src/core/server/graph/tenant/mutators/Users.ts
```diff
@@ -140,7 +142,18 @@ export const Users = (ctx: TenantContext) => ({
...
+    updateEmailByID(ctx.mongo, ctx.tenant, input.userID, input.email),
+  updateEmail: async (input: GQLUpdateEmailInput) =>
+    updateEmail(
+      ctx.mongo,
+      ctx.tenant,
+      ctx.mailerQueue,
+      ctx.config,
+      ctx.signingConfig!,
+      ctx.user!,
+      input.email,
```
wyattjoh:
```suggestion
      input.email.toLowerCase(),
```

src/core/server/graph/tenant/mutators/Users.ts
```diff
@@ -140,7 +142,18 @@ export const Users = (ctx: TenantContext) => ({
       ctx.user!
     ),
   updateUserEmail: async (input: GQLUpdateUserEmailInput) =>
-    updateEmail(ctx.mongo, ctx.tenant, input.userID, input.email),
+    updateEmailByID(ctx.mongo, ctx.tenant, input.userID, input.email),
```
wyattjoh:
```suggestion
    updateEmailByID(ctx.mongo, ctx.tenant, input.userID, input.email.toLowerCase()),
```

src/core/server/models/user/user.ts
```diff
@@ -622,50 +622,52 @@ export async function updateUserProfileEmail(
...
+
+      throw new Error("an unexpected error occurred");
     }
 
-    throw new Error("an unexpected error occurred");
+    return result.value || null;
+  } catch (err) {
+    if (err instanceof MongoError && err.code === 11000) {
+      // Check if duplicate index was about the email.
+      if (err.errmsg && err.errmsg.includes("tenantID_1_email_1")) {
```
wyattjoh: This is a different index (unfortunately, I don't know the correct one), but the only reason a `11000` could be thrown in this context is a uniqueness violation on the email, so I'd just throw the `DuplicateEmailError` for any error with code `11000`.

src/core/server/errors/index.ts
```diff
@@ -679,3 +679,11 @@ export class PasswordIncorrect extends CoralError {
     });
   }
 }
+
+export class CannotUpdateLocalProfile extends CoralError {
```
wyattjoh: I wouldn't bother throwing a new `CoralError` for this, as it's only going to occur if a user tries to call this API directly. Throwing `new Error("cannot update profile, settings disallowed")` or something should be fine.

src/core/server/models/user/user.ts
```diff
@@ -965,44 +901,45 @@ export async function updateUserEmail(
...
-  if (!result.value) {
-    // Try to get the current user to discover what happened.
-    user = await retrieveUser(mongo, tenantID, id);
-    if (!user) {
-      throw new UserNotFoundError(id);
+    return result.value;
+  } catch (err) {
+    if (err instanceof MongoError && err.code === 11000) {
+      // Check if duplicate index was about the email.
+      if (err.errmsg && err.errmsg.includes("tenantID_1_email_1")) {
```
wyattjoh: Reference above mentioned comment on the error message.

src/core/client/stream/tabs/Profile/ChangeUsername/ChangeUsernameContainer.tsx
```diff
@@ -50,14 +51,29 @@ const FREQUENCYSCALED = reduceSeconds(ALLOWED_USERNAME_CHANGE_FREQUENCY, [
...
 
 interface FormProps {
   username: string;
   usernameConfirm: string;
 }
 
-const ChangeUsernameContainer: FunctionComponent<Props> = ({ viewer }) => {
+type AuthIntegrations = "local" | "sso" | "google" | "facebook" | "oidc";
+
+function enabledAuthenticationIntegrations(tenant: SettingsData): string[] {
```
tessalt: ok so I have this function on both components (change email and change username) and I'm not sure how to share this logic without things getting fairly tangled with types and imports. The function needs to know what a tenant is, but `SettingsData` is specific to the fragment requested in each component, so I'm not sure how to abstract this out. On the server side there are generic types shared across the codebase but on the client the types are all defined based on the fragments, and these two components don't have the same fragments. I was thinking that I'd need to have a separate component that just does the show/not show logic that requests the specific fragment it needs? but then it seems like overkill to have to pass that fragment and its type around between this component and the next one up. Any thoughts?

src/core/server/locales/en-US/errors.ftl
```diff
@@ -55,5 +55,4 @@ error-inviteTokenExpired = Invite link has expired.
 error-inviteRequiresEmailAddresses = Please add an email address to send invitations.
 error-passwordIncorrect = Password provided was incorrect.
 error-usernameAlreadyUpdated = You may only change your username once every { framework-timeago-time }.
-error-persistedQueryNotFound = The persisted query with ID { $id } was not found.
-error-rawQueryNotAuthorized = You are not authorized to execute this query.
```
wyattjoh: This looks like a merge error.

wyattjoh (changes requested):

wyattjoh (approved):

PR #2470, opened 2019-08-13
[CORL-476] Add badges to user from SSO token

tessalt:
## What does this PR do?
- adds a new "badges" field for Users
- pulls badges from SSO token if present
- displays badges in comment stream and user drawer

## How to test?
- generate an SSO token with `badges: ["badgeName"]` added to the user claim and log in


wyattjoh (changes requested):
Ensure you update the `README.md` to reflect the new `user.badges` parameter available in the token.

tessalt:
readme updated!

wyattjoh (approved):

PR #2473, opened 2019-08-13
[CORL-529] Proxy

wyattjoh:
## What does this PR do?

Adjust the parsing of the proxy parsing options to better allow configuration based on the express documentation.


immber (approved):
👍

PR #2475, opened 2019-08-14
[CORL-476] add user role from SSO token

tessalt:
## What does this PR do?
- Assign SSO user a role based on role passed in on sso token

## How do I test this PR?
- generate an SSO token with `role: GQLUSER_ROLE` added to user claim

PR is vs CORL-172 because I'm pretty sure they'd conflict otherwise. Will change base branch to next after 172 is merged. 

Question: can we assume role will always be passed through in uppercase or should we allow lowercase and transform?

### DIFF
--- a/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
+++ b/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
@@ -133,10 +133,7 @@ export async function findOrCreateSSOUser(
       {
         id,
         username,
-        role:
-          role && Object.values(GQLUSER_ROLE).includes(role)
-            ? role
-            : GQLUSER_ROLE.COMMENTER,
+        role: role || GQLUSER_ROLE.COMMENTER,
         badges,
         email,
         profiles: [profile],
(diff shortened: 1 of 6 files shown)


src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
```diff
@@ -131,7 +133,10 @@ export async function findOrCreateSSOUser(
       {
         id,
         username,
-        role: GQLUSER_ROLE.COMMENTER,
+        role:
+          role && Object.values(GQLUSER_ROLE).includes(role)
```
wyattjoh: I'd prefer the role validating moving to the Joi model as mentioned above.

src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
```diff
@@ -60,8 +61,9 @@ export const SSOUserProfileSchema = Joi.object()
       .required(),
     username: Joi.string().required(),
     badges: Joi.array().items(Joi.string()),
+    role: Joi.string(),
```
wyattjoh: You can add a role specifier here:

```suggestion
    role: Joi.string().only(Object.values(GQLUSER_ROLE)),
```

wyattjoh (changes requested):
Ensure you update the `README.md` to reflect the new `user.role` parameter available in the token.

wyattjoh (approved):

PR #2477, opened 2019-08-15
[CORL-503] More stable CSS classNames

alinhle:
## What does this PR do?
- Adds more stable CSS class names see `src/core/client/stream/classes.ts`
- Attach `variant` and `color` data attributes to `buttons` so that they can be addressed by css selectors.

## Breaking custom styles
- Removed `.coral-createComment-box`, use `.coral-createComment .coral-rte` instead or any of the more specific rte selectors like `.coral-rte-content`.
- `.coral-tabBar-allComments` is now `.coral-tabBar-comments`


cristiandean:
Hey @cvle, great job!

What do you think if we had a class which references whether the user is authenticated or not? Maybe it would be a root class, because sometimes we need to change the styles based on that.

If you want I can make a PR with that.

kgardnr:
@cvle we shouldn't change classnames, we should support both and then deprecate over time cc @wyattjoh

wyattjoh:
Considering we're still in the beta @kgardnr, I think we should permit the swapping around of the class names as long as we're clear in the release notes about the change. Once we release the golden 5.0.0 I completely agree.

alinhle:
@cristiandean thanks for pointing that out, it will follow in another PR!

wyattjoh (approved):

PR #2478, opened 2019-08-15
[CORL-505] Perspective API Strict Config

wyattjoh:
## What does this PR do?

Adds some more validation and strict control over the configuration passed for the perspective API URL attributes.


immber (commented):
Confirmed that when the value of `tenants.integrations.perspective.endpoint` is `NULL` or correctly set to a string, the unescaped char error is not thrown!

kgardnr (approved):

PR #2479, opened 2019-08-15
[CORL-500] Install Copy

wyattjoh:
## What does this PR do?

Adjusted the copy for the installations screen.

## TODO:

- [ ] provide es translation


kgardnr (approved):

kgardnr:
@wyattjoh I'll do missing es translations later, it's not priority since we don't have any es newsrooms on v5 yet

PR #2482, opened 2019-08-16
[CORL-543] Persisted Query Fix

wyattjoh:
## What does this PR do?

Handles case where the `src/core/server/graph/tenant/persisted/__generated__` folder is not generated, which occurs in development.

## How do I test this PR?

```sh
rm -rf src/core/server/graph/tenant/persisted/__generated__
npm run start:development
```

Notice no error! Also notice the nice warning in the logs mentioning that no persisted queries were loaded.


tessalt (approved):
yay

PR #2484, opened 2019-08-16
[CORL-548] Cache Issue

wyattjoh:
## What does this PR do?

Hash size was too small and was causing caching bugs, this expands the hash size to avert the issue.


immber (approved):
This looks like its resolved the issue! 👍

PR #2487, opened 2019-08-19
[CORL 490] Don't show logout link for sso

tessalt:
## What does this PR do?
Literally enforces the rule "When Coral auth is disabled and SSO is enabled for stream, don't show logout" but I'm not 100% sure that's what it should be doing. 

## How do I test this PR?
- enable sso for tenant
- disable local login for tenant
- sign in with sso
- you should not see "Not you? Sign out" link on comments tab

### DIFF
--- a/src/core/client/stream/common/UserBox/UserBoxContainer.tsx
+++ b/src/core/client/stream/common/UserBox/UserBoxContainer.tsx
@@ -49,7 +49,7 @@ export class UserBoxContainer extends Component<Props> {
       this.props.settings.auth,
       "stream"
     );
-    return integrations.length === 1  && integrations[0] === "sso";
+    return integrations.length === 1 && integrations[0] === "sso";
   }
 
   private get supportsLogout() {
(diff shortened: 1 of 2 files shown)


src/core/client/stream/common/UserBox/UserBoxContainer.tsx
```diff
@@ -43,11 +44,20 @@ export class UserBoxContainer extends Component<Props> {
   private handleRegister = () => this.props.showAuthPopup({ view: "SIGN_UP" });
   private handleSignOut = () => this.props.signOut();
 
+  private get authTypePreventsLogout() {
+    const integrations = getAuthenticationIntegrations(
+      this.props.settings.auth,
+      "stream"
+    );
+    return integrations.includes("sso") && !integrations.includes("local");
```
wyattjoh: This should instead be:

```suggestion
    return integrations.length === 1  && integrations[0] === "sso";
```

wyattjoh (changes requested):

wyattjoh (approved):

PR #2493, opened 2019-08-21
[CORL-489] change routing for auth and auth/callback

tessalt:
## What does this PR do?
Fixes https://vmproduct.atlassian.net/browse/CORL-489?atlOrigin=eyJpIjoiZWY1M2MyNzRhNjc0NDkxNDkxYTg4MDhjMzBlZWQ3YmEiLCJwIjoiaiJ9

This was a weird one. Turns out the route /embed/auth/callback was actually displaying the html and js for /embed/auth, because of two separate bugs in dev and in prod. 

## How do I test this PR?
- enable google or fb login
- attempt to log in to admin via google or fb

### DIFF
--- a/src/core/server/app/router/client.ts
+++ b/src/core/server/app/router/client.ts
@@ -122,19 +122,19 @@ export function mountClientRoutes(
     })
   );
   router.use(
-    "/embed/auth",
-    createClientTargetRouter({
-      staticURI,
-      cacheDuration: false,
-      entrypoint: entrypoints.get("auth"),
-    })
-  );
-  router.use(
     "/embed/auth/callback",
     createClientTargetRouter({
       staticURI,
       cacheDuration: false,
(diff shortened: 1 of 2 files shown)


src/core/server/app/router/client.ts
```diff
@@ -122,19 +122,19 @@ export function mountClientRoutes(
     })
   );
   router.use(
-    "/embed/auth",
+    "/embed/auth/callback",
```
tessalt, line 125: because `router.use` delegates all routes matching the top-level path (in this cause `/auth`), the `/embed/auth/callback` route was hitting the router for `/auth` instead of `/auth/callback`. Changing the order should fix the problem, but its a bit of a hacky solution.

kgardnr (approved):
Tested and works! Will roll out and confirm on client sites

PR #2495, opened 2019-08-21
[CORL-236] Allow user to delete their account

nick-funk:
## What does this PR do?

Adds options to the My Profile > Settings area in the comment stream to allow a user to schedule deletion of their account.

## How do I test this PR?

- Create/use an account you're cool with being deleted
- Head to comment stream
- Select My Profile > Settings > Request account deletion
- Follow modal steps
- See that account is scheduled for deletion
- Test around for state while in "scheduled for deletion" state

### Forcing a deletion

- Update line 19 in `src/core/server/cron/accountDeletion.ts` to `"0,30 * * * * *"` so that deletions are scheduled every 30 seconds
- Update account using Azure: Cosmos DB (or direct mongo queries) so that scheduled deletion is prior to now (should have already happened)

```
"scheduledDeletionDate": {
    "$date": "2019-05-20T20:40:31.627Z" <-- set months ago to account for any UTC day diffs
  },
```

- Run the server `npm run watch`
- Watch as at some point it says it's deleting the account you scheduled for deletion
- Test around to see that the comments are deleted in both stream and moderation area
- Test around the moderation community area to see that the user account is deleted and user drawer is unavailable to be opened


nick-funk:
More delete account/comment tests incoming, also working on API checks to prevent users from commenting when they are scheduled for extermination.

nick-funk:
API authorization for users who are scheduled to be deleted are completed.
As well, I've added several tests to try and cover the new deleted states and the deletion feature flow.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -28,6 +28,12 @@ enum USER_AUTH_CONDITIONS {
   SUSPENDED is provided when the User is currently under an active suspension.
   """
   SUSPENDED
+
+  """
+  DELETION is provided when the User is scheduled for deletion or actively
```
tessalt: kinda picky but it feels like this should be an adjective instead of a noun to fit with the other conditions. Like `PENDING_DELETION` or something?

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,229 @@
...
+        },
+      }
+    );
+
+    if (
+      !userResult.ok ||
+      userResult.value === null ||
+      userResult.value === undefined
+    ) {
+      logger.info("No users found that were scheduled for deletion.");
```
tessalt: will this log get written even if several users have been deleted in previous loops?

tessalt (commented):
I'm a bit curious about the reasoning for using a loop for the actual user deletion rather than using batches?

src/core/server/cron/accountDeletion.ts
wyattjoh: Maybe `no more users were scheduled for deletion`?

src/core/server/graph/tenant/schema/schema.graphql
nick-funk: Please be as picky as you can, I much prefer comments to no comments. Thank you for noticing this!

src/core/server/cron/accountDeletion.ts
nick-funk: This log will appear once per scheduled deletion operation. That will be either:
- No users found initially, outputs message
- Finished deleting many users, no more users available to delete

src/core/client/stream/tabs/Comments/Stream/DeleteAccount/StreamDeletionRequestCalloutContainer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  Icon,
+  Typography,
+} from "coral-ui/components";
+
+import { StreamDeletionRequestCalloutContainer_viewer } from "coral-stream/__generated__/StreamDeletionRequestCalloutContainer_viewer.graphql";
+
+import CancelAccountDeletionMutation from "coral-stream/mutations/CancelAccountDeletionMutation";
+
+import { SCHEDULED_DELETION_TIMESPAN_DAYS } from "coral-common/constants";
+import { DateTime } from "luxon";
```
wyattjoh: Mind import ordering here.

src/core/client/stream/tabs/Comments/Stream/DeleteAccount/StreamDeletionRequestCalloutContainer.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }, [cancelAccountDeletion]);
+
+  const deletionDate = viewer.scheduledDeletionDate
+    ? formatter(locales, new Date(viewer.scheduledDeletionDate))
+    : null;
+
+  const requestDate = viewer.scheduledDeletionDate
+    ? formatter(
+        locales,
+        DateTime.fromISO(viewer.scheduledDeletionDate)
```
wyattjoh: Not a comment to change, but this is the first time in the client I think we've used luxon before, this adds around 15KB's to the gzipped code.

src/core/client/stream/tabs/Comments/Stream/DeleteAccount/StreamDeletionRequestCalloutContainer.tsx
```diff
@@ -0,0 +1,123 @@
...
+              </Typography>
+            </Localized>
+          </HorizontalGutter>
+          <HorizontalGutter className={styles.gutter}>
+            <Localized
+              id="comments-stream-deleteAccount-callOut-cancelDesc"
+              $date={deletionDate}
+            >
+              <Typography variant="bodyCopy">
+                {`If you would like to continue leaving comments, replies or reactions, you may cancel your request to delete your account before ${deletionDate}.`}
```
wyattjoh: Remove string template, preferring instead the inline version:

```suggestion
                If you would like to continue leaving comments, replies or reactions, you may cancel your request to delete your account before {deletionDate}.
```

src/core/client/stream/tabs/Profile/DeletionRequest/DeletionRequestCalloutContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+  CallOut,
+  Flex,
+  HorizontalGutter,
+  Icon,
+  Typography,
+} from "coral-ui/components";
+
+import { DeletionRequestCalloutContainer_viewer } from "coral-stream/__generated__/DeletionRequestCalloutContainer_viewer.graphql";
+
+import CancelAccountDeletionMutation from "coral-stream/mutations/CancelAccountDeletionMutation";
```
wyattjoh, line 19: Mind import ordering here.

src/core/client/stream/tabs/Profile/DeletionRequest/DeletionRequestCalloutContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+        day: "numeric",
+        hour: "numeric",
+        minute: "numeric",
+        second: "numeric",
+      }).format(new Date(viewer.scheduledDeletionDate))
+    : null;
+
+  return (
+    <>
+      {deletionDate && (
```
wyattjoh: To simplify, I'd suggest prior to the `const deletionDate = viewer.scheduledDeletionDate` line, you just do:

```js
if (!viewer.scheduledDeletionDate) {
  return null;
}
```

Then all the conditional logic after that line goes away.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountContainer.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useCoralContext } from "coral-framework/lib/bootstrap";
+import {
+  graphql,
+  useMutation,
+  withFragmentContainer,
+} from "coral-framework/lib/relay";
+import { Icon, Typography } from "coral-ui/components";
+import { Button } from "coral-ui/components/Button";
+
+import { DeleteAccountContainer_settings } from "coral-stream/__generated__/DeleteAccountContainer_settings.graphql";
```
wyattjoh, line 15: Mind import ordering here.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountContainer.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <Button variant="outlined" size="small" onClick={showPopover}>
+          <Icon size="sm" className={styles.icon}>
+            cancel
+          </Icon>
+          <Localized id="profile-settings-deleteAccount-requestDelete">
+            <span>Request account deletion</span>
+          </Localized>
+        </Button>
+      )}
+      {deletionDate && (
```
wyattjoh: Prefer a nested ternary instead of multiple `&&` statements, reduces readability.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountContainer.tsx
```diff
@@ -0,0 +1,138 @@
...
+        <>
+          <Localized
+            id="profile-settings-deleteAccount-cancelDelete-description"
+            $date={deletionDate}
+          >
+            <Typography variant="bodyCopy" className={styles.section}>
+              {`You have already submitted a request to delete your account.
```
wyattjoh: Same as above, avoid the use of string templates.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountContainer.tsx
```diff
@@ -0,0 +1,138 @@
...
+              Your account will be deleted on ${deletionDate}.
+              You may cancel the request until that time.`}
+            </Typography>
+          </Localized>
+          <Button variant="filled" size="small" onClick={cancelDeletion}>
+            <Localized
```
wyattjoh: Localization wrapper should be around the entirety of the Button to provide the `title` attribute rather than multiple localization wrappers inside the button. Maybe @cvle has some idea to simplify this?

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+  }, [step, setStep, closeModal]);
+
+  return (
+    <>
+      <Modal
+        open={open}
+        onClose={closeModal}
+        data-testid="delete-account-modal"
+      >
+        <Box className={styles.root}>
```
wyattjoh: Suggest breaking the modal contents into a separate component so it's current step can be stored in that component, and you wouldn't need to set it to `0` when you close it (as the Modal content component wouldn't be mounted when the modal is closed).

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/Pages/Common/PageStepBar.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+interface Props {
+  step: number;
+}
+
+const PageStepBar: FunctionComponent<Props> = ({ step }) => {
+  return (
+    <div>
+      <StepBar currentStep={step} className={styles.stepBar}>
+        <Step lineClassName={styles.line}>{""}</Step>
```
wyattjoh: Check with @IAmSamHankins around why this is different than the install wizard component that uses actual string contents inside the step components instead of `{""}`.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/Pages/CompletionPage.tsx
```diff
@@ -0,0 +1,120 @@
...
+          <Typography variant="bodyCopyBold" className={styles.sectionHeader}>
+            Tell us why.
+          </Typography>
+        </Localized>
+        <Localized
+          id="profile-settings-deleteAccount-pages-completeSec2Content"
+          $email={organizationEmail}
+        >
+          <Typography variant="bodyCopy" className={styles.sectionContent}>
+            {`We'd like to know why you chose to delete your account. Send us
```
wyattjoh: Avoid use of string templates inside React components.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/Pages/ConfirmPage.tsx
```diff
@@ -0,0 +1,241 @@
...
+  HorizontalGutter,
+  InputLabel,
+  PasswordField,
+  TextField,
+  Typography,
+} from "coral-ui/components";
+
+import RequestAccountDeletionMutation from "./RequestAccountDeletionMutation";
+
+import PageStepBar from "./Common/PageStepBar";
```
wyattjoh, line 28: Mind import ordering.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/Pages/ConfirmPage.tsx
```diff
@@ -0,0 +1,241 @@
...
+      pristine,
+      hasValidationErrors,
+      hasSubmitErrors,
+      dirtySinceLastSubmit,
+    } = formState;
+    return (
+      pristine ||
+      hasValidationErrors ||
+      (hasSubmitErrors && !dirtySinceLastSubmit)
+    );
```
wyattjoh: Object destructuring is unnecessary.

wyattjoh: Also, this does not need to exist in the render function has it has no shared scope.

src/core/client/ui/components/Steps/Step.tsx
```diff
@@ -12,6 +12,7 @@ interface StepProps {
   completed?: boolean;
   last?: boolean;
   hidden?: boolean;
+  lineClassName?: string;
```
wyattjoh: Other components use a `classes: { line: string }` shape instead.

src/core/server/app/router/client.ts
```diff
@@ -138,6 +138,15 @@ export function mountClientRoutes(
     })
   );
 
+  router.use(
```
wyattjoh: This should not be required.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,223 @@
...
+  mongo: Db,
+  mailer: MailerQueue
+): CronJob {
+  const job = new CronJob({
+    cronTime: "0,30 * * * *",
+    timeZone: "America/New_York",
+    start: true,
+    runOnInit: false,
+    onTick: async () => {
+      deleteScheduledAccounts(mongo, mailer);
```
wyattjoh: Instead, turn `deleteScheduledAccounts` into a function that receives the mongo and mailer references and returns a `CronCommand`.

```js
function deleteScheduledAccounts(mongo: Db, mailer: MailerQueue): CronCommand {
  return async () => {
```

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,223 @@
...
+  if (job.running) {
+    logger.info("Account deletion scheduler now running.");
+  }
+
+  return job;
+}
+
+async function deleteScheduledAccounts(mongo: Db, mailer: MailerQueue) {
+  logger.info(
+    "Scheduled account deletion: checking for accounts that require deletion..."
```
wyattjoh: Log messages should never contain punctuation or specifiers, file location and log line content provides enough specificity.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,223 @@
...
+        $set: {
+          scheduledDeletionDate: rescheduledDeletionDate,
+        },
+      }
+    );
+
+    if (
+      !userResult.ok ||
+      userResult.value === null ||
+      userResult.value === undefined
```
wyattjoh: Simpler to check `if (!userResult.value)`.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,223 @@
...
+    ) {
+      logger.info("No more users were scheduled for deletion.");
+      break;
+    }
+
+    const userToDelete = userResult.value;
+
+    logger.info(`Deleting ${userToDelete.username}...`);
```
wyattjoh: Never use string templates in log messages, add structured fields if necessary.

src/locales/da/framework.ftl
```diff
@@ -31,7 +31,7 @@ framework-validation-emailsDoNotMatch = E-mails stemmer ikke overens. Prøv igen
 framework-validation-notAWholeNumberBetween = Indtast et helt tal mellem { $min } og { $max }.
 framework-validation-notAWholeNumberGreaterThan = Indtast et helt tal større end { $x }
 framework-validation-notAWholeNumberGreaterThanOrEqual = Indtast et helt tal større end eller lig med { $x }
-
+framework-validation-deleteConfirmationInvalid = Incorrect confirmation. Try again.
```
wyattjoh: Remove english translation.

src/core/server/services/users/index.ts
```diff
@@ -312,6 +315,92 @@ export async function updatePassword(
...
+  mongo: Db,
+  mailer: MailerQueue,
+  tenant: Tenant,
+  user: User
+) {
+  if (!user.email) {
+    throw new EmailNotSetError();
+  }
+
+  const updatedUser = await clearDeletionDate(mongo, tenant.id!, user.id);
```
wyattjoh: Remove `!` on tenant's ID.

src/core/server/services/users/index.ts
```diff
@@ -312,6 +315,92 @@ export async function updatePassword(
...
+  });
+
+  const updatedUser = await scheduleDeletionDate(
+    mongo,
+    tenant.id!,
+    user.id,
+    deletionDate.toJSDate()
+  );
+
+  const formattedDate = Intl.DateTimeFormat([tenant.locale], {
```
wyattjoh: Array isn't required:

```suggestion
  const formattedDate = Intl.DateTimeFormat(tenant.locale, {
```

src/core/server/models/user/user.ts
```diff
@@ -671,6 +682,63 @@ export async function updateUserPassword(
...
+
+  return result.value || null;
+}
+
+export async function clearDeletionDate(
+  mongo: Db,
+  tenantID: string,
+  userID: string
+) {
+  const users = collection(mongo);
```
wyattjoh: Avoid creating collection variables like this, prefer inline use of `collection(mongo)`.

src/core/server/models/user/user.ts
```diff
@@ -671,6 +682,63 @@ export async function updateUserPassword(
...
+      id: userID,
+      tenantID,
+    },
+    {
+      $set: {
+        scheduledDeletionDate: deletionDate,
+      },
+    },
+    {
+      returnOriginal: false,
```
wyattjoh, line 702: Ensure you retains the comment about the `returnOriginal` variable, as it's pretty unclear.

src/core/server/models/user/user.ts
```diff
@@ -671,6 +682,63 @@ export async function updateUserPassword(
...
+  const users = collection(mongo);
+  const result = await users.findOneAndUpdate(
+    {
+      id: userID,
+      tenantID,
+    },
+    {
+      $set: {
+        scheduledDeletionDate: null,
+      },
```
wyattjoh, line 726: Prefer unset if possible.

src/core/server/index.ts
```diff
@@ -307,6 +310,8 @@ class Server {
     );
 
     logger.info({ port }, "now listening");
+
+    this.scheduledTasks = startCronJobs(this.mongo, this.tasks.mailer);
```
wyattjoh: Saving variables in `scheduledTasks` yet function is called `startConJobs`?

wyattjoh: Also, prefer object for variables rather than positional ones (requirements likely to expand).

src/core/server/graph/tenant/resolvers/Mutation.ts
```diff
@@ -185,12 +185,19 @@ export const Mutation: Required<GQLMutationTypeResolver<void>> = {
...
+  requestUserCommentsDownload: async (source, { input }, ctx) => ({
     archiveURL: await ctx.mutators.Users.requestUserCommentsDownload(input),
     clientMutationId: input.clientMutationId,
   }),
+  requestAccountDeletion: async (source, { input }, ctx) => ({
+    user: await ctx.mutators.Users.requestAccountDeletion(input),
+    clientMutationId: input.clientMutationId,
+  }),
+  cancelAccountDeletion: async (source, { input }, ctx) => ({
+    user: await ctx.mutators.Users.cancelAccountDeletion(input),
```
wyattjoh, line 201: Missing `clientMutationId`.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,223 @@
...
+  const user = await users.findOne({ id });
+  if (!user) {
+    logger.warn(`Unable to delete user ${id} as they could not be found.`);
+    return;
+  }
+
+  const tenant = await tenants.findOne({ id: user.tenantID });
+  if (!tenant) {
+    logger.warn(
+      `Unable to delete user ${id} as we could not find their tenant.`
```
wyattjoh: Never use string templates in log entries.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,223 @@
...
+      },
+    }
+  );
+}
+
+async function deleteUser(db: Db, mailer: MailerQueue, id: string, now: Date) {
+  const users = createCollection<User>("users")(db);
+  const tenants = createCollection<Tenant>("tenants")(db);
+
+  const user = await users.findOne({ id });
```
wyattjoh: Never query for any collection without the tenant ID.

wyattjoh (changes requested):

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/Pages/Common/PageStepBar.tsx
nick-funk: It uses strings as it allows us to optionally use material icons inside the step circle elements.

This allows us to show a difference between "completed and checked step" and "inside step but not completed", both with appropriate colouring.

nick-funk: I've updated the Step to be able to take an optional `children`.
This allows it to show a child React element, string, or nothing.

Since we optionally use nothing when the steps aren't fulfilled, this removes the silly looking `{""}` child element.

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountContainer.tsx
nick-funk: I tried taking it outside to encapsulate the button, it does as you might expect, setting the contents with the localized text, but blowing away the child icon.

Looking elsewhere in the code, we seem to do this in other places, sometimes localizing around both the icon and the inner text.

I can do another localization that sets the title for the button if you want? I can also build an icon control for the button, or an `IconButton`. Lots of options, up to you what you'd like.

Would be good to get a standard on this, if you decide what you want to do, I will add to `docz` docs.

src/core/server/models/user/user.ts
nick-funk: I'm not sure I agree with this comment.

For one, it reads more like this in TS: `const users = createCollection<User>("users");`

Secondly, I'm doing a query and then some updates. I don't want multiples of `createCollection<User>("users");` everywhere.

Does that make sense? Or maybe I’m missing something?

src/core/client/stream/tabs/Comments/Stream/DeleteAccount/StreamDeletionRequestCalloutContainer.tsx
nick-funk: Removed for now with equivalent native JS Date logic.
When you decide whether you want Luxon in there and if that 15KB is worth it, we can bring it back in.

src/core/server/app/router/client.ts
nick-funk: Fixed

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/Pages/ConfirmPage.tsx
nick-funk: I'm not sure what you mean by object de-structuring not being necessary, it's a nice way to pull out all the FormProps and evaluate them.

As for not being in the render function, it seems to work appropriately, I also modeled this off of our recent changes in ChangeEmailContainer. Was this also incorrect?

I don't see another way to disable the Delete button on this form without this utility function?

src/core/client/stream/tabs/Profile/Settings/DeleteAccount/DeleteAccountContainer.tsx
wyattjoh: Yeah, for this one, I'd sync with @cvle if you can to find a good solution!

nick-funk: Yup, we're on it!

nick-funk: @cvle has indicated that we shouldn't be using `title` attribute for this use case.

If the button text is unclear, or the button is an icon on its own, we should use `aria-label` on the button for accessibility support.

Specifically, in this case, he suggests that we leave the internal `Localized` for the text, but remove the `Localized` around the `Icon`.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,221 @@
...
+  return job;
+}
+
+function deleteScheduledAccounts(mongo: Db, mailer: MailerQueue): CronCommand {
+  return async () => {
+    logger.info("Checking for accounts that require deletion...");
+
+    const users = createCollection<User>("users");
+
+    while (true) {
```
wyattjoh: Add `TODO` to iterate over tenants.

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  return job;
+}
+
+function deleteScheduledAccounts(mongo: Db, mailer: MailerQueue): CronCommand {
+  return async () => {
+    logger.info("Checking for accounts that require deletion...");
+
+    const users = createCollection<User>("users");
```
wyattjoh: Replace with something like:

```ts
const collection = (mongo: Db) => ({
  users: createCollection<User>("users")
});

// ...

// await collection(mongo).users.findOneAnyModify
// await collection.users(mongo).findOneAnyModify
```

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,221 @@
...
+  if (job.running) {
+    logger.info("Account deletion scheduler now running.");
+  }
+
+  return job;
+}
+
+function deleteScheduledAccounts(mongo: Db, mailer: MailerQueue): CronCommand {
+  return async () => {
+    logger.info("Checking for accounts that require deletion...");
```
wyattjoh: Include try/catch, and remove punctuation please!

src/core/server/services/users/index.ts
```diff
@@ -312,6 +315,92 @@ export async function updatePassword(
...
+  });
+
+  const updatedUser = await scheduleDeletionDate(
+    mongo,
+    tenant.id!,
+    user.id,
+    deletionDate.toJSDate()
+  );
+
+  const formattedDate = Intl.DateTimeFormat(tenant.locale, {
```
wyattjoh, line 350: Add a `TODO` to extract common formats.

src/locales/en-US/stream.ftl
```diff
@@ -196,6 +218,78 @@ profile-settings-download-comments-recentRequest =
...
+
+profile-settings-deleteAccount-pages-descriptionHeader = Delete my account?
+profile-settings-deleteAccount-pages-descriptionText =
+  You are attempting to delete your account. This means:
+profile-settings-deleteAccount-pages-list1 =
+  All of your comments are removed from this site
+profile-settings-deleteAccount-pages-list2 =
+  All of your comments are deleted from our database
+profile-settings-deleteAccount-pages-list3 =
+  Your username and email address are removed from our system
```
wyattjoh: Confirm with sam on copy.

src/locales/en-US/stream.ftl
```diff
@@ -196,6 +218,78 @@ profile-settings-download-comments-recentRequest =
...
+profile-settings-deleteAccount-pages-completeHeader = Account deletion requested
+profile-settings-deleteAccount-pages-completeDescript =
+  Your request has been submitted and a confirmation has been sent to the email
+  address associated with your account.
+profile-settings-deleteAccount-pages-completeTimeHeader =
+  Your account will be deleted on: { $date }
+profile-settings-deleteAccount-pages-completeSec1Header = Changed your mind?
+profile-settings-deleteAccount-pages-completeSec1Content =
+  Simply sign in to your account again before this time and select
+  <strong>“Cancel Account Deletion Request.”</strong>
```
wyattjoh: Check quotes?

wyattjoh (changes requested):

src/core/server/cron/accountDeletion.ts
```diff
@@ -0,0 +1,221 @@
...
+      );
+
+      if (!userResult.value) {
+        logger.info("No more users were scheduled for deletion.");
+        break;
+      }
+
+      const userToDelete = userResult.value;
+
+      logger.info(`Deleting user...`);
```
wyattjoh:
```suggestion
      logger.info({ userID: userToDelete.id, tenantID: userToDelete.tenantID }, "deleting user");
```

src/locales/en-US/stream.ftl
nick-funk: removed quotes as they look weird

src/locales/en-US/stream.ftl
nick-funk: Sent Sam some screenshots and explained we don't delete usernames.
Updated the message to say we delete email address pre-emptively so that we can get this in and tweak later when Sam can get back to us.

wyattjoh (approved):

PR #2498, opened 2019-08-22
[CORL-498, CORL-495, CORL-539, CORL-496, CORL-494] Email Notifications Support & Framework

wyattjoh:
## What does this PR do?

Provides the base framework that we will build the notifications on top of.

### Migration

```js
db.users.updateMany(
  { notifications: { $exists: false } },
  {
    $set: {
      notifications: {
        onReply: false,
        onFeatured: false,
        onModeration: false,
        onStaffReplies: false,
        digestFrequency: "NONE",
      },
    },
  }
);
db.users.updateMany(
  { digests: { $exists: false } },
  {
    $set: {
      digests: [],
    },
  }
);
```


src/core/server/cron/notificationDigesting.ts
```diff
@@ -0,0 +1,131 @@
...
+          } else {
+            acc.push({
+              template: entry.template.name,
+              partial: path.basename(entry.template.name),
+              contexts: [entry.template.context],
+            });
+          }
+
+          return acc;
+        },
```
tessalt, line 107: this reads kind of weird as a `reduce`, I don't love that elements of `acc` are being mutated but that's probably a bit pedantic!

tessalt (approved):
seems good!

src/core/server/cron/notificationDigesting.ts
wyattjoh: I've typically used `acc` as "accumulated", but totally can change it.

tessalt: oh i get that, me too, I just mean the fact that the array items are being mutated

nick-funk (approved):

nick-funk (approved):

src/core/server/cron/notificationDigesting.ts
wyattjoh: Reduce does not modify the underlying array, it just bases a reduction operation to return a new array. `acc` is the "new" array that we're modifying, which is the point of that part here as we want to.

PR #2500, opened 2019-08-22
[CORL-508] api downloads by user id

tessalt:
## What does this PR do?
- adds a mutation for admins to request archive URL for a user's comment history
- adds a GET request for archive accepting token as query param

## How do I test this PR?

```
mutation requestComments($input: RequestUserCommentsDownloadInput!) {
  requestUserCommentsDownload(input:$input) {
    clientMutationId
    archiveURL
  }
}
```

### DIFF
--- a/src/core/server/app/handlers/api/account/download.ts
+++ b/src/core/server/app/handlers/api/account/download.ts
@@ -27,6 +27,8 @@ export type DownloadOptions = Pick<
   "mongo" | "redis" | "signingConfig" | "config"
 >;
 
+export type AdminDownloadOptions = Pick<AppOptions, "mongo" | "signingConfig">;
+
 async function sendExport(
   mongo: Db,
   tenant: Tenant,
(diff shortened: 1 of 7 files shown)


src/core/server/app/handlers/api/account/download.ts
```diff
@@ -216,6 +218,46 @@ export const downloadHandler = ({
   };
 };
 
+export const adminDownloadHandler = ({
+  mongo,
+  signingConfig,
+}: AdminDownloadOptions): RequestHandler => {
+  return async (req: Request, res, next) => {
```
tessalt, line 225: the only difference between this method and the user-facing one is that it doesn't have the rate-limiting, I'm not mad about the duplicated code, but I could understand if someone was!

src/core/server/app/router/api/account.ts
```diff
@@ -34,6 +35,7 @@ export function createNewAccountRouter(
   router.put("/invite", jsonMiddleware, inviteHandler(app));
 
   router.get("/downloadcheck", downloadCheckHandler(app));
+  router.get("/download", adminDownloadHandler(app));
```
tessalt, line 38: I know @wyattjoh had opinions on the `/downloadcheck` endpoint but I'm happy with `/download?token=` for the GET request, so in that context maybe `/downloadcheck` makes more sense?

src/core/server/app/handlers/api/account/download.ts
nick-funk: I think we follow rule of 3, so this is acceptable.
As well, this is a pretty short handler, easy to mentally grok and maintain. :)

src/core/server/app/router/api/account.ts
nick-funk: Alternatively, if we have the requesting user from our auth token, maybe we can check if the user asking to download is an admin.

If they are admin, ignore rate limiting, else, rate limit.

nick-funk (approved):

src/core/server/app/router/api/account.ts
tessalt: I was just thinking a POST request doesn't make as much sense for the admin endpoint

kgardnr:
Can I merge this one @tessalt / @nick-funk ?

tessalt:
@kgardnr should be good!

PR #2502, opened 2019-08-23
update durationfield component to not use lifecycle methods

tessalt:
## What does this PR do?
We only have a couple internal components that use soon-to-be deprecated lifecycle methods, this updates DurationField to use hooks instead. 

## How do I test this PR?
- try adjusting the comment edit window duration from the config panel


wyattjoh (approved):
Made a small tweak, let me know if it looks ok to you!

PR #2503, opened 2019-08-23
update authcheck component to not use to-be-deprecated lifecycle methods

tessalt:
## What does this PR do?
This is the other internal component that uses to-be-deprecated methods, required a bit less refactoring. However I'm a bit more nervous about this one because it's harder to test, I think it's covering a lot of edge cases that I don't really know how to test. 

## How do I test this PR?
- I'd be interested to know...
- I tested it by trying to hit /admin with the wrong privileges, trying to hit /admin/configure with the wrong privileges, trying to hit admin routes with no auth token


wyattjoh (dismissed):

wyattjoh (changes requested):
Visiting http://localhost:8080/admin/configure/general while not logged in does not redirect to a sign in page.

Issue #2512, opened 2019-08-28
[Next] wrong reactions number in the badge


bderen:
Hi there,

on the pages:
/admin/moderate/pending
/admin/moderate/unmoderated

we see a wrong number of reactions in the badge, for example, it says 262 but there is none on the page.

Coral Beta 7

kgardnr:
@sgtram can you check the caching/CDN/server configuration on your end and report back?

kgardnr:
Hey @sgtram we were able to confirm this is a bug, we have a fix for it here going out in the next release: https://github.com/coralproject/talk/pull/2516

PR #2516, opened 2019-09-03
[CORL-572] Filter rejected comments from reply counts on user comment history

nick-funk:
## What does this PR do?

Update the replyCount resolver on comments to filter out rejected and pre-mod status replies.

## How do I test this PR?

- Create a comment using a commenter account
- Have another user reply at least twice
- Have a moderator account reject one of the replying comments
- Head back to the original, root commenter account, check their My Profile > My Comments > Comment History
- Reply count under the comment should not include the rejected comment

### DIFF
--- a/src/core/server/models/comment/comment.ts
+++ b/src/core/server/models/comment/comment.ts
@@ -1135,23 +1135,3 @@ export async function retrieveRecentStatusCounts(
   ]);
   return counts[0];
 }
-
-export async function countPublishedComments(
-  mongo: Db,
-  tenantID: string,
-  commentIDs: string[]
-) {
-  if (commentIDs.length <= 0) {
-    return 0;
-  }
-
-  const count = await collection(mongo)
-    .find({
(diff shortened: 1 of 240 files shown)


kgardnr:
Fixes #2512

src/core/server/models/comment/comment.ts
```diff
@@ -1135,3 +1135,23 @@ export async function retrieveRecentStatusCounts(
...
+    return 0;
+  }
+
+  const count = await collection(mongo)
+    .find({
+      tenantID,
+      id: { $in: commentIDs },
+      status: { $in: PUBLISHED_STATUSES },
+    })
+    .count();
```
wyattjoh: Unfortunately, `.count()` can never be used in MongoDB for performance reasons. We'll have to find an alternative solution.

wyattjoh (changes requested):

wyattjoh (approved):

PR #2517, opened 2019-09-04
[CORL-561, CORL-562] Comment stream and account profile styling cleanup

tessalt:
## What does this PR do?
ok this is a lot of smaller styling and UI changes, mainly: 

- redesign of the account settings tab
- making a lot of typography darker
- adding lines between comments
- new button style that transitions from outline to filled on hover


src/core/client/stream/tabs/Profile/CommentHistory/HistoryComment.css
```diff
@@ -1,13 +1,33 @@
...
+
 .icon {
   color: var(--palette-text-secondary);
   margin-right: calc(0.5 * var(--mini-unit));
 }
 
 .replies {
   composes: button from "coral-ui/shared/typography.css";
+  display: inline-flex;
+  margin-right: var(--spacing-3)!important;
```
alinhle: You seem to use `!important` here to overwrite the gutter that is provided the `Flex` layouting component. You can remove the gutter form the `Flex` Component instead of using the `!important` here.

alinhle:
We should be very careful with using `!important` and only resort to it if absolute necessary.
An excerpt from: https://www.lifewire.com/what-does-important-mean-in-css-3466876.
> 
> The !important directive is helpful when you are testing and debugging a website. If you are not sure why a style is not being applied and think it may be a specificity conflict, add the !important declaration to your style to see if that fixes it — and if it does, change the order of the selectors and remove the !important directives from your production code.
> 
> If you lean too heavily on the !important declaration to achieve your desired styles, you will eventually have a style sheet littered with !important styles. You will be fundamentally changing the way that page's CSS is processed. It is a lazy practice that is not good from a long-term management standpoint.

alinhle:
It'll also make custom css harder to apply their styles as they usually rely on `!important`.

alinhle:
Most ui components exposes a `classes` prop that lets you customize any css styles and might help with avoiding specificity issues.

alinhle:
`margins` sometimes don't apply because they are managed by a Layouting component like `<HorizontalGutter />` or `<Flex itemGutter />` (with gutter turned on). You can remove the layouting component if it's not purposeful.

src/core/client/ui/components/Button/Button.css
```diff
@@ -573,3 +580,33 @@
     }
   }
 }
+
+.variantTextUnderlined {
+  font-weight: 700;
+  &:not(.disabled) {
+    &.colorRegular {
+      &.mouseHover {
+        color: var(--palette-primary-main)!important;
```
alinhle: Is the  `!important` (and the other) here really necessary? My quick tests seems to work also without it?

src/core/client/stream/tabs/Profile/CommentHistory/HistoryComment.css
tessalt: Oh I didn't realize flex had a configurable gutter.

src/core/client/stream/tabs/Comments/Comment/PermalinkButton/PermalinkButton.css
```diff
@@ -2,3 +2,7 @@
   width: 350px;
   max-width: 80%;
 }
+
+.icon {
+  text-decoration: none!important;
```
alinhle: Is the  `!important`(and the others) here really necessary? My quick tests seems to work also without it? Could be me though. Just checking ways not to use it because `!important` should only be a last resort.

src/core/client/ui/components/Button/Button.css
```diff
@@ -573,3 +580,33 @@
     }
   }
 }
+
+.variantTextUnderlined {
+  font-weight: 700;
+  &:not(.disabled) {
+    &.colorRegular {
```
alinhle: Hmm how changing this to `.colorPrimary`? 
And have `.colorRegular` to not change colors (e.g. appear grey).
Following the other variants, regular is grey and only primary gets the primary color.

alinhle: In this case the buttons would need to be changed to have color="primary" then.

nick-funk (approved):

PR #2538, opened 2019-09-09
[CORL-595] move notification settings update button

tessalt:
## What does this PR do?
- moves "update" button to bottom of notifications settings form (was previously floated to the right)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@coralproject/talk",
-  "version": "5.0.0-beta.8",
+  "version": "5.0.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


nick-funk:
I could be wrong, but I think the placement might have been more like this rather than at the very bottom.

image

Making it just aligned under the header and floated right makes it part of that area without forcing it directly underneath the other control items.

Please note, the underline is just for alignment demonstration, I don't think they want a horizontal rule under the header title.

tessalt:
@nick-funk ohh i thought we decided on underneath?

nick-funk:
> @nick-funk ohh i thought we decided on underneath?

I'm not sure anymore. I remember discussing this with Sam and she thought it was a good idea the way I've suggested.

She could have changed her mind and I could be completely wrong on this.

tessalt:
@nick-funk your solution makes more sense wrt the jira ticket 👍

tessalt:
@nick-funk updated to match your illustration!

nick-funk (approved):

wyattjoh (approved):

PR #2543, opened 2019-09-09
[CORL-559] Add new da translations

kgardnr:
## What does this PR do?

- Adds new da contributions

### DIFF
--- a/src/locales/da/admin.ftl
+++ b/src/locales/da/admin.ftl
@@ -174,6 +174,7 @@ configure-email-smtpPasswordLabel = Adgangskode
 
 ### Authentication
 
+configure-auth-authIntegrations = Autentificeringsintegrationer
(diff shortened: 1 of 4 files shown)


src/locales/da/admin.ftl
```diff
@@ -174,6 +174,7 @@ configure-email-smtpPasswordLabel = Adgangskode
 
 ### Authentication
 
+configure-auth-authIntegrations = Autentificeringsintegrationer
```
wyattjoh, line 177: Can you double check this translation?

wyattjoh (changes requested):

src/locales/da/admin.ftl
wyattjoh: Confirmed that it is correct!

wyattjoh (approved):

PR #2544, opened 2019-09-09
[CORL-523] Document GraphQL mutations for managing accounts

nick-funk:
## What does this PR do?

- Adds documentation for how to perform GraphQL queries and mutations as well as describing some critical GraphQL mutations for account management when using SSO. 
- Adds a `deleteUserAccount` mutation to allow an administrator token to delete an account without a waiting period.


nick-funk:
@wyattjoh, I am unsure how to test the admin API download comments endpoint.

Would you mind including the instructions for doing so to the README as I think you were the one who wrote that functionality?

wyattjoh:
So @nick-funk, I think that the original JIRA ticket was a bit unclear, so this is a revision of that.

I've documented the preexisting GraphQL mutations available for account management that can be used by integrators when attempting to comply with GDPR. I've assigned @cvle to review the changes, as I also had to introduce a new GraphQL mutation for deleting a user account immediately as I noticed we missed that one.

src/core/server/services/users/delete.ts
```diff
@@ -0,0 +1,180 @@
...
+  const tenant = await collections.tenants(mongo).findOne({ id: tenantID });
+  if (!tenant) {
+    throw new Error("could not find tenant by ID");
+  }
+
+  // Delete the user's action counts.
+  await deleteUserActionCounts(mongo, userID, tenantID);
+
+  // Delete the user's comments.
+  await deleteUserComments(mongo, userID, tenantID);
```
alinhle, line 158: Just something to think about later: If the database breaks down in the middle of these operations, we'll have an inconsistent state right?

README.md
```diff
@@ -364,6 +392,207 @@ npm run docz -- dev
...
+curl --request POST \
+  --url "http://localhost:8080/api/graphql" \
+  --header "content-type: application/json" \
+  --data '{"query":"query GetComments($url: String!) {story(url: $url) { id metadata { title } url title comments { nodes { id body author { id username } } } } }","variables":{"url":"http://localhost:8080/"},"operationName":"GetComments"}'
+```
+
+When you unpack that, it's really quite simple. We're executing a `POST` request
+to the `/api/graphql` route of the local Talk server with the GraphQL
+request we want to make. It's composed of the `query`, `variables`, and
+`operationName`.
```
alinhle, line 419: Is it even necessary to mention `operationName` here? It seems like an edge case. It's an optional value only required when having multiple queries, which I think is an  edge case no?

alinhle (approved):

README.md
alinhle: I see it's helpful for debugging.

src/core/server/services/users/delete.ts
nick-funk: Unfortunately, yes, this is true. We don't have a transaction-esque way of doing this right now.

PR #2546, opened 2019-09-09
[CORL-527] English copy tweaks

kgardnr:
## What does this PR do?
- Clarifies some of the English copy


src/locales/en-US/admin.ftl
```diff
@@ -145,18 +145,18 @@ configure-general-commentEditing-seconds = Seconds
...
   Your organization name will appear on emails sent by { -product-name } to your community and organization members.
 configure-organization-email = Organization email
 configure-organization-emailExplanation =
-  This email address will be used as in emails and across
-  the platform for community members to get in touch with
-  the organization should they have any questions about the
-  status of their accounts or moderation questions.
+  Open or close comment streams for new comments sitewide. When new comments
+  are turned off sitewide, new comments cannot be submitted, but existing
+  comments can continue to receive “Respect” reactions, be reported, and be
```
alinhle: Maybe leave the `"Respect"` out as it can be customised to anything?

kgardnr: Nice catch! Fixing now

alinhle (approved):

PR #2555, opened 2019-09-10
[CORL-554] Un-feature a comment when it is rejected

nick-funk:
Also decrements the featured count for the story if rejected using the
in-stream moderation.

## How to test

- Sign in as a regular user
- Create a comment
- Sign in as moderator
- Go to stream
- Feature the comment
- Reject the comment
- See how comment is rejected (disappears from stream)
- See how featured count goes down by 1


src/core/client/stream/tabs/Comments/Comment/ModerationDropdown/ModerationActionsContainer.tsx
```diff
@@ -48,8 +48,12 @@ const ModerationActionsContainer: FunctionComponent<Props> = ({
     if (!comment.revision) {
       return;
     }
-    reject({ commentID: comment.id, commentRevisionID: comment.revision.id });
-  }, [approve, comment]);
+    reject({
```
wyattjoh: Should use `await` here and make the callback an `async` function.

src/core/server/models/comment/comment.ts
```diff
@@ -805,6 +805,16 @@ export async function updateCommentStatus(
   revisionID: string,
   status: GQLCOMMENT_STATUS
 ): Promise<UpdateCommentStatus | null> {
+  const update: any = {
+    $set: { status },
+  };
+
+  if (status === GQLCOMMENT_STATUS.REJECTED) {
+    update.$pull = {
+      tags: { type: GQLTAG.FEATURED },
```
wyattjoh: This should not be handled in the model, and should instead be handled in the mutation layer as https://github.com/coralproject/talk/blob/master/src/core/server/graph/tenant/mutators/Comments.ts#L169 is, where the comment is checked if it has a featured tag then and then removing it. This would allow you to use the top level function `removeTag` to handle it.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -4063,6 +4063,12 @@ input RejectCommentInput {
   """
   commentRevisionID: ID!
 
+  """
+  storyID is the ID of the story the comment belongs to. This is used to
+  update featured counts if comment was previously featured.
+  """
+  storyID: ID
```
wyattjoh: `storyID` is not needed on the server, so it should not be added.

wyattjoh (changes requested):

src/core/client/stream/tabs/Comments/Comment/ModerationDropdown/RejectCommentMutation.ts
```diff
@@ -11,15 +11,36 @@ import { RejectCommentMutation as MutationTypes } from "coral-stream/__generated
...
+  if (!storyRecord) {
+    return;
+  }
+  const commentCountsRecord = storyRecord.getLinkedRecord("commentCounts");
+  if (!commentCountsRecord) {
+    return;
+  }
+  const tagsRecord = commentCountsRecord.getLinkedRecord("tags");
+  if (tagsRecord) {
+    tagsRecord.setValue(tagsRecord.getValue("FEATURED") - 1, "FEATURED");
```
wyattjoh: Unless I'm misreading this, it decrements the featured count every time a comment is rejected, which isn't the case for all tagged comments (such as staff comments).

nick-funk: Updated to pull back FEATURED comment counts appropriately rather than tryin to increment/decrement.

Went this route, because at this point, the feature flag is already cleared from the comment and the `store`, so we can't do a safe feature flag check before decrement.

wyattjoh (approved):

PR #2559, opened 2019-09-12
[CORL-605] remove user agent margin on safari search bar

tessalt:
The admin search bars looked weird in safari because safari applies a margin-top and margin-bottom of 2px to input elements. We may want to look into resetting this globally, but for now I've just added a specific rule to each text input.

### DIFF
--- a/src/core/client/admin/routes/Community/UserTableFilter.css
+++ b/src/core/client/admin/routes/Community/UserTableFilter.css
@@ -6,6 +6,7 @@
 .textField {
   height: 31px;
   width: calc(45 * var(--mini-unit));
+  margin: 0;
 }
 
 .adornment {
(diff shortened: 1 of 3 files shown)


wyattjoh:
Screen Shot 2019-09-12 at 10 18 26 AM
Screen Shot 2019-09-12 at 10 17 58 AM

I'm still seeing issues in Safari 12.1.2.

wyattjoh (approved):
This does resolve the noted issue, but there are larger problems with these forms as now outlined by the following:

Screen Shot 2019-09-13 at 2 17 31 PM
Screen Shot 2019-09-13 at 2 17 43 PM

We should probably address this separately as we tackle these styles globally.

PR #2561, opened 2019-09-12
[CORL 581] toggleable timestamp

tessalt:
Adds a new Timestamp component in ui/components that wraps the relative timestamp in a button which toggles between absolute and relative time. Removed now unused Timestamp component from admin. updated a loooot of specs. 
timestamp


src/core/client/admin/components/Timestamp.tsx
```diff
@@ -0,0 +1,42 @@
...
+      )}
+    </div>
+  );
+};
+
+Timestamp.defaultProps = {
+  toggleAbsolute: true,
+};
+
+export default Timestamp;
```
wyattjoh: Why does this component exist here and `src/core/client/ui/components/Timestamp/Timestamp.tsx`?

wyattjoh (changes requested):

src/core/client/admin/components/Timestamp.tsx
tessalt: 🤦‍♀ i meant to delete the one in /admin

wyattjoh (approved):

PR #2565, opened 2019-09-12
[CORL-540] Logging improvements

wyattjoh:
This adds improvements to the logging on the server:

- Improved performance using the "simple child" mode
- Added additional context to loggers that are propagated in the application
- Added better error reporting around story creation


src/core/server/app/handlers/api/account/confirm.ts
```diff
@@ -100,11 +100,14 @@ export const confirmRequestHandler = ({
...
-        requestingUserID: requestingUser.id,
-        tenantID: tenant.id,
-      });
+      const log = coral.logger.child(
+        {
+          targetUserID,
+          requestingUserID: requestingUser.id,
+          tenantID: tenant.id,
+        },
+        true
```
nick-funk, line 109: I'm trying to figure out what this `simple` option does, but can't find it in the documentation on the github repo.

src/core/server/models/story/index.ts
```diff
@@ -120,14 +123,14 @@ export interface UpsertStoryInput {
 export async function upsertStory(
   mongo: Db,
   tenantID: string,
-  { id, url }: UpsertStoryInput,
+  { id = uuid.v4(), url }: UpsertStoryInput,
```
nick-funk, line 126: I've noticed we're using `uuid.v1()` in some places and `uuid.v4()` in others. Would it make sense to pick one and standardize across the codebase?

nick-funk (approved):

src/core/server/models/story/index.ts
wyattjoh: `uuid.v1()` is used when a true "secure random" source isn't required, this is for things like log entries and such. This should be preferred because of the lower cost on the underlying machine versus `uuid.v4()`, which has a higher cost to call.

src/core/server/app/handlers/api/account/confirm.ts
wyattjoh: You can see the description here: https://github.com/trentm/node-bunyan/blob/master/lib/bunyan.js#L346

Essentially, it's possible to override options for the logger using `child()`, passing `true` as the second parameter tells bunyan that you're only passing properties to log, which lowers the cost of calling it.

nick-funk: Thanks for clarification, I knew there had to be a reason you were doing this, very good to know!

src/core/server/models/story/index.ts
nick-funk: I was aware of differing cost to generate for v4 versus v1, wasn't aware it was of a large enough impact we were worried about it. I would have used v4 everywhere and not worried about it until it became a bottleneck.

Good to know, will keep in mind for future.

PR #2569, opened 2019-09-13
[CORL-621] Auth Fixes

wyattjoh:
This resolves an issue that happens when authenticating using a single redirect (for newsrooms with only one auth integration enabled, where that integration is an OAuth2 based login). This also improves some of the resizing logic around the popup window.


src/core/client/auth/dom/resizePopup.ts
```diff
@@ -1,9 +1,6 @@
 function resizePopup() {
   const innerHeight = window.document.body.offsetHeight;
-  window.resizeTo(
-    window.outerWidth,
-    innerHeight + window.outerHeight - window.innerHeight
-  );
+  window.resizeTo(350, innerHeight + window.outerHeight - window.innerHeight);
```
nick-funk, line 3: Maybe break 350 out to a named constant? If this is only place this number is used, maybe leave it, but I have a feeling 350 is used elsewhere to initialize a width/height somewhere.

nick-funk (approved):

PR #2571, opened 2019-09-16
[CORL-623] add live updates for comments moving from premod to approved

tessalt:
ok this is an extremely naive implementation but it works so should be a good place to start a discussion. 

I've literally just copied all the code for the `commentCreated` subscription and created a `commentReleased` subscription that gets called when a comments status changes from `PREMOD` to `APPROVED`. 

Questions/conversations:
- does it make sense to have this as a new event, or should i be reusing the commentStatusChanged event? (I chose not to bc that one seemed to be for admins/mods)
- should I be trying to re-use more of the code from the commentCreated event and subscription instead of just copying? 
- should the logic of when to publish the event go in the event itself or stay in the moderate fn like i have it?
- should this be just an entirely new event that replaces `commentCreated` with something like `commentAppearsInStream` or something which gets published to in either situation? 
- i had other questions but I forgot...


src/core/server/services/comments/moderation/index.ts
```diff
@@ -111,6 +112,9 @@ const moderate = (
     result.comment.id,
     input.moderatorID
   );
+  if (result.oldStatus === "PREMOD" && status === "APPROVED") {
```
wyattjoh: I'd check against any unpublished status instead, as `SYSTEM_WITHHELD` could move to `APPROVED`.

src/core/server/services/events/comments.ts
```diff
@@ -57,6 +57,21 @@ export function publishCommentCreated(
   }
 }
 
+export function publishCommentReleased(
+  publish: Publisher,
+  comment: Pick<Comment, "id" | "storyID" | "parentID" | "status">
+) {
+  if (!comment.parentID && hasPublishedStatus(comment)) {
```
wyattjoh, line 64: This seems to filter by only top level comments, is this intentional?

src/core/server/services/comments/moderation/index.ts
tessalt: i was worried about getting double-notified in the case of a stream that is not premod but I guess that wouldn't happen?

src/core/server/services/events/comments.ts
tessalt: yeah because the event for commentCreated filters by top-level so I figured we'd want to keep that experience the same

wyattjoh (approved):

PR #2572, opened 2019-09-16
[CORL-174] premod user

tessalt:
Still needs copy review. 

## What this does
- add "always premoderate user" as an option in the user status dropdown
- when set to true, comments by this user will be flagged as PREMOD and will need to be approved
- includes confirmation modal, filtering by premod status, and account history entires

## How to test
- select "always premoderate user" from the status dropdown for a user
- log in as the premoderated user, attempt to comment on a stream _that isn't already set to premod_
- the comment should be held for moderation


src/core/client/admin/components/UserStatus/UserStatusChange.tsx
```diff
@@ -94,6 +100,32 @@ const UserStatusChange: FunctionComponent<Props> = ({
...
+                  onClick={() => {
+                    onPremod();
+                    toggleVisibility();
+                  }}
+                >
+                  Always Premod User
+                </DropdownButton>
+              </Localized>
+            )}
+            {premod && (
```
wyattjoh: Maybe using `{premod ? <truthy> : <falsy>}` is a bit clearer here?

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -1448,6 +1455,44 @@ type SuspensionStatus {
...
+  """
+  createdBy: User! @auth(roles: [ADMIN, MODERATOR])
+
+  """
+  createdAt is the time the user was set to pre-mod 
+  """
+  createdAt: Time!
+    @auth(
+      roles: [ADMIN, MODERATOR]
+      userIDField: "userID"
```
wyattjoh: The current user doesn't need to know when they were set to premod.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -1448,6 +1455,44 @@ type SuspensionStatus {
   history: [SuspensionStatusHistory!]! @auth(roles: [ADMIN, MODERATOR])
 }
 
+type PremodStatusHistory {
+  """
+  active when true, indicates that the given user is premodded.
+  """
+  active: Boolean!
```
wyattjoh, line 1454: The current user doesn't need to know if they are in premod or not.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -1448,6 +1455,44 @@ type SuspensionStatus {
...
+}
+
+type PremodStatus {
+  """
+  active when true, indicates that the given user is set to pre-mod.
+  """
+  active: Boolean!
+    @auth(
+      roles: [ADMIN, MODERATOR]
+      userIDField: "userID"
```
wyattjoh: The current user doesn't need to know if they are in premod or not.

src/core/server/models/user/user.ts
```diff
@@ -1307,6 +1351,129 @@ async function retrieveConnection(
...
+  );
+  if (!result.value) {
+    // Get the user so we can figure out why the ban operation failed.
+    const user = await retrieveUser(mongo, tenantID, id);
+    if (!user) {
+      throw new UserNotFoundError(id);
+    }
+
+    // Check to see if the user is already banned.
+    const ban = consolidateUserPremodStatus(user.status.premod);
```
wyattjoh: Variable name is `ban`, should be `premod` 😉

wyattjoh (changes requested):

src/core/client/admin/components/UserStatus/UserStatusChange.tsx
```diff
@@ -43,7 +49,7 @@ const UserStatusChange: FunctionComponent<Props> = ({
         <ClickOutside onClickOutside={toggleVisibility}>
           <Dropdown>
             {!banned && (
-              <Localized id="community-userStatus-banUser">
+              <Localized id="community-userStatus-ban">
```
wyattjoh: Ensure you modify the default text (below) to reflect the copy translation change (it still reads `Ban User`).

wyattjoh (approved):

PR #2574, opened 2019-09-17
[CORL-395] Update COMMENT_DETECTED_SPAM marker text to Spam Detected

nick-funk:
Since this is used by Akismet and other spam detection tools, this
wording matches what Wordpress users will expect when seeing comments
detected as spam.


tessalt (approved):

wyattjoh (changes requested):
changes to language keys require new keys, not updating translation text

nick-funk:
> changes to language keys require new keys, not updating translation text

Old keys have been replaced, new keys added with a new name.

PR #2576, opened 2019-09-17
[CORL-395] Update COMMENT_DETECTED_SPAM marker text to Spam Detected

nick-funk:
Since this is used by Akismet and other spam detection tools, this
wording matches what Wordpress users will expect when seeing comments
detected as spam.


src/locales/da/admin.ftl
```diff
@@ -355,6 +355,7 @@ moderate-marker-link = Link
 moderate-marker-bannedWord = Forbudt ord
 moderate-marker-suspectWord = Mistænkte ord
 moderate-marker-spam = Spam
```
wyattjoh, line 357: Old keys should be removed.

nick-funk: It's still being used for the `COMMENT_REPORTED_SPAM` marker flag. If we remove it, that functionality will no longer work.

wyattjoh (approved):

PR #2581, opened 2019-09-20
[CORL-181] Comment Count Injection

alinhle:
- Adds a new script `count.js` that injects comment count into selected elements.

## Docs 

Add the `count.js` script to your `html` tree. On a page using `Coral.createStreamEmbed` this is done for you automatically, however for best performance we recommend to include it into the `<head>` tag.

```html
<script href="//{{ CORAL_DOMAIN_NAME }}/assets/js/count.js" defer></script>
```

> **NOTE:** Replace the value of `{{ CORAL_DOMAIN_NAME }}` with the location of your running instance of Coral.

Insert an html element with the class `coral-count` and configure it using `data-coral-*` attributes. 

```html
<span class="coral-count" data-coral-url="http://example.com/blog-entry-1/"></span>
```

After successful injection it will become:

```html
<span class="coral-count" data-coral-url="http://example.com/blog-entry-1/">
   <span class="coral-count-number">5</span>
  <span class="coral-count-text">Comments</span>
</span>
```

### Available `data-coral-*` attributes

Set the class of your html element to `coral-count` in order to get story counts. The following `data-coral-*` attributes will configure the output:

- `data-coral-id` – The id of the story of which counts should be injected.
- `data-coral-url` – The url of the story of which counts should be injected.
- `data-coral-notext` – If set to `"true"`, only the count number will be injected

Either `data-coral-id` or `data-coral-url` should be set. If none are provided the story url
will be retrieved from the [canonical url reference](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Choosing_between_www_and_non-www_URLs#Using_%3Clink_relcanonical%3E) `<link rel="canonical" href="...">` or inferred using the current page url.

## Test
- Go to https://localhost:8080/count.html for some count integrations
- Go to https://localhost:8080/story.html as it now has counts integrated.

## TODO
- [x] Integrate with coral embed for live comment count
- [x] Docs (separate PR)
- [x] Tests
- [x] Configurable cache settings?
- [x] Auto-inject `count.js` script when using the stream embed.


src/core/common/countQuery.ts
```diff
@@ -0,0 +1,9 @@
+export interface CountQueryArgs {
+  id?: string;
+  url?: string;
+  notext?: boolean;
+}
+
+export function createCountQueryRef(args: CountQueryArgs) {
+  return btoa(`${JSON.stringify(!!args.notext)};${args.id || args.url}`);
```
wyattjoh: This appears to be repeated from `src/core/client/count/index.ts` as well.

Also, while your solution is more compact, wouldn't using `btoa(JSON.stringify(args))` be more readable?

alinhle: Oh the duplicate is actually not used, will remove this function. Otherwise it makes sense to simplify it to your version :-)

alinhle: Oh actually I remember why I did it that way... it was to save characters 🤔

alinhle: I would prefer the current version that saves quite a few characters

wyattjoh (approved):

PR #2583, opened 2019-09-20
[CORl-640] User Registration Race

wyattjoh:
This corrects an issue where a race may cause a user to get a duplicate user error during registration.


tessalt (approved):

src/core/server/models/user/user.ts
```diff
@@ -522,38 +522,54 @@ export async function insertUser(
   }
 
   // Guard against empty login profiles (they need some way to login).
-  if (input.profiles.length === 0) {
-    throw new Error("users require at least one profile");
+  if (input.profiles.length !== 1) {
+    throw new Error("users require one profile");
```
nick-funk: Was saying "users require one profile" intended? This implies that they can have only exactly 1 profile and not multiples.

Unsure if that was the intention of this message.

src/core/server/models/user/user.ts
```diff
@@ -468,7 +468,7 @@ function hashPassword(password: string): Promise<string> {
   return bcrypt.hash(password, 10);
 }
 
-export type InsertUserInput = Omit<
+export type FindOrCreateUserInput = Omit<
```
nick-funk: I like the fact that you renamed this to make it match the functionality it was actually doing.

However, I am not a fan of functions that do multiple things. It tends to violate single responsibility principle.

Particularly, in this case its name implies both returning a user via "Find" and also updating/creating via "CreateUser".

I feel these could maybe be broken into two separate operation tasks and the multi-operation function calls each individually? This adds difficulty, as it's not as atomic (hinders race condition handling) but it also helps mitigate what could be a code smell developing.

nick-funk (changes requested):

nick-funk (approved):

PR #2584, opened 2019-09-20
v4 - Adding live updates for accepted comments

simonjm:
This PR is for opening up the `commentAccepted` subscription for all users. This is very similar to this PR https://github.com/coralproject/talk/pull/2571 for v5.

I'm opening this up as a draft as I'm not sure if this is the best approach. Is there a reason `commentAccepted` was only allowed for admin/moderator users? Should all of this go under a separate subscription? Any feedback is appreciated!


wyattjoh:
Thanks @simonjm for the feedback! We decided not to provide accepted comments as a live subscription because we didn't want to leak information about moderator activity to abusive users which could be used to game the moderation system.

We'll take your feedback back to our team to see if it's something we want to include into core.

simonjm:
Will that still be the case in v5? I didn't notice any restrictions in that PR.

wyattjoh (approved):

PR #2588, opened 2019-09-24
[CORL-469] keyboard shortcuts

tessalt:
Adds a bunch of keyboard shortcuts to the moderation views. 

Makes comments selectable/focusable to allow actions to be taken on comments. Selected comments have a more prominent box shadow. 

- **?** brings up shortcuts modal
- **j** selects next comment
- **k** selects previous comment
- **ctrl-f** opens search
- **1-4** switches to respective queue
- **t** switch to next queue
- **z** toggle single-comment view
- **d** approve comment
- **f** reject comment
- **b** ban comment author (opens ban modal)


package.json
```diff
@@ -98,6 +98,8 @@
     "jsonwebtoken": "^8.3.0",
     "juice": "^5.2.0",
     "jwks-rsa": "^1.3.0",
+    "keymaster": "^1.6.2",
+    "linkify-it": "^2.1.0",
```
alinhle: Was this one added by mistake, or is it used somewhere?

tessalt: it came through in a merge I think? must be something that was added then removed from master? will remove

src/core/client/admin/components/ModerateCard/CommentAuthorContainer.tsx
```diff
@@ -0,0 +1,45 @@
...
+        comment.author.status &&
+        comment.author.status.current.includes(GQLUSER_STATUS.BANNED) && (
+          <Localized id="commentAuthor-status-banned">
+            <div className={styles.authorStatus}>
+              <Tag color="error">BANNED</Tag>
+            </div>
+          </Localized>
+        )}
+    </>
+  );
```
wyattjoh: Rather than using an in-line check here, maybe instead just do something like:

```jsx
const CommentAuthorContainer: FunctionComponent<Props> = ({ comment }) => {
  if (!comment.author || !comment.author.ban.active) {
    return null;
  }

  return (
    <Localized id="commentAuthor-status-banned">
      <div className={styles.authorStatus}>
        <Tag color="error">BANNED</Tag>
      </div>
    </Localized>
  );
};
```

And querying instead directly for the ban status rather than inspecting the current status which is typically used by authentication helpers and not display components.

src/core/client/admin/components/ModerateCard/ModerateCard.tsx
```diff
@@ -83,11 +97,45 @@ const ModerateCard: FunctionComponent<Props> = ({
...
+      key(HOTKEYS.REJECT, id, onReject);
+      key.setScope(id);
+      return () => {
+        key.deleteScope(id);
+      };
+    } else {
+      key.deleteScope(id);
+    }
+    return () => null;
+  }, [selected, comment]);
```
wyattjoh: It looks like this affect uses `id`, shouldn't it also be listed in the effect dependencies?

src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
```diff
@@ -146,6 +168,35 @@ const ModerateCardContainer: FunctionComponent<Props> = ({
...
 
+  const onFocusOrClick = useCallback(() => {
+    if (setSelected) {
+      setSelected();
+    }
+  }, [selected, comment]);
+
+  const handleBanModalClose = useCallback(() => {
+    setShowBanModal(false);
+  }, [setShowBanModal]);
```
wyattjoh: TIL that the dispatch function (like the `setState` method from `useState`) is actually guaranteed to be static, so you can omit the `setShowBanModal` from the dependencies list. This applies to any `useCallback` or `useEffect` hook.

src/core/client/admin/constants.ts
```diff
@@ -1 +1,12 @@
 export const REDIRECT_PATH_KEY = "adminRedirectPath";
+export const HOTKEYS = {
+  NEXT: "J",
```
wyattjoh: Does this imply that you need a capital `j`, `J` to trigger the hotkey?

src/core/client/admin/routes/Moderate/Moderate.tsx
```diff
@@ -24,20 +32,34 @@ const Moderate: FunctionComponent<Props> = ({
...
-  </div>
-);
+}) => {
+  const [showHotkeysModal, setShowHotkeysModal] = useState(false);
+  const closeModal = useCallback(() => {
+    setShowHotkeysModal(false);
+  }, [setShowHotkeysModal]);
+  const openModal = useCallback(() => {
+    setShowHotkeysModal(true);
+  }, [setShowHotkeysModal]);
```
wyattjoh: `setShowHotkeysModal` does not need to be in the dependencies list.

src/core/client/admin/routes/Moderate/ModerateNavigation/Navigation.tsx
```diff
@@ -11,55 +14,95 @@ interface Props {
...
+        router.replace(moderationLinks[i]);
+      });
+    }
+    return () => {
+      key.unbind(HOTKEYS.SWITCH_QUEUE);
+      for (let i = 0; i < moderationLinks.length; i++) {
+        key.unbind(`${i + 1}`);
+      }
+    };
+  }, [match]);
```
wyattjoh: The effect uses `moderationLinks`, which changes based on the `storyID`. It should include `moderationLinks` in the dep array.

src/core/client/admin/routes/Moderate/ModerateSearchBar/Bar.tsx
```diff
@@ -39,13 +46,21 @@ const Bar: FunctionComponent<Props> = ({ title, options, onSearch }) => {
...
     keyboardNavigationHandlers,
   ] = useComboBox("moderate-searchBar-listBoxOption", options);
 
+  useEffect(() => {
+    key(HOTKEYS.SEARCH, () => {
+      if (searchInput && searchInput.current) {
+        searchInput.current.focus();
+      }
+    });
+  }, []);
```
wyattjoh: Because `useRef` will return the same object every render, but `searchInput.current` will return the current referenced object, you should probably tie this dep to the current ref:

```suggestion
  }, [searchInput.current]);
```

src/core/client/admin/routes/Moderate/Queue/Queue.tsx
```diff
@@ -41,6 +43,45 @@ const Queue: FunctionComponent<Props> = ({
...
   const [userDrawerId, setUserDrawerID] = useState("");
+  const [selectedComment, setSelectedComment] = useState<number | null>(0);
+  const [singleView, setSingleView] = useState(false);
+
+  const toggleView = useCallback(() => {
+    if (!singleView) {
+      setSelectedComment(0);
+    }
+    setSingleView(!singleView);
+  }, [singleView, setSingleView, setSelectedComment]);
```
wyattjoh: You can remove `setSingleView` and `setSelectedComment` from the deps here too.

wyattjoh (changes requested):

src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
tessalt: ah ok I've seen conflicting info about this.

tessalt:
updated!

wyattjoh (approved):

PR #2597, opened 2019-09-26
[CORL-649] Migrations

wyattjoh:
This adds support for the new migrations framework within Coral.

When migrations have to be run as a part of code changes, the server will execute those changes automatically. These migrations are also skipped when a new installation is setup.

This PR also adds the user pre-moderation migration as a sample with the requisite changes to the code reflecting the introduction of the new migration framework.

## Creating a new Migration

To create a new migration, you have to run:

```sh
npm run migration:create "<migration name>"
```

And a migration file will be generated.


tessalt (approved):
yay!

nick-funk (approved):

nick-funk (approved):

cristiandean:
Hey @wyattjoh, if we run Talk in distributed mode and there are pending migrations, will each instance perform the migration? If so, should this happen?

wyattjoh:
Hi @cristiandean! That indeed would happen. As we introduce migrations however, they will be written in a way that will allow multiple of them to run in parallel.

For example, if we are adding a feature (as we are in this PR as an example) to add a premod users status to every user, then we are filtering only documents that do not have that field set. This should allow multiple cursors not to conflict during the migration process.

We're still investigating ways to improve the migration process, and are open to any suggestions! Let me know if you have any further questions!

README.md
```diff
@@ -128,7 +128,7 @@ If you don't already have these databases running, you can execute the following
 assuming you have Docker installed on your local machine:
 
 ```bash
-docker run -d -p 27017:27017 --restart always --name mongo mongo:3.6
+docker run -d -p 27017:27017 --restart always --name mongo mongo:4.2
```
nick-funk, line 131: Whoo! 4.2!

cristiandean:
@wyattjoh It seems like a good idea to perform the document filter, my fear is multiple migrations running in parallel with a high volume of data in collections like comments, which is our case. 

I believe that in the future we can think about something like a lock, which can be done in redis or mongo itself with an atomic query once by whoever is responsible for performing the migration. Anyway, it's something that needs to be very well thought out, as it would require failover, retry, integrity checking, etc.

I believe that the way it is already serves us for a while.

Thank you very much for the clarification =)

nick-funk (approved):

tessalt (approved):

wyattjoh:
Thanks for the feedback @cristiandean! We've gone and implemented a sort of lock using MongoDB's atomic operations to help lock things down when it comes to migration time:

https://github.com/coralproject/talk/pull/2597/files#diff-7f9ff7f994b6cac779849b2208300d4eR166-R169

Always open to feedback!

PR #2601, opened 2019-09-30
[CORL-90] moderator notes

tessalt:
Adds a `moderatorNotes` field to `User` and interface for creating and deleting notes in user drawer. 

Screen Shot 2019-09-30 at 9 05 57 AM
Screen Shot 2019-09-30 at 9 08 56 AM


nick-funk (commented):
Went through and code changes look good.

Looks like a migration is needed, assuming that was held off as we need @wyattjoh 's migration PR to be merged in first. Will hold off further testing until then.

tessalt:
@nick-funk @wyattjoh added migration

src/core/client/admin/components/UserHistoryDrawer/UserDrawerNotesContainer.tsx
```diff
@@ -0,0 +1,129 @@
...
+  );
+  const onSubmit = useCallback(
+    async ({ body }, form: FormApi) => {
+      await createNote({
+        userID: user.id,
+        body,
+      });
+      form.reset();
+    },
+    [createNote, user]
```
wyattjoh: You can omit the `createNote` here.

src/core/server/models/user/user.ts
```diff
@@ -2323,3 +2351,78 @@ export async function retrieveUserScheduledForDeletion(
...
+    },
+    {
+      $pull: {
+        moderatorNotes: { id, createdBy },
+      },
+    },
+    {
+      returnOriginal: false,
+    }
+  );
```
wyattjoh, line 2426: Probably good to add the guard clause from the `createModeratorNote` function that returns the `UserNotFoundError` on the falsy response of `findOneAndUpdate`.

src/core/server/services/migrate/migrations/1569947670260_add_moderator_notes_to_user.ts
```diff
@@ -0,0 +1,49 @@
...
+      },
+      "added empty moderatorNotes array"
+    );
+  }
+
+  public async down(mongo: Db, tenantID: string) {
+    const result = await collections.users(mongo).updateMany(
+      {
+        tenantID,
+        $exists: "moderatorNotes",
```
wyattjoh: The syntax for the `$exists` operator is actually:

```suggestion
        moderatorNotes: { $exists: true },
```

wyattjoh (changes requested):

tessalt:
updated

src/core/client/admin/components/UserHistoryDrawer/UserDrawerNotesContainer.css
```diff
@@ -0,0 +1,30 @@
...
+  padding: calc(0.5 * var(--mini-unit));
+  border-radius: 2px;
+  margin-bottom: var(--spacing-2);
+  padding: var(--spacing-3);
+  font-weight: var(--font-weight-regular);
+  font-family: var(--font-family-sans-serif);
+  font-size: calc(16rem / var(--rem-base));
+  line-height: 1;
+  letter-spacing: calc(0.2em / 16);
+  color: var(--palette-text-primary);
```
wyattjoh, line 20: Firefox adds a `border-width: 2px;` when unspecified, perhaps add a `border-width: 1px;` to have it match chrome?

wyattjoh (approved):

PR #2605, opened 2019-10-01
Remove X-Powered-By express header

cristiandean:
http://expressjs.com/en/advanced/best-practice-security.html#at-a-minimum-disable-x-powered-by-header


wyattjoh (approved):
Great catch! Been on my radar for a while as well, just haven't gotten around to it!

PR #2616, opened 2019-10-03
[CORL-668] User Comment Rate Limiting

wyattjoh:
This adds a rate limit to users so that they can only write one comment every 30 seconds.


tessalt (approved):
I feel like its a bit unintuitive that the return value of the update timestamp method is used to determine if a user should be rate limited... but I can't think of a solution that doesn't have a downside. LGTM

PR #2620, opened 2019-10-03
[CORL-681] Configurable Staff Badges

wyattjoh:
This allows the staff badge text to be configured similar to the reactions labels.

### DIFF
--- a/src/core/client/admin/routes/Configure/sections/General/StaffConfig.css
+++ b/src/core/client/admin/routes/Configure/sections/General/StaffConfig.css
@@ -1,0 +1,8 @@
+.textInput {
+  min-width: 300px;
+}
+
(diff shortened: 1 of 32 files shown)


src/core/client/admin/routes/Configure/sections/General/StaffConfig.css
```diff
@@ -0,0 +1,8 @@
+.textInput {
+  min-width: 300px;
+}
+
+.staffTag {
+  display: inline-block;
```
tessalt: this is probably something the flex component could do

tessalt (commented):
one small comment but it doesn't bug me that much

tessalt (approved):

PR #2631, opened 2019-10-08
Remove canChangeStatus condition

cristiandean:
Does this condition make sense? Compare `user id` with `story id`...


wyattjoh (approved):
Good catch... Not sure where that came from.

PR #2632, opened 2019-10-10
[CORL-8] prevent repeat comments

tessalt:
- stores ID of user's most recent comment in redis for 6 hours 
- adds moderation step to check the body of that comment and compare it against the new comment
- throws an error if we're nudging, flags comment and witholds if not
- display "Duplicate comment" flag on comments marked as duplicates

**Question:** the expiration of the last id in redis... it makes sense in this context but I'm wondering if we might want this to be more general-purpose and have the ID _not_ expire and instead compare the createdAt in the moderation step? But then we're storing one entry indefinitely per user in redis and that could be a lot? 

How to test: 
1. log in as commenter on non-premod stream
2. post a comment
3. attempt to post the same comment
4. observe nudge, post comment anyways
5. observe message that comment will be held for moderation
6. log in as admin, comment should be marked "Duplicate comment"


src/core/server/services/comments/pipeline/phases/repeatPost.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+    true
+  );
+
+  // We can only check for spam if this comment originated from a graphql
+  // request via an HTTP call.
+  if (!req) {
+    log.debug("request was not available");
+    return;
+  }
```
wyattjoh: Not required as `req` isn't used in this phase

src/core/server/services/users/users.ts
```diff
@@ -31,6 +32,10 @@ import {
   GQLUSER_ROLE,
 } from "coral-server/graph/tenant/schema/__generated__/types";
 import logger from "coral-server/logger";
+import {
+  Comment as CommentModel,
```
wyattjoh: You should be able to import this as `Comment` instead of aliasing it to `CommentModel`.

src/core/server/services/users/users.ts
```diff
@@ -1232,3 +1244,50 @@ export async function updateUserLastWroteCommentTimestamp(
...
+  user: User,
+  commentID: string
+) {
+  const key = userLastCommentIDKey(tenant, user);
+
+  await redis
+    .multi()
+    .set(key, commentID)
+    .expire(key, COMMENT_REPEAT_POST_TIMESPAN)
+    .exec();
```
wyattjoh: As you aren't depending on the different results, you can instead use:

```js
await redis.set(key, commentID, "EX", COMMENT_REPEAT_POST_TIMESPAN);
```

src/core/server/services/comments/pipeline/phases/repeatPost.ts
```diff
@@ -0,0 +1,93 @@
...
+      tenant,
+      author
+    );
+
+    if (!lastComment) {
+      return;
+    }
+
+    const revision = getLatestRevision(lastComment);
+    const isRepeatComment = revision.body === comment.body;
```
wyattjoh, line 51: Maybe worth it for us to add a `TODO` to possibly use a package that can compare similarity for the future: https://www.npmjs.com/package/damerau-levenshtein

src/core/common/errors.ts
```diff
@@ -316,4 +316,6 @@ export enum ERROR_CODES {
   RAW_QUERY_NOT_AUTHORIZED = "RAW_QUERY_NOT_AUTHORIZED",
 
   USER_ALREADY_PREMOD = "USER_ALREADY_PREMOD",
+
+  REPEAT_POST = "REPEAT_POST",
```
wyattjoh, line 324: Should add a small comment above the constant similar to the other entries in this object.

src/core/common/constants.ts
```diff
@@ -63,3 +63,5 @@ export const COMMENT_LIMIT_WINDOW_SECONDS = 3;
  * DEFAULT_SESSION_LENTTH is the length of time in seconds a session is valid for unless configured in tenant.
  */
 export const DEFAULT_SESSION_LENGTH = 7776000;
+
+export const COMMENT_REPEAT_POST_TIMESPAN = 21600;
```
wyattjoh, line 71: Should include comment

wyattjoh (changes requested):

src/core/server/services/comments/pipeline/phases/repeatPost.ts
tessalt: the body should be processed by an earlier step in the moderation pipeline at this point so it should match with what was saved previously?

wyattjoh: That's right, mine was more of a feature suggestion/wish to support catching: "Yes" versus "yes" possibly in the future.

tessalt:
updated, the "re-request" button isn't showing up

wyattjoh (approved):

PR #2636, opened 2019-10-10
Add auto redirect when there is just one admin provider

allanassis:
Hi, guys.
One step less to login.


wyattjoh:
Hey @allanassis! Sorry for the delay on this one. We decided to keep the experience consistent, we wanted to prevent the auto redirect for the admin. Much appreciated though!

PR #2637, opened 2019-10-15
[CORL_653] throw error if inviting existing users

tessalt:
- throws an error if you try to invite an existing user, instead of silently not sending an email. 

copy needs to be confirmed

Screen Shot 2019-10-15 at 8 49 04 AM


src/core/server/services/users/auth/invite.ts
```diff
@@ -196,8 +197,7 @@ export async function invite(
     // Check to see if the user with the specified email already has an account.
     const userAlready = await retrieveUserWithEmail(mongo, tenant.id, email);
     if (userAlready) {
-      payloads.push({ email });
-      continue;
+      throw new InviteIncludesExistingUser(email);
```
wyattjoh, line 201: This could possibly create a few invites before failing. Perhaps doing a loop before hand to test each if a user already exists before actually creating the invites?

wyattjoh (changes requested):

wyattjoh (approved):

PR #2643, opened 2019-10-16
[CORL-680] Persist Fragment Access Tokens in sessionStorage

wyattjoh:
When access tokens are passed via a fragment (like in SSO), the tokens should be persisted to `sessionStorage` if they haven't expired.

### DIFF
--- a/src/core/client/admin/local/initLocalState.ts
+++ b/src/core/client/admin/local/initLocalState.ts
@@ -1,8 +1,9 @@
 import { commitLocalUpdate, Environment } from "relay-runtime";
 
-import { REDIRECT_PATH_KEY } from "coral-admin/constants";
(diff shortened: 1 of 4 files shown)


src/core/client/admin/local/initLocalState.ts
```diff
@@ -12,27 +13,60 @@ export default async function initLocalState(
...
+
+  // Get all the parameters from the hash.
+  const params = getParamsFromHash();
+  if (params) {
+    // If there were params in the hash, then clear them!
     clearHash();
 
-    // Keep redirect path as we are in the middle of an auth flow.
+    // If there was an error, add it.
+    if (params.errors) {
```
alinhle: Why is this `errors`, instead of `error` like we use in our OAUTH?

alinhle: Shouldn't we harmonize it and use the same?

nick-funk (approved):

PR #2644, opened 2019-10-17
[CORL-694] use new styles for decision history

tessalt:
two things i'm not sure about: 

- the `mini-unit`... what's the deal with those?
- I didn't move the `Popover` component because I couldn't find styles for it in airtable

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24266,7 +24266,7 @@
       "dependencies": {
         "async": {
           "version": "1.5.2",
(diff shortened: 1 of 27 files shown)


nick-funk:
No need to move the popover if we don’t have styles for it. We’ll get to that eventually when and if we have to.

The mini-units appear to be an old way of spacing things prior to the spacing tokens. You did the right thing by replacing them with appropriate spacing tokens.

nick-funk (approved):

PR #2651, opened 2019-10-18
admin branding: tables

tessalt:
- updating the stories and users tables 
- update underlying `Table` component


src/core/client/admin/components/UserRole/UserRoleText.css
```diff
@@ -1,9 +1,11 @@
...
-  font-weight: var(--font-weight-regular);
-  color: var(--palette-grey-darkest);
+  font-family: var(--v2-font-family-primary);
+  font-weight: var(--v2-font-weight-primary-semi-bold);
+  color: var(--v2-palette-table-body-text);
 }
 
 .commenter {
+  color: var(--v2-palette-table-body-text);
   font-weight: var(--font-weight-regular);
```
nick-funk: This looks to be a duplicate font weight.

src/core/client/admin/routes/Community/UserRow.css
```diff
@@ -16,18 +16,14 @@
...
 .statusColumn {
-  vertical-align: top;
 }
 
 .usernameButton {
   padding: 0px;
-  font-family: var(--font-family-sans-serif);
-  font-weight: var(--font-weight-regular);
-  font-size: calc(14rem / var(--rem-base));
-  line-height: calc(16em / 14);
```
nick-funk, line 30: Do we still want these defined from variables, or did Sam remove them in AirTable as well?

I believe we have a `font-size` 14rem/var(--rem-base) => 0.875rem variable in our `variables2` definitions.
Similarly for `line-height` being 1.14.

src/core/client/admin/routes/Community/UserRow.css
```diff
@@ -37,6 +33,5 @@
 .deleted {
   display: inline-block;
   padding-left: var(--spacing-2);
-
   color: var(--palette-warning-main);
```
nick-funk, line 38: We can discuss this colouring with Sam so we can get an appropriate colour variable generated for it.

src/core/client/admin/ui/components/Table/TableBody.css
```diff
@@ -0,0 +1,2 @@
+.root {
```
nick-funk, line 1: Are we keeping this so custom CSS can style against it? Or can we delete this file?

nick-funk (changes requested):

src/core/client/admin/ui/components/Table/TableBody.css
tessalt: yeah i left it for future use because that seems to be a pattern but i'm ok with deleting it

src/core/client/admin/routes/Community/UserRow.css
tessalt: not sure what this is referring to, are you commenting on the deleted lines?

nick-funk: I'm commenting on how there was once a `font-size` and a `line-height` and wanted to make sure that you wanted to delete (and not update) these values. Also, that Sam hadn't perhaps missed a couple items in AirTable.

If this works and is fine, please resolve this and ignore my comment. Was only being wary about deleting a line height and font size values and there not being any in the resultant change set.

tessalt: ahhh fair enough I'll check

src/core/client/admin/ui/components/Table/TableBody.css
nick-funk: No worries, if you're seeing it elsewhere, then I'm fine with this too. I usually delete empty classes, but this isn't going to break any camel's backs.

nick-funk (approved):

PR #2655, opened 2019-10-18
[CORL-726] Repeat Comment Fix

wyattjoh:
This strips the HTML when comparing comments for repeat posts to ensure that formatting does not cause a false negative.

### DIFF
--- a/src/core/server/services/comments/pipeline/pipeline.ts
+++ b/src/core/server/services/comments/pipeline/pipeline.ts
@@ -1,8 +1,9 @@
 import { Db } from "mongodb";
+import striptags from "striptags";
 
 import { Omit, Promiseable, RequireProperty } from "coral-common/types";
 import { Config } from "coral-server/config";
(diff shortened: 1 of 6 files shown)


src/core/server/services/comments/pipeline/pipeline.ts
```diff
@@ -44,8 +48,15 @@ export interface ModerationPhaseContext {
   req?: Request;
 }
 
+export interface ModerationPhaseContext extends ModerationPhaseContextInput {
+  /**
+   * stripped is the HTML stripped version of the comment body.
+   */
+  strippedTags: string;
```
tessalt: I find this naming a bit confusing, it suggests that it stores the tags that were stripped, not the remaining string. Maybe `strippedString` or `htmlStripped` or `tagStripped`?

tessalt (commented):
looks good, but I had a comment about variable naming!

tessalt (approved):

PR #2658, opened 2019-10-21
update design tokens for text inputs and selects

tessalt:
updates the textfield and select components, and implements them for the table filters for users and stories.

### DIFF
--- a/src/core/client/admin/routes/Community/UserTableFilter.css
+++ b/src/core/client/admin/routes/Community/UserTableFilter.css
@@ -4,7 +4,6 @@
 }
 
 .textField {
-  height: 31px;
   width: calc(45 * var(--mini-unit));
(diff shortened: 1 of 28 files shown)


src/core/client/admin/ui/components/Label/Label.css
```diff
@@ -0,0 +1,8 @@
+.root {
+  color: var(--v2-palette-input-label);
+  font-size: var(--v2-font-size-2);
+  line-height: var(--v2-lh-reset);
```
nick-funk: Did we add a variant to allow use of `lh` as well as `line-height`?
If you want to use this more shorthand term, go ahead, but make sure to add it to the `variables2`.

nick-funk (approved):

PR #2667, opened 2019-10-23
Bring new branding styles to button components

nick-funk:
Adds new v2 button control that resides under `coral-ui/components/v2`.

Uses the new post-css advanced variables to define it's design tokens, this will be pattern moving forward for UI components as we migrate to new branding.

### DIFF
--- a/src/core/client/ui/components/v2/Icon/Icon.css
+++ b/src/core/client/ui/components/v2/Icon/Icon.css
@@ -1,0 +1,59 @@
+.root {
+  composes: icon from "coral-ui/shared/icon.css";
+  font-family: "Material Icons";
+  speak: none;
+  font-style: normal;
+  font-weight: normal;
+  font-variant: normal;
+  text-transform: none;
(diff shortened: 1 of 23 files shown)


src/core/client/ui/components/v2/Icon/Icon.css
```diff
@@ -0,0 +1,59 @@
+.root {
+  composes: icon from "coral-ui/shared/icon.css";
```
tessalt: do we want this to be referencing styles outside of coral-admin?

nick-funk: I missed this, thank you for catching it, will get it updated. :)

nick-funk: It turned out I had already copied over the styles from shared into `root` I simply forgot to remove the `composes` declaration! Fixed up, thank you again!

tessalt (approved):

PR #2687, opened 2019-11-01
branding cleanup

tessalt:
okay! this is the big terrible PR that replaces the majority of the remaining ui components in the /admin routes. It's A LOT of files but the majority are just adding `/v2` and updating the props on buttons. Will add comments where it may be unclear. 

Remaining legacy components are all modals (suspend/ban, invite, hotkeys) which are awaiting design.

This does not include the auth, invite, and forgot password routes as those don't have designs.


src/core/client/admin/routes/Moderate/SingleModerate/SingleModerate.css
```diff
@@ -1,11 +1,15 @@
+$single-moderate-background: var(--v2-colors-grey-100);
```
tessalt, line 1: we don't have designs for the single comment moderate view but i just updated colours + fonts.

src/core/client/ui/components/v2/TextField/TextField.css
```diff
@@ -41,18 +52,23 @@
...
-    border-color: var(--v2-palette-input-disabled-foreground);
+  border: 1px solid $input-border;
+  &:focus {
+    border: 1px solid $input-regular-focused-border;
+    box-shadow: inset 0px 0px 0px 1px $input-regular-focused-border;
+  }
+}
+
+.colorDark {
+  border: 1px solid $input-border;
```
tessalt, line 63: the teal focus border + blue adornment button looked weird, so I added a `dark` variant for the text field where the focus border is blue.

src/core/client/admin/routes/Moderate/SingleModerate/SingleModerate.css
nick-funk: We can work with Sam to get these specc'ed and updated in another ticket. This is good for now. :)

nick-funk (approved):

PR #2693, opened 2019-11-05
GDPR and SSO docs updates

kgardnr:
## What does this PR do?
- Updated SSO docs
- Added GDPR docs


kgardnr:
@wyattjoh @immber will you add onto the GDPR docs for v5? I also updated some of the SSO pieces based on some changes we've made. Thank you!

PR #2706, opened 2019-11-13
[CORL-765] Fix delete account modal closing prematurely

nick-funk:
It turned out that modal click events were propagating out to the backdrop and triggering the close click event. This resulted in the modal closing whenever you clicked on anything.

This appears to also fix CORL-767.

## How to test?

- Go to the stream
- Login
- Head to your Account settings
- Try to Delete your account
- See that you can now click `Proceed` and other controls without the modal closing on you

### DIFF
--- a/src/core/client/ui/components/Modal/Modal.tsx
+++ b/src/core/client/ui/components/Modal/Modal.tsx
@@ -98,6 +98,9 @@ const Modal: FunctionComponent<Props> = ({
     },
     [onBackdropClick, onClose]
   );
+  const handleWrapperClick = useCallback((e: React.MouseEvent) => {
+    e.stopPropagation();
+  }, []);
 
   if (open && modalDOMNode) {
     return ReactDOM.createPortal(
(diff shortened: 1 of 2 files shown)


src/core/client/ui/components/Modal/Modal.tsx
```diff
@@ -115,7 +118,11 @@ const Modal: FunctionComponent<Props> = ({
           >
             <div className={styles.alignContainer1}>
               <div className={styles.alignContainer2}>
-                <div className={styles.wrapper}>
+                <div
+                  className={styles.wrapper}
+                  role="presentation"
```
nick-funk, line 123: Unsure if this is the correct role, but seems to be in-line with the role we used on the upper content wrapper that handles the escape key and scrolling.

wyattjoh: This seems correct. From my understanding, it just notes that the wrapper should not be mapped to a a11y interface (as it's just presentational).

kgardnr (approved):

PR #2709, opened 2019-11-13
[CORL-773] Rate Limiting

wyattjoh:
## What does this PR do?

Added new `@rate` directive to schema to support generalized rate limiting against mutations/queries in the schema.

## How do I test this PR?

Try to comment more than once within a three second window, notice the error!


nick-funk:
If I spam switching stories back and forth, I get the following error:

image

nick-funk (commented):
Appears to work except for the edge case I commented on.

nick-funk (approved):
The issue I found has turned out to be unrelated to these changes. Approving.

PR #2713, opened 2019-11-15
[CORL-645] Add Slack support for v5

nick-funk:
You can find the configuration for Slack channels at `Admin > Configure > Slack`.

You will want to follow the Tettra docs on creating a Slack App to hook this up and test it: 
https://app.tettra.co/teams/coralproject/pages/slack-apps


package.json
```diff
@@ -274,6 +274,7 @@
     "eventemitter2": "^5.0.1",
     "farce": "^0.2.6",
     "final-form": "4.11.0",
+    "final-form-arrays": "1.1.2",
```
nick-funk, line 280: I'm leaving this because we are still using arrays.

I didn't want to hack away all the array logic and make it hard to update to multi-channel support later. This way we can easily update when we get our final-form update in.

src/core/server/services/slack/publisher.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  const storyTitle = story.metadata ? story.metadata.title : "";
+  const commentBody =
+    comment.revisions.length > 0
+      ? comment.revisions[comment.revisions.length - 1].body
+      : "";
+  const body = commentBody.replace(new RegExp("<br>", "g"), "\n");
+  const moderateLink = createModerationLink(orgURL, comment.id);
+
+  const data = {
```
nick-funk, line 96: When building this, do we want to have some kind of templating?
Currently, this will only support English and I could easily see a way to template this with string interpolation and substitution.

src/locales/en-US/admin.ftl
```diff
@@ -380,6 +381,27 @@ decisionHistory-rejectedCommentBy = Rejected comment by <Username></Username>
...
+  Automatically send comments from Coral moderation queues to Slack
+  channels. You will need Slack admin access to set this up.
+
+configure-slack-channel-name = Channel
+configure-slack-channel-enabled = Enabled
+configure-slack-channel-hookURL-label = Webhook URL
+configure-slack-channel-hookURL-description =
+  Slack provides a channel-specific URL to activate webhook
+  connections. To find the URL for one of your Slack channels,
+  follow these instructions: url
```
nick-funk: These are in flux, as we need to discuss adding Slack details to our docs.

package.json
wyattjoh: I would suggest removing these from the `package.json` for now, and commenting out parts of the code that included this library until we actually use it.

src/core/client/admin/routes/Configure/sections/Slack/SlackChannel.tsx
```diff
@@ -0,0 +1,182 @@
...
+              Receive notifications in this Slack channel for
+            </InputLabel>
+          </Localized>
+          <Field
+            name={`${channel}.triggers.allComments`}
+            type="checkbox"
+            parse={parseBool}
+          >
+            {({ input }) => {
+              setAllComments(input.value);
```
wyattjoh: Instead of using this, you should use a wrapped field to propagate the field value down:

```
<Field name="allComments">
  {({ input: allCommentsInput }) => (
  ...
  <Field name="reportedComments">
    // use allCommentsInput.value here to disable the input
  </Field>
  ...
  )}
</Field>
```

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -931,6 +931,53 @@ type EmailConfiguration {
...
 
+################################################################################
+## Slack Configuration
+################################################################################
+
+type SlackConfiguration {
+  """
+  channels is the set of Slack channels configured to receive comments
+  """
+  channels: [SlackChannel]
```
wyattjoh: This should return instead:

```suggestion
  channels: [SlackChannel!]!
```

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -931,6 +931,53 @@ type EmailConfiguration {
...
+  """
+  reportedComments: Boolean
+  """
+  pendingComments is whether this channel will receive pending comments
+  """
+  pendingComments: Boolean
+  """
+  featuredComments is whether this channel will receive featured comments
+  """
+  featuredComments: Boolean
```
wyattjoh:
```suggestion
  featuredComments: Boolean!
```

Same for other edges in the `SlackChannelTriggers` type.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -1212,6 +1259,11 @@ type Settings {
   """
   email: EmailConfiguration! @auth(roles: [ADMIN, MODERATOR])
 
+  """
+  slack is the configuration for slack communication
+  """
+  slack: SlackConfiguration @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh:
```suggestion
  slack: SlackConfiguration! @auth(roles: [ADMIN, MODERATOR])
```

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -3392,6 +3444,49 @@ input CommenterAccountFeaturesInput {
...
+  the correlated channel
+  """
+  triggers: SlackTriggersConfigurationInput
+}
+
+input SlackConfigurationInput {
+  """
+  channels are the channels configured for Slack integration
+  """
+  channels: [SlackChannelConfigurationInput]
```
wyattjoh:
```suggestion
  channels: [SlackChannelConfigurationInput!]
```

src/core/server/services/slack/publisher.ts
```diff
@@ -0,0 +1,214 @@
...
+    p.queue === GQLMODERATION_QUEUE.PENDING
+  );
+}
+
+function createModerationLink(rootURL: string, commentID: string) {
+  return `${rootURL}/admin/moderate/comment/${commentID}`;
+}
+
+function createCommentLink(storyURL: string, commentID: string) {
+  return `${storyURL}?commentID=${commentID}`;
```
wyattjoh: As with `src/core/server/services/users/download/download.ts:107`, you should modify this to support story URL's that contain query params:

```suggestion
  const url = new URL(storyURL);
  url.searchParams.set('commentID', commentID);
  return url.href;
```

`URL` is from `import { URL } from 'url';`.

src/core/server/services/slack/publisher.ts
wyattjoh: I think in a future iteration we can tie this into i18n, but for now this should meet our requirements.

src/core/server/services/slack/publisher.ts
```diff
@@ -0,0 +1,214 @@
...
+  if (author === null) {
+    return;
+  }
+
+  const storyTitle = story.metadata ? story.metadata.title : "";
+  const commentBody =
+    comment.revisions.length > 0
+      ? comment.revisions[comment.revisions.length - 1].body
+      : "";
+  const body = commentBody.replace(new RegExp("<br>", "g"), "\n");
```
wyattjoh:
```suggestion
  const body = commentBody.replace(/<br\/?>/g, "\n");
```

src/core/server/services/slack/publisher.ts
```diff
@@ -0,0 +1,214 @@
...
+      method: "POST",
+      mode: "cors",
+      cache: "no-cache",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      redirect: "follow",
+      referrer: "no-referrer",
+      body: JSON.stringify(data),
+    });
```
wyattjoh, line 134: Why is `cors` specified? This shouldn't apply server side. Also is there a reason `redirect` and `referrer` are set?

src/core/server/services/slack/publisher.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    if (response.status !== 200) {
```
wyattjoh: Should check if `response.ok` instead (which is any non 200-299 status code).

wyattjoh (changes requested):

src/core/client/admin/routes/Configure/sections/Slack/SlackChannel.tsx
wyattjoh: See: https://github.com/coralproject/talk/blob/master/src/core/client/admin/routes/Configure/sections/Email/EmailConfigContainer.tsx

nick-funk: Makes sense, fixed!

wyattjoh (approved):

PR #2714, opened 2019-11-15
[CORL-645] Create Slack integration documentation

nick-funk:
Explains how to create a Slack App and hook it into
Coral as well as how to re-find your webhook URLs if you have lost them.


wyattjoh (commented):
Please replace images with link: https://slack.com/intl/en-ca/help/articles/115005265063-incoming-webhooks-for-slack :)

kgardnr (approved):

PR #2715, opened 2019-11-16
[CORL-750] SSO Migration Script Bug

wyattjoh:
## What does this PR do?

This corrects a bug in the migration script as well as updates the sso key schema to reflect the terminology better.

The problem will only affect instances that install new tenants after or on 5.3.0 which will wipe the generated SSO key during the migration. This will not impact installations prior to 5.3.0 that have upgraded to 5.3.0 or beyond.

If your installation is affected, you need to click "Regenerate" under the SSO authentication configuration to generate a new signing secret.

## How do I test this PR?

- Run the migrations on a new install
- Create a new tenant on a fresh install


tessalt (approved):
makes sense!

PR #2722, opened 2019-11-20
[chore] NodeJS 12

wyattjoh:
- Upgraded from 10 LTS to 12 LTS
- Added log line with userID to GraphQL log lines


src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
```diff
@@ -84,11 +84,6 @@ export async function findOrCreateSSOUser(
   token: SSOToken,
   now = new Date()
 ) {
-  if (!token.user) {
-    // TODO: (wyattjoh) replace with better error.
-    throw new Error("token is malformed, missing user claim");
```
nick-funk, line 89: Why was this removed?

wyattjoh: The `token.user` is actually always available in a `SSOToken`, and is checked in the `supports` method elsewhere in the class.

nick-funk (approved):

PR #2725, opened 2019-11-21
[CORL-719] A11y runtime checks using react-axe

alinhle:
## What does this PR do?
**Caution: Automatic accessibility checks can only discover a number of a11y issues. It does not replace a proper accessibility audit.**

Screenshot 2019-11-21 at 18 52 24

- Integrates https://github.com/dequelabs/react-axe into development that checks for a11y issues using https://github.com/dequelabs/axe-core at runtime. 

- Due to considerable performance hits the a11y checks are only turned on when adding`axe` into the url query like `http://localhost:8080?axe`.

- Only available during development.
- Not available when running on mobile.
- This PR also fixes a number of a11y issues that was discovered by `axe` in the `stream`, `auth`, `install` and `account` bundle.

## How do I test this PR?
- `npm run watch`
- Goto `http://localhost:8080?axe`


src/core/client/account/index.tsx
```diff
@@ -11,6 +12,8 @@ import localesData from "./locales";
 import "coral-ui/theme/variables.css";
 
 async function main() {
+  // Potentially inject react-axe for runtime a11y checks.
+  await potentiallyInjectAxe();
```
nick-funk, line 16: Does this mean that `axe` is in our bundle even if we build for production?

wyattjoh: Nope! It gets cleaned out via:

https://github.com/coralproject/talk/pull/2725/files#diff-93f19631eb4cc786929200be9c9bf1edR28-R31

wyattjoh (approved):

src/core/client/account/index.tsx
nick-funk: Makes sense!

nick-funk (approved):

PR #2726, opened 2019-11-21
[CORL-575] allow line breaks in stream tabs

tessalt:
Allows line breaks in the stream tabs so that longer text (for example, in spanish) does not get cut off 
Screen Shot 2019-11-21 at 10 29 29 AM
@IAmSamHankins

### DIFF
--- a/src/core/client/stream/tabs/Comments/Stream/StreamContainer.css
+++ b/src/core/client/stream/tabs/Comments/Stream/StreamContainer.css
@@ -23,6 +23,13 @@
 .featuredCommentsTabContainer {
   position: relative;
 }
+
+.streamTab {
+  max-width: 200px;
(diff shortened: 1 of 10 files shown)


nick-funk (approved):

wyattjoh (changes requested):
Looks a bit odd with the line breaking too early, and the filter button popping up earlier now?

Screen Shot 2019-11-22 at 11 36 17 AM
Screen Shot 2019-11-22 at 11 35 51 AM

tessalt:
@wyattjoh ok i've come up with a slightly different solution that keeps the sort options bottom-aligned and prevents the tab from breaking lines if it has space not to

wyattjoh (approved):

PR #2731, opened 2019-11-25
[CORL-353] show error if network request fails

tessalt:
I had to hook into the retryable request middleware and make it abort retries after x attempts in order to show an error message... I went with 2 but we could change it? Also not sure what copy we want here.

### DIFF
--- a/src/core/client/admin/routes/AuthCheck/NetworkError.tsx
+++ b/src/core/client/admin/routes/AuthCheck/NetworkError.tsx
@@ -1,0 +1,20 @@
+import { Localized } from "fluent-react/compat";
+import React, { FunctionComponent } from "react";
+
+import { Message } from "coral-ui/components";
+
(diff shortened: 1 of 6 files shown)


src/core/client/admin/routes/AuthCheck/NetworkError.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Message } from "coral-ui/components";
+
+import styles from "./NetworkError.css";
+
+const NetworkError: FunctionComponent<{}> = () => {
+  return (
+    <div className={styles.root}>
+      <Message color="error">
+        <Localized id="authcheck-network-error">
+          <span>A network error ocurred. Please refresh the page</span>
```
wyattjoh:
```suggestion
          <span>A network error occurred. Please refresh the page</span>
```

wyattjoh (approved):

PR #2736, opened 2019-11-29
Security Headers

wyattjoh:
## What does this PR do?

Adds security headers suggested by [`helmet`](https://www.npmjs.com/package/helmet).

This also introduces a `DISABLE_FORCE_SSL` that can be used to disable the SSL enforcement in testing environments. Setting this variable to `true` will print a warning to the console when starting if also in a production environment.


src/core/server/config.ts
```diff
@@ -255,6 +255,14 @@ const config = convict({
     env: "SCRAPE_TIMEOUT",
     arg: "scrapeTimeout",
   },
+  disable_force_ssl: {
+    doc:
+      "Disables forcing SSL in production environments. Should not be used except for testing.",
```
tessalt, line 260: is this being used?

tessalt (approved):

src/core/server/config.ts
wyattjoh: Yes, right here: https://github.com/coralproject/talk/pull/2736/files#diff-cf797d405858c20b0420663894bf0c4bR133

Mostly designed to be used that when we're testing production branches locally, we might need a way to force off SSL redirects.

PR #2738, opened 2019-11-29
[CORL-687] Webhooks

wyattjoh:
## What does this PR do?

Adds support for webhooks on the server for subscribing to events emitted by Coral.

## How do I test this PR?

TBD.


WEBHOOKS.md
```diff
@@ -0,0 +1,205 @@
...
+signatures prefixed by `sha256=`.
+
+If you receive a signature containing multiple signatures, it is typically when
+you have rolled the signing secret from the administrative panel, and chosen to
+keep the previous secret active for a duration of time.
+
+### How to verify the signature(s)
+
+```js
+// Set your secret secret here from the administratio n panel.
```
tessalt: typo

src/core/client/admin/routes/Configure/sections/WebhookEndpoints/ConfigureWebhookEndpoint/EndpointDangerZone.tsx
```diff
@@ -0,0 +1,181 @@
...
+  }, [webhookEndpoint, disableWebhookEndpoint, router]);
+
+  return (
+    <>
+      <Localized id="configure-webhooks-dangerZone">
+        <Subheader>Danger Zone</Subheader>
+      </Localized>
+      <FormField>
+        <Localized id="configure-webhooks-rollSigningSecret">
+          <Label>Roll signing secret</Label>
```
tessalt: isn't it usually "rotate"?

src/core/client/admin/routes/Configure/sections/WebhookEndpoints/ConfigureWebhookEndpoint/RollSigningSecretModal.tsx
```diff
@@ -0,0 +1,154 @@
...
+        <Card className={styles.root}>
+          <Flex justifyContent="flex-end">
+            <CardCloseButton onClick={onHide} ref={firstFocusableRef} />
+          </Flex>
+          <Form onSubmit={onRollSecret} initialValues={{ inactiveIn: 0 }}>
+            {({ handleSubmit, submitting, submitError }) => (
+              <form onSubmit={handleSubmit}>
+                <HorizontalGutter size="double">
+                  <Localized id="configure-webhooks-rollSigningSecret">
+                    <Typography variant="header2">
```
tessalt: I thought we weren't using `Typography` anymore post-redesign

src/core/client/framework/lib/validation.tsx
```diff
@@ -153,6 +154,15 @@ export const validateEqualPasswords = createValidator(
   PASSWORDS_DO_NOT_MATCH()
 );
 
+/**
+ * validateEventSelection is a Validator that checks for a valid combination of
+ * event selections for webhook endpoints.
+ */
+export const validateEventSelection = createValidator(
```
tessalt: Can we indicate that this is specific to webhooks in the name of this validator?

package.json
```diff
@@ -388,7 +388,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "FAST_LINT=true lint-staged"
+      "pre-commit": "lint-staged"
```
alinhle: The `FAST_LINT` is intentional because it's soooo much slower. It disables type information for linting. The idea was, that small changes, shouldn't make you wait so long for the linting when you commit. Obviously it makes sense when you change a lot of things, so it was a trade-off decision that favors convenience and speed over thorough linting, which still can be achieved by manually running `npm run lint`. Up to you :-)

src/core/client/admin/routes/Configure/sections/WebhookEndpoints/ConfigureWebhookEndpoint/DeleteWebhookEndpointMutation.ts
```diff
@@ -0,0 +1,38 @@
...
+          deleteWebhookEndpoint(input: $input) {
+            endpoint {
+              id
+            }
+          }
+        }
+      `,
+      variables: {
+        input: {
+          ...input,
```
alinhle: It's usually better to write out the individual props instead of using `...` because it has been a constant source of bugs, e.g. `{ id: input.id }`.

alinhle:
image
Can we have a similar solution to what @tessalt did with the multisite feature in order to avoid nested `<form>`?

alinhle:
One note from me: I feel like rotating the secret needs some kind of visually confirmation that it was successfully rotated.

src/core/server/events/events.ts
```diff
@@ -0,0 +1,87 @@
...
+  CoralEventType.STORY_CREATED,
+  {
+    storyID: string;
+    storyURL: string;
+  }
+>;
+
+export const StoryCreatedCoralEvent = createCoralEvent<
+  StoryCreatedCoralEventPayload
+>(CoralEventType.STORY_CREATED);
```
alinhle, line 87: I think it should be possible to instead of:
```ts
export type CommentReleasedCoralEventPayload = CoralEventPayload<
  CoralEventType.COMMENT_RELEASED,
  CommentReleasedInput
>;

export const CommentReleasedCoralEvent = createCoralEvent<
  CommentReleasedCoralEventPayload
>(CoralEventType.COMMENT_RELEASED);
```
To do:
```ts
export const CommentReleasedCoralEvent = createCoralEvent<
  CommentReleasedInput
>(CoralEventType.COMMENT_RELEASED);
```
and have a Type function that returns the Payload whenever you need:
```ts
type Payload = CoralEventPayload<typeof CommentReleasedCoralEvent>;
```
But I haven't actually validated it yet.

wyattjoh: I've looked into this too, but there's not really a clean way to extract the generic type there, and your proposal loses the binding of the data type to the type name, meaning we don't get the free type switching provided by the current system. It's not as pretty, but nothing wrong defining a type like that.

I was actually considering naming them the same:

```ts
export type CommentLeftModerationQueueCoralEventPayload = CoralEventPayload<
  CoralEventType.COMMENT_LEFT_MODERATION_QUEUE,
  CommentLeftModerationQueueInput
>;

export const CommentLeftModerationQueueCoralEvent = createCoralEvent<
  CommentLeftModerationQueueCoralEventPayload
>(CoralEventType.COMMENT_LEFT_MODERATION_QUEUE);
```

becomes:

```ts
export type CommentLeftModerationQueueCoralEvent = CoralEventPayload<
  CoralEventType.COMMENT_LEFT_MODERATION_QUEUE,
  CommentLeftModerationQueueInput
>;

export const CommentLeftModerationQueueCoralEvent = createCoralEvent<
  CommentLeftModerationQueueCoralEvent
>(CoralEventType.COMMENT_LEFT_MODERATION_QUEUE);
```

And Typescript would select the type/value depending on the usage automatically.

kgardnr (approved):

PR #2740, opened 2019-12-02
[CORL-758] Show new comments in stream when sorting by oldest

nick-funk:
## What does this PR do?

Push approved/new comments to the end of the stream when sorting by oldest.

## How do I test this PR?

1. Go to an non-pre-modded stream
2. Select Sort By: Oldest
3. Create a new comment that won't get flagged for toxicity
4. See the comment at bottom of stream above the "Load More" button


src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
```diff
@@ -59,14 +77,18 @@ function addCommentToStory(
...
   const connectionKey = "Stream_comments";
-  const filters = { orderBy: "CREATED_AT_DESC" };
 
-  if (streamProxy) {
-    const con = ConnectionHandler.getConnection(
-      streamProxy,
-      connectionKey,
-      filters
-    );
+  if (input.commentsOrderBy === "CREATED_AT_ASC") {
```
nick-funk, line 84: `commentsOrderBy` can be null, so we only do this if it's an exact match.

Hence the duplicate `getConnection(...)` rigamarole.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -2761,6 +2761,12 @@ input CreateCommentInput {
   """
   body: String!
 
+  """
+  commentOrderBy is how the comments are ordered when the
+  comment was created.
+  """
+  commentsOrderBy: String
+
```
wyattjoh: This should not be passed to the server, this is purely a client side parameter.

wyattjoh (changes requested):

wyattjoh (approved):

PR #2744, opened 2019-12-06
[CORL-688] Add user comment count tracking

nick-funk:


wyattjoh:
I've made some changes, will discuss after the holidays!

wyattjoh (approved):

PR #2752, opened 2019-12-13
[CORL-664] Threshold moderate new commenters

nick-funk:
## What does this PR do?

This is the same changes @tessalt previously wrote with two minor changes.

- Updates the old branch to latest master with all the fragment and style changes around the moderation/config area
- Replaces the prior counting calculation with a simple lookup from the user's cached `commentCounts` introduced in CORL-688

If there is a `don't merge` label on this PR, it's due to the fact that `CORL-688` likely hasn't been merged yet.

## How do I test this PR?

- Login as an admin/mod
- Head to the Moderation config and enable new commenter approval feature
- Default approves is likely set to 2
- Create a new commenter and sign in
- Make a couple comments on the stream
- See that they get sent for moderation
- Check back in the admin side with your mod account in the pending queue
- Approve 2 of their comments (or whatever your approval threshold setting was)
- Go back to the stream and post a comment as the commenter
- Their comments should be showing up in stream and not be held for moderation

NOTE: make sure you turn auto pre-mod off on the stream you choose to test with


src/core/server/models/user/user.ts
```diff
@@ -471,6 +471,17 @@ export interface User extends TenantResource {
    */
   deletedAt?: Date;
 
+  /**
+   * isNew is whether the user is a new user with less approved
+   * comments than the new commenter moderation config allows to
+   * be considered no longer a new user.
+   */
+  isNew: boolean;
```
wyattjoh: This cannot be stored on the user model document as it is dynamically determined by inspecting the counts on a user.

src/core/server/models/user/user.ts
```diff
@@ -2460,3 +2472,31 @@ export async function deleteModeratorNote(
...
+    const user = await retrieveUser(mongo, tenantID, id);
+    if (!user) {
+      throw new UserNotFoundError(id);
+    }
+
+    throw new Error("an unexpected error occurred");
+  }
+
+  return result.value;
+}
```
wyattjoh: This is not required when computing dynamically.

src/core/server/services/comments/pipeline/phases/premodNewCommenter.ts
```diff
@@ -0,0 +1,50 @@
...
+  "author" | "tenant" | "now" | "mongo"
+>): Promise<IntermediatePhaseResult | void> => {
+  // Ensure this mode is enabled.
+  if (!tenant.newCommenters.premodEnabled) {
+    return;
+  }
+
+  if (!author.isNew) {
+    return;
+  }
```
wyattjoh: Should be removed.

src/core/server/services/comments/pipeline/phases/premodNewCommenter.ts
```diff
@@ -0,0 +1,50 @@
...
+          metadata: {
+            count: author.commentCounts.status.APPROVED,
+          },
+        },
+      ],
+    };
+  } else {
+    await markUserNotNew(mongo, tenant.id, author.id);
+  }
+};
```
wyattjoh: Should be removed.

src/core/server/models/user/user.ts
```diff
@@ -525,6 +536,7 @@ async function findOrCreateUserInput(
     moderatorNotes: [],
     digests: [],
     createdAt: now,
+    isNew: true,
```
wyattjoh: Should be removed.

src/core/server/graph/tenant/schema/schema.graphql
```diff
@@ -1921,6 +1945,12 @@ type User {
   deletedAt is the time when the User was deleted.
   """
   deletedAt: Time
+
+  """
+  isNew is whether the commenter is a new user as defined
+  by the new commenter moderation rules (if enabled).
+  """
+  isNew: Boolean!
```
wyattjoh: Should be removed.

wyattjoh (changes requested):

tessalt (approved):

wyattjoh (approved):

PR #2753, opened 2019-12-13
[CORL-688] Consolidate tenant and common contexts

nick-funk:
Merges these two context classes together since we use TenantContext everywhere.

Don't merge on there cause I branched off CORL-688, will get merged first.


wyattjoh:
This should also merge anything inside `src/core/server/graph/tenant/*` and `src/core/server/graph/common/*` into `src/core/server/graph/*`:

```
src/core/server/graph
  directives
  resolvers
  ...etc
```

wyattjoh:
Closing to move work to another PR.

PR #2756, opened 2019-12-18
[CORL-824] Feature Flags

wyattjoh:
## What does this PR do?

Adds support for feature flags within Coral to support experiments with non-standard options. These features are designed to be removed in the future as they would be incorporated into core or removed.


nick-funk (approved):

nick-funk:
Updated the CLI tool to match these changes:
https://github.com/coralproject/coral-cli-plugin-cloud/pull/1

PR #2759, opened 2019-12-19
[CORL-825] Allow sending moderation feedback to perspective

nick-funk:
If there is a `don't merge` label on this, it's because we haven't merged in `CORL-688` yet and this branch is based off of it.

## What does this PR do?

Adds a configuration option to send approved and rejected comment moderation actions to perspective to help improve the tracking model.

## How do I test this PR?

- Head to Admin > Config > Moderation > Toxic Comment Filter
- Enable/allow the `Allow Coral to send moderation actions to Google` option
- Create a comment with a commenter account
- Approve/Reject it
- It'll get sent to Perspective, feel free to check logs and debug into `src/core/server/services/perspective/index.ts` functions


src/core/client/admin/routes/Configure/OnOffField.tsx
```diff
@@ -15,6 +15,8 @@ interface Props {
   offLabel?: React.ReactNode;
   format?: (value: any, name: string) => any;
   parse?: (value: any, name: string) => any;
+  onTestId?: string;
+  offTestId?: string;
```
wyattjoh: Perhaps something like this instead?

```ts
interface Props {
  // ...
  testIDs?: {
    on: string;
    off: string;
  }
}
```

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+) {
+  if (!config.enabled) {
+    return;
+  }
+  if (!config.key) {
+    return;
+  }
+  if (!config.sendFeedback) {
+    return;
+  }
```
wyattjoh: Consolidate to:

```ts
if (!config.enabled || !config.key || !config.sendFeedback) {
  return;
}
```

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const endpoint = config.endpoint
+    ? config.endpoint
+    : TOXICITY_ENDPOINT_DEFAULT;
+  const apiKey = config.key;
+
+  const communityId = `Coral:${domain}`;
```
wyattjoh: `domain != ROOT_URL`, this requires a fully resolved URL including scheme and port. See usage of `reconstructTenantURL` here:

https://github.com/coralproject/talk/blob/0f456330c5755c1faac3e7846204be3cad2bb73d/src/core/server/app/url.ts#L32-L48

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  const clientToken = `comment:${comment.id}`;
+  const latestRevision = comment.revisions[comment.revisions.length - 1];
```
wyattjoh: The latest revision in this case should not be used. Because approve and reject requires you to specify the revision ID being approved/rejected, that revision ID should be used to extract the correct revision to send to perspective instead.

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const story = await findStory(mongo, comment.tenantID, {
```
wyattjoh: Replace with `retrieveStory`, which already searches by ID.

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+    logger.error({ storyID: comment.storyID }, "could not find story");
+    return;
+  }
+
+  const url = getURLWithCommentID(story.url, comment.id);
+
+  const commentStatus = computeStatus(status);
+  if (!commentStatus) {
+    return;
+  }
```
wyattjoh: This can short circuit expensive database calls, move it right after the guard clauses referencing the `config`.

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const communityId = `Coral:${domain}`;
+  const clientToken = `comment:${comment.id}`;
+  const latestRevision = comment.revisions[comment.revisions.length - 1];
+
+  const story = await findStory(mongo, comment.tenantID, {
+    id: comment.storyID,
+  });
+  if (!story) {
+    logger.error({ storyID: comment.storyID }, "could not find story");
```
wyattjoh:
```suggestion
    logger.warn({ storyID: comment.storyID }, "could not find story");
```

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+            value: 1,
+          },
+        },
+      },
+      languages: ["EN"],
+      communityId,
+      clientToken,
+    };
+
+    const result = await send(endpoint, apiKey, "comments:suggestscore", body);
```
wyattjoh, line 141: Add a `logger.debug` about sending the feedback to perspective.

src/core/server/services/perspective/index.ts
```diff
@@ -0,0 +1,146 @@
...
+
+  const commentStatus = computeStatus(status);
+  if (!commentStatus) {
+    return;
+  }
+
+  try {
+    const body = {
+      comment: {
+        text: latestRevision.body,
```
wyattjoh: We only send the HTML stripped version of the body to perspective, ensure you strip the HTML prior to sending:

https://github.com/coralproject/talk/blob/0f456330c5755c1faac3e7846204be3cad2bb73d/src/core/server/services/comments/pipeline/pipeline.ts#L90

wyattjoh (changes requested):

PR #2767, opened 2019-12-20
[CORL-827/828] Scrub downloaded comments of html tags, rename archive

nick-funk:
## What does this PR do?

- Prefix the downloaded comments archive with coral instead of talk
- Scrub the downloaded comment bodies of their html tags

## How do I test this PR?

- Download the comments of a user
- Check that the archive is prefixed with `coral-`
- Check that the archive comment text does not include html tags (common ones are `<br/>` and `<href ...>`)


kgardnr (approved):
LGTM, I'll just let @wyattjoh have a quick peek before we merge 👍

wyattjoh (approved):

PR #2777, opened 2019-12-30
[CORL-839] Passport Upgrades

wyattjoh:
## What does this PR do?

Upgraded some packages to latest. This should resolve #2774. Fixes were
also applied after types upgrades that helped discover other errors.

PR #2778, opened 2019-12-31
Changing passport-google-oauth2 to 0.2.0 #2777

albertly:
## What does this PR do?
Google auth. gives   an error: "Cannot fetch user profile".
I've changed "passport-google-oauth2": "^0.1.6" to 0.2.0 (latest now).
See #2777

## How do I test this PR?
The project was deployed on Azure VM and checked. Seems work fine. No more error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -24716,9 +24716,9 @@
       }
     },
     "passport-google-oauth2": {
-      "version": "0.1.6",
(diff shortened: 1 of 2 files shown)


albertly:
Sorry, seems to me that somebody made the same PR #2777 92d81afea2c2640c54a973553101d2335db8c23a

PR #2779, opened 2020-01-06
[CORL-829] Bump line-height on select fields

nick-funk:
Bump line-height on select to account for hanging font characters

## What does this PR do?

Prevents characters like "g" or "j" being cut off at the bottom of the select text area.

## How do I test this PR?

Head to Admin > Configure > General.
See that the select drop down is no longer cutting off the text.

image


tessalt (commented):
I was thinking we might want to add another variable for this line height, since it's come up for me in a few more places

nick-funk:
> I was thinking we might want to add another variable for this line height, since it's come up for me in a few more places

Makes sense to me, I'll try and pick a reasonably generic name for the line height.

wyattjoh (approved):

PR #2787, opened 2020-01-10
[CORL-836] Create activeStories endpoint

nick-funk:
## What does this PR do?

Creates a new endpoint to query the most recently commented upon stories.

This adds a field that is set whenever a comment is created/edited on stories to store the date the last comment was created. We then use this to query active stories from our endpoint. The date field is partial indexed to improve performance.

## Behavior

`activeStories` endpoint returns (in one query):

- 10 most recently commented on stories, can request a max of 25 stories
- for each of those stories, you can pull default 20 comments, or request max 50

## How do I test this PR?

- Spin up coral
- Create some new comments on various stories
- Using a GraphQL request, hit this endpoint
```
query {
  activeStories {
    id
    lastCommentedAt
  }
}
```
- You should see a summary from newest to oldest, the stories you commented upon.

### DIFF
--- a/src/core/server/graph/resolvers/Query.ts
+++ b/src/core/server/graph/resolvers/Query.ts
@@ -21,4 +21,6 @@ export const Query: Required<GQLQueryTypeResolver<void>> = {
   debugScrapeStoryMetadata: (source, { url }, ctx) =>
     ctx.loaders.Stories.debugScrapeMetadata.load(url),
   moderationQueues: moderationQueuesResolver,
+  activeStories: (source, { limit = 10 }, ctx) =>
+    ctx.loaders.Stories.activeStories(limit),
 };
--- a/src/core/server/models/story/index.ts
+++ b/src/core/server/models/story/index.ts
@@ -7,7 +7,6 @@ import {
   DuplicateStoryIDError,
   DuplicateStoryURLError,
 } from "coral-server/errors";
-import { Logger } from "coral-server/logger";
 import {
   Connection,
   ConnectionInput,
@@ -471,6 +470,11 @@ export async function retrieveActiveStories(
   const stories = await collection(mongo)
     .find({
       tenantID,
+      // We limit this query to stories that have the following field. This
+      // allows us to use the index.
+      lastCommentedAt: {
+        $exists: true,
+      },
     })
     .sort({ lastCommentedAt: -1 })
     .limit(limit)
@@ -483,10 +487,9 @@ export async function updateStoryLastCommentedAt(
   mongo: Db,
   tenantID: string,
   storyID: string,
-  now: Date,
-  log: Logger
+  now: Date
 ) {
-  const result = await collection(mongo).findOneAndUpdate(
+  await collection(mongo).updateOne(
     {
       tenantID,
       id: storyID,
@@ -497,8 +500,4 @@ export async function updateStoryLastCommentedAt(
       },
     }
   );
-
-  if (!result.ok) {
-    log.error({ tenantID, storyID }, "unable to set lastCommentedAt on story");
-  }
 }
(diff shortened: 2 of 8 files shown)


src/core/server/services/stories/index.ts
```diff
@@ -368,3 +369,37 @@ export async function merge(
...
+export async function updateStoryLastCommentedAt(
+  mongo: Db,
+  tenant: Tenant,
+  story: Story,
+  now: Date
+) {
+  // Create a logger for this operation.
+  const log = logger.child({ tenantID: tenant.id, storyID: story.id }, true);
+
+  const result = await collections.stories(mongo).findOneAndUpdate(
```
wyattjoh: Database interactions should be contained to the `src/core/server/models` directory. These model functions should always return the result (the story) that was updated.

src/core/server/models/story/index.ts
```diff
@@ -455,3 +465,16 @@ async function retrieveConnection(
...
 }
+
+export async function compileActiveStories(mongo: Db, tenantID: string) {
+  const stories = await collections
+    .stories(mongo)
+    .find({
+      tenantID,
+    })
+    .sort({ lastCommentedAt: -1 })
+    .limit(ACTIVE_STORIES_LIMIT)
```
wyattjoh: The limit should be variable, not a constant.

src/core/server/models/story/index.ts
```diff
@@ -455,3 +465,16 @@ async function retrieveConnection(
   // Return a connection.
   return resolveConnection(query, input, story => story.createdAt);
 }
+
+export async function compileActiveStories(mongo: Db, tenantID: string) {
+  const stories = await collections
+    .stories(mongo)
```
wyattjoh: You should instead use the already available `collection` that is an alias of `collections.stories`.

src/core/server/models/story/index.ts
```diff
@@ -30,6 +33,8 @@ import {
 export * from "./counts";
 export * from "./helpers";
 
+const ACTIVE_STORIES_LIMIT = 25;
+
```
wyattjoh: This maximum is already reflected via the provided constraint on the GraphQL schema.

src/core/server/models/story/index.ts
```diff
@@ -455,3 +465,16 @@ async function retrieveConnection(
   // Return a connection.
   return resolveConnection(query, input, story => story.createdAt);
 }
+
+export async function compileActiveStories(mongo: Db, tenantID: string) {
```
wyattjoh: `compile*` is not used elsewhere, instead use `retrieve*`

src/core/server/graph/resolvers/Query.ts
```diff
@@ -21,4 +21,5 @@ export const Query: Required<GQLQueryTypeResolver<void>> = {
   debugScrapeStoryMetadata: (source, { url }, ctx) =>
     ctx.loaders.Stories.debugScrapeMetadata.load(url),
   moderationQueues: moderationQueuesResolver,
+  activeStories: (source, args, ctx) => ctx.loaders.Stories.activeStories(),
```
wyattjoh: Limit should be passed in from the provided arguments:

```suggestion
  activeStories: (source, { limit = 10 }, ctx) => ctx.loaders.Stories.activeStories(limit),
```

src/core/server/stacks/createComment.ts
```diff
@@ -175,6 +176,7 @@ export default async function create(
   );
 
   await updateUserLastCommentID(redis, tenant, author, comment.id);
+  await updateStoryLastCommentedAt(mongo, tenant, story, now);
```
wyattjoh: Because the ordering of these is not important, you can optimize this a bit with:

```ts
await Promise.all([
  updateUserLastCommentID(redis, tenant, author, comment.id),
  updateStoryLastCommentedAt(mongo, tenant, story, now),
]);
```

wyattjoh (changes requested):

wyattjoh (approved):

PR #2792, opened 2020-01-14
[CORL-787] Show no comments when story closed

nick-funk:
## What does this PR do?

Show "There are no comments on this story." instead of "There are no comments yet. Why don't you write one?" when the story is closed and there were never any comments.

## How do I test this PR?

- Set up a story with no comments visible
- Close the stream
- Check that the `All Comments` tab shows "There are no comments on this story."

### DIFF
--- a/src/core/client/stream/test/comments/stream/postReply.spec.tsx
+++ b/src/core/client/stream/test/comments/stream/postReply.spec.tsx
@@ -328,34 +328,32 @@ it("handle disabled commenting error", async () => {
 });
 
 it("handle story closed error", async () => {
-  let returnStory = stories[0];
-  const { rte, form } = await createTestRenderer(
-    {
-      Mutation: {
-        createCommentReply: sinon.stub().callsFake(() => {
-          throw new InvalidRequestError({
-            code: ERROR_CODES.STORY_CLOSED,
-          });
-        }),
-      },
-      Query: {
-        story: sinon.stub().callsFake(() => returnStory),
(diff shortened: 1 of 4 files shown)


src/core/client/stream/test/comments/stream/postReply.spec.tsx
```diff
@@ -326,36 +326,3 @@ it("handle disabled commenting error", async () => {
   expect(rte.props.disabled).toBe(true);
   expect(within(form).getByText("Submit").props.disabled).toBe(true);
 });
-
-it("handle story closed error", async () => {
```
nick-funk, line 330: I deleted this test because we can no longer see the stream when it is closed. This will never appear to the user and be accessible.

wyattjoh: If the story is closed while the user has the comment stream opened, it will occur.

wyattjoh (changes requested):

src/core/client/stream/test/comments/stream/postReply.spec.tsx
nick-funk: Looking into this, think you're right. Not sure why I was getting a node not mounted search error with this.

nick-funk: Is fixed.

wyattjoh (approved):

PR #2797, opened 2020-01-15
[CORL-867] Version SSL Bypass

wyattjoh:
## What does this PR do?

Much like #2796 bypassed the SSL wrap for `/api/health`, this moves the `/api/version` outside as well. This is a temporary fix to address some deployment issues around updating K8s healthchecks that prevent us from updating them post creation (See https://github.com/kubernetes/ingress-gce/issues/39).

## How do I test this PR?

```
curl -v http://localhost:3000/api/version
```

Notice that it returns 200, not a 301.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@coralproject/talk",
-  "version": "5.4.1",
+  "version": "5.4.2",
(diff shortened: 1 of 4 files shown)


immber (approved):
👍

PR #2799, opened 2020-01-16
Support multisite

tessalt:
## What does this PR do?
- **updates install process (CORL-816)** installation process will now create a single site and community
- **add a new site (CORL-812)** Add a new site form creates new site and sets multisite: true on tenant
- **view sites (CORL-813)** lists sites with links to edit
- **update sites (CORL-814)** update basic site settings
- **filter stories by site (CORL-841)** adds filter dropdown to stories table
- **filter moderation by site (CORL-817)** adds filter dropdown to moderation search bar
- migration to make existing tenants compatible
- stream-side updates to determine site of a story from domain
- adds new global notification component

## How do I test this PR?
- test migration by trying to run with existing tenant 
- try to add a new site, make sure moderation, stories table, and stream still work
- test installation process against empty db


src/core/client/admin/App/GlobalNotification/GlobalNotificationContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+  useContext,
+  useMemo,
+  useReducer,
+} from "react";
+
+interface State {
+  message: ReactNode | null;
+  visible: boolean;
+}
+
```
tessalt, line 7: had to create a global notification component using Context because the component has to appear at a higher level in the nesting order than routing would permit otherwise.

src/core/client/admin/App/GlobalNotification/GlobalNotificationContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const clearMessage = () => dispatch({ type: "CLEAR_MESSAGE" });
+  return {
+    state,
+    dispatch,
+    setMessage,
+    clearMessage,
+  };
+}
+
```
tessalt: this all is apparently standard boilerplate for using context. Allows components to use `setMessage` and `clearMessage` as hooks.

src/core/client/admin/routeConfig.tsx
```diff
@@ -45,17 +48,49 @@ export default makeRouteConfig(
         <Route path="moderate" {...ModerateRoute.routeConfig}>
           <Redirect from="/" to="/admin/moderate/reported" />
           <Route path="reported" {...ReportedQueueRoute.routeConfig} />
-          <Route path="reported/:storyID" {...ReportedQueueRoute.routeConfig} />
+          <Route
```
tessalt, line 51: note new URLs for moderation queues. Need to be able to distinguish between site ID and story ID.

src/core/client/admin/routes/Moderate/ModerateContainer.tsx
```diff
@@ -24,18 +25,43 @@ class ModerateContainer extends React.Component<Props> {
 
   public render() {
     const allStories = !this.props.match.params.storyID;
+    // TODO: (tessalt) get active route in a better way
+    const queueName = [
```
tessalt, line 29: this feels sloppy but I need to know the name of the active route so I can pass it down to the `SiteSelector` component which needs to be able to construct a URL

src/core/server/models/story/counts/shared.ts
```diff
@@ -53,21 +66,30 @@ export async function recalculateSharedModerationQueueQueueCounts(
   mongo: Db,
   redis: AugmentedRedis,
   tenantID: string,
+  siteID: string | null,
   now = new Date()
 ) {
-  const key = commentCountsModerationQueueQueuesKey(tenantID);
+  const key = commentCountsModerationQueueQueuesKey(tenantID, siteID);
```
tessalt: this is probably the wrong solution for getting counts by site.

src/core/client/admin/App/GlobalNotification/GlobalNotificationContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+    visible: false,
+  });
+  const value = useMemo(() => [state, dispatch], [state]);
+  return <NotificationContext.Provider value={value} {...props} />;
+}
+
+function useNotification() {
+  const context = useContext(NotificationContext);
+  if (!context) {
+    throw new Error(`useCount must be used within a CountProvider`);
```
wyattjoh: Mind the error text here. Also maybe extract the hook into another file?

src/core/client/admin/routes/Configure/ConfigureLinks.tsx
```diff
@@ -0,0 +1,38 @@
+import { Localized } from "@fluent/react/compat";
+import React, { FunctionComponent } from "react";
+
+import Link from "./Link";
+import Navigation from "./Navigation";
+
+const ConfigureLinks: FunctionComponent<{}> = () => {
```
wyattjoh, line 7:
```suggestion
const ConfigureLinks: FunctionComponent = () => {
```

src/core/client/admin/routes/Configure/sections/Organization/EmptySitesMessage.tsx
```diff
@@ -0,0 +1,12 @@
+import { Localized } from "@fluent/react/compat";
+import React, { FunctionComponent } from "react";
+
+const EmptySitesMessage: FunctionComponent = props => (
+  <Localized id="sites-emptyMessage">
+    <div>
+      We could not find anyone in your community matching your criteria.
```
wyattjoh: Is this copy correct given this component name referrers to a missing site?

src/core/client/admin/routes/Configure/sections/Organization/SitesConfigContainer.tsx
```diff
@@ -0,0 +1,94 @@
...
+const SitesConfigContainer: React.FunctionComponent<Props> = props => {
+  const sites = props.query
+    ? props.query.sites.edges.map(edge => edge.node)
+    : [];
+  const [loadMore, isLoadingMore] = useLoadMore(props.relay, 10);
+  const [, isRefetching] = useRefetch<
+    SitesConfigContainerPaginationQueryVariables
+  >(props.relay);
+  return (
+    <IntersectionProvider>
```
wyattjoh, line 29: Why is the intersection provider used here?

src/core/client/admin/routes/Configure/sections/Sites/AddSiteRoute.tsx
```diff
@@ -0,0 +1,77 @@
...
+  }
+  return (
+    <ConfigBox
+      title={
+        <Localized
+          id="configure-sites-add-new-site"
+          $site={props.data.settings.organization.name}
+        >
+          <Header>
+            {" "}
```
wyattjoh: Extra space.

src/core/client/admin/routes/Moderate/ModerateContainer.tsx
wyattjoh: We can probably look into moving the queue selector as a query param on the URL, making it easier to pull out.

src/core/client/admin/routes/Moderate/SiteSelector/SiteSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <>
+          {site && <SiteSelectorSelected site={site} />}
+
+          {!site && (
+            <Localized id="site-selector-all-sites">
+              <span className={styles.buttonText}>All sites</span>
+            </Localized>
+          )}
+        </>
+      }
```
wyattjoh, line 52: Why not use a `site ? <SiteSelectorSelec... /> : <Localized .... />` instead? Then you can remove the fragment.

src/core/client/admin/routes/Moderate/SiteSelector/SiteSelectorSite.tsx
```diff
@@ -0,0 +1,40 @@
...
+const SiteSelectorSite: FunctionComponent<Props> = ({ site, link, active }) => {
+  return (
+    <Link
+      className={cn(styles.root, {
+        [styles.active]: active,
+      })}
+      to={link || ""}
+    >
+      {site && site.name}
+      {!site && "All sites"}
```
wyattjoh: Missing translation. Should combine `{site ...` and `{!site ...` into a single ternary expression `{site ? site.name : <Localized... />}`

src/core/client/stream/local/initLocalState.ts
```diff
@@ -27,6 +27,9 @@ export default async function initLocalState(
 
     // Parse query params
     const query = parseQuery(location.search);
+    if (query.siteID) {
+      localRecord.setValue(query.siteID, "siteID");
+    }
```
wyattjoh: The siteID must always be inferred from the story connection, it shouldn't be directly referenced.

src/core/server/graph/loaders/Sites.ts
```diff
@@ -0,0 +1,33 @@
...
+    createManyBatchLoadFn((id: string) =>
+      retrieveSite(ctx.mongo, ctx.tenant.id, id)
+    )
+  ),
+  site: new DataLoader<string, Site | null>(
+    ids => retrieveManySites(ctx.mongo, ctx.tenant.id, ids),
+    {
+      cache: !ctx.disableCaching,
+    }
+  ),
```
wyattjoh: Both `find` and `site` seem to do the same thing, use `site` only.

src/core/server/graph/resolvers/ModerationQueues.ts
```diff
@@ -92,10 +116,18 @@ export const moderationQueuesResolver:
   | QueryToModerationQueuesResolver
   | ApproveCommentPayloadToModerationQueuesResolver
   | RejectCommentPayloadToModerationQueuesResolver = async (
-  source,
-  args,
-  ctx
+  source: any,
+  args: any,
+  ctx: GraphContext
```
wyattjoh: Types should be preserved here from the provided function types.

src/core/server/graph/schema/schema.graphql
```diff
@@ -3722,6 +3774,11 @@ input SettingsInput {
   newCommenters is the configuration for how new commenters comments are treated.
   """
   newCommenters: NewCommentersConfigurationInput
+
+  """
+  multisite is whether multiple sites are enabled.
+  """
+  multisite: Boolean
```
wyattjoh: This should remain as a computed property.

src/core/server/models/story/index.ts
```diff
@@ -432,13 +447,23 @@ export async function removeStories(
...
+  input: StoryConnectionInput,
+  siteID?: string
 ): Promise<Readonly<Connection<Readonly<Story>>>> {
   // Create the query.
-  const query = new Query(collection(mongo)).where({ tenantID });
+  const where: StoryConnectionWhere = { tenantID };
+  if (siteID) {
+    where.siteID = siteID;
+  }
+  const query = new Query(collection(mongo)).where(where);
```
wyattjoh: `siteID` should be included in the `input.filter` instead.

src/core/server/models/tenant/tenant.ts
```diff
@@ -202,6 +202,7 @@ export async function createTenant(
     slack: {
       channels: [],
     },
+    multisite: false,
```
wyattjoh: Should be a computed property.

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/__snapshots__/PostCommentFormContainer.spec.tsx.snap
```diff
@@ -23,6 +23,7 @@ exports[`renders correctly 1`] = `
           "enabled": false,
         },
       },
+      "siteID": "site-id",
```
wyattjoh: Check origin of this param.

src/core/server/errors/index.ts
```diff
@@ -259,11 +259,7 @@ export class URLInvalidError extends CoralError {
 }
 
 export class StoryURLInvalidError extends CoralError {
-  constructor(properties: {
-    storyURL: string;
-    allowedDomains: string[];
-    tenantDomain?: string;
-  }) {
+  constructor(properties: { storyURL: string; tenantDomain?: string }) {
```
wyattjoh, line 246: Revisit if we can include this

src/core/server/graph/mutators/Stories.ts
```diff
@@ -35,7 +35,9 @@ export const Stories = (ctx: GraphContext) => ({
         ctx.config,
         input.story.id,
         input.story.url,
-        omitBy(input.story, isNull),
+        {
+          ...omitBy(input.story, isNull),
+        },
```
wyattjoh: Revert

src/core/server/errors/index.ts
tessalt: so the way we find out if a URL is valid has changed a bit, previously we would check tenant.allowedDomains, and then create the story only if the ULR matched one of the domains, and throw this error otherwise. The new approach is to look for a site where site.allowedDomains matches the URL, if there's no match, it returns null. To populate `allowedDomains` on the error message we'd have to query all sites and then concat all the site.allowedDomains together, which we can definitely do, if it will be valuable for debugging in prod.

src/core/server/graph/resolvers/ModerationQueues.ts
tessalt: I get a `parameter has an implicit any type` error if I do this 🤔

src/core/client/admin/App/GlobalNotification/useNotification.ts
```diff
@@ -0,0 +1,31 @@
...
+  }
+  const [state, dispatch] = context;
+
+  const setMessage = (message: ReactNode, timeout?: number) => {
+    dispatch({ type: "SET_MESSAGE", message });
+    if (timeout) {
+      setTimeout(() => {
+        dispatch({ type: "CLEAR_MESSAGE" });
+      }, timeout);
+    }
```
alinhle, line 23: I think the timeout needs to be cleared when the component unmounts.

alinhle: This can bite you at somepoint with ugly React warnings especially in tests.

You can use the following to know it's being unmounted.
```ts
useEffect(() => {
  return () => {
    // unmounting.
  }
}, [])
```

src/core/client/admin/App/GlobalNotification/NotificationContainer.tsx
```diff
@@ -0,0 +1,12 @@
+import React, { FunctionComponent } from "react";
+import useNotification from "./useNotification";
+
+const NotificationContainer: FunctionComponent<{}> = () => {
+  const { state } = useNotification();
+  if (!state.visible) {
+    return null;
+  }
+  return <div>{state.message}</div>;
```
alinhle: You can read this for accessibility related stuff:
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_status_role

src/core/client/admin/App/GlobalNotification/useNotification.ts
tessalt: I thought that was just if you were using a timer inside a `useEffect` hook?

src/core/server/graph/resolvers/ModerationQueues.ts
wyattjoh: The following diff will fix the error :)

```diff
diff --git a/src/core/server/graph/resolvers/ModerationQueues.ts b/src/core/server/graph/resolvers/ModerationQueues.ts
index 3fb2d759d..a3724dedd 100644
--- a/src/core/server/graph/resolvers/ModerationQueues.ts
+++ b/src/core/server/graph/resolvers/ModerationQueues.ts
@@ -12,10 +12,8 @@ import {
 } from "coral-server/services/comments/moderation/counts";
 
 import {
-  ApproveCommentPayloadToModerationQueuesResolver,
   GQLModerationQueuesTypeResolver,
   QueryToModerationQueuesResolver,
-  RejectCommentPayloadToModerationQueuesResolver,
 } from "coral-server/graph/schema/__generated__/types";
 
 import GraphContext from "../context";
@@ -112,13 +110,10 @@ export const sharedModerationInputResolver = async (
  * @param args the args of the payload containing potentially a Story ID
  * @param ctx the TenantContext for which we can use to retrieve the shared data
  */
-export const moderationQueuesResolver:
-  | QueryToModerationQueuesResolver
-  | ApproveCommentPayloadToModerationQueuesResolver
-  | RejectCommentPayloadToModerationQueuesResolver = async (
-  source: any,
-  args: any,
-  ctx: GraphContext
+export const moderationQueuesResolver: QueryToModerationQueuesResolver = async (
+  source,
+  args,
+  ctx
 ): Promise<ModerationQueuesInput | null> => {
   if (args.siteID) {
     const site = await ctx.loaders.Sites.site.load(args.siteID);
```

nick-funk:
Found an issue with error reporting on `Site Permitted Domains` under Configure > Organization > Sites > Add Site.

Steps to reproduce:
- Create a site with the `http://localhost:3000` under "Site permitted domains"
- Attempt to create another site with `http://localhost:3000` under "Site permitted domains"
- The form will not proceed and will not add the site
- Check the error logs and you will see the server has correctly rejected the request due to a uniqueness requirement on site permitted domains
- However, the user now has no idea why the page is rejecting their add site request. Probably an issue with validation/error messages for the form?

image

image

nick-funk:
image

I'm seeing an error trying to load "My Profile" as a commenter using the comment stream. Looks like it's trying to find the Organization.contactEmail on a site as if it were loading the old tenant info.

Checking my site, there is a `contactEmail`, so I think this should be a quick query and update fix?

```
{
  "_id": {
    "$oid": "5e2f0451fbb0286ae4f10d50"
  },
  "id": "51cc83c1-816d-4b51-a0f0-e2a1c722f7a1",
  "createdAt": {
    "$date": "2020-01-27T15:40:01.361Z"
  },
  "name": "Polygon",
  "contactEmail": "polygon@voxmedia.com",
  "url": "http://localhost:8000",
  "allowedDomains": [
    "http://localhost:3000",
    "http://localhost:8000"
  ],
  "tenantID": "36c20ccb-ae77-400c-a3e8-918ad423299c"
}
```

nick-funk:
Documents under `Sites` in mongo appear to be using the site name as the document name instead of the id like other document records. Is this intended?

It might also be Azure: Cosmos DB plugin trying to be clever.

Either way, I was able to make a very confusing setup with two `The Verge`'s in my db.

image

nick-funk:
I've create two sites and successfully added comments to their streams. I headed over to moderation and the queues show a count of 2 for unmoderated, but the comments never seem to load.

I think something is erroring out, but I'm not seeing any feedback in the console or logs to indicate why it's not loading anything.

Mod queues stuck on loading spinner:

image

Comments existing in stream:

image

nick-funk (approved):

src/core/client/admin/App/GlobalNotification/useNotification.ts
alinhle: In this case it's unrelated. Whenever a component unmounts, state changes should no longer be triggered. In this case, the notification component is always there, so you don't see warnings. But you might run into issues in tests or when the assumption that it's always there changes.

alinhle: `useRef` hook is useful here to keep a reference of the timeout handle to clear it. https://reactjs.org/docs/hooks-reference.html#useref.

PR #2813, opened 2020-01-29
[CORL 547] org-wide announcements

tessalt:
## What does this PR do?
- adds config section to "General" pane allowing creation and deletion of announcements
- displays announcements on stream for a specified duration
- allows users to dismiss announcements

## How do I test this PR?
- navigate to configuration -> Community announcement
- create an announcement
- check that announcement is present above stream
- dismiss announcement from stream
- remove announcement from admin
- create a new announcement
- new announcement should display on stream


src/core/client/admin/routes/Configure/sections/General/Announcement.tsx
```diff
@@ -0,0 +1,66 @@
...
+    .toJSDate();
+  const formattedDate = useMemo(() => {
+    return new Intl.DateTimeFormat(locales, {
+      day: "2-digit",
+      month: "2-digit",
+      year: "numeric",
+      hour: "2-digit",
+      minute: "2-digit",
+    }).format(new Date(disableAt));
+  }, [disableAt]);
```
alinhle: This looks like `disableAt` changing everytime which defeats the purpose of `useMemo`. I think line `26-28` should be inside the callback as well.

src/core/client/admin/routes/Configure/sections/General/AnnouncementConfigContainer.tsx
```diff
@@ -0,0 +1,49 @@
...
+}) => {
+  const createAnnouncement = useMutation(CreateAnnouncementMutaiton);
+  const deleteAnnouncement = useMutation(DeleteAnnouncementMutaiton);
+  return (
+    <>
+      <AnnouncementConfig
+        settings={settings}
+        createAnnouncement={createAnnouncement}
+        deleteAnnouncement={deleteAnnouncement}
+      />
```
alinhle, line 85: A small suggestion would be to combine `AnnouncementConfigContainer.tsx` and `AnnouncementConfig.tsx` together into `AnnouncementConfigContainer.tsx`.

src/core/client/ui/hooks/useLocalStorage.ts
```diff
@@ -0,0 +1,32 @@
...
+      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+    } catch (error) {
+      // console.log(error);
+    }
+  };
+
+  return [storedValue, setValue];
+}
+
+export default useLocalStorage;
```
alinhle: We have abstracted localStorage into CoralContext: `src/core/client/framework/lib/bootstrap/CoralContext.tsx`, so we should probably reuse that.

src/core/client/admin/routes/Configure/ConfigureRoute.tsx
```diff
@@ -28,7 +29,8 @@ class ConfigureRoute extends React.Component<Props, State> {
     data: Parameters<Props["updateSettings"]>[0]["settings"],
     form: FormApi
   ) => {
-    await this.props.updateSettings({ settings: data });
+    const payload = omit(data, "announcement");
```
alinhle: I'm a bit worried about this, let's have a chat tomorrow.

alinhle (changes requested):
I have some concerns about the implementation of the announcement config as it conflicts with the top level form save button which makes it feel unintuitive and foreign. Let's discuss.

image

src/core/client/admin/routes/Configure/ConfigureRoute.tsx
tessalt: we had a chat about this in #coral-tech-talk a few days ago and couldn't come up with a good solution that didn't involve re-architecting the whole configuration section. Love to hear your suggestions on this!

src/core/client/stream/tabs/Comments/Stream/Announcement/AnnouncementContainer.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const AnnouncementContainer: FunctionComponent<Props> = ({
+  settings,
+}) => {
+  const { localStorage } = useCoralContext();
+  const [dismissed, setDismissed] = useState(false);
+  useEffect(() => {
+    async function getDismissedStatus() {
+      if (!settings.announcement) {
+        setDismissed(true);
```
alinhle: You can just ignore changing state when `settings.announcement` is not set right, because it's rendered as `null` anyway.

alinhle (approved):
One last comment: https://github.com/coralproject/talk/pull/2813#discussion_r373253263.
Otherwise looks good, great work!

alinhle:
Maybe add tests if you feel like.

PR #2817, opened 2020-02-03
[CORL-896] Perspective Timeout

wyattjoh:
## What does this PR do?

Allows reconfiguration of the perspective timeout for all tenants on an installation via the `PERSPECTIVE_TIMEOUT` environment variable.

Example:

```sh
# One second.
PERSPECTIVE_TIMEOUT=1 second

# Two thousand milliseconds.
PERSPECTIVE_TIMEOUT=2000
```

## How do I test this PR?

Try setting the configuration variable and inspect using a breakpoint to see the new value picked up.

### DIFF
--- a/src/core/server/config.ts
+++ b/src/core/server/config.ts
@@ -51,9 +51,15 @@ convict.addFormat({
 convict.addFormat({
   name: "ms",
   validate: (val: number) => {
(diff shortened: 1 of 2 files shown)


immber (approved):
👍

PR #2818, opened 2020-02-03
[CORL-845] Account Linking

wyattjoh:
## What does this PR do?

Adds account linking to support when there is more than one login method enabled. When local and another social login is enabled, it will prompt the user to provide a password when an email address is already in use.

## How do I test this PR?

- Create a user with a email that you use for an facebook account
- Login with facebook with an account that uses the same email
- Follow link account flow

## TODO

- [x] Add support for FE to catch duplicate email error on add email form to instead attempt linking


src/core/client/admin/routes/Login/AccountCompletionContainer.tsx
```diff
@@ -28,14 +30,22 @@ type Props = {
...
     viewer,
     auth,
     setAuthView,
+    setDuplicateEmail,
   } = props;
   if (viewer) {
     if (!viewer.email) {
-      if (authView !== "ADD_EMAIL_ADDRESS") {
+      if (authDuplicateEmail || viewer.duplicateEmail) {
+        if (authView !== "ADD_EMAIL_ADDRESS" && authView !== "LINK_ACCOUNT") {
```
tessalt, line 46: this logic is getting a bit intense, might be worth adding a comment

tessalt (approved):
very cool! just one comment about some of the client side logic getting a bit hard to follow. Should those components be refactored as functional components at some point?

src/core/client/admin/routes/Login/AccountCompletionContainer.tsx
cvle: True!

alinhle: True!

PR #2824, opened 2020-02-04
resovle import error by removing useContext from ui component

tessalt:
Adds `locales` to `UIContext` so that the `AbsoluteTime` component doesn't have to refer to components outside the folder boundaries.

### DIFF
--- a/src/core/client/ui/components/AbsoluteTime/AbsoluteTime.tsx
+++ b/src/core/client/ui/components/AbsoluteTime/AbsoluteTime.tsx
@@ -1,19 +1,21 @@
 import React, { FunctionComponent, useMemo } from "react";
 
-import { Typography } from "coral-ui/components";
+import { Typography, useUIContext } from "coral-ui/components";
 import { PropTypesOf } from "coral-ui/types";
 
 interface Props {
   date: string;
   className?: string;
-  locales: string[];
+  locales?: string[];
 }
 
 const AbsoluteTime: FunctionComponent<Props> = ({
   date,
   className,
-  locales,
+  locales: localesFromProps,
 }) => {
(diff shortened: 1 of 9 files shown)


src/core/client/ui/components/AbsoluteTime/AbsoluteTime.tsx
```diff
@@ -1,16 +1,19 @@
...
 }
 
-const AbsoluteTime: FunctionComponent<Props> = ({ date, className }) => {
-  const { locales } = useCoralContext();
+const AbsoluteTime: FunctionComponent<Props> = ({
+  date,
+  className,
+  locales,
+}) => {
   const formatted = useMemo(() => {
```
alinhle, line 19: What do you think of this way?
```ts
const { locales: localesFromContext } = useUIContext();
const locales = localesFromProps || localesFromContext || ["en-US"];
```
`useUIContext()` doesn't exist but you can easily construct it, just look at how `useCoralContext` is created.

src/core/client/ui/components/v2/Timestamp/Timestamp.tsx
```diff
@@ -19,19 +24,24 @@ const Timestamp: FunctionComponent<TimestampProps> = props => {
...
-        <AbsoluteTime
-          date={props.children}
-          className={cn(styles.text, props.className)}
-        />
-      ) : (
-        <RelativeTime
-          className={cn(styles.text, props.className)}
-          date={props.children}
-        />
+    <UIContext.Consumer>
```
alinhle, line 27: I saw you reverted this in the v1 `Timestamp`. Should this be reverted too?

PR #2825, opened 2020-02-05
[CORL-900] Added support for customUserAgent for scraping

wyattjoh:
## What does this PR do?

Some newsrooms require a custom user agent header to be sent to bypass the paywall for integrations, this adds a new config option for configuring the custom user agent header.

## How do I test this PR?

Set the custom user agent header, observe requests being made to the upstream host with the new user agent header.


immber (approved):
👍

nick-funk (approved):

Issue #2819, opened 2020-02-04
Comments sometimes get approved even if they have banned words, and sometimes they go to a mistaken moderation queue


HectorNM:
Hi everyone! Congratulations for this project!

I'm having some issues with banned words. We expected that, if a comment has a banned word, it goes automatically to "Rejected" queue on Admin. However, after running this system and receiving a lot of comments, we noticed a few behaviours regarding this:

- Comments that had banned words but were being approved;
- Comments that had banned words but went to "Pending" queue on Admin;
- Comments that had banned words but went to "Not moderaded" queue on Admin; 

To reproduce this problem, it's really "random". I have to enter a comment with a banned word a lot of times (but with small variations on text so the comment is not repeated), and then I could see one time or another it going to the wrong queue (or being approved, without going to any queue).

For me, while debugging and putting some breakpoints inside the application, it looked like a race condition anywhere inside the application, because sometimes the method `containsMatchingPhraseMemoized` on line `26` of file `src/core/server/services/comments/pipeline/phases/wordList.ts` would return `false` for a comment with a banned word, although the expected behaviour would be it returning `true`.
However, I still couldn't understand the underlying conditions which this race condition happened - if it was realy a race condition. 

I couldn't find a similar issue, so I'm opening this one.

Thanks in advance for any help!

**Expected behavior:** Comments with banned words being rejected automatically

**Actual behavior:** In a few cases, randomly, comments with banned words go approved or go the other moderation queues (non-moderated and pending)

**Related Issues:** 

**Versions:**

- NodeJS: 10.16.0
- NPM: 6.12.1
- MongoDB: 3.6
- Redis: 3.2
- Browser: Google Chrome 79.0.3945.130 64 bits
- OS: macOS Catalina 10.15.2

kgardnr:
Hey Hector, could you explain more what you're seeing step by step, so we can try to reproduce the issue?

HectorNM:
Hi @kgardnr , sorry, I accidentally pressed "Enter" while still typing details for the issue. I've just edited with more information. Thank you for the help!

kgardnr:
Update on this:

@HectorNM has tracked this down to be a bug in the regex for our wordlists, and I've confirmed the issue

HectorNM:
I found a better way to reproduce and understand the problem. 

[...]

PR #2828, opened 2020-02-07
Changed memoized RegExp to not be global

HectorNM:
## What does this PR do?
This PR exchanges .test() to test regular expression matching with .match(). It fixes the problem described and reproduced on [this issue](https://github.com/coralproject/talk/issues/2819).

## How do I test this PR?
You can create a blacklist with words "banned" and "banned2". Then, write the following comments: "banned 1", "banned 2", "banned 3", "banned 4". Without this fix, only "banned 1" and "banned 3" will be rejected. With this fix, all those comments will be rejected.

### DIFF
--- a/src/core/server/services/comments/pipeline/wordList.spec.ts
+++ b/src/core/server/services/comments/pipeline/wordList.spec.ts
@@ -1,4 +1,7 @@
-import { containsMatchingPhrase } from "coral-server/services/comments/pipeline/wordList";
+import {
+  containsMatchingPhrase,
+  containsMatchingPhraseMemoized,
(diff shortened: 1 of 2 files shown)


wyattjoh:
Thanks for the info @HectorNM! I seem to have found the root cause.

It turns out that by [using a global flag on a RegExp](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test#Using_test_on_a_regex_with_the_global_flag) we ran into an issue where subsequent calls to `.test()` were returning the first strings result, not the new string's result. Removing the global flag from this one resolved the issue.

wyattjoh (approved):

PR #2834, opened 2020-02-11
[CORL-761] stream-side account tab for sso

tessalt:
## What does this PR do?
- adds a new claim to the jwt `user` object called `url` with which a provider can specify a link to account management features
- adds `ssoURL` field to user model
- if an SSO user has an `ssoURL` field, the "account" tab in the stream profile tab will be an external link to the provided URL
- if an SSO user **does not** have `ssoURL`, the account tab does not appear
- non-SSO users will still see the account tab
- additional re-arranging and renaming of the tabs in stream profile:
  - "download comment history" has been moved to the "my comments" tab under comment history
  - the "notifications" tab has been re-named "preferences" and the "ignored users" management has been moved to this tab

## How do I test this PR?
- generate an SSO token with a `user.url` claim. From the stream, ensure account tab is present and redirects to this URL.


src/core/client/stream/tabs/Profile/ProfileContainer.tsx
```diff
@@ -17,37 +18,44 @@ interface ProfileContainerProps {
...
-      ...CommentHistoryContainer_viewer
       ...AccountSettingsContainer_viewer
-      ...ChangeUsernameContainer_viewer
-      ...ChangeEmailContainer_viewer
+      ...MyCommentsContainer_viewer
       ...DeletionRequestCalloutContainer_viewer
-      ...NotificationSettingsContainer_viewer
+      ...PreferencesContainer_viewer
+      profiles {
+        __typename
```
nick-funk, line 49: Was not aware we could query this, that's neat!

nick-funk (approved):

PR #2837, opened 2020-02-12
[CORL-917] Autofill in Chrome

wyattjoh:
## What does this PR do?

Chrome incorrectly assumes all password fields on a page are associated with a user login, this causes issues where chrome automatically auto-fills logins without interaction. This PR serves to mark fields that are not user password fields as `new-password` fields, which is documented [here](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion#Preventing_autofilling_with_autocompletenew-password) as the most reasonable solution to preventing this behaviour.

## How do I test this PR?

Try to visit the moderation page when on the `master` branch to see if chrome autofills your email/password into the perspective API settings. Try again on this branch and notice that it does not do that.


tessalt (approved):
TIL!

PR #2845, opened 2020-02-18
Q&A Beta

nick-funk:
## What does this do?

Adds full Q&A features to the stream.

Figma Spec: https://www.figma.com/file/IJu2JNZFNvYiolJWEdNeY7al/Talk-v5---Stream?node-id=4211%3A0

## How to test?

- Enable Q&A using the new configure tab
- Add an expert
- Go back to the stream and ask questions as a commenter
- Respond to them as the expert user
- See how the functionality works with the 3 new tabs


src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -210,6 +224,16 @@ export class CommentContainer extends Component<Props, State> {
             </Localized>
           </Tag>
         )}
+        {hasAnsweredTag && (
+          <Tag variant="regular" color="primary" className={styles.answeredTag}>
+            <Flex alignItems="center">
+              <Icon size="xs" className={styles.tagIcon}>
+                check
+              </Icon>
+              ANSWERED
```
tessalt: are we inside a localization tag here?

src/core/client/stream/tabs/Comments/Comment/ReactionButton/ReactionButtonContainer.tsx
```diff
@@ -80,8 +81,8 @@ class ReactionButtonContainer extends React.Component<Props> {
         onClick={this.handleClick}
         totalReactions={totalReactions}
         reacted={reacted}
-        label={label}
-        labelActive={labelActive}
+        label={this.props.isQA ? "Vote" : label}
```
tessalt: should "vote" and "voted" be localized?

src/core/client/stream/tabs/Comments/Comment/UserTagsContainer.tsx
```diff
@@ -1,26 +1,49 @@
...
   return (
-    <>{staffTag && <Tag className={className}>{settings.staff.label}</Tag>}</>
+    <Flex alignItems="center">
+      {expertTag && (
+        <Tag variant="regular" color="primary" className={styles.tag}>
+          <Flex alignItems="center">
+            <Icon size="xs" className={styles.icon}>
+              star
+            </Icon>
+            EXPERT
```
tessalt: missing Localization tag

src/core/client/stream/tabs/Comments/Stream/AnsweredCommentsTab/AnsweredCommentContainer.css
```diff
@@ -0,0 +1,29 @@
...
+.replies,
+.gotoConversation {
+  font-size: calc(14rem / var(--rem-base));
+  font-weight: var(--font-weight-medium);
+  font-family: var(--font-family-sans-serif);
+  line-height: 0;
+}
+
+.gotoArrow {
+  display: inline-block;
```
tessalt, line 22: `inline-block` often indicates we could be using flexbox.

src/core/client/stream/tabs/Comments/Stream/AnsweredCommentsTab/AnsweredCommentContainer.tsx
```diff
@@ -0,0 +1,213 @@
...
+          className={CLASSES.featuredComment.authorBar.$root}
+        >
+          {comment.author && (
+            <UsernameWithPopoverContainer
+              className={CLASSES.featuredComment.authorBar.username}
+              comment={comment}
+              viewer={viewer}
+            />
+          )}
+          <Box ml={1} container="span">
```
tessalt: lot of uses of `mt` props in this component that could maybe be simplified with `HorizontalGutter` but no promises.

src/core/client/stream/tabs/Comments/Stream/AnsweredCommentsTab/AnsweredCommentsContainer.tsx
```diff
@@ -0,0 +1,182 @@
...
+      loadMoreEvent.success();
+    } catch (error) {
+      loadMoreEvent.error({ message: error.message, code: error.code });
+      // eslint-disable-next-line no-console
+      console.error(error);
+    }
+  }, [loadMore, beginLoadMoreEvent, props.story.id]);
+  const comments = props.story.featuredComments.edges.map(edge => edge.node);
+  return (
+    <>
```
tessalt: i don't think this fragment wrapper is needed.

src/core/client/stream/tabs/Comments/Stream/UnansweredCommentsTab/UnansweredCommentsTabContainer.tsx
```diff
@@ -0,0 +1,329 @@
...
+  const onViewMore = useCallback(() => viewMore({ storyID: props.story.id }), [
+    props.story.id,
+    viewMore,
+  ]);
+  const comments = props.story.comments.edges.map(edge => edge.node);
+  const viewNewCount =
+    (props.story.comments.viewNewEdges &&
+      props.story.comments.viewNewEdges.length) ||
+    0;
+  return (
```
tessalt, line 138: i think this all makes sense but it is a lot

src/core/client/stream/tabs/Configure/ConfigureStream/ConfigureStreamContainer.tsx
```diff
@@ -23,6 +23,8 @@ class ConfigureStreamContainer extends React.Component<Props> {
     data: UpdateStorySettingsInput["settings"],
     form: FormApi
   ) => {
+    window.console.log(data);
```
tessalt: 🚒

src/core/client/stream/tabs/Configure/Q&A/ExpertListItem.css
```diff
@@ -0,0 +1,126 @@
...
+  flex-wrap: wrap;
+  justify-content: space-between;
+  align-items: center;
+
+  max-width: 80%;
+}
+
+@media (min-width: 550px) {
+  .usernameEmail {
+    display: inline-block;
```
tessalt: again this miiiight make more sense as a flexbox thing.

src/core/client/stream/tabs/Configure/Q&A/ExpertListItem.css
```diff
@@ -0,0 +1,126 @@
...
+
+  color: var(--palette-text-primary);
+  font-size: calc(16rem / var(--rem-base));
+  font-weight: var(--font-weight-regular);
+  font-family: var(--font-family-sans-serif);
+  line-height: 1.375em;
+  letter-spacing: 0.0125em;
+}
+
+.username {
```
tessalt: did these few rules get repeated a few times here?

src/core/client/stream/tabs/Configure/Q&A/ExpertListItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+    onClickRemove(id);
+  }, [id, onClickRemove]);
+
+  return (
+    <li key={id} className={styles.root}>
+      <div className={styles.usernameEmail}>
+        {username && <span className={styles.username}>{username}</span>}
+        {email && (
+          <span className={styles.email}>
+            {"("}
```
tessalt: can we put this all one one line somehow?

src/core/client/stream/tabs/Configure/Q&A/ExpertSearchItem.css
```diff
@@ -0,0 +1,28 @@
+.button {
+  padding-top: var(--v2-spacing-1);
+  padding-bottom: var(--v2-spacing-1);
+}
+
+.username {
+  display: inline-block;
```
tessalt: is this the same styling as was used in the List Item?

src/core/client/stream/tabs/Configure/Q&A/ExpertSelectionContainer.tsx
```diff
@@ -0,0 +1,353 @@
...
+  const [tempSearchFilter, setTempSearchFilter] = useState<string>("");
+  const loading = !query || isRefetching;
+  const hasMore = !isRefetching && relay.hasMore();
+
+  const addExpertMutation = useMutation(AddExpertMutation);
+  const removeExpertMutation = useMutation(RemoveExpertMutation);
+
+  const clearSearchFilter = useCallback(() => {
+    const searchInputNode: any = searchFieldRef.current;
+    searchInputNode.value = "";
```
tessalt: can we manage the field state in a more react-y way?

src/core/server/services/comments/pipeline/phases/expert.ts
```diff
@@ -0,0 +1,31 @@
...
+  IntermediateModerationPhase,
+  IntermediatePhaseResult,
+} from "coral-server/services/comments/pipeline";
+
+import {
+  GQLSTORY_MODE,
+  GQLTAG,
+} from "coral-server/graph/schema/__generated__/types";
+
+export const expert: IntermediateModerationPhase = ({
```
tessalt: can we add a comment here or change the function name? "phases/expert" sounds ambiguous/vague imo.

tessalt (dismissed):
missing some localization tags + a couple UI questions & nitpicks.

src/core/client/stream/tabs/Comments/Stream/UnansweredCommentsTab/UnansweredCommentsTabContainer.tsx
tessalt: this is a useless comment lol ignore.

src/core/server/graph/schema/schema.graphql
```diff
@@ -2590,6 +2619,20 @@ type StorySettings {
   messageBox stores settings related to the Story Message Box.
   """
   messageBox: StoryMessageBox!
+
+  """
+  mode is whether the story stream is in commenting or Q&A mode.
+  This will determine the appearance of the stream and how it functions.
+  This is an optional parameter and if unset, defaults to commenting.
+  """
+  mode: STORY_MODE
```
wyattjoh: Instead of the documentation describing the default, apply the default in the resolver.

```suggestion
  mode: STORY_MODE!
```

src/core/server/graph/schema/schema.graphql
```diff
@@ -5847,13 +6003,45 @@ type Mutation {
...
+  enableQAOnStory will enable Q&A mode on a story.
+  """
+  enableQAOnStory(input: EnableQAInput!): EnableQAPayload!
+    @auth(roles: [ADMIN, MODERATOR])
+
+  """
+  disableQAOnStory will disable Q&A mode on a story.
+  """
+  disableQAOnStory(input: DisableQAInput!): DisableQAPayload!
+    @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh: Instead of discrete `enable/disableQA` mutations, instead use `updateStoryMode`.

src/core/server/graph/schema/schema.graphql
```diff
@@ -5847,13 +6003,45 @@ type Mutation {
...
+  addExpertToStory(
+    input: AddExpertInput!
+  ): AddExpertPayload! @auth(roles: [ADMIN, MODERATOR])
+
+  """
+  removeExpertFromStory removes an expert from a story.
+  """
+  removeExpertFromStory(
+    input: RemoveExpertInput!
+  ): RemoveExpertPayload! @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh: Instead of using `from/to` in mutation, prefer `addStoryExpert` or `removeStoryExpert`.

src/core/server/models/comment/comment.ts
```diff
@@ -929,7 +929,7 @@ export async function retrieveStoryCommentTagCounts(
     // we should switch this out to something like
     // `"tags.type": { $exists: true }` to ensure that we are using the
     // specified index.
-    "tags.type": GQLTAG.FEATURED,
+    "tags.type": { $exists: true },
```
wyattjoh, line 931: Adjust the documentation above this line.

src/core/server/models/story/index.ts
```diff
@@ -501,3 +521,137 @@ export async function updateStoryLastCommentedAt(
...
+  const story = await collection(mongo).findOne({ tenantID, id: storyID });
+  if (!story) {
+    throw new StoryNotFoundError(storyID);
+  }
+  if (
+    story.settings.expertIDs &&
+    story.settings.expertIDs.some(e => e === userID)
+  ) {
+    return story;
+  }
```
wyattjoh: This check should be completed in the services layer, and not here.

src/core/server/models/story/index.ts
```diff
@@ -501,3 +521,137 @@ export async function updateStoryLastCommentedAt(
...
+  const story = await collection(mongo).findOne({ tenantID, id: storyID });
+  if (!story) {
+    throw new StoryNotFoundError(storyID);
+  }
+  if (
+    story.settings.expertIDs &&
+    !story.settings.expertIDs.some(e => e === userID)
+  ) {
+    return story;
+  }
```
wyattjoh: Same as before, should be moved to services layer.

src/core/server/services/comments/pipeline/phases/approve.ts
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  // If user is an expert and we are in Q&A mode
+  // approve and feature the comment
+  if (tags.some(tag => tag.type === GQLTAG.EXPERT)) {
+    result.status = GQLCOMMENT_STATUS.APPROVED;
+    result.tags.push({
+      type: GQLTAG.FEATURED,
+      createdAt: now,
+    });
```
wyattjoh: This tag assignment should be moved to the `expert.ts` phase, which would simplify this phase to instead return a status `APPROVED` instead of editing the shared `result` from above.

src/core/server/services/comments/pipeline/phases/unanswered.ts
```diff
@@ -0,0 +1,52 @@
...
+          type: GQLTAG.UNANSWERED,
+          createdAt: now,
+        },
+      ],
+    };
+  }
+
+  // If we have experts and the user is not one
+  // of them, this is an unanswered question.
+  if (!story.settings.expertIDs.some(id => id === comment.authorID)) {
```
wyattjoh:
```suggestion
  if (story.settings.expertIDs.every(id => id !== author.id)) {
```

wyattjoh: Also, because the above logical test and this one both are making a decision based on the same domain (both are checking to see if this user is not an expert), it makes sense to me to combine these into the same logic test. It would instead read:

> If we have no experts, or the current author is not an expert, than this is an `UNANSWERED` comment.

src/core/server/services/comments/pipeline/phases/expert.ts
```diff
@@ -0,0 +1,31 @@
...
+  author,
+  now,
+  story,
+}): IntermediatePhaseResult | void => {
+  if (
+    author &&
```
wyattjoh: This check isn't required.

src/core/server/graph/schema/schema.graphql
```diff
@@ -4242,6 +4286,8 @@ input UpdateStorySettings {
   messageBox stores settings related to the Story Message Box.
   """
   messageBox: StoryMessageBoxInput
+
+  mode: STORY_MODE
```
wyattjoh: Because the story mode changes are done in another mutation, this should be removed.

src/core/server/graph/schema/schema.graphql
```diff
@@ -2653,6 +2696,7 @@ type Story {
     first: Int = 20 @constraint(max: 50)
     orderBy: COMMENT_SORT = CREATED_AT_DESC
     after: Cursor
+    tag: TAG
```
wyattjoh, line 2835: I would add a line in the docs that we should deprecate this `tag` in the future if we go deeper down Q&A by way of adding a `unansweredComments` edge to the story as we did for `featuredComments`.

src/core/server/graph/loaders/Comments.ts
```diff
@@ -200,12 +200,20 @@ export default (ctx: Context) => ({
...
     retrieveCommentStoryConnection(ctx.mongo, ctx.tenant.id, storyID, {
       first: defaultTo(first, 10),
       orderBy: defaultTo(orderBy, GQLCOMMENT_SORT.CREATED_AT_DESC),
       after,
       filter: {
+        tags: tag
+          ? {
+              $elemMatch: { type: tag },
+            }
+          : undefined,
```
wyattjoh: Instead prefer:

```ts
...tagFilter(tag),
```

As seen: https://github.com/coralproject/talk/blob/29742efb4b516ca9c6d9a5ada6cc2279a282c4c8/src/core/server/graph/loaders/Comments.ts#L144

src/core/client/stream/tabs/Configure/Q&A/QAConfigContainer.tsx
```diff
@@ -0,0 +1,58 @@
...
+      if (story.settings.mode === "COMMENTS") {
+        enableQA({ storyID: story.id });
+      } else {
+        disableQA({ storyID: story.id });
+      }
+      setWaiting(false);
+    }
+  }, [waiting, setWaiting, story, disableQA, enableQA]);
+
+  return story.settings.mode === "QA" ? (
```
wyattjoh: Should use enum constants instead of inlined strings.

src/core/client/stream/tabs/Configure/Q&A/QAConfigContainer.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const QAConfigContainer: FunctionComponent<Props> = ({ story }) => {
+  const [waiting, setWaiting] = useState(false);
+  const enableQA = useMutation(EnableQAMutation);
+  const disableQA = useMutation(DisableQAMutation);
+
+  const handleOnClick = useCallback(async () => {
+    if (!waiting) {
+      setWaiting(true);
+      if (story.settings.mode === "COMMENTS") {
```
wyattjoh: Should use enum constants instead of inlined strings.

src/core/client/stream/tabs/Configure/Q&A/AddExpertMutation.ts
```diff
@@ -0,0 +1,51 @@
...
+
+let clientMutationId = 0;
+
+const AddExpertMutation = createMutation(
+  "addExpert",
+  (
+    environment: Environment,
+    input: MutationInput<AddExpertMutation> & {
+      username: string;
+      email: string;
```
wyattjoh: Why does this mutation accept a username/email address?

src/core/client/stream/tabs/Comments/Stream/StreamContainer.tsx
```diff
@@ -240,12 +291,29 @@ export const StreamContainer: FunctionComponent<Props> = props => {
...
+                <FeaturedComments />
+              </TabPane>
+            )}
+            {props.story.settings.mode === GQLSTORY_MODE.QA && (
+              <TabPane
+                className={CLASSES.allCommentsTabPane.$root}
+                tabID="UNANSWERED_COMMENTS"
+              >
+                <UnansweredCommentsTab />
+              </TabPane>
```
wyattjoh, line 333: The same logic test is performed above checking for Q&A mode, perhaps moving this into a fragment in the previous section?

src/core/client/stream/tabs/Comments/Stream/StreamContainer.tsx
```diff
@@ -139,6 +142,12 @@ export const StreamContainer: FunctionComponent<Props> = props => {
       } else {
         onChangeTab("FEATURED_COMMENTS", false);
       }
+
+      // If we are in Q&A mode, we default to most voted
+      // sorting by default
+      if (props.story.settings.mode === GQLSTORY_MODE.QA) {
+        setLocal({ commentsOrderBy: "REACTION_DESC" });
+      }
```
wyattjoh, line 158: These should be included in the `useEffect` dependencies, otherwise updates to the story will not cause this to be hit again.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabContainer.tsx
```diff
@@ -153,18 +159,38 @@ export const AllCommentsTabContainer: FunctionComponent<Props> = props => {
...
+          )}
+        {props.story.settings.mode === GQLSTORY_MODE.QA &&
+          comments.length <= 0 &&
+          !props.story.isClosed && (
+            <Localized id="qa-noQuestionsYet">
+              <CallOut fullWidth>
+                There are no questions yet. Why don't you ask one?
+              </CallOut>
+            </Localized>
+          )}
```
wyattjoh: Given the deep nesting of related checks, it makes more sense to me to nest these into an external component:

```tsx
const NoQuestions = ({ mode, isClosed }) => {
  if (mode === GQLSTORY_MODE.QA) {
    if (isClosed) {
      return <Localized /* ... */ />;
    } else {
      return <Localized /* ... */ />;
    }
  } else if (mode === GQLSTORY_MODE.COMMENTS) {
    if (isClosed) {
      return <Localized /* ... */ />;
    } else {
      return <Localized /* ... */ />;
    }
  }

  return null;
};

// ...

{comments.length <= 0 && <NoComments mode={props.story.settings.mode} isClosed={props.story.isClosed} />}
```

wyattjoh (dismissed):

src/core/client/stream/tabs/Comments/Stream/AnsweredCommentsTab/AnsweredCommentContainer.tsx
nick-funk: I'll be honest, this was copy pasted from the original Featured comment stream. Going to prioritize the other more serious comments on the PR, but I'll see if this can be fixed. 👍

src/core/client/stream/tabs/Comments/Stream/UnansweredCommentsTab/UnansweredCommentsTabContainer.tsx
nick-funk: Eh, I agree, will do what I can. Was original copy pasted from Featured stream, but that doesn't mean we can't clean it up! 😄

src/core/server/models/story/index.ts
nick-funk: I disagree, this is important to ensure that we don't add a duplicate expert ID onto the story.

I don't want safety checks spread across a service layer as well as the model (repository). This is a data specific problem, the service does not care about duplicates, it cares about the adding/removing. These checks I believe belong in the data layer.

src/core/server/models/story/index.ts
nick-funk: Disagree as per previous comment.

src/core/client/stream/tabs/Configure/Q&A/ExpertSelectionContainer.tsx
nick-funk: I thought ref's were part of React? 😆 
I get what you mean though, I'll try and find a cleaner way to clear this field.

src/core/server/models/story/index.ts
wyattjoh: So, given that the below database code is using a `findOneAndUpdate` method, it actually performs the find operation for you, for free without any additional database access being required. The use of the `$addToSet` operation also actually prevents duplicates from being added to the set, thereby ensuring that the userID is not duplicated.

I would check to see the remainder of the function additions in this file do not query the database unnecessarily for these operations. You can instead just do:

```ts
if (!result.value) {
  throw new StoryNotFoundError(storyID);
}
```

Which can accomplish the same check without any additional database access.

The idea/drive behind this suggestion is that we want to ensure that we use atomic update operations when making document updates because we can then remove the pre-checks before these updates are performed.

I would also ensure that new functions added to this file include the name `Story` in the function name to stay consistent with the existing function naming.

kgardnr (approved):

PR #2848, opened 2020-02-19
[CORL-919] approved queue

tessalt:
## What does this PR do?
adds "approved" tab to moderation queue navigation and route to /approved. Behaves in all ways as the inverse of "rejected" queue. Used the circle + checkmark icon for the tab as it wasn't specified. 

## How do I test this PR?
Approve one or more comments and navigate to the new "approved" queue. 

Screen Shot 2020-02-19 at 10 08 30 AM


alinhle (approved):
Tests would be awesome, other than that: Looks really good!

_Btw. if you look at the older `rejected` queue tests, don't use as many `toMatchSnapshot` calls as we did back then. Instead just look for some text that you expect to appear._

PR #2850, opened 2020-02-20
[CORL-916] conversation view

tessalt:
## What does this PR do?
Adds a "view conversation" button to moderate cards that shows the context surrounding a comment. Shows the direct parent, and shows an option to load replies if they exist

## How do I test this PR?
Click "view conversation" on moderate cards for comments that are part of reply threads.


src/core/client/ui/components/v2/Button/Button.tsx
```diff
@@ -55,6 +55,9 @@ interface Props extends Omit<BaseButtonProps, "ref"> {
 
   /** Internal: Forwarded Ref */
   forwardRef?: Ref<HTMLButtonElement>;
+
+  /* if true adds a link-style underline */
+  underline?: boolean;
```
nick-funk, line 60: Since underlining is specific to certain button variants (i.e. the `textUnderlined` button variant). I would suggest overriding one of those in CSS and passing it into the `className` on the button.

If you have trouble overriding the parent styles via `className`, you can force it via the `classes` prop on the button.

I know this is an annoying way to do it. However, it's a stop-gap as Sam, Kiwi, and I are trying to make button configuration easier and more adaptable to custom button styles like you're trying to do here.

nick-funk (approved):

PR #2866, opened 2020-02-26
Q&A Fixes

nick-funk:
## What does this PR do?

Addresses bugs found during Coral Party 6.0.0 around Q&A.

### Fix various Q&A moderation/tagging issues:

- Allow Staff members to have their questions
answered and appropriately tagged
- Properly filter answering for only top level
questions (comments)
- Add documentation around various moderation
phases and comment creation steps
- Remove unnecessary status filter when setting
the status for a comment

None

## How do I test this PR?

- Ask questions on Q&A stream as staff
- Answer with expert account
- See they are answered correctly

- Ask a question as a commenter
- Reply to the commenter with more commenter questions
- Reply to nested reply on root question with expert
- Is not listed as answered because expert didn't answer the top level question


src/core/server/models/comment/comment.ts
```diff
@@ -734,9 +734,6 @@ export async function updateCommentStatus(
       id,
       tenantID,
       "revisions.id": revisionID,
-      status: {
-        $ne: status,
-      },
```
nick-funk, line 739: Is this being done for index reasons @wyattjoh? It causes issues when we want to "approve" a comment that has already been approved.

i.e
- Staff asks a question
- Is auto-approved cause they're staff
- Expert answers their question
- System tries to approve parent question cause it's an answered question
- This filter causes the `$ne: APPROVED` on that pre-approved comment to return a not found for this comment update
- Error occurs preventing the comment from being replied to/answered

src/core/server/services/comments/pipeline/phases/staff.ts
```diff
@@ -12,7 +12,13 @@ export const staff: IntermediateModerationPhase = ({
...
   now,
 }): IntermediatePhaseResult | void => {
-  if (author.role !== GQLUSER_ROLE.COMMENTER) {
+  const staffRoles = [
+    GQLUSER_ROLE.ADMIN,
+    GQLUSER_ROLE.MODERATOR,
+    GQLUSER_ROLE.STAFF,
+  ];
+
+  if (staffRoles.some(r => r === author.role)) {
```
nick-funk: Feel like it's better if we clarify this since now an EXPERT is a thing... so we want to be more specific.

Let me know if there is a helper method/constant list we could use instead?

src/core/server/services/comments/pipeline/phases/tagExpertAnswers.ts
```diff
@@ -24,10 +27,6 @@ export const tagExpertAnswers: IntermediateModerationPhase = ({
           type: GQLTAG.EXPERT,
           createdAt: now,
         },
-        {
-          type: GQLTAG.FEATURED,
-          createdAt: now,
-        },
```
nick-funk, line 30: Doing this later as we want to only feature if it's a top level comment and we don't have enough comment data during phases to calculate that.

If we only check on parentID, nested comments will be marked as answered via featuring. 
If we try and use ancestorID lists, those aren't populated yet since this "comment" still doesn't exist yet during mod phases.

src/core/server/stacks/createComment.ts
```diff
@@ -93,11 +94,29 @@ const markCommentAsAnswered = async (
...
+    // Expert's answers are featured so they appear
+    // in the featured (answered) tab
+    await addTag(
+      mongo,
+      tenant,
+      comment.id,
+      revisionID,
+      author,
+      GQLTAG.FEATURED
+    );
```
nick-funk: I'm doing the addTag now since we've confirmed this is a top level question/answer pair via the previous if check on the ancestors.

This prevents nested answers being listed as q/a pairs.

wyattjoh (approved):

PR #2868, opened 2020-02-27
[CORL-933] Add additional error handling around logQuery

nick-funk:
We are seeing that sometimes the context has an undefined logger when we log queries from GraphQL.

We want to catch this case and ensure we don't fail. If it happens again, we will catch everything we need to know and that will help us resolve the issue in the future.

### DIFF
--- a/src/core/server/graph/extensions/LoggerExtension.ts
+++ b/src/core/server/graph/extensions/LoggerExtension.ts
@@ -48,7 +48,10 @@ export function logQuery(
     if (!ctx || !ctx.logger) {
       // Catch the context and log it out, we need to know what is
       // going on here
-      logger.warn({ ctx }, "unable to find logger on GraphContext");
+      logger.warn(
+        { ctx: Boolean(ctx) },
+        "unable to find logger on GraphContext"
+      );
 
       // Continue to log the query.
       // We don't want to lose the precious metrics.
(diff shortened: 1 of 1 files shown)


src/core/server/graph/extensions/LoggerExtension.ts
```diff
@@ -33,9 +39,36 @@ export function logQuery(
...
+    };
+
+    // CORL-933: We are checking we have a logger as sometimes
+    // we are seeing no logger on the graph context. We're not
+    // sure why this is the case, but we want to catch when this
+    // is happening and not throw an error.
+    if (!ctx || !ctx.logger) {
+      // Catch the context and log it out, we need to know what is
+      // going on here
+      logger.warn({ ctx }, "unable to find logger on GraphContext");
```
wyattjoh: Avoid logging the context directly (it _will_ try to serialize everything) and instead just do a lighter reference:

```suggestion
      logger.warn({ ctx: Boolean(ctx) }, "unable to find logger on GraphContext");
```

src/core/server/graph/extensions/LoggerExtension.ts
```diff
@@ -33,9 +39,36 @@ export function logQuery(
...
+  } catch (err) {
+    // Since we introduced all these checks, we don't know
+    // the state of the context or what other data may be
+    // undefined.
+    //
+    // If while compositing this log data we fail for other
+    // reasons, let's catch that to prevent other errors from
+    // occurring and log everything we can.
+    logger.error({ ctx, document }, "unable to log graphQL query");
+    logger.error(err);
```
wyattjoh: Same feedback from above, as well as noting that the document node is also pretty heavy..

Noting as well that calling `logger.error(err)` does not work, and instead you should use:

```ts
logger.error({ err }, "unable to log GraphQL query");
```

Note the `{ err }` as an object. This is required because bunyan (our logger) uses the key names to assist with serialization, and passing `logger.error(err)` provides no keys.

wyattjoh (changes requested):

wyattjoh (approved):

PR #2887, opened 2020-03-12
[CORL-963] Story sorting fix

wyattjoh:
## What does this PR do?

Fixes issues where when combined with auto story closing, the stories tab filters appeared to not work as intended.

None

## How do I test this PR?

Set some stories `createdAt` in the DB way into the past. Try to filter by closed/opened when the story auto close is on/off.

### DIFF
--- a/src/core/server/graph/loaders/Stories.ts
+++ b/src/core/server/graph/loaders/Stories.ts
@@ -1,8 +1,10 @@
 import DataLoader from "dataloader";
 import { defaultTo } from "lodash";
+import { DateTime } from "luxon";
 
 import GraphContext from "coral-server/graph/context";
 import { Connection } from "coral-server/models/helpers";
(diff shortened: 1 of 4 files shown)


nick-funk (approved):
Looks good, makes sense. Would `autoCloseComments` and `autoClosedAt` make sense for naming instead of `closeCommenting` and `considerClosedAt`?

Either way, this looks good to go. Comments clarify what's going on. Appreciate the `$or` array of conditions!

wyattjoh:
Good feedback!

Unfortunately, `autoCloseComments` can't be used as `closeCommenting` is provided by the tenant. I was able to rename the `close` in the status filter object to `closeCommenting` to make it more consistent.

The naming for `considerClosedAt` is used because it isn't actually associated with the `closedAt` field as `autoClosedAt` might imply, `considerClosedAt` to me at least reads as "consider the story closed if the created at date is outside this window".

PR #2888, opened 2020-03-13
[DOCS] 6.0.0 Release

wyattjoh:
## What does this PR do?

Adds documentation for upgrading to version 6.0.0.

N/A

## How do I test this PR?

Review documentation changes.


immber (approved):
There are a couple other places we'll want to change from "version 5" to "5+", but I think this is good for now, and we can address those later!

PR #2899, opened 2020-03-19
Add tenantDomain to StoryCreatedEventPayload

cristiandean:
## What does this PR do?
When you have multiple Talk instances running, it seems to make sense to specify from which tenant a given event was sent.

This PR add `tenantDomain`  to `StoryCreatedCoralEventPayload`.

## How do I test this PR?
Create a webhook, access some story and look at the request body.

### DIFF
--- a/src/core/server/services/stories/index.ts
+++ b/src/core/server/services/stories/index.ts
@@ -66,7 +66,7 @@ export async function findOrCreate(
   scraper: ScraperQueue,
   now = new Date()
 ) {
-  let siteID = "";
+  let siteID = null;
   if (input.url) {
     const site = await findSiteByURL(mongo, tenant.id, input.url);
     // If the URL is provided, and the url is not associated with a site, then refuse
@@ -94,8 +94,7 @@ export async function findOrCreate(
     StoryCreatedCoralEvent.publish(broker, {
       storyID: story.id,
       storyURL: story.url,
-      tenantID: tenant.id,
-      siteID,
+      siteID: story.siteID,
     });
   }
 
@@ -224,7 +223,6 @@ export async function create(
   StoryCreatedCoralEvent.publish(broker, {
     storyID: story.id,
     storyURL: story.url,
-    tenantID: tenant.id,
     siteID: site.id,
   });
 
--- a/src/core/server/services/stories/index.ts
+++ b/src/core/server/services/stories/index.ts
@@ -66,7 +66,7 @@ export async function findOrCreate(
   scraper: ScraperQueue,
   now = new Date()
 ) {
-  let siteID = null;
+  let siteID = "";
   if (input.url) {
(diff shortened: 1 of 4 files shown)


cristiandean:
I've been thinking that maybe the tenant domain could be placed in the header. If you agree I can change.

src/core/server/services/stories/index.ts
```diff
@@ -66,7 +66,7 @@ export async function findOrCreate(
   scraper: ScraperQueue,
   now = new Date()
 ) {
-  let siteID = null;
+  let siteID = "";
```
wyattjoh: I would revert this change here, and instead below in the call to `StoryCreatedCoralEvent.publish` you use `siteID: story.siteID` instead.

src/core/server/services/stories/index.ts
```diff
@@ -94,6 +94,8 @@ export async function findOrCreate(
     StoryCreatedCoralEvent.publish(broker, {
       storyID: story.id,
       storyURL: story.url,
+      tenantID: tenant.id,
```
wyattjoh: I want to move the `tenantID` out of the publish call, I'll be submitting a PR shortly that will add the tenantID to the payload instead at a higher level.

wyattjoh (changes requested):

wyattjoh:
I've gone ahead and added the `tenantID` to the payload for all webhook payloads now and added to the webhook documentation. Feel free to comment on those changes!

src/core/server/services/stories/index.ts
wyattjoh: Note that this is now available on the event object itself (not the `data` member).

cristiandean:
@wyattjoh I had made a few more changes but I hadn't submitted.

In addition to tenantID, I think it would be good if we also sent tenantDomain. For example, we have several instances of talk running often on different servers and databases. We need to know from which of these instances we received the event.

If we have the domain of the tenant within the webhook we can run queries on it on its _/api/graphql_ route.


What do you think?

wyattjoh:
I agree that adding the tenant domain could be helpful here as well. If you can you push/merge your additional changes, I can add the `tenantDomain` using a similar method to `tenantID`!

src/core/server/services/stories/index.ts
cristiandean: Great! I already removed it from there.

cristiandean:
I ended up resolving the siteID issue and removing the tenantID from StoryCreatedCoralEvent.

I had also put tenantDomain into the data, but what you said makes perfect sense, it should be one level up and available from all hooks.

cristiandean:
I also added tenantDomain to the webhook schema.

wyattjoh (approved):
Looks awesome!

Issue #2840, opened 2020-02-13
Regular expression for banned words is not matching properly with accented characters


HectorNM:
Hi everyone!

In Portuguese we have a lot of accented characters. We just found that that the regular expression for matching banned words thinks that accented characters are word separators (like blank spaces), and they end up marking the whole word as a banned word. 

Let me give you an example:

**Banned words**: ["ba", "bi"]
**Generated regular expression**: (^|[^\w])(ba|bi)(?=[^\w]|$)
**Comment text:** Os biólogos são importantes. 

Even though "**biólogo**" is a single word, **bi** is a match for this regular expression. So, even though the comment doesn't have a banned word, it thinks it has. The problem is because "ó" is an accented character, and so the regular expression thinks "bi" and "logo" are separated words.

![image](https://user-images.githubusercontent.com/5360796/74441391-7f659200-4e4e-11ea-9ea7-b5180253e3e3.png)

Please note that this happens when the accented character comes before the banned word as well:

![image](https://user-images.githubusercontent.com/5360796/74441556-cd7a9580-4e4e-11ea-9ed2-5985e42db5ab.png)

It seems that `[^\w]` includes spaces and accented characters. The fix for this problem would probably make this rule more strict, to accept only spaces as word separators.

**Expected behavior:** Banned words separated by accented characters shouldn't ban a comment.

**Actual behavior:** Banned words separated by accented characters are banning comments, as described above.

**Related Issues:** 

**Versions:**

- NodeJS: 12.16.0
- NPM: 6.13.4
- MongoDB: 3.6
- Redis: 3.2
- Browser: Chrome 79.0.3945.130 64 bits
- OS: Catalina 10.15.2

HectorNM:
Just a comment about this issue: I was thinking about a solution to this problem, and I thought that maybe it doesn't affect only Portuguese, but other languages as well. I think that other characters from other languages which are not in the A-Za-z range could be affected too and be considered word separators.

What do you think of defining specific regular expressions for each language? That way, each language could define in the regular expression their respective accented characters. Besides, since the regular expression won't consider all possible accented words from all different languages, each regular expression would be more simple and so be more performatic.

kgardnr:
Hey @HectorNM check out this PR when you get a moment; this will be going into our 6.0.0 release, which will be cut this week.

https://github.com/coralproject/talk/pull/2869

PR #2906, opened 2020-03-25
Add Brazilian Portuguese case to generate Regular Expression of banned/suspect word list

HectorNM:
## What does this PR do?
This PR adds the a custom rule to generate a regular expression of banned/suspect words for Brazilian Portuguese comments. It's needed to accented characters are not considered word separators, as commented in issue #2840 .

None.

## How do I test this PR?
You can set up Coral Talk to be in "pt-BR" language. Add some banned words and check comments with them being rejected. Also, try to reproduce the problem described in the issue #2840 , as it won't happen anymore.

### DIFF
--- a/src/core/client/admin/components/ModerateCard/CommentContent.spec.tsx
+++ b/src/core/client/admin/components/ModerateCard/CommentContent.spec.tsx
@@ -38,3 +38,37 @@ it("renders empty words correctly", () => {
   renderer.render(<CommentContent {...props} />);
   expect(renderer.getRenderOutput()).toMatchSnapshot();
 });
+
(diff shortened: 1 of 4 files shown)


wyattjoh:
Thank you for the work! I've copied some of the tests over that you added for specific cases used in `pt-BR` to #2920 as that PR adds much more extended unicode support which captures accented characters properly.

Let me know if that branch works correctly for you!

HectorNM:
Great work @wyattjoh ! Thank you for your help! I'll check that branch by today and I'll tell you! 
Since tests are passing there, I think everything should be working properly 😃 

EDIT: accented banned words seems to be working properly!! :)

Issue #2908, opened 2020-03-25
Consecutive banned words in a comment text will make this comment text display incorrectly in Moderation section


HectorNM:
While testing regular expressions, I tried to change a test existing on `CommentContext.spec.tsx` to include consecutive banned words, line breaks, word in the beginning and end of the line. Like in the image below:

![image](https://user-images.githubusercontent.com/5360796/77564316-484cbc80-6ea1-11ea-91ae-99b1a5a42ff3.png)

However, when running this test, I found that once I put consecutive banned words, the HTML marking would be placed mistakenly in non-banned words in that comment, like in blank spaces. Please note the `Received` in red in the image below. Besides, as you can see, banned words are not marked as banned:

![image](https://user-images.githubusercontent.com/5360796/77564393-67e3e500-6ea1-11ea-9dde-afb193972f82.png)

I was trying to understand the problem, and I ended up coding a solution. I'm opening a PR for this solution, please check if it makes sense for you.

**Expected behavior:**
When typing two consecutive words that are banned in a comment, it'll mark these banned words correctly in moderation section.

**Actual behavior:**
Consecutive banned words are not marked properly in moderation section.

**Related Issues:** 

**Versions:**

- NodeJS: 12.16.0
- NPM: 6.13.4
- MongoDB: 3.6
- Redis: 3.2
- Browser: Chrome 80.0.3987.149 64 bits
- OS: macOS Catalina 10.15.2

HectorNM:
I opened a PR for a possible solution for this problem, please see if it makes sense for you: https://github.com/coralproject/talk/pull/2909

PR #2909, opened 2020-03-25
Fix banned words RegExp so consecutive banned words display properly in moderation view

HectorNM:
## What does this PR do?
It changes regular expression for  matching banned/suspect words in comments, as detailed in #2908 . The current regular expression does match correctly banned/suspect words, however, it captures (returns as a match) separators around banned words. Take this regular expression for banned words `bad` and `worse`:

`/(^|[^\w])(bad|worse)($|[^\w])/iu`

It has two three groups that returns their matches: `(^|[^\w])`, `(bad|worse)` and `($|[^\w])`, being the first and last one for word separators (spaces, line breaks, etc).

A problem arises when we call `split` (as it happens in `CommentContent`, line 33) to separate comment text  in smaller parts according to this regular expression, because since these are _groups_, they'll return their matches for word separators as well. Let's suppose we have the following comment:

`Hey, I'm bad bad, very bad`

Now let's try to reproduce this `split` logic: 

image

The array returned isn't in the format expected in the expected format `[STRING_WITH_NO_MATCH, NEW_WORD_DELIMITER, MATCHED_WORD, ...]` in `CommentContent`. This is happening because spaces are being returned as well.

To overcome this limitation, I changed the logic a little bit, so instead of using groups, I'm using a positive lookahead (?= ...) and positive lookbehind (?<= ...) in regular expression. For this example, that would be:

`(?<=^|[^\w])(bad|worse)(?=$|[^\w])`

This way, I'm not returning word separators as a match, but just using them to check if banned words are surrounded by word separators. If I do this, `split` would work like this:

image

No matter if I have consecutive words or not, I'll always have this structure in array: `[STRING_WITH_NO_MATCH, MATCHED_WORD, ...]`. That's what this PR does, it marks banned words according to this new regular expression / split logic.

None.

## How do I test this PR?
Try typing consecutive banned words and see if they are marked correctly in Moderate section.

### DIFF
--- a/src/core/client/admin/components/ModerateCard/CommentContent.spec.tsx
+++ b/src/core/client/admin/components/ModerateCard/CommentContent.spec.tsx
@@ -22,6 +22,24 @@ it("renders correctly", () => {
   expect(renderer.getRenderOutput()).toMatchSnapshot();
 });
 
+it("renders correctly even if it has consecutive banned words on comments", () => {
+  const props: PropTypesOf<typeof CommentContent> = {
+    phrases: {
+      locale: "en-US",
+      wordList: {
+        suspect: ["worse"],
+        banned: ["bad"],
+      },
+    },
+    className: "custom",
+    children:
+      "This is a very long comment with bad words. Let's try bad and bad. Now bad bad.\nBad BAD bad.\n",
+  };
+  const renderer = createRenderer();
+  renderer.render(<CommentContent {...props} />);
+  expect(renderer.getRenderOutput()).toMatchSnapshot();
+});
+
 it("renders empty words correctly", () => {
   const props: PropTypesOf<typeof CommentContent> = {
     phrases: {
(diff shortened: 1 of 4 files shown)


wyattjoh:
Hey @HectorNM! After reviewing all the amazing work you submitted here, I was able to find a solution that resolved the issue without the addition of positive lookahead/lookbehind. Please check out #2920 when you got a chance!

wyattjoh:
Hi again @HectorNM! Apologies for the toss around.

Turns out the proposed solution I had in #2920 using the optional capture groups for white space did not actually resolve the issue, and instead caused problems with substring detection.

Unfortunately however, we can't accept the solution involving positive lookahead/lookbehind because Firefox [still hasn't implemented lookbehind](https://bugzilla.mozilla.org/show_bug.cgi?id=1225665) yet. This is a particular implementation detail that can't be filled by polyfills as far as I know.

At this time, I'm going to close this PR for now, but will make a note to come back here if that implementation is completed in other browsers.

HectorNM:
Hi @wyattjoh . Thank you for all the support on this issue! It seems this was a particularly challenging problem to solve!

Oh, so Firefox doesn't support it, I see :( When I have free time later, I'll try to find a solution that doesn't have a browser limitation. If I have any updates, I'll say here. Again, thank you so much!!

PR #2919, opened 2020-03-30
[CORL-997] Removed crossorigin attribute for custom CSS

wyattjoh:
## What does this PR do?

Removes the [`crossorigin`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) attribute for custom css stylesheets. It was incorrectly assumed that the `crossorigin` attribute would assist in the stylesheet requests, but it actually muddles the behaviour to cause CORS errors.

Note that this does not address cross-origin fonts, this is an ongoing issue.

None

## How do I test this PR?

Add a stylesheet from an external domain and notice no errors.

### DIFF
--- a/src/core/server/app/views/client.html
+++ b/src/core/server/app/views/client.html
@@ -11,7 +11,7 @@
     {% endfor %}
   {% endif %}
   {% if enableCustomCSS and tenant and tenant.customCSSURL %}
-    {{ macros.preload(tenant.customCSSURL, "style", crossorigin = true) }}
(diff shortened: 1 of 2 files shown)


nikrich:
Ahh great @wyattjoh ! Is this aimed to address the CORS issues on the CSS files?

kgardnr:
Yes @nikrich! We think this should resolve that. Let's merge this and we can see if it works for you.

kgardnr (approved):

PR #2920, opened 2020-03-31
[CORL-1001] Wordlist Fixes

wyattjoh:
## What does this PR do?

This PR was submitted in response to the **amazingly documented** PR from @HectorNM (#2909). Special call out to @michalnicp for the assistance in debugging the regular expressions used in this fix!

The cliff-notes from the PR submitted by @HectorNM (#2909) was that repeated banned or suspect words were not being highlighted in the moderation pane. The proposed solution used a feature in regular expressions called positive lookahead `(?= ...)` and positive lookbehind `(?<= ...)`. While these did resolve the issues with detection, it also introduced a performance hit with their use because those patterns involve backtracking on the parsed string during the iteration process. This is especially concerning because the same regular expression that is used by the frontend client is also used by the server to flag comments. Working with @michalnicp, we were able to resolve the issue and attain the desired behaviour by transitioning the previous pattern:

```js
/(^|[^\w])(bad|worse)($|[^\w])/iu
```

To the following:

```js
/([^\w]?)(bad|worse)([^\w]?)/iu
```

Note the replacement of the leading and trailing capture groups with the `?`. Swapping these out allowed us to handle the situations where we encounter a word at beginning of string, end of string, or when we encounter sequential matching phrases.

Further, digging into the (in my opinion) poorly documented [String.prototype.split()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split) method, we discovered that @HectorNM's discovery that our code related to handling the output from `split` was wrong. What we found however from our analysis was that the returned value from `split` based on the changes to the pattern was:

```
[STRING_WITH_NO_MATCH, NEW_WORD_DELIMITER, MATCHED_WORD, NEW_WORD_DELIMITER, ...]
```

With the pattern repeating for subsequent matches in the same source text. This allowed us to fix the mapping to rely on a `i % 4 === 2` rule to mark matching phrases.

This PR also addresses some performance related issues on the client by adjusting how the wordlists were pre-processed before being formed into a regular expression (by preventing overlaps). More streamlined highlighting logic was added to skip any comments that were not tagged as containing matching words as well, and stripping of tags was also performed to ensure highlighting was successful even when formatting directives were used inbetween.

**TLDR**

- Closes #2909 (huge shout out to @HectorNM for the work! 🎉 🎉 🎉 )

None.

## How do I test this PR?

Add the following to your suspect word list:

```
bad
```

Post a comment such as the one below that contains consecutive suspect/banned words:

```
Hey, I'm bad bad, very bad
```

And see that the moderation panel does indeed highlight the correct words without whitespace.


wyattjoh:
Found a bug actually with some substring detection going on... Working.

wyattjoh:
So, some research. Turns out that the optional whitespace capture groups are a bit too greedy when grabbing text around word matches, and also ends up grabbing strings that start with or end with these capture groups.

I'll be reverting that particular change, but will still be planning to merge this PR because it also includes some performance improvements to the comment marking.

wyattjoh:
Based on some more discussions with @michalnicp, I was able to improve the performance of the Regular Expressions quite a bit by replacing the usage of `[^\w]` with `[^\p{L}]` which has full Unicode support.

Many more automated tests were written to verify that the word list behaviour stays consistent as to ensure that the splitting beheviour does not break too much.

The issue still persists unfortunately as highlighted in #2909 and #2908, without adding the positive lookahead/lookbehinds, we can't support highlighting consecutive words. As noted in #2909, Firefox does not support these. Next steps to resolving this issue would likely involve moving the highlighting operation to the server (which uses NodeJS/v8 that _does_ support these expressions) and then the client can just overlay the `<mark>` elements at the marked sections. This is out of scope at the moment, as we don't have the capacity.

As this supports Unicode characters better now, and the tests from #2906 are currently passing, this should close #2906.

kgardnr:
Thank you for your help @michalnicp! 😻

nick-funk (approved):

PR #2931, opened 2020-04-10
[CORL-1002] Queue Improvements

wyattjoh:
## What does this PR do?

- Fixes issues with queue growth when failed jobs accumulate
- Replaced `performance-now` with `Date.now()`
- Increased log level for queuing operations
- Added `Query.queues` for looking up global queue state
- Fixed bug relating to story counts for stories not found

Added:

- `Query.queue`
- `type Queues`
- `type Queue`
- `type QueueCounts`

## How do I test this PR?

Start the server, notice no errors during startup.


src/core/server/graph/resolvers/Query.ts
```diff
@@ -28,4 +28,5 @@ export const Query: Required<GQLQueryTypeResolver<void>> = {
   sites: (source, args, ctx) => ctx.loaders.Sites.connection(args),
   site: (source, { id }, ctx) => (id ? ctx.loaders.Sites.site.load(id) : null),
   webhookEndpoint: (source, { id }, ctx) => getWebhookEndpoint(ctx.tenant, id),
+  queues: () => ({}),
```
nick-funk, line 31: Why are we exposing queues here if we're returning nothing?

Is it so it can be overridden by the QueuesResolver below?

src/core/server/helpers/createTimer.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * createTimer will create a new timer that can be called again to get the
+ * milliseconds since the timer was created.
+ *
+ */
+const createTimer = () => {
+  const start = Date.now();
+  return () => {
+    const finish = Date.now();
+    return finish - start;
```
nick-funk, line 10: nit: In the past, when following this scoped pattern, we returned:

```
{
   started: start,
   elapsed: finish - start,
   finished: finish
}
```

Sometimes you want all the details of the timespan that went by.

src/core/server/queue/tasks/scraper.ts
```diff
@@ -39,22 +39,13 @@ const createJobProcessor = ({
...
   log.debug("starting to scrape the story");
-
-  try {
-    await scrape(mongo, config, tenantID, storyID, storyURL);
-  } catch (err) {
-    log.error({ err }, "could not scrape the story");
-
-    throw err;
-  }
+  await scrape(mongo, config, tenantID, storyID, storyURL);
```
nick-funk, line 45: I've walked the ref's on this all the way up to server's `index.ts` and it will no longer be caught if this throws an error. That would take down the node instance. Is that what we want?

nick-funk (approved):

PR #2934, opened 2020-04-13
[CORL-1033] don't show view conversation link if callback is not defined

tessalt:
## What does this PR do?

avoid showing "view conversation" button if the callback is not defined.

none

## How do I test this PR?
View moderation card in user drawer

### DIFF
--- a/src/core/client/admin/components/ModerateCard/ModerateCard.tsx
+++ b/src/core/client/admin/components/ModerateCard/ModerateCard.tsx
@@ -59,7 +59,7 @@ interface Props {
   onReject: () => void;
   onFeature: () => void;
   onUsernameClick: (id?: string) => void;
-  onConversationClick: ((id?: string) => void) | null;
+  onConversationClick: (() => void) | null;
   onFocusOrClick: () => void;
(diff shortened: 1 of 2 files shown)


src/core/client/admin/components/ModerateCard/ModerateCard.tsx
```diff
@@ -231,14 +233,16 @@ const ModerateCard: FunctionComponent<Props> = ({
...
-              <Button iconLeft variant="text" onClick={viewConversationClick}>
-                <ButtonIcon>question_answer</ButtonIcon>
-                <Localized id="moderate-comment-viewConversation">
-                  <span>View conversation</span>
-                </Localized>
-              </Button>
-            </div>
+            {onConversationClick && (
+              <div className={styles.viewContext}>
+                <Button iconLeft variant="text" onClick={viewConversationClick}>
```
wyattjoh: No need to wrap this in a `useCallback` because it's already passed in the right form. Just use `onClick={onConversationClick}` instead!

wyattjoh (approved):

PR #2944, opened 2020-04-24
[CORL-1048] Cookie Deprecation

wyattjoh:
## What does this PR do?

New third party cookie policies created by Safari in their [recent 13.1 release](https://webkit.org/blog/10218/full-third-party-cookie-blocking-and-more/) have been causing issues for organizations using Coral with our non-SSO based (email/password) login methods.

This is closely related to the issue described https://github.com/privacycg/storage-access/issues/2#issuecomment-582214857. Our hopes is that the Storage Access API takes these issues into account during the development to provide guidelines for first-party SaaS social integrations while maintaining the security boundary of the iframe.

Other comment platforms render directly into the DOM of the top level frame which provide inferior security boundaries when compared to the isolation granted by the iframe.

This PR removes cookies entirely as they are transparent to the JS and therefore causes issues described in the below scenario. This PR instead tries to save access token details to `localStorage` whenever possible to ensure admin logins persist after login.

### Problem Scenario

1. The user first arrives at `organization.com`, where Coral is embedded from `organization.coral.coralproject.net`.
2. The user scrolls to the comment widget, clicks "Sign In" to open a popup, thereby granting temporary storage access.
3. The user signs in with valid credentials, and we set a cookie containing the access token.
4. We then send the access token back through `postMessage` to the iframe, and the user can proceed logged in.
5. The user refreshes or navigates on the top frame (on `organization.com`), and scrolls back to the comment section.

Because the user navigated, the storage access is removed, and the iframe no longer has storage access, and therefore does not show that the user is signed in. The issue we're experiencing now is when the user again clicks "Sign In":

1. A popup opens for login.
2. The popup (now granted storage access), sends a request to validate the current user session (where we now have the cookie attached).
3. The response says we're logged in via cookie, and the popup sends a message to the iframe regarding the valid credentials.

Because we don't have access to the access token after the refresh (because of the `HttpOnly` flag on the cookie), we can't send it back to the iframe after the second popup. This means that the user is stuck, the iframe does not have permission to access or send cookies along with the request, and we don't want to expose the cookie to the JS.

None.

## How do I test this PR?

Try to use in Safari where the Coral domain differs from the embedder domain.


tessalt (approved):
works!

src/core/client/framework/lib/bootstrap/createManaged.tsx
```diff
@@ -149,50 +140,48 @@ function createManagedCoralContextProvider(
 
     // This is called every time a user session starts or ends.
     private clearSession = async (nextAccessToken?: string | null) => {
-      // Clear session storage.
-      this.state.context.sessionStorage.clear();
```
alinhle, line 153: If we don't clear the session storage, then all the comment drafts will stay after logout and can be exposed to the next user who logs in.

alinhle (changes requested):
Putting a merge block for now, needs discussion

alinhle (changes requested):
Actionlist from discussion with @wyattjoh :
- Restore Relay Storage as source of truth
- Remove `Auth`, `JWTManager`, `cleanupCallbacks` and simplify into functions (or a class if it really needs to^^)
- clear session storage during `clearSession`
- No need to deal with about-to-expire-tokens for now 
- Remove automatic pause/resume in `createManagedSubscriptionClient`

src/core/client/framework/lib/network/createManagedSubscriptionClient.ts
```diff
@@ -47,7 +47,7 @@ export interface ManagedSubscriptionClient {
   /** Resume all subscriptions eventually causing websocket to start with new connection parameters */
   resume(): void;
   /** Sets access token and restarts the websocket connection */
```
alinhle, line 49: Can we add a comment like: `Sets access token. Note: You have to pause() and resume() the client manually for it to take effect.`

src/core/client/framework/lib/auth/auth.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (process.env.NODE_ENV !== "production") {
+      // eslint-disable-next-line no-console
+      console.error("could not get access token from storage", err);
+    }
+
+    return;
+  }
+}
+
+export function updateAccessToken(accessToken: string) {
```
alinhle: Can we rename this to `storeAccessToken`?

src/core/client/framework/lib/auth/auth.ts
```diff
@@ -0,0 +1,93 @@
...
+    const accessToken = storage.getItem(ACCESS_TOKEN_KEY);
+    if (!accessToken) {
+      // Looks like the access token wasn't in storage.
+      return;
+    }
+
+    // Return the parsed access token.
+    return parseAccessToken(accessToken);
+  } catch (err) {
+    if (process.env.NODE_ENV !== "production") {
```
alinhle: Just a question: What harm is there to output these errors in production?

wyattjoh: From a users perspective, it's not likely that this would be used to debug issues. If we're concerned about tracking these errors, we should consider these types of guards as indicators to include error reporters.

alinhle (approved):

Issue #2951, opened 2020-05-04
Deleted comments in moderation queues


allanassis:
Hi guys, i know it is an expected behavior, but why is it necessary to keep deleted comments in moderation queues?

I can not see any reasons for that, there are no information about these comments, it is just the relation about a blank box with the story title.

What do you guys think about removing these comments from these queues?

<img width="699" alt="Captura de Tela 2020-05-04 às 18 38 12" src="https://user-images.githubusercontent.com/23871115/81016584-1b43e000-8e37-11ea-9945-5391fa53e2af.png">

HectorNM:
Another idea instead of not showing at comments at all, maybe it could exist an extra queue called "Deleted" just to keep deleted comments. What do you think?

tessalt:
should be resolved by https://github.com/coralproject/talk/pull/3638

PR #2952, opened 2020-05-05
Add flag deleted with value "true" to comments of deleted users

HectorNM:
## What does this PR do?

We noticed that when user deleted their accounts, their deleted comments would appear like this on admin:

image

Moderator users were confused because these looked like empty comments. Besides, they couldn't approve or reject those comments to move them away.

After debugging, we noticed the problem was that the component `ModerateCard` expect a `deleted` flag to determine if the comment was deleted or not. However, deleted comments were not setting this flag. This PR sets this flag to `true` as the default behaviour. After doing this, we could see the message of deleted comment:

image

Just to be clear, this PR is not directly related to issue https://github.com/coralproject/talk/issues/2951 created recently. It would be nice if deleted comments had a different behaviour, actually. But assuming this is the expected behaviour, this PR solves this bug :) 

None. Actually, `deleted` field was already specified at GraphQL Schema. 

## How do I test this PR?
Delete a user account without this PR to see the bug happening. Then, apply this PR changes and try to delete a user account or a comment again, and you can see comments appearing with a default "comment deleted" message

### DIFF
--- a/src/core/server/graph/resolvers/Comment.ts
+++ b/src/core/server/graph/resolvers/Comment.ts
@@ -54,6 +54,7 @@ export const Comment: GQLCommentTypeResolver<comment.Comment> = {
       : null,
   revisionHistory: (c) =>
     c.revisions.map((revision) => ({ revision, comment: c })),
+  deleted: ({ deletedAt }) => !!deletedAt,
   editing: ({ revisions, createdAt }, input, ctx) => ({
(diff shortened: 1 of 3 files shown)


wyattjoh:
Great find @HectorNM! This is actually retroactively resolved by configuring the `deleted` resolver edge as I've done. Let me know if this works for your use case!

alinhle (approved):

PR #2955, opened 2020-05-06
[CORL-1075] Metrics Cleanup

wyattjoh:
## What does this PR do?

This moves prometheus metrics to another configurable port (default `9000`) to serve both metrics and cluster metrics from.

All breaking changes are:

- Remove `CLUSTER_METRICS_PORT` option
- Add `METRICS_PORT` option (default `9000`)

Metrics now cannot be served on the same port as standard traffic to secure.

None

## How do I test this PR?

Start the server, and curl http://localhost:9000/metrics to see the metrics print.

### DIFF
--- a/src/core/server/app/index.ts
+++ b/src/core/server/app/index.ts
@@ -1,6 +1,7 @@
+import cluster from "cluster";
 import cons from "consolidate";
 import cors from "cors";
-import { Express } from "express";
+import express, { Express } from "express";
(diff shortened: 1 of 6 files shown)


nick-funk:
image

I'm unable to `npm run start:development`, it says that 9000 is already in use. To be safe, I restarted my entire comp (killing any other node processes running) and tried again. Still no dice.

nick-funk (approved):

PR #2974, opened 2020-06-01
[SPIKE] [CORL-1012] Twitter and YouTube embeds

nick-funk:


src/core/server/graph/schema/schema.graphql
```diff
@@ -1240,6 +1240,22 @@ type StoryMessageBox {
...
+type EmbedLinksConfiguration {
+  """
+  twitterEnabled when true will enable twitter link embeds in the stream.
+  """
+  twitterEnabled: Boolean!
+
+  """
+  youtubeEnabled when true will enable youtube link embeds in the stream.
+  """
+  youtubeEnabled: Boolean!
```
wyattjoh: Suggestion: Rename fields to `twitter: Boolean!` and `youtube: Boolean!`. The "Enabled" part of the property can be inferred from the type.

src/core/server/app/handlers/api/oEmbed/oEmbed.ts
```diff
@@ -0,0 +1,64 @@
...
+    return `https://publish.twitter.com/oembed?url=${url}`;
+  }
+  if (type === "youtube") {
+    return `https://www.youtube.com/oembed?url=${url}`;
+  }
+
+  return "";
+};
+
+export const oEmbedHandler = (): RequestHandler => {
```
wyattjoh: Handlers should be reflective of their mounting points. Coral enforces a strict lowercase on URL's, this should be renamed to:

```suggestion
export const oembedHandler = (): RequestHandler => {
```

src/core/server/app/handlers/api/oEmbed/oEmbed.ts
```diff
@@ -0,0 +1,64 @@
...
+const fetchURL = (url: string, type: string) => {
+  if (type === "twitter") {
+    return `https://publish.twitter.com/oembed?url=${url}`;
+  }
+  if (type === "youtube") {
+    return `https://www.youtube.com/oembed?url=${url}`;
+  }
+
+  return "";
+};
```
wyattjoh: All URL parameters should be encoded with `encodeURIComponent`. While this is handled in the handler, it should instead be moved here to ensure that it's protected.

src/core/server/graph/schema/schema.graphql
```diff
@@ -1240,6 +1240,22 @@ type StoryMessageBox {
   content: String
 }
 
+################################################################################
+## Embed Links
+################################################################################
+
+type EmbedLinksConfiguration {
```
wyattjoh: As this can likely be the integration point for other embedded content, suggest renaming:

```suggestion
type EmbedConfiguration {
```

src/core/server/graph/schema/schema.graphql
```diff
@@ -2461,6 +2483,30 @@ type CommentRevisionMetadata {
...
+  """
+  YOUTUBE
+
+  """
+  TWITTER represents a twitter link source.
+  """
+  TWITTER
+}
+
+type CommentEmbedLink {
```
wyattjoh: Same suggestion as above:

```suggestion
type CommentEmbed {
```

src/core/server/graph/schema/schema.graphql
```diff
@@ -2485,6 +2531,12 @@ type CommentRevision {
   """
   body: String
 
+  """
+  embedLinks is the collection of links that were identified as supported
+  embedded content in the comment body
+  """
+  embedLinks: [CommentEmbedLink!]
```
wyattjoh:
```suggestion
  embed: [CommentEmbed!]
```

src/core/server/app/handlers/api/oEmbed/oEmbed.ts
nick-funk: I have done this, but had to remove the previous encoding on the url from the request. Reason being, you cannot call `encodeURIComponent` on a url twice... it will garble the url string.

nick-funk:
Closing as @tessalt has migrated this into another branch that combines GIF + YouTube + Twitter together into one consolidated project.

tldr; the Twitter/YouTube spike is over.

PR #2975, opened 2020-06-02
[CORL 1031] Story actions

tessalt:
## What does this PR do?

Adds "rescrape story" option to stories table, under an "actions" popover, which also moves the open/close story button to the action dropdown

None

## How do I test this PR?
Change a story title/other metadata then click "re-scrape story" to see metadata update


nick-funk (approved):

nick-funk:
Noticed this is merging into `stream-rebranding` but when I went to check the stream, it does not seem to be rebranded. Was this originally based off of master?

PR #2981, opened 2020-06-08
[CORL-1120] Fix SSO Sign Out

wyattjoh:
## What does this PR do?

When using SSO, the logout shouldn't appear unless the `jti` and `exp` claims are specified. This fixes these issues caused by setting them to undefined instead of null. We changed it to null here to match with the generated typescript type better.

None.

## How do I test this PR?

Try signing in with SSO, excluding the `exp` or `jti` claims. Notice the lack of sign out button.


immber:
Confirmed this resovled issue and sign out option was no longer displayed. 👍

immber (approved):

PR #2983, opened 2020-06-09
Replace nodemailer with emailjs

wyattjoh:
## What does this PR do?

This replaces nodemailer with emailjs for email sending.

None.

## How do I test this PR?

Perform an action in Coral that causes an email to be sent, see the email!


nick-funk (approved):

wyattjoh:
This is pending a fix: https://github.com/eleith/emailjs/pull/255

PR #2985, opened 2020-06-09
Added moderation queues detail to docs

losowsky:
Text additions. 

Question for @kgardnr:  Should these go into a separate 'Moderation' menu item below 'Admin Settings', as these aren't Admin only?


wyattjoh:
CI is being difficult @kgardnr, feel free to approve + merge, it should be handled automatically.

kgardnr (approved):

kgardnr:
@losowsky I'll merge this as-is and we can break it out separately - but good idea!

PR #2986, opened 2020-06-10
[CORL-1121] Remove username validation when updated by admin

nick-funk:
## What does this PR do?

Applies no validation when an admin updates a username.

None

## How do I test this PR?

Use the following mutation:

```
mutation ($id: ID!, $name: String!, $clientMutationId: String!) {
  updateUserUsername(
    input: {
      userID: $id,
      username: $name,
    })
      {
        user {
          id
          username
        }
      }
}
```

where:
- `$id` is some valid userID.
- `$name` is a username with spaces in it.
- `$clientMutationId` is `"1"`

### DIFF
--- a/src/core/server/services/users/users.ts
+++ b/src/core/server/services/users/users.ts
@@ -8,6 +8,7 @@ import {
   DOWNLOAD_LIMIT_TIMEFRAME_DURATION,
   SCHEDULED_DELETION_WINDOW_DURATION,
 } from "coral-common/constants";
+import { SSO_USERNAME_REGEX } from "coral-common/helpers";
 import { Config } from "coral-server/config";
 import {
   DuplicateEmailError,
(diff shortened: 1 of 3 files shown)


src/core/server/services/users/users.ts
```diff
@@ -631,7 +632,7 @@ export async function updateUsernameByID(
   createdBy: User
 ) {
   // Validate the username.
-  validateUsername(username);
+  validateUsername(username, SSO_USERNAME_REGEX);
```
nick-funk: This is the update call used by the `updateUserUsername` mutator.

@wyattjoh is this sufficient? Or should we load up the user and confirm they are an SSO account before passing this different RegEx?

If we do want to load the user and confirm they are SSO, what fields should we be using to determine the user is an SSO user?

wyattjoh: For this mutation, I'd just remove the username validation here all together. It can only be used by the admin's, and if they want to change it or allow it through their interfaces, we can't stop it.

```suggestion

```

wyattjoh (approved):

PR #2988, opened 2020-06-11
[CORL-1126] Run regex in a sandbox so it can time out

nick-funk:
## What does this PR do?

Isolates the wordList RegEx inside of a sandbox container so that it can be timed out when it runs for too long.

When it runs for too long, we kill it and flag it with `SYSTEM_WITHHELD`.

None

## How do I test this PR?

- Set some banned/suspect words in admin config
- Create some comments with those words
- Create comments without those words
- See the right comments are flagged for review in the moderation queues


nick-funk:
@wyattjoh did you just rewrite every line of code I wrote? Please feel free to merge and approve it at this point, cause it's not really my PR anymore.

wyattjoh (approved):
Amazing find for the vm work @nick-funk 👯‍♂️

kgardnr:
🚀

PR #3019, opened 2020-07-15
[CORL-1179] Enforce max width/auto scroll on mod card embeds

nick-funk:
Some gifs are too wide for the mod cards. Set
a max-width and allow it to scroll if it exceeds
the set width.

image


wyattjoh (changes requested):
Prefer adjusting the maximum width to allow resizing to occur rather than constraining with an overflow.

wyattjoh (approved):

PR #3020, opened 2020-07-15
[CORL-1180] Moderation + Editing Improvements

wyattjoh:
## What does this PR do?

When a comment is edited after already appearing in the moderation queue, this will ensure that the comment updates to the edited version after a moderation action is taken. This is a stop-gap solution that should be addressed when we add support for live comment edit notifications in the moderation stream. This prevents the scenario where a comment is approved in the moderation queue while the moderator is viewing an older revision of the comment.

None.

## How do I test this PR?

1. In one context, as a commenter, write a comment
2. In another context, as a administrator, view the comment in the moderation panel
3. Back in your first context, edit the comment
4. Back in your second context, try to approve that comment, note how it does not actually get approved, and instead updates to the edited comment that we made in step 3
5. Finally approve the comment after viewing the edited comment

### DIFF
--- a/src/core/client/admin/mutations/ApproveCommentMutation.ts
+++ b/src/core/client/admin/mutations/ApproveCommentMutation.ts
@@ -47,7 +47,7 @@ const ApproveCommentMutation = createMutation(
                   }
                 }
               }
-              ...ModeratedByContainer_comment
+              ...ModerateCardContainer_comment
             }
             moderationQueues(
               storyID: $storyID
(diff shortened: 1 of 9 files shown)


wyattjoh:
Note that this also fixes a bug where the moderation cards would temporarily show "Moderated by System" after a user moderated a comment, this is because the `viewerDidModerate` is cleared after the server response is received, this corrects that be repeating the mutation in the `updater`.

nick-funk (approved):

tessalt (approved):

PR #3026, opened 2020-07-16
[CORL-1185] translate error from iframe embeds

tessalt:
## What does this PR do?

send back a translated error message in the media iframe if oembed request returns nothing. 

none


src/core/server/app/handlers/api/oembed/oembed.ts
```diff
@@ -46,14 +39,27 @@ export const oembedHandler = (): RequestHandler => {
...
+            font-family: sans-serif;
+          }
+          .container * {
+            margin: 0!important;
+          }
+      `;
       const response = await fetchOEmbedResponse(type, url, maxWidth);
       if (response === null || !response.html) {
+        const bundle = i18n.getBundle(tenant.locale);
+        const error = new EmbedNotFound().serializeExtensions(bundle);
```
wyattjoh: Clever solution.

I'd just recommend instead just including this translation message in the `src/core/server/locales/en-US/common.ftl` file and using the `translate` function, as we don't throw this error, and the `serializeExtensions` is typically only used in the GraphQL responses and logs.

src/core/server/app/handlers/api/oembed/oembed.ts
```diff
@@ -46,14 +39,31 @@ export const oembedHandler = (): RequestHandler => {
...
+            margin: 0!important;
+          }
+      `;
       const response = await fetchOEmbedResponse(type, url, maxWidth);
       if (response === null || !response.html) {
+        const bundle = i18n.getBundle(tenant.locale);
+        const message = translate(
+          bundle,
+          "Requested media could not be found",
+          "commen-embedNotFound"
```
nick-funk: typo here `commen-...`, I think you meant `comment-...`.

nick-funk (changes requested):

nick-funk (approved):

PR #3035, opened 2020-07-20
[CORL-1172] Send staff comments to Slack with all comments enabled

nick-funk:
## What does this PR do?

When we set up a Slack channel, you can select to have
"All comments" sent to the slack channel. If we do this
we want staff comments to come through as well.

This wasn't happening before because there is a `staffCreated`
flag instead of `created` when a staff member comments.

So we change the trigger return type to an array, and when
a staff comment appears, return `staffCreated` as well as
`created` so the all filter lets it through.

I also changed `getBlocks(...)` to `getContent(...)` as it
now returns a JSON object with both a `blocks` array and a
`text` string field. This fixes the problem where Slack would
notifications would say "this content cannot be displayed".
Now it will say the text provided.

None

## How do I test this PR?

- Create or update an existing slack channel
- Set it to only have `All comments` checked under the channel options
- Post a comment as a staff user
- See it show up in Slack

### DIFF
--- a/src/core/server/events/listeners/slack/publishEvent.ts
+++ b/src/core/server/events/listeners/slack/publishEvent.ts
@@ -14,7 +14,12 @@ import {
   GQLUSER_ROLE,
 } from "coral-server/graph/schema/__generated__/types";
 
-export type Trigger = "reported" | "pending" | "featured" | "created";
+export type Trigger =
+  | "reported"
+  | "pending"
+  | "featured"
+  | "created"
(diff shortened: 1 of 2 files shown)


wyattjoh:
May I suggest modifying the `getTrigger` method to `getTriggers` and have it return an array? That way you can instead check in `shouldPublishToChannel` if `trigger.includes("created")` and other events which might be a bit clearer?

wyattjoh (approved):

PR #3039, opened 2020-07-22
pass custom stream css url through to html

tessalt:
## What does this PR do?

allow a custom CSS URL to be passed through via embed code. This is a naive implementation, should we be validating this URL? csp and cors should prevent loading of anything undesirable... 

none

### DIFF
--- a/src/core/server/app/router/client.ts
+++ b/src/core/server/app/router/client.ts
@@ -54,6 +54,11 @@ export interface ClientTargetHandlerOptions {
    * available on the Tenant.
    */
   enableCustomCSS?: boolean;
+  /**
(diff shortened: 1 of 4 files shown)


src/core/server/app/router/client.ts
```diff
@@ -120,6 +126,8 @@ const clientHandler = ({
     enableCustomCSS,
     locale,
     config,
+    enableCustomCSSQuery,
+    customCSSURL: req.query.customCSSURL,
```
wyattjoh: Just a small suggestion!

```suggestion
    customCSSURL: enableCustomCSSQuery ? req.query.customCSSURL : null,
```

Which might simplify things a tad in the template!

wyattjoh (approved):

PR #3040, opened 2020-07-22
[CORL-1149] Active Stories

wyattjoh:
## What does this PR do?

This introduces a new jsonp edge designed to return data on active stories with their comment counts in the following shape:

```ts
interface ActiveHandlerResponse {
  stories: Array<{
    id: string;
    url: string;
    title: string | null;
    image: string | null;
    publishedAt: string | null;
    count: number;
  }>;
}
```

None.

## How do I test this PR?

TBD


tessalt (approved):
👍

PR #3041, opened 2020-07-23
Stream UI bugfixes

alinhle:
## What does this PR do?

Address issues like:
- UserBox Authenticated: Should be `align-items: flex-end`
- Read more of this conversation has weird padding.
- Indent missing for last level reply
- Topbar Left should be "flex alignItems center"
- `aria-pressed` for reaction buttons needed to style pressed states
- share button missing classname
- counter missing classname
- title of verify email address has no styling
- Simplify styling some of the buttons
- Add more stable css classnames
- RTE radius 
- Remove `line-height` form css variables
- Make `line-height` values relative


nick-funk (approved):

kgardnr (approved):
Got approval from Kiwi

PR #3047, opened 2020-07-27
ensure preloaded css is actually used

tessalt:
## What does this PR do?

Forgot to include the macro to actually use the custom stylesheet after its been preloaded

none


wyattjoh (approved):
Nice catch!

PR #3049, opened 2020-07-28
don't save clientMutationId when creating announcement

tessalt:
## What does this PR do?

I don't think this fixes the bug we're seeing with announcements but we shouldn't be storing clientMutationId in the db. 

none


wyattjoh:
Instead of doing this in the mutator, you could do this in the Mutation resolver!

https://github.com/coralproject/talk/blob/1f72cdbfb36a5435360f16f67d25e8dbb89eeb68/src/core/server/graph/resolvers/Mutation.ts#L34-L41

wyattjoh (approved):

PR #3050, opened 2020-07-28
[CORL-1150] Discussions Tab

wyattjoh:
## What does this PR do?

Adds support for the discussions tab with active stories across the site and conversations that the user has made.

- Added `Site.topStories`
- Added `User.ongoingDiscussions`

## How do I test this PR?

visit the "discussions" tab from stream as a commenter who has made multiple comments on multiple stories.


src/core/server/graph/schema/schema.graphql
```diff
@@ -2429,6 +2442,7 @@ type User {
   ssoURL is the url for managing sso account
   """
   ssoURL: String
+    @auth(userIDField: "id", permit: [SUSPENDED, BANNED, PENDING_DELETION])
```
tessalt, line 2450: is this related or just a fix while you're here?

wyattjoh: Yes. It's only accessed via the `viewer` in our codebase, and shouldn't be hit unless the user is logged in, which this `@auth` directive is just enforcing.

nick-funk (approved):

kgardnr (approved):

PR #3056, opened 2020-07-31
Docs update- Jul 2020

immber:
## What does this PR do?

Documentation updates

## How do I test this PR?

build docs locally or verify netlify preview

### DIFF
--- a/docs/source/01-03-administration.md
+++ b/docs/source/01-03-administration.md
@@ -60,7 +60,7 @@ characters. Commenters will be alerted that they have gone over that number and 
 Set a limit on how long commenters have to edit their comments sitewide. Edited comments are marked as (Edited) on the comment stream and the moderation panel.
 
 #### **Closing comment streams**
-Set comment streams to close after a defined period of time after a story’s publication
+Set comment streams to close after a defined period of time after a story’s publication. When this feature is enabled, existing stories older than the defined period will also be updated and set as closed. 
 
 #### **Closed comment stream message**
 A message to appear when a story is closed for commenting.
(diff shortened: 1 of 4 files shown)


docs/source/01-03-administration.md
```diff
@@ -60,7 +60,7 @@ characters. Commenters will be alerted that they have gone over that number and
 Set a limit on how long commenters have to edit their comments sitewide. Edited comments are marked as (Edited) on the comment stream and the moderation panel.
 
 #### **Closing comment streams**
-Set comment streams to close after a defined period of time after a story’s publication
+Set comment streams to close after a defined period of time after a story’s publication. When this feature is enabled, existing stories older than the defined period will also be updated and set as closed.
```
wyattjoh: This is only the case when the story has never been closed/opened manually. If it has been, it will be unaffected.

docs/source/02-02-sso.md
```diff
@@ -129,3 +125,8 @@ var embed = Coral.createStreamEmbed({
   }
 });
 ```
+## Troubleshooting JWT Validation Errors
+
+Each `user.id`/`user.email` combination must be unique inside Coral. This is true for both Single Sign On users created by JWT tokens, as well as users that register/login using Coral’s built in “Login with email authentication”. You can not share authentication strategies for a single user; thus if a user logs in with SSO, they can not also login with email and vice versa. When you attempt to authenticate a JWT token with an email address that already exists in Coral with a different `user.id` than was passed on the token Coral will throw a Duplicate User error.
```
wyattjoh: It’s not just the combination that is unique, it’s each individually.

docs/source/01-03-administration.md
immber: Will change to clarify

docs/source/02-02-sso.md
immber: Clarified the constraints better.

wyattjoh (approved):

PR #3059, opened 2020-07-31
add react hooks eslint plugin

tessalt:
## What does this PR do?

adds an eslint plugin to enforce hooks rules, some of which are weird and cause unexpected runtime errors like the one fixed in https://github.com/coralproject/talk/pull/3057. Plugin does catch this issue: 

Screen Shot 2020-07-31 at 10 47 59 AM

none


nick-funk (approved):

.eslintrc.js
```diff
@@ -175,6 +175,7 @@ module.exports = {
     "eslint:recommended",
     "plugin:jsdoc/recommended",
     "plugin:prettier/recommended",
+    "plugin:react-hooks/recommended",
```
wyattjoh: Instead of this, can you instead do:

```js
{
  "plugins": [
    // ...
    "react-hooks"
  ],
  "rules": {
    // ...
    "react-hooks/rules-of-hooks": "warn",
    "react-hooks/exhaustive-deps": "warn"
  }
}
```

That way we'll get the warnings to keep us in check, and when we've eventually fixed all those warnings we can switch it back to this.

wyattjoh (changes requested):

wyattjoh (approved):

PR #3065, opened 2020-07-31
Add support to request Perspective API via proxy

desk467:
## What does this PR do?

This PR introduces the possibility to make requests to Perspective API in an environment where it's necessary to make additional proxy configurations.

The `proxyUrl` field has been added to `PerspectiveExternalIntegration`, which will save the proxy server configuration when it's set via admin interface.

## How do I test this PR?

Through the admin interface dedicated for moderation (at the URL /admin/configure/moderation) you need to activate Toxic Comment Filter. Then, you can add a proxy server URL at the end of this section, in the  textfield `Perspective proxy URL`. After that, when server requests Perspective API, these requests will go through the proxy server provided.


wyattjoh:
Closing for now as we don't have the resources to test/validate this feature unfortunately. Please re-open if there's a dire need!

PR #3067, opened 2020-08-03
[CORL-1233] Mobile display fixes

tessalt:
## What does this PR do?

corrects some mobile display issues for comments that are 3+ replies deep, text and buttons were being cut off at < 400px wide screens

none

### DIFF
--- a/src/core/client/stream/tabs/Comments/Comment/Comment.css
+++ b/src/core/client/stream/tabs/Comments/Comment/Comment.css
@@ -2,6 +2,7 @@ $commentTimestampColor: var(--palette-grey-500);
 
 .root {
   flex: 1;
+  overflow: hidden;
 }
 
 .highlight {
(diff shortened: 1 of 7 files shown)


nick-funk:
I'm noticing that the media query changes are prematurely shrinking the respect/reply/share butttons to their icon format even when we have the space to show them fully.

Is this what we want? Maybe we need to rethink our fixed media query sizes and use more granular values?

image

tessalt:
@nick-funk yeah it seems like there's enough space to show the full word, but there isn't if the comment is nested more than 2 levels. To get the best of both worlds we'd need to either a) add another media query between xs and sm (though this wouldn't fix 100% of cases, just get us closer) b) show/hide text based on the level of indent c)... invent container queries? any other ideas?

nick-funk:
> @nick-funk yeah it seems like there's enough space to show the full word, but there isn't if the comment is nested more than 2 levels. To get the best of both worlds we'd need to either a) add another media query between xs and sm (though this wouldn't fix 100% of cases, just get us closer) b) show/hide text based on the level of indent c)... invent container queries? any other ideas?

To be honest, my only fix was to add an extra media query width than our fixed set we currently have.

This is stepping into the "huge refactor" or "small duct tape fix" trade-off moment. I'd bring it up with Kim and Wyatt then see what they want to have done given our schedule.

nick-funk (approved):

PR #3069, opened 2020-08-03
[CORL-1227] Fix scroll-able modal position popping

nick-funk:
## What does this PR do?

**Adds a `disableScroll` option onto modals and uses it for the `SiteModeratorModal`.**

The site moderator modal already has a scroll bar inside it and its height is less than 480px. We don't need a scrollbar on the modal background since the modal window will always be fully visible on even the smallest screens.

This eliminates the bug of the modal sliding away when there are enough sites in the org to cause the background scroll to be activated.

**Fixes the input element positioning on our custom Checkboxes**

The checkboxes use a `::before` to create a custom styled checkbox that matches our theming. They also hide away the actual input element, which was not being positioned relative to the checkbox root. Because of this, they were appearing relative to the scrolled modal root, sometimes far below the modal's scroll-able area, which shoved the modal way up on the screen.

This positions the checkbox input element relative to the checkbox root and fixes it to the positioning of the stylized checkbox. This prevents it from appearing in weird places due to being inside a scroll-able element or paginated, scroll-able element.

None

## How do I test this PR?

- Enable the `SITE_MODERATOR` feature flag
- Create a lot of sites under your org (10-20 should do)
- Shrink the height of your window to roughly 600px
- Go to the community page
- Go to change a user's role to site moderator
- Scroll down through your massive list of sites
- See that the modal background does not have a scroll bar
- See that the modal does not scroll up and away through the top of the screen

### DIFF
--- a/src/core/client/admin/components/UserRole/SiteModeratorModal.tsx
+++ b/src/core/client/admin/components/UserRole/SiteModeratorModal.tsx
@@ -56,7 +56,7 @@ const SiteModeratorModal: FunctionComponent<Props> = ({
   );
 
   return (
-    <Modal open={open} onClose={onCancel}>
+    <Modal open={open} onClose={onCancel} disableScroll>
       {({ firstFocusableRef, lastFocusableRef }) => (
         <Card className={styles.root}>
           <Flex justifyContent="flex-end">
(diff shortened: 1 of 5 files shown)


wyattjoh:
This still seems like an issue on Google Chrome when selecting a site that's further than the first page.

Screen Shot 2020-08-03 at 6 11 26 PM

wyattjoh (changes requested):
This still seems to be an issue on Google Chrome:

Screen Shot 2020-08-03 at 6 11 26 PM

tessalt (approved):
this makes sense to me

wyattjoh (approved):

PR #3070, opened 2020-08-03
[CORL-1159] Fix deleting old SSO keys

nick-funk:
## What does this PR do?

Fixes the deletion of rotated out SSO keys by replacing `keys` with the correct name `signingSecrets` in the `$pull` mongo call.

Looks like this was missed during the migration of `auth.integrations.sso.keys` renamed to `auth.integrations.sso.signingSecrets`.

None

## How do I test this PR?

- Go to `Admin > Config > Authentication`
- Enable SSO 
- Create/rotate out a key
- Delete the old expired key
- See that it updates and key is deleted

### DIFF
--- a/src/core/server/models/tenant/sso.ts
+++ b/src/core/server/models/tenant/sso.ts
@@ -31,8 +31,8 @@ export async function deactivateTenantSSOSigningSecret(
     { id },
     {
       $set: {
-        "auth.integrations.sso.keys.$[keys].inactiveAt": inactiveAt,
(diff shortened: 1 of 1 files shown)


wyattjoh (changes requested):
Looks like you missed the `...sso.keys` in the `deactivateTenantSSOSigningSecret` function. Can you update those three lines as well with this fix? Thanks!

wyattjoh (approved):

PR #3073, opened 2020-08-04
[CORL-794] Create a common date formatter

nick-funk:
## What does this PR do?

Updates all uses of Intl.DateTimeFormat(...)
with re-usable date time formatter functions.

The formatter functions have default formatting
options, but these can be overridden.

None

## How do I test this PR?

Look around the site at various dates (comment stream time stamps is a good area) and see that the new consolidated date formatting is equivalent to before.

### DIFF
--- a/src/core/client/ui/hooks/useDateTimeFormat.ts
+++ b/src/core/client/ui/hooks/useDateTimeFormat.ts
@@ -1,11 +1,9 @@
 import { useMemo } from "react";
 
+import { createDateFormatter } from "coral-common/date";
 import { useUIContext } from "coral-ui/components/v2";
 
 export default function useDateTimeFormat(options: Intl.DateTimeFormatOptions) {
   const { locales } = useUIContext();
-  return useMemo(() => new Intl.DateTimeFormat(locales, options), [
-    locales,
-    options,
-  ]);
+  return useMemo(() => createDateFormatter(locales, options), [locales]);
 }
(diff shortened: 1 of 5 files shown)


src/core/client/ui/hooks/useDateTimeFormat.ts
```diff
@@ -1,11 +1,9 @@
...
+import { createDateFormatter } from "coral-common/date";
 import { useUIContext } from "coral-ui/components/v2";
 
 export default function useDateTimeFormat(options: Intl.DateTimeFormatOptions) {
   const { locales } = useUIContext();
-  return useMemo(() => new Intl.DateTimeFormat(locales, options), [
-    locales,
-    options,
-  ]);
+  return useMemo(() => createDateFormatter(locales, options), [locales]);
```
wyattjoh:
```suggestion
  return useMemo(() => createDateFormatter(locales, options), [locales, options]);
```

`options` must be included as a dependency in `useMemo` in case the value ever changes.

src/core/common/date.ts
```diff
@@ -0,0 +1,20 @@
+export function createDateFormatter(
+  locales: string[] | undefined,
```
wyattjoh: suggestion: because technically `Intl.DateTimeFormat` can accept `string | string[] | undefined`, modifying this type will allow the other uses of it (like in `src/core/server/services/users/download/download.ts`) to just pass the `string` instead.

Same thing with the `formatDate` function here (that's used in `src/core/server/services/users/users.ts`).

wyattjoh (changes requested):

wyattjoh (approved):

PR #3074, opened 2020-08-05
[CORL-1243] Subscription Errors

wyattjoh:
## What does this PR do?

In the process of debugging this issue, I replaced the `CommentContainer` `React.Component` with a `React.FunctionComponent`, which replaced all the private methods on the class with calls to `useCallback` and `useEffect`. This simplification yielded a few fixes to the component and reduced the duplication of logic within the component.

The actual problem was solved when I checked the `pending` property on the comment before subscribing to events on the comment.

None.

## How do I test this PR?

Write a few comments and notice in the console logs that it never unsubscribes from any events (only adding).


src/core/client/stream/tabs/Comments/Comment/IndentedComment.tsx
```diff
@@ -21,38 +21,33 @@ export interface IndentedCommentProps
...
+}) => {
+  return (
     <Indent
       level={indentLevel}
       collapsed={rest.collapsed}
       className={cn(
         {
           [styles.open]: !rest.collapsed,
-          [styles.blur]: props.blur,
+          [styles.blur]: rest.blur,
```
alinhle: `blur` shouldn't be passed down to `CommentToggle` nor `Comment` as rest.

src/core/client/stream/tabs/Comments/ReplyList/ReplyListContainer.tsx
```diff
@@ -79,27 +81,54 @@ export const ReplyListContainer: React.FunctionComponent<Props> = (props) => {
...
+
     return () => {
       disposable.dispose();
     };
   }, [
     subcribeToCommentReplyCreated,
     props.comment.id,
     props.indentLevel,
+    // TODO: (wyattjoh) investigate why this is here?
     props.relay.hasMore(),
```
alinhle: Dependency list should really contain all variables that is used inside the effect callback:
```ts
[
    subcribeToCommentReplyCreated,
    props.comment.id,
    props.indentLevel,
    props.comment.pending,
    props.settings.disableCommenting.enabled,
    props.liveDirectRepliesInsertion,
    props.story.isClosed,
    props.story.closedAt,
    props.story.settings.live.enabled,
  ]
```

I think `props.relay.hasMore()` can be removed, it seems to be an copy&paste leftover.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabContainer.tsx
```diff
@@ -67,56 +74,82 @@ export const AllCommentsTabContainer: FunctionComponent<Props> = (props) => {
...
   }, [
     commentsOrderBy,
     subscribeToCommentCreated,
     subscribeToCommentReleased,
-    props.story.id,
-    props.relay.hasMore(),
-    props.story.settings.live.enabled,
+    story.id,
+    relay.hasMore(),
+    story.settings.live.enabled,
```
alinhle, line 147: See above:
```ts
[
    commentsOrderBy,
    subscribeToCommentCreated,
    subscribeToCommentReleased,
    story.id,
    story.isClosed,
    story.closedAt,
    story.settings.live.enabled,
    settings.disableCommenting.enabled,
    relay.hasMore(),
  ]
```

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -82,438 +89,421 @@ interface Props {
...
-    }
-  }
+      void setCommentID({ id: comment.id });
 
-  public componentWillUnmount() {
-    clearTimeout(this.uneditableTimer);
-  }
+      return false;
+    },
+    [eventEmitter, comment, setCommentID]
```
alinhle: instead of comment, here it's more optimised to set `comment.id` into the dependency array.

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -82,438 +89,421 @@ interface Props {
...
+      ShowEditFormEvent.emit(eventEmitter, {
+        commentID: comment.id,
       });
+      setShowEditDialog(true);
     } else {
-      void this.props.showAuthPopup({ view: "SIGN_IN" });
+      void showAuthPopup({ view: "SIGN_IN" });
     }
-  };
+  }, [viewer, eventEmitter, comment, showAuthPopup]);
```
alinhle: Possible optimisation: `comment.id` instead of `comment`. 
Turn viewer into a boolean before and use that in deps:

```
const isLoggedIn = Boolean(viewer);
const openEditDialog = useCallback(() => {
  if (isLoggedIn) {
   ...
}, [isLoggedIn, ...]);
```

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -82,438 +89,421 @@ interface Props {
...
-      new Date(this.props.comment.editing.editableUntil!).getTime() -
-      Date.now();
-    if (ms > 0) {
-      return setLongTimeout(() => this.setState({ editable: false }), ms);
+      void showAuthPopup({ view: "SIGN_IN" });
     }
-    return;
-  }
+  }, [
+    viewer,
```
alinhle: See above

alinhle (changes requested):
Great work! Here are some suggestions for even more improvements :-)

alinhle (approved):

PR #3080, opened 2020-08-05
[CORL-1242] Ensure SSO Access Tokens are not persisted

wyattjoh:
## What does this PR do?

This ensures that access tokens provided to the stream for single sign on are not persisted after navigation.

None.

## How do I test this PR?

Provide an access token via the `embed.login(<access token>)` method, and refresh the page. See that you are logged out after the refresh.


alinhle (commented):
Is it expected that SSO users when they click on `moderate` on the stream to get to the admin, that the token gets stored into localstorage?

wyattjoh:
That is expected then, yes.

alinhle (approved):

alinhle:
Good job! 👍

PR #3101, opened 2020-08-12
[CORL-1277] Replace fetch for axios on perspective requests

nick-funk:
## What does this PR do?

Swaps out Perspective's createFetch requests with axios requests that natively support time outs. 

If this mitigates issues of failed requests and doesn't fail in other ways, this means we can switch over to this version of time out based fetch requests and call it a day.

None

## How do I test this PR?

Locally:
- setting up perspective with an API key
- creating some bad comments
- see they get flagged and notify you when posting

In production:

- watch and see if we get reduced perspective network errors over time


src/core/server/services/axios.ts/axios.ts
```diff
@@ -0,0 +1,79 @@
+import axios, { Method } from "axios";
```
wyattjoh: Please move the `axios.ts/axios.ts` to `fetch` instead and export it from `fetch/index.ts`.

src/core/server/services/axios.ts/axios.ts
```diff
@@ -0,0 +1,79 @@
...
+      keepAlive: true,
+    }),
+  };
+
+  return async (
+    url: string,
+    { method = "GET", headers = {}, body = {} }: AxiosOptions,
+    timeout?: number
+  ) => {
+    const response = await axios.request({
```
wyattjoh: Should add the following parameters to match the config to existing fetch implementation:

- `maxContentLength: 2e6`
- `maxRedirects: 0`

src/core/server/services/axios.ts/axios.ts
```diff
@@ -0,0 +1,79 @@
...
+    }),
+    http: new http.Agent({
+      keepAlive: true,
+    }),
+  };
+
+  return async (
+    url: string,
+    { method = "GET", headers = {}, body = {} }: AxiosOptions,
+    timeout?: number
```
wyattjoh: Should enforce a default timeout of `10000ms`, and I'd suggest moving the timeout into the `AxiosOptions` to keep it similar to the fetch API.

src/core/server/services/perspective/perspective.ts
```diff
@@ -3,10 +3,12 @@ import { URL } from "url";
 
 import { TOXICITY_ENDPOINT_DEFAULT } from "coral-common/constants";
 import { LanguageCode } from "coral-common/helpers";
+import logger from "coral-server/logger";
 import { getURLWithCommentID } from "coral-server/models/story";
-import { createFetch } from "coral-server/services/fetch";
 
-const fetch = createFetch({ name: "perspective" });
+import { createAxios } from "../axios.ts";
```
wyattjoh: This'll need updates after the file is moved to `coral-server/services/fetch`.

src/core/server/services/perspective/perspective.ts
```diff
@@ -169,6 +176,10 @@ export async function sendToPerspective(
       );
     }
 
+    if (err.message && err.message.startsWith("axios request timed out")) {
+      logger.error("perspective request timed out");
```
wyattjoh: Instead of logging here, throw an error that matches the same error text as the fetch API's error message when the timeout occurs, then @immber doesn't need to update the existing reporting flows to track incidents.

wyattjoh (changes requested):

kgardnr (approved):

wyattjoh (approved):

PR #3103, opened 2020-08-12
[CORL-1291] Logging Improvements

wyattjoh:
## What does this PR do?

Whenever settings are updated, log the fields being updated as well as the user performing the action.

None.

## How do I test this PR?

Update settings, see the new log entry: `settings update audit`.


tessalt (commented):
Because of the way our settings forms work, this is logging every key from whatever page you were on: 
image

(In this case I only edited the media rating setting)

wyattjoh:
That’s true. This is a stop gap until we’re able to replace our settings update mutations with more discrete mutations that would let us apply auditing on the mutation/schema level.

tessalt (approved):

PR #3104, opened 2020-08-12
[CORL-1281] Improve Type Handling

wyattjoh:
## What does this PR do?

Various areas of the front end we access the relay cache without property typechecking controls (using `thing!.other!.thing` to bypass the strict null checks). This modifies some of those critical paths to add checks to prevent errors from occuring.

None.

## How do I test this PR?

Try to interact with the moderation page when comments are coming in and see no errors occur.


alinhle (changes requested):
I think a lot of these cases, are correct. If it's not in the relay cache, then there is a bigger problem and needs investigation. Silently returning would just mask the underlying problem.

wyattjoh:
Closing for now till we can handle this better.

PR #3105, opened 2020-08-13
[CORL-1285] Support older browser

alinhle:
## What does this PR do?
- Polyfills are now using feature detection instead of parsing the user agent, which allows better support for older browsers.
- Add `"iOS >= 9", "Android >= 4.4.4", "Edge >= 17", "Firefox >= 68", "Chrome >= 49"` to browserlist.
- Polyfill `Intl` on demand


src/core/client/framework/lib/browserInfo.ts
```diff
@@ -1,20 +1,44 @@
...
+  version: number;
   ios: boolean;
   msie: boolean;
   mobile: boolean;
+  supportsCSSVariables: boolean;
+  supportsFetch: boolean;
+  supportsProxyObject: boolean;
+  supportsIntl: boolean;
+  supportsIntlPluralRules: boolean;
+  supportsIntersectionObserver: boolean;
```
wyattjoh: May I suggest putting this all under a `supports` key?

wyattjoh (approved):

PR #3111, opened 2020-08-17
ensure getlinkedRecord is truthy before retrieving value in queue subscription

tessalt:
Ok I'm not sure this is actually the _real_ problem but this addresses the `cannot call getValue on undefined` error was seeing in prod on clicking "view x new comments" in the moderation queue. 
Screen Shot 2020-08-17 at 12 43 38 PM

why was the `!` used in the first place, is there a reason `getLinkedRecord` should always be truthy here that I'm missing? If it's undefined, what could that mean? much to discuss, maybe

### DIFF
--- a/src/core/client/admin/routes/Moderate/Queue/QueueCommentLeftSubscription.tsx
+++ b/src/core/client/admin/routes/Moderate/Queue/QueueCommentLeftSubscription.tsx
@@ -33,9 +33,12 @@ function handleCommentLeftModerationQueue(
   if (connection) {
     const linked = connection.getLinkedRecords("viewNewEdges") || [];
     connection.setLinkedRecords(
-      linked.filter(
-        (r) => r.getLinkedRecord("node")!.getValue("id") !== commentID
-      ),
+      linked.filter((r) => {
+        return (
+          r.getLinkedRecord("node") &&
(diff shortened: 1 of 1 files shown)


src/core/client/admin/routes/Moderate/Queue/QueueCommentLeftSubscription.tsx
```diff
@@ -33,9 +33,12 @@ function handleCommentLeftModerationQueue(
   if (connection) {
     const linked = connection.getLinkedRecords("viewNewEdges") || [];
     connection.setLinkedRecords(
-      linked.filter(
-        (r) => r.getLinkedRecord("node")!.getValue("id") !== commentID
-      ),
+      linked.filter((r) => {
+        return (
+          r.getLinkedRecord("node") &&
+          r.getLinkedRecord("node")!.getValue("id") !== commentID
```
wyattjoh, line 39: You can actually write this as:

```suggestion
          r.getLinkedRecord("node")?.getValue("id") !== commentID
```

Though this is related already to the issue that @cvle is already assigned (CORL-1281), so we might need to wait for this fix.

alinhle:
I'm looking into this already, there is probably more to it.

PR #3115, opened 2020-08-18
[CORL-1281] Speed up comment contents loading by removing wordlist from queries

tessalt:
## What does this PR do?
makes wordlist optional for rendering comment contents, loading wordlist is slowing things down for conversation modal and edit history. 

none


wyattjoh (changes requested):
I'd suggest modifying this part:

https://github.com/coralproject/talk/blob/0f4a4e3617c3575580f126f1ef1f43f775e0c185/src/core/client/admin/components/Comment/CommentContent.tsx#L83-L85

To instead read:

```ts
    // Sanitize the input for display.
    const node = getSanitize(highlight && !!phrases)(children);
```

wyattjoh (approved):

PR #3133, opened 2020-08-24
[CORL-1320] Add badges back to CommentContainer fragment

nick-funk:
## What does this PR do?

The `badges` were missing from `author` on the `CommentContainer`'s comment fragment.

This is despite the fact that it was defined in the `AuthorBadgesContainer_comment` fragment. 

I am hesitant to call it so, but there appears to be a bug with React being unable to process how we've nested the rendering of the user badges down multiple conditional levels of `CommentContainer`, `IndentedComment`, `CommentToggle`, `Comment`, and then `AuthorBadgesContainer`.

Even switching the `AuthorBadgesContainer_comment` to be `AuthorBadgesContainer_user` and nesting it on the `author` of the  `CommentContainer_comment` doesn't work.

I don't know what to say, but we have a pretty gnarly React rendering bug.

All that being said, this will fix the rendering issue and cause the badges to appear on both the collapsed and non-collapsed state of a comment.

None

## How do I test this PR?

- Set badges on a commenter
  - To do this, simply set `badges: [ "some-badge" ]` on a commenter in Mongo
- Go to stream
- Sign in and create a comment as the above commenter
- Collapse and open the comment you just created
- See that the badges you set prior are visible in both states


wyattjoh:
Upon some further discussions with @nick-funk, we were able to simplify and solve the problem. Remember that the `any` type is evil!

wyattjoh (approved):

PR #3149, opened 2020-08-27
[CORL-1296] New Comment Count

wyattjoh:
## What does this PR do?

With the `NEW_COMMENT_COUNT` feature flag, comment streams viewed will record the current comment count. When used in conjunction with [`count.js`](https://docs.coralproject.net/coral/v5/integrating/counts/), it will show `/ 5 New` when the number of comments exceeds what was previously recorded.

- Added `FEATURE_FLAG.NEW_COMMENT_COUNT`

## How do I test this PR?

1. Enable the `NEW_COMMENT_COUNT` feature flag.
2. View comments on a story.
3. In another browser, add a comment to that story.
4. In the first browser, view the comment count widget and notice the new `/ 1 New` added to the comment count text.


src/core/server/services/giphy/giphy.ts
```diff
@@ -6,7 +6,7 @@ import { LanguageCode } from "coral-common/helpers";
 import {
   GiphyGifRetrieveResponse,
   GiphyGifSearchResponse,
-} from "coral-common/rest/external/giphy";
+} from "coral-common/types/giphy";
```
nick-funk, line 9: Why were Giphy files modified in this PR?

nick-funk:
I am only seeing the `2 comments / 1 new` counter update if I refresh the page. Is this expected behaviour?

nick-funk (approved):

src/core/server/services/giphy/giphy.ts
wyattjoh: I moved types around to keep things consistent!

PR #3156, opened 2020-09-01
[CORL-1327] allow configuration of admin and moderator comment tags

tessalt:
## What does this PR do?

Adds new more specific tags to comments based on user role: instead of tagging all comments from staff, moderator, and admin users as "staff", they are now tagged with their respective roles. Also allows customization of display labels for these tags, defaulting all three to "staff". 

- Adds `ADMIN` and `MODERATOR` tags to `Tag` enum
- adds `staff.adminLabel` and `staff.moderatorLabel` to tenant settings 

## How do I test this PR?

- under general configuration, change the labels for moderator or admin roles to something other than "staff"
- make a comment as a moderator or admin
- observe the comment being tagged with the appropriate role


src/core/server/graph/schema/schema.graphql
```diff
@@ -4579,6 +4603,16 @@ input StaffConfigurationInput {
   label is the string used when displaying the STAFF badge to users.
   """
   label: String
+
```
wyattjoh, line 4609: I'd suggest the introduction of a new `staffLabel`. Mark the existing `label` as `@deprecated` and replace usage with `staffLabel`.

wyattjoh (changes requested):

src/core/server/graph/schema/schema.graphql
```diff
@@ -1514,6 +1514,18 @@ type StaffConfiguration {
...
+  moderatorLabel is the string used when displaying the MODERATOR badge to users. Defaults
+  to value of `staff.label`
+  """
+  moderatorLabel: String
+
+  """
+  adminLabel is the string used when displaying the ADMIN badge to users. Defaults
+  to value of `staff.label`
+  """
+  adminLabel: String
```
wyattjoh: These should be `String!` and not `String`.

wyattjoh: Remove the notice about the default value as well because that's applied only until the first save is hit.

src/core/server/graph/schema/schema.graphql
tessalt: would this require a migration?

wyattjoh: Nope! We just adjust it on the resolver level, not in the database.

src/core/server/graph/resolvers/StaffConfig.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const StaffConfiguration: GQLStaffConfigurationTypeResolver<settings.StaffConfiguration> = {
+  // MIGRATE: plan to remove this in 7.0.0.
+  label: (config, args, { tenant }) => tenant.staff.label,
+  adminLabel: (config, args, { tenant }) =>
+    tenant.staff.adminLabel || tenant.staff.label,
+  moderatorLabel: (config, args, { tenant }) =>
+    tenant.staff.moderatorLabel || tenant.staff.label,
+  staffLabel: (config, args, { tenant }) =>
+    tenant.staff.staffLabel || tenant.staff.label,
```
wyattjoh: These can be accessed via `config.staffLabel` and `config.label` instead of on the tenant because we're in that component already.

src/core/server/graph/resolvers/Settings.ts
```diff
@@ -66,4 +66,5 @@ export const Settings: GQLSettingsTypeResolver<Tenant> = {
   rte: ({ rte = defaultRTEConfiguration }) => rte,
   media: ({ media = {} }) => media,
   live: ({ live }): LiveConfigurationInput => live,
+  staff: ({ staff }) => staff,
```
wyattjoh: This isn't required anymore because this is the default!

wyattjoh (changes requested):
Two more small tweaks and it's perfect!

wyattjoh (approved):

PR #3158, opened 2020-09-01
[CORL-1334] Keyboard shortcuts

nick-funk:
## What does this PR do?

Allows you to use the `C` and `Shift + C` key to traverse through the comment stream.

## Changes to schema?

None

## How to test

- Go to All Comments in stream
- Press `C` to traverse through visible comment tree
- Press `Shift + C` to reverse traverse through comments


src/core/client/stream/App/useKeys.ts
```diff
@@ -0,0 +1,48 @@
+import { useEffect, useState } from "react";
+
+function useKeys() {
+  const [key, setKey] = useState(null);
+
+  function handleKeyMessage(e: any) {
```
wyattjoh: This function should be wrapped in a `useCallback`.

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -0,0 +1,70 @@
...
+  if (!container) {
+    return;
+  }
+
+  container.scrollIntoView();
+  KeyedComments.setCurrentIndex(index);
+}
+
+const KeyboardShortcuts: FunctionComponent = () => {
+  const handleKeyMessage = useCallback((e: any) => {
```
wyattjoh: Should use concrete types instead of `any`.

src/core/client/embed/PymControl.ts
```diff
@@ -15,7 +15,7 @@ export const defaultPymControlFactory: PymControlFactory = (config) =>
   new PymControl(config);
 
 export default class PymControl {
-  private pym: pym.Parent;
+  public pym: pym.Parent;
```
wyattjoh: Revert to `private` as it doesn't seem it's used elsewhere.

src/core/client/embed/StreamEmbed.ts
```diff
@@ -167,6 +168,8 @@ export class StreamEmbed {
       decorators: streamDecorators,
       url,
     });
+
+    hookUpWindowEvents(this.pymControl);
```
wyattjoh: Use the existing `streamDecorators` array above and rewrite the `hookUpWindowEvents` as a decorator so it supports a unsubscribe function.

See https://github.com/coralproject/talk/blob/master/src/core/client/embed/decorators/withClickEvent.ts for an example.

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -0,0 +1,191 @@
...
+    [pym]
+  );
+
+  const getKeyStops = useCallback(() => {
+    const matches = document.querySelectorAll(`[data-keystop="true"]`);
+    return matches;
+  }, []);
+
+  const toKeyStop = useCallback((el: Element) => {
+    const id = el.attributes.getNamedItem("data-keyid");
```
wyattjoh: Data attributes can be retrieved instead via:

```
el.dataset.keyid
```

wyattjoh: Note that you must use `HTMLElement` instead of `Element` here to allow this.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabContainer.tsx
```diff
@@ -264,6 +266,8 @@ export const AllCommentsTabContainer: FunctionComponent<Props> = ({
               disabled={isLoadingMore}
               aria-controls="comments-allComments-log"
               className={CLASSES.allCommentsTabPane.loadMoreButton}
+              data-keystop={true}
+              data-isLoadMore={true}
```
wyattjoh: Data props typically are written in kebab case:

```
data-key-stop
data-is-load-more
```

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
wyattjoh: Note that if you rewrite access for data attributes to use kebab case, access via `.dataset` is done via camel case:

```
el.dataset.isLoadMore
```

wyattjoh (changes requested):

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabContainer.tsx
alinhle: I don't think that there is a convention.
`aria` props does not separate words:
```
aria-labelledby
aria-describedby
```
either way is fine for me though.

wyattjoh (approved):

alinhle (approved):

PR #3174, opened 2020-09-02
[CORL-1331] Always use our font versions

alinhle:
## What does this PR do?

We have a bug report that shows that the material icons couldn't be loaded:

Screenshot 2020-09-02 at 20 17 01

I wasn't able to reproduce this. I have a theory that this could be caused by corrupt local font installations. In order to rule out that issue, this patch enforces that our provided font versions are always used.


kgardnr (approved):

kgardnr:
That's what I thought too @cvle! I asked the user to delete custom fonts on their local machine, so we'll see if anything comes back on that

PR #3182, opened 2020-09-04
[CORL-1358] bios

tessalt:
## What does this PR do?

Adds support for commenter bios, which must be enabled through config. 

Adds new `settings.memberBios` config option on tenant, and `bio` field on users. 

## How do I test this PR?

- enable bios from Configuration -> General
- add a bio from stream by going to the preferences tab
- observe bio is visible when clicking on username in stream, in user drawer in admin.
- observe new "Bio is offensive or abusive" option for reporting a comment


nick-funk:
We may need to set a max image width here haha

That or disable images because it looks like they aren't being wrapped in an iframe when they're embedded on the page

Also, giant Gandalf says hi

image

nick-funk:
It looks like URL's are going to be problematic with the 100 character limit. Not sure whether we want to lengthen the limit or just not allow images.

image

nick-funk (changes requested):
The code looks good, but I think there are some design considerations to run by Sam that may keep this from going out. See my previous comments for details on that. 😃

tessalt:
@nick-funk I figured bios probably don't need links or images. I updated the markdown editor component to make the toolbar customizable and removed buttons we don't need.

kgardnr (approved):

nick-funk (approved):

wyattjoh (changes requested):
This poses to much of a security risk, holding back for additional review.

nick-funk:
@wyattjoh, I've replaced the RTE with a TextArea, have a look and tell me what you think

wyattjoh (approved):

PR #3186, opened 2020-09-09
[CORL-1361] Add badge class names to root of comments

nick-funk:
## What does this PR do?

If we have badges, use them to set a classname so we can bind to it.

None

## How do I test this PR?

- edit badges on a user
- post a comment as the user
- check the DOM and see that `badge-<your-badge>` is set on the root div for the comment


tessalt (approved):

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -321,6 +321,10 @@ export const CommentContainer: FunctionComponent<Props> = ({
     comment.author.badges.length > 0
       ? comment.author.badges
       : null;
+  const badgesJoin = badges
+    ?.map((b: string) => `badge-${b.replace(" ", "-")}`)
```
alinhle: Because every "official" classname so far has a `coral-` prefix, maybe we want to do that here too?

nick-funk: Agreed, it is done!

alinhle (approved):

PR #3187, opened 2020-09-09
[CORL-1341] Moderation Client Performance improvements

alinhle:
## What does this PR do?

- Batch subscription data, so that the client remains responsive even when bombarded with live updates.
- Only load newest 20 comments when clicking on "View new comments" and let infinite scrolling load the rest.


tessalt (approved):
👍

PR #3190, opened 2020-09-11
Site moderator promotions

tessalt:
## What does this PR do?

Implements changes needed to allow site moderators limited role change permissions, allowing moderator's to update a user's role in the following cases:
- moderators can only change role *to* moderator
- scoped moderators can only create scoped moderators
- scoped moderators can only add moderators to sites they already moderate
- moderators cannot demote admins

The complexity of these conditions may suggest another solution is required, such as the creation of "site admins". PR as a discussion point more than proposed feature.


src/core/server/graph/schema/schema.graphql
```diff
@@ -7716,7 +7722,7 @@ type Mutation {
   updateUserRole will update a given User's role.
   """
   updateUserRole(input: UpdateUserRoleInput!): UpdateUserRolePayload!
-    @auth(roles: [ADMIN])
+    @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh: We should not re-use the `updateUserRole` for this operation.

src/core/server/graph/schema/schema.graphql
```diff
@@ -7725,7 +7731,7 @@ type Mutation {
   """
   updateUserModerationScopes(
     input: UpdateUserModerationScopesInput!
-  ): UpdateUserModerationScopesPayload! @auth(roles: [ADMIN])
+  ): UpdateUserModerationScopesPayload! @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh: We should not re-use the `updateUserModerationScopes` for this operation.

src/core/server/services/users/users.ts
```diff
@@ -655,21 +656,50 @@ export async function updateUsernameByID(
...
   if (user.id === userID) {
     throw new Error("cannot update your own user role");
   }
 
-  return updateUserRole(mongo, tenant.id, userID, role);
+  // technically a site moderator can promote to organization moderator...
+  if (user.role === GQLUSER_ROLE.MODERATOR) {
+    if (role !== GQLUSER_ROLE.MODERATOR) {
+      throw new Error("Moderators can only promote to moderator");
+    }
```
wyattjoh: Authorization must generally not be completed within the service functions.

wyattjoh (changes requested):

src/core/server/graph/schema/schema.graphql
```diff
@@ -7772,7 +7811,13 @@ type Mutation {
   updateUserRole will update a given User's role.
   """
   updateUserRole(input: UpdateUserRoleInput!): UpdateUserRolePayload!
-    @auth(roles: [ADMIN])
+    @auth(roles: [ADMIN, MODERATOR])
+
+  """
+  promoteUser will allow you to promote a user to a moderator.
+  """
+  promoteUser(input: PromoteUserInput!): PromoteUserPayload!
```
wyattjoh, line 7895:
```suggestion
  promoteUserRole(input: PromoteUserRoleInput!): PromoteUserRolePayload!
```

src/core/server/services/users/users.ts
```diff
@@ -658,21 +692,133 @@ export async function updateUsernameByID(
...
+    );
+  }
+
+  return await updateUserRole(mongo, tenant.id, userID, role);
+}
+
+export async function promoteUser(
+  mongo: Db,
+  tenant: Tenant,
+  user: Pick<User, "id" | "role" | "moderationScopes">,
```
wyattjoh: I'd suggest calling this `viewer` to disassociate this from the user we're promoting a bit further.
```suggestion
  viewer: Pick<User, "id" | "role" | "moderationScopes">,
```

src/core/server/services/users/users.ts
```diff
@@ -658,21 +692,133 @@ export async function updateUsernameByID(
...
+) {
+  // you cannot update your own user role for whatever reason
+  if (user.id === userIDToPromote) {
+    throw new Error("cannot update your own user role");
+  }
+
+  // you can only promote users if you're currently an admin or moderator
+  if (![GQLUSER_ROLE.ADMIN, GQLUSER_ROLE.MODERATOR].includes(user.role)) {
+    throw new Error("you have insufficient permissions to promote users");
+  }
```
wyattjoh: This is not required as it's already handled by the `@auth` directive in the schema.

src/core/server/services/users/users.ts
```diff
@@ -658,21 +692,133 @@ export async function updateUsernameByID(
...
+
+  // you can't promote a user that is already an admin
+  const userToPromote = await retrieveUser(mongo, tenant.id, userIDToPromote);
+  if (userToPromote?.role === GQLUSER_ROLE.ADMIN) {
```
wyattjoh: This seems in error. If you need to use `?` in `userToPromote?.role`, it implies you should check for `if (!userToPromote) { throw new UserNotFound }` instead.

src/core/server/graph/schema/schema.graphql
```diff
@@ -6738,6 +6738,45 @@ type UpdateUserRolePayload {
...
+input PromoteUserInput {
+  """
+  userID is the ID of the User that should have their role updated.
+  """
+  userID: ID!
+
+  """
+  role is the `USER_ROLE` that the User should be set to.
+  """
+  role: USER_ROLE!
```
wyattjoh: This promotion edge is already exclusively used to promote users to moderator roles. As such it shouldn't accept a role and should assume moderator instead.

src/core/server/graph/schema/schema.graphql
```diff
@@ -6738,6 +6738,45 @@ type UpdateUserRolePayload {
...
+  """
+  clientMutationId is required for Relay support.
+  """
+  clientMutationId: String!
+
+  """
+  moderationScopes defines the scopes that a given moderator should be
+  limited to
+  """
+  moderationScopes: ModerationScopesInput
```
wyattjoh: These should also be implied, as the user should only be able to promote other users to the same level as them.

src/core/server/services/users/users.ts
```diff
@@ -658,21 +692,133 @@ export async function updateUsernameByID(
...
+    user.role === GQLUSER_ROLE.MODERATOR &&
+    isModerationScoped(user.moderationScopes)
+  ) {
+    throw new Error(
+      "scoped moderators cannot promote users to organization level moderators"
+    );
+  }
+
+  return await updateUserRole(mongo, tenant.id, userID, role);
+}
```
wyattjoh: These changes should be reverted because we should only be affecting the `promoteUser` instead.

src/core/server/services/users/users.ts
```diff
@@ -658,21 +692,133 @@ export async function updateUsernameByID(
...
+    hasFeatureFlag(tenant, GQLFEATURE_FLAG.SITE_MODERATOR)
+  ) {
+    await updateModerationScopes(
+      mongo,
+      tenant,
+      user,
+      userIDToPromote,
+      moderationScopes
+    );
+  }
```
wyattjoh: One DB call instead of two please!

src/core/server/services/users/users.ts
```diff
@@ -682,21 +772,38 @@ export async function updateModerationScopes(
     });
   }
 
-  if (user.id === userID) {
+  // you can only change scopes if you're currently an admin or moderator
+  if (![GQLUSER_ROLE.ADMIN, GQLUSER_ROLE.MODERATOR].includes(viewer.role)) {
+    throw new Error("you have insufficient permissions to promote users");
+  }
```
wyattjoh: This check is completed by the schema level and should not be duplicated here.

src/core/server/services/users/users.ts
```diff
@@ -682,21 +772,38 @@ export async function updateModerationScopes(
...
+
+  // scoped moderators (site mods) should not be using updateModerationScopes
+  // they should be using promoteUserRole
+  if (
+    viewer.role === GQLUSER_ROLE.MODERATOR &&
+    viewer.moderationScopes &&
+    viewer.moderationScopes.siteIDs?.length !== 0
+  ) {
+    throw new Error("you have insufficient permissions to set roles");
+  }
```
wyattjoh: This check should be completed at the schema level and should not be duplicated here.

src/core/server/services/users/users.ts
```diff
@@ -658,21 +659,110 @@ export async function updateUsernameByID(
...
+    mongo,
+    tenant.id,
+    userID,
+    viewer.moderationScopes
+  );
+  if (
+    !_.isEqual(
+      _.sortBy(viewer.moderationScopes?.siteIDs),
+      _.sortBy(scopesResult.moderationScopes?.siteIDs)
+    )
```
wyattjoh: You should not use the default import here for lodash and instead import the named functions.

src/core/server/graph/schema/schema.graphql
```diff
@@ -7825,7 +7853,13 @@ type Mutation {
   updateUserRole will update a given User's role.
   """
   updateUserRole(input: UpdateUserRoleInput!): UpdateUserRolePayload!
-    @auth(roles: [ADMIN])
+    @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh: Add `unscoped` to require moderators to be unscoped when using this mutation:

```suggestion
    @auth(roles: [ADMIN, MODERATOR], unscoped: true)
```

src/core/server/graph/schema/schema.graphql
```diff
@@ -7834,7 +7868,7 @@ type Mutation {
   """
   updateUserModerationScopes(
     input: UpdateUserModerationScopesInput!
-  ): UpdateUserModerationScopesPayload! @auth(roles: [ADMIN])
+  ): UpdateUserModerationScopesPayload! @auth(roles: [ADMIN, MODERATOR])
```
wyattjoh: Add `unscoped` to require moderators to be unscoped when using this mutation:

```suggestion
  ): UpdateUserModerationScopesPayload! @auth(roles: [ADMIN, MODERATOR], unscoped: true)
```

src/core/server/services/users/users.ts
```diff
@@ -658,21 +659,110 @@ export async function updateUsernameByID(
...
 
+  // only admins can promote to admin
+  if (viewer.role !== GQLUSER_ROLE.ADMIN && role === GQLUSER_ROLE.ADMIN) {
+    throw new Error("only admins can update users to administrators");
+  }
+
+  // you can only update users if you're currently an admin or moderator
+  if (![GQLUSER_ROLE.ADMIN, GQLUSER_ROLE.MODERATOR].includes(viewer.role)) {
+    throw new Error("you have insufficient permissions to set roles");
+  }
```
wyattjoh: Not required and should be removed as it's currently duplicating logic from the schema.

src/core/server/services/users/users.ts
```diff
@@ -658,21 +659,110 @@ export async function updateUsernameByID(
...
+
+  // scoped moderators (site mods) should not be using updateRole
+  // they should be using promoteUserRole
+  if (
+    viewer.role === GQLUSER_ROLE.MODERATOR &&
+    viewer.moderationScopes &&
+    viewer.moderationScopes.siteIDs?.length !== 0
+  ) {
+    throw new Error("you have insufficient permissions to set roles");
+  }
```
wyattjoh: This should be removed when `unscoped: true` is added to the `@auth` directive instead.

src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
```diff
@@ -80,22 +137,52 @@ const UserRoleChangeContainer: FunctionComponent<Props> = ({
...
+    case GQLUSER_ROLE.MODERATOR:
+      return GQLUSER_ROLE.MODERATOR;
+    case GQLUSER_ROLE.STAFF:
+      return GQLUSER_ROLE.STAFF;
+    case GQLUSER_ROLE.COMMENTER:
+      return GQLUSER_ROLE.COMMENTER;
+    default:
+      return GQLUSER_ROLE.COMMENTER;
+  }
+}
```
wyattjoh: Using the `GQLUSER_ROLE_RL` is important here, and should not be converted to a `GQLUSER_ROLE`. When referring to the _value_, you should always use `GQLUSER_ROLE` (Like when you're saying `viewer.role !== GQLUSER_ROLE.ADMIN`). When _typeing_ it, you should always use `GQLUSER_ROLE_RL`, as it ensures the type system will be forwards compatible with new role types.

src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
```diff
@@ -80,22 +137,52 @@ const UserRoleChangeContainer: FunctionComponent<Props> = ({
...
       moderationScopesEnabled={moderationScopesEnabled}
       query={query}
+      viewerScoped={
+        viewer.moderationScopes ? viewer.moderationScopes?.scoped : false
+      }
+      viewerSites={
+        viewer.moderationScopes
+          ? viewer.moderationScopes.sites?.map((s) => s.id)
+          : null
+      }
```
wyattjoh: It feels like there's a lot of leakage and duplication of logic in this container component and it's associated control component. Ideally, we should be passing more explicitly what modes to enable for the `UserRoleChange` component rather than giving it all the context to make it's own decisions.

wyattjoh (changes requested):

src/core/server/graph/schema/schema.graphql
nick-funk: This doesn't work. Admin's and org mods can also use this. They will not be scoped.

src/core/client/admin/components/UserRole/PromoteUserRoleMutation.ts
```diff
@@ -0,0 +1,45 @@
...
+  createMutation,
+  MutationInput,
+} from "coral-framework/lib/relay";
+
+import { PromoteUserRoleMutation as MutationTypes } from "coral-admin/__generated__/PromoteUserRoleMutation.graphql";
+
+let clientMutationId = 0;
+
+const PromoteUserRoleMutation = createMutation(
+  "updateUserModerationScopes",
```
wyattjoh: Update to `promoteUserRole`

src/core/client/admin/components/UserRole/SiteModeratorModal.tsx
```diff
@@ -90,7 +94,11 @@ const SiteModeratorModal: FunctionComponent<Props> = ({
                       and issue suspensions on the sites they are assigned.
                     </ModalBodyText>
                   </Localized>
-                  <SiteModeratorModalSiteFieldContainer query={query} />
+                  <SiteModeratorModalSiteFieldContainer
+                    query={query}
+                    viewerScoped={viewerScoped}
+                    viewerSites={viewerSites}
```
wyattjoh: Query from `query` instead

src/core/client/admin/components/UserRole/UserRoleChange.tsx
```diff
@@ -22,53 +25,196 @@ import styles from "./UserRoleChange.css";
...
+export interface RoleDescription {
+  id?: string;
+  role: GQLUSER_ROLE_RL;
+  scoped: boolean | undefined;
+}
+
+export function canChangeRole(
+  viewerRole: RoleDescription,
+  existingRole: RoleDescription,
+  newRole: RoleDescription
```
wyattjoh:
```suggestion
  viewer: RoleDescription,
  user: RoleDescription,
  target: RoleDescription
```

src/core/client/admin/components/UserRole/UserRoleChange.tsx
```diff
@@ -22,53 +25,196 @@ import styles from "./UserRoleChange.css";
...
     togglePopoverVisibility,
   ] = useToggleState();
 
+  const viewerSites = useMemo(() => {
+    if (!viewer.moderationScopes) {
+      return [];
+    }
+
+    return viewer.moderationScopes.sites ? viewer.moderationScopes.sites : [];
+  }, [viewer]);
```
wyattjoh:
```suggestion
  const viewerSites = viewer.moderationScopes?.sites || [];
```

wyattjoh (approved):

Issue #2624, opened 2019-10-04
When SSO token expire stream stop to work


cristiandean:
Hello, we found a bug when using SSO Token and it expires. When this happens all interactions stop working.
 
Expected behavior:
Open login window

From what I realized we need to set `authMiddleware.tokenRefreshPromise` ([createNetwork](https://github.com/coralproject/talk/blob/721b2e7d67272dea898be55d935f4c9dbf821f6d/src/core/client/framework/lib/network/createNetwork.ts#L61))  with an event to call login. What is the best way to do this?

yannick-fernand:
Hello, do you plan to do something  about access token renewal ?

alinhle:
We are looking into it!

kgardnr:
Hey @cristiandean and @yannick-fernand if you want to take a look here: https://github.com/coralproject/talk/pull/3200

cristiandean:
@kgardnr Looks good to me, thank you :)

kgardnr:
Fixed in #3200

PR #3200, opened 2020-09-29
[CORL-774] Refresh expired access token

alinhle:
## What does this PR do?

- Upgrade `relay-network-modern`
- Introduce refresh access token mechanism when encountering an expired access token.
  - ~Refresh access token callback configurable through the Coral Embed using `refreshAccessToken` config option, which can return `string | Promise<string>`.~
  - Refresh access token callback configurable through the `Coral.createStreamEmbed` using `refreshAccessToken` config option:
```js
  Coral.createStreamEmbed({
     ...options,
     refreshAccessToken: function(nextAccessToken) {
       const token = retrieveNewAccessToken();
       nextAccessToken(token);
     }
  });
```
  - If Coral controls the auth, it'll open the auth popup to let the user reauthenticate

## How do I test this PR?
- Set session expiry (`sessionDuration`) on server to a short amount
- Let session expire and perform an action.
- Reauthenticate

https://github.com/coralproject/talk/issues/2624


yannick-fernand (approved):

src/core/client/embed/Coral.ts
```diff
@@ -15,6 +15,8 @@ export interface Config {
   autoRender?: boolean;
   events?: (eventEmitter: EventEmitter2) => void;
   accessToken?: string;
+  /** refreshAccessToken is called to obtain a new access token when the current one has expired. */
+  refreshAccessToken?: () => Promise<string> | string;
```
alinhle: Use callback syntax

src/core/client/framework/lib/auth/auth.ts
```diff
@@ -19,25 +26,29 @@ export interface AuthState {
...
     }
   }
 
   return { accessToken, claims };
 }
 
+export function isLoggedIn(environment: Environment) {
+  return Boolean(lookup<GQLQuery>(environment, "client:root")?.viewer);
+}
+
```
alinhle: Put in helpers?

src/core/client/framework/helpers/onPymMessageOnce.ts
```diff
@@ -0,0 +1,18 @@
...
+
+function onPymMessageOnce(
+  pym: Child,
+  messageType: string,
+  callback: MessageCallback
+) {
+  const dispose = onPymMessage(pym, messageType, (raw) => {
+    callback(raw);
+    // Dispose next frame to not disrupt pym handling messages.
+    setTimeout(() => dispose());
```
alinhle, line 13: Discussion about moving `setTimeout` directly into `dispose()`

alinhle: I'm against it, because sometimes I want `dispose` to happen right away so to be sure that another event callback is not called. We should investigate that error in sentry why and when it's happening and target that specifically.

.eslintrc.js
```diff
@@ -80,7 +80,7 @@ const typescriptOverrides = {
       "@typescript-eslint/no-use-before-define": "off", // TODO: (cvle) Should be on?
       "@typescript-eslint/no-use-before-declare": "off",
       "@typescript-eslint/no-var-requires": "error",
-      "@typescript-eslint/prefer-for-of": "error",
+      "@typescript-eslint/prefer-for-of": "off", // Different looping mechanism has their own valid performance characteristics.
```
wyattjoh: Only place I see this used like this is here:

https://github.com/coralproject/talk/pull/3200/files#diff-405c5dece9d29d580eb0a4365a06d905R23

Which honestly still could benefit from the `for-of` syntax instead.

wyattjoh: I'd suggest reverting this change to the linting rules and instead preferring the `for-of` syntax.

alinhle: Looked at it again and it seems you are right. I remember that `for-of` was translated to a very complicated es2015 code in the past, but it seems to be quite good now. I'll revert this change.

wyattjoh (approved):

PR #3216, opened 2020-10-09
Create app.json

e-talent2018:


wyattjoh:
With mLab currently on the slate to be deprecated (see https://devcenter.heroku.com/changelog-items/1823)  We're unfortunately deprecating the Heroku button with the associated `app.json` file. We know there is still a reference to it in the documentation, but that issue will be patched soon!

PR #3228, opened 2020-10-19
[CORL-1431] Tenant Cache Fix 

wyattjoh:
## What does this PR do?

In javascript, the default behaviour of the JSON library does not perform automatic translation to Date formats. This differs from how the MongoDB driver translates to/from a BSON format.

This PR serves to enhance the JSON support with Date support for use with the Tenant Cache which relies on this translation when publishing and subscribing to updates via JSON. We mirror the format used by BSON by storing date objects as a nested object:

```json
{ "date": { "$date": "2020-10-19T18:35:31.151Z" } }
```

Which when parsed now yields:

```js
{ date: new Date("2020-10-19T18:35:31.151Z") }
```

None.

## How do I test this PR?

This is difficult to simulate the failure state, but a side affect of incorrect parsing here can be seen though a related bug.

1. Create a new announcement from the admin panel to expire an hour from now.
2. Check to see the announcement on the stream (without dismissing it).
3. Stop the server.
4. Edit the document for the Tenant in MongoDB where the document may resemble:

```json
  "announcement": {
    "id": "71060aa5-6951-4369-8843-2cccfa0c5547",
    "content": "Happy Monday!",
    "duration": 3600,
    "createdAt": {
      "$date": "2020-10-20T17:36:14.462Z"
    }
  }
```

To instead resemble:

```json
  "announcement": {
    "id": "71060aa5-6951-4369-8843-2cccfa0c5547",
    "content": "Happy Monday!",
    "duration": 3600,
    "createdAt": "2020-10-20T17:36:14.462Z"
  }
```

_Note how we changed the `createdAt` field to remove the nesting and the related `$date` key.

5. Start the server.
6. Try to view the announcement from the stream again, you shouldn't see one!

This is exactly what was happening when running Coral in production, dates were not being converted back to Date objects when published through the tenant cache service.

While this doesn't directly show how this particular fix solves the problem, refer to the included unit tests to verify functionality.


nick-funk:
Please move the `🚨 Important Release Note 🚨` section to the top of the description, we don't want anyone to miss it.

Other than that, nice tests, looks good!

nick-funk (approved):

PR #3232, opened 2020-10-20
[CORL-1407] Improve Subscriptions Performance

alinhle:
## What does this PR do?
- Create new Subscription `commentEntered` and deprecated `commentCreated`, `commentReleased`, `commentReplyCreated`
- Reduce amount of subscriptions
- Moderating a Reply is now updated live
- Fix Q&A Subscriptions in Unanswered tab.
- Remove forced sort to Most Voted in Q&A mode
- Close websocket client only after an inactivity period

- Created a new Subscription `commentEntered` and deprecated `commentCreated`, `commentReleased`, 

## How do I test this PR?

- Test subscription features on stream (create comment, create coment reply, permalink view, moderate comments, q&a)


src/core/client/stream/tabs/Comments/PermalinkView/PermalinkViewContainer.tsx
```diff
@@ -1,110 +1,197 @@
...
+        ancestorID: comment.id,
+        liveDirectRepliesInsertion: true,
+        storyConnectionKey: "Stream_comments",
+      });
+      return () => {
+        disposable.dispose();
+      };
     }
-  }
+    return;
```
wyattjoh: Typically would recommend something where you early return instead:

```ts
if (!comment?.id) {
  return;
}

// do the stuff that returns a function..
```

src/core/client/stream/tabs/Comments/PermalinkView/PermalinkViewContainer.tsx
```diff
@@ -1,110 +1,197 @@
...
-        showAllCommentsHref={this.getShowAllCommentsHref()}
-        onShowAllComments={this.showAllComments}
-      />
-    );
-  }
-}
+  useEffect(() => {
+    if (pym) {
+      setTimeout(() => pym.scrollParentToChildPos(0), 100);
+    }
```
wyattjoh: Same here as above:

```suggestion
    if (!pym) {
      return
    }
    
    setTimeout(() => pym.scrollParentToChildPos(0), 100);
```

src/core/client/stream/tabs/Comments/PermalinkView/PermalinkViewContainer.tsx
```diff
@@ -1,110 +1,197 @@
...
+        commentID: comment && comment.id,
+      });
+      void setCommentID({ id: null });
+      e.preventDefault();
+    },
+    [comment, eventEmitter, setCommentID]
+  );
+
+  const showAllCommentsHref = useMemo(() => {
+    const url = (pym && pym.parentUrl) || window.location.href;
```
wyattjoh: More easily written:

```suggestion
    const url = pym?.parentUrl || window.location.href;
```

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/CommentEnteredSubscription.tsx
```diff
@@ -0,0 +1,266 @@
...
+    // could not trace back to ancestor, discard.
+    return;
+  }
+
+  // Comment is just outside our visible depth.
+  if (depth >= 4) {
+    // Inform last comment in visible tree about the available replies.
+    // This will trigger to show the `Read More of this Conversation` link.
+    const replyCount = parentProxy.getValue("replyCount") || 0;
+    parentProxy.setValue((replyCount as number) + 1, "replyCount");
```
wyattjoh: Might be better to set the result of `.getValue` to a number instead?

```suggestion
    const replyCount = (parentProxy.getValue("replyCount") as number) || 0;
    parentProxy.setValue(replyCount + 1, "replyCount");
```

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/CommentEnteredSubscription.tsx
```diff
@@ -0,0 +1,266 @@
...
+ */
+function determineDepthTillAncestor(
+  comment: RecordProxy,
+  ancestorID?: string | null
+) {
+  let depth = 0;
+  let cur: RecordProxy | null | undefined = comment;
+  while (cur) {
+    cur = cur.getLinkedRecord("parent");
+    if (cur && cur.getValue("createdAt")) {
```
wyattjoh: Why use `createdAt` instead of `id`?

src/core/server/events/listeners/subscription.ts
```diff
@@ -30,10 +32,12 @@ export class SubscriptionCoralEventListener
     CoralEventType.COMMENT_LEFT_MODERATION_QUEUE,
     CoralEventType.COMMENT_STATUS_UPDATED,
     CoralEventType.COMMENT_REPLY_CREATED,
+    CoralEventType.COMMENT_ENTERED,
     CoralEventType.COMMENT_CREATED,
     CoralEventType.COMMENT_FEATURED,
     CoralEventType.COMMENT_RELEASED,
   ];
+  public readonly composite = [];
```
wyattjoh: Remove unused property.

src/core/server/events/events.ts
```diff
@@ -106,7 +120,26 @@ export type CommentReleasedCoralEventPayload = CoralEventPayload<
...
->(CoralEventType.COMMENT_RELEASED);
+>(CoralEventType.COMMENT_RELEASED, {
+  forward: [CommentEnteredCoralEvent],
+});
+
+export interface CommentReplyReleasedInput {
+  ancestorIDs: string[];
+  commentID: string;
+  storyID: string;
+}
```
wyattjoh: This type is not used outside this invocation below for the `CommentReplyReleasedCoralEventPayload`. Inline this within the `CommentReplyReleasedCoralEventPayload` as others in this file have done instead.

src/core/server/events/events.ts
```diff
@@ -106,7 +120,26 @@ export type CommentReleasedCoralEventPayload = CoralEventPayload<
...
+export type CommentReplyReleasedCoralEventPayload = CoralEventPayload<
+  CoralEventType.COMMENT_REPLY_RELEASED,
+  CommentReplyReleasedInput
+>;
+
+export const CommentReplyReleasedCoralEvent = createCoralEvent<
+  CommentReplyReleasedCoralEventPayload
+>(CoralEventType.COMMENT_REPLY_RELEASED, {
+  forward: [CommentEnteredCoralEvent],
+});
```
wyattjoh: Besides forwarding this event to the `CommentEnteredCoralEvent`, it does not appear that it's published events are handled anywhere. Maybe only calling `CommentEnteredCoralEvent` is better in this circumstance?

wyattjoh (changes requested):

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/CommentEnteredSubscription.tsx
alinhle: I'm adding this comment
```ts
  // Because we request the parent including its id in the the subscription below
  // we can't just check for truthiness nor check the `id` field of the record to determine
  // that it already exists in our cache. Therefore we check for another field that
  // is part of the `CommentContainer_comment` fragment.
```

src/core/server/events/events.ts
alinhle: Hmm yeah maybe

wyattjoh (approved):

PR #3242, opened 2020-10-23
[CORL-1455] Only show login when enabled

wyattjoh:
## What does this PR do?

This updates the few other instances we missed checking if we need to show the login popup from #3231.

None.

## How do I test this PR?

With nothing but the SSO login enabled on the stream, try to interact with the stream to trigger the login popup, you shouldn't be able to!

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@coralproject/talk",
-  "version": "6.3.11",
+  "version": "6.3.12",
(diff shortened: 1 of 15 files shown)


nick-funk (approved):

nick-funk (changes requested):
Noticing that the "loginPrompt" callback isn't being called with local auth disabled on stream.

nick-funk (approved):

PR #3253, opened 2020-10-27
[CORL-1467] Auth Flow Fixes

wyattjoh:
## What does this PR do?

This improves the reliability of the `AccountCompletionContainer` by removing callbacks from within the render cycle and into hooks. This also changes the redirect path to use local storage instead of session storage to additionally improve reliability under some redirect conditions.

*Additional changes (@cvle):
Access tokens are not stored during the auth completion process until it's completed. This prevents users using an incomplete account to query for the stream which would result in a `USER_NOT_ENTITLED` error. Also a user can just cancel the auth completion process now by just closing the window and signing in with a different user, e.g. in case he used the wrong account or did not intend to create a user.

Another change is that access tokens are now saved over the pym bridged local storage instead of the local storage inside the iframe. This allow safari browsers to retain the access token.

## How do I test this PR?

Perform sign in and sign up with different authentication configurations:

- Only _Local_ enabled
- Only _OpenID Connect_ enabled
- Both _Local_ and _OpenID Connect_ enabled

With both the stream and admin login areas.


alinhle (approved):
All yours @wyattjoh!

PR #3255, opened 2020-10-28
[CORL-1412] Integrate giphy web SDK

tessalt:
## What does this PR do?

Turns out giphy released a web sdk complete with react components shortly after we built our own custom UI. Theirs is better, but there's a couple tradeoffs.

Pros:
- slicker UI that will be familiar to users ("masonry" grid)
- infinite scroll for free
- easier to adapt UI to mobile

Cons:
- giphy API key must be public (giphy says this is intentional and that they monitor for abuse on their end... https://github.com/Giphy/giphy-js/issues/120)
- bundling in a bunch of deps that some clients won't use

giphy rating and api key are now public

## How do I test this PR?

search and select gifs on desktop and mobile


package.json
```diff
@@ -136,11 +138,13 @@
     "prom-client": "^12.0.0",
     "proxy-agent": "^3.1.1",
     "querystringify": "^2.1.1",
+    "react-use": "^15.3.4",
```
tessalt: this library is just being used for a nice `useDebounce` (imported independently) but if we don't want to add this dep i'm happy to write my own.

src/core/client/stream/tabs/Comments/GiphyInput/GiphyInput.tsx
```diff
@@ -110,19 +74,23 @@ const GiphyInput: FunctionComponent<Props> = ({ onSelect }) => {
     e.preventDefault();
   }, []);
 
+  const gridColumns = useMemo(() => {
+    if (width < APPROX_COL_WIDTH * 2) {
```
tessalt, line 90: number of columns determined by width of screen

package.json
wyattjoh: These dependencies are great, but can we move them to `devDependencies`? They aren't used by the server so we don't need to include it in `dependencies`.

package.json
```diff
@@ -136,11 +138,13 @@
...
     "source-map-support": "^0.5.16",
     "stack-utils": "^2.0.1",
     "striptags": "^3.1.1",
     "tsscmp": "^1.0.6",
     "url-regex": "^5.0.0",
+    "use-resize-observer": "^6.1.0",
```
wyattjoh: We actually already have our own resize observer provided by:

```ts
import { useResizeObserver } from "coral-framework/hooks";
```

wyattjoh: If this provides a better API though, I'd replace our implementation with the package one.

src/core/client/stream/tabs/Comments/GiphyInput/GiphyInput.tsx
```diff
@@ -1,107 +1,71 @@
...
+  const inputRef = useRef<HTMLInputElement>(null);
+  const { ref, width = 1 } = useResizeObserver<HTMLDivElement>();
+  const gf = new GiphyFetch(apiKey);
+  const fetchGifs = async (offset: number) =>
+    gf.search(query, {
+      offset,
+      limit: 10,
+      rating: maxRating as SearchOptions["rating"],
+      sort: "relevant",
+    });
```
wyattjoh: This should be wrapped in a `useMemo`:

```suggestion
  const fetchGifs = useMemo(() => {
    const gf = new GiphyFetch(apiKey);
    return async (offset: number) =>
      gf.search(query, {
        offset,
        limit: 10,
        rating: maxRating as SearchOptions["rating"],
        sort: "relevant",
      });
    }, [apiKey, maxRating]);
```

wyattjoh (changes requested):

package.json
tessalt: It does provide a better API in my opinion, I've replaced our implementation in the one place it was used.

wyattjoh (approved):

PR #3270, opened 2020-11-02
[CORL-1461] show media in comment revisions

tessalt:
## What does this PR do?

shows any media embedded in previous comment revisions in an edited comment

none

## How do I test this PR?
view a comment whose previous revisions include media in admin


src/core/client/admin/components/MediaContainer/MediaContainer.tsx
```diff
@@ -20,46 +17,18 @@ const MediaContainer: FunctionComponent<Props> = ({ comment }) => {
...
   }
+
+  return (
+    <Media
+      media={media}
+      type={media.__typename}
+      id={comment.id}
+      siteID={comment.site.id}
+    />
+  );
```
wyattjoh: The disadvantage to doing this instead of what was here before is that you can't rely on the type system enough, the connection is a lot more "loose".

Passing the type name information into the `<Media />` component relies on that component to understand the prop requirements and shape of the items based on the passed `__typename`.

src/core/client/admin/components/MediaContainer/YouTubeMedia.tsx
```diff
@@ -6,7 +6,7 @@ import { BaseButton, Flex, Icon } from "coral-ui/components/v2";
 
 import styles from "./Media.css";
 
-interface Props {
+export interface Props {
```
wyattjoh: Typically if you're exporting `Props`, I'd name it after the component to give it a unique name (i.e. `YouTubeMediaProps` instead).

wyattjoh (changes requested):

tessalt:
@wyattjoh updated to remove the new component and just duplicate the code.

wyattjoh (approved):

PR #3274, opened 2020-11-04
[CORL-1495] Improve Cache Busting

wyattjoh:
## What does this PR do?

When the iframe loads, it adds a time sensitive value to the URL to aid in cache busting. The hash length of assets was also increased.

With these changes I've also modified the embed to remove the requirement of the `Object.assign` polyfill, which saved us some precious KB on the `embed.js` file.

None.

## How do I test this PR?

1. Run `npm run build` and see the increased length of built assets.
2. Load a page with the iframe embed on it and see a time sensitive value on the URL for it.
3. Load the page on IE 11 and ensure it still works.


src/core/client/embed/StreamEmbed.ts
```diff
@@ -43,139 +40,176 @@ export interface StreamEmbedConfig {
...
+
+    const query = stringifyQuery({
+      storyID: config.storyID,
+      storyURL: config.storyURL,
+      commentID: config.commentID,
+      customCSSURL: config.customCSSURL,
+
+      // Add the version to the query string to ensure that every new version of
+      // the stream will cause stream pages to cache bust.
+      v: process.env.TALK_VERSION ? process.env.TALK_VERSION : "dev",
```
alinhle, line 105: Is it okay to expose our version from a security pov?

wyattjoh: The version is already exposed in the application releases as it is, this isn't revealing anything new.

alinhle (approved):

PR #3280, opened 2020-11-06
[CORL-378] AMP Support

alinhle:
## What does this PR do?

- Adds AMP support

## How do I test this PR?
- In Development you'll see another story page called `AMP`
- You can embed inside amp using an url pattern like `http://mycoralserver.com/embed/stream/amp?storyURL=http://story.url` or instead `?storyURL=` you can also use `?storyID=`


src/core/server/app/middleware/csp/tenant.ts
```diff
@@ -69,30 +72,56 @@ async function retrieveSiteFromEmbed(
...
-  const origins = site ? site.allowedOrigins : [];
+  // If the frame ancestors isn't being forced, and the tenant has AMP enabled,
+  // then don't add any headers!
+  if (
+    !frameAncestorsDeny &&
+    req.coral.tenant &&
+    hasFeatureFlag(req.coral.tenant, GQLFEATURE_FLAG.ENABLE_AMP)
+  ) {
+    return next();
+  }
```
alinhle: Isn't this going to let **_anyone_** embed Coral, when amp is enabled?

wyattjoh: Yes, though it will still require to use the permitted domains list for the story creation.

src/core/server/config.ts
```diff
@@ -377,6 +389,12 @@ const config = convict({
     default: ms("1m"),
     env: "NODEJS_HEADERS_TIMEOUT",
   },
+  amp_cache_domains: {
+    doc: "Specifies the amp cache domains as a comma seperated list",
+    format: "list",
+    default: ["cdn.ampproject.org", "www.bing-amp.com"],
```
alinhle: I really think we don't need to add `www.bing-amp.com` it looks like their format used to be different e.g. `https://www.bing-amp.com/c/s/yourdomain.com/yourpath`, and it seems that it's not being used anymore. Have you found any evidence that this still exists?

src/core/server/app/middleware/csp.ts
```diff
@@ -193,9 +193,18 @@ export function generateFrameOptions(
     return "deny";
   }
 
-  // If the requester origin does not exist in the allowed origins, then don't
-  // do anything!
+  // If the requester origin is the tenant origin, then allow it!
+  if (
+    req.coral.tenant &&
+    requesterOrigin ===
+      prefixSchemeIfRequired(req.secure, req.coral.tenant.domain).toLowerCase()
```
alinhle, line 202: Add a comment that this will only work for standard ports here and we're good!

wyattjoh (approved):

PR #3286, opened 2020-11-11
[CORL-1396] order story search results by relevance

tessalt:
## What does this PR do?

when stories are searched by query, they should be ordered by relevance instead of by createdAt

none

## How do I test this PR?
search for a story by url (may require changing the URL on a story to something unique)


wyattjoh (approved):
Thanks @tessalt! We needed to also add support for adjusting the pagination parameters. Let's do a quick review together before we merge this!

PR #3302, opened 2020-11-25
[CORL-1449] Handle when actionCounts are unavailable

nick-funk:
## What does this PR do?
Throws an error with comment id if we don't see action counts so we can find the root cause of this.

This is a bit puzzling as to how it could occur, because the schema validation won't let the action counts ever be undefined/null.

None

## How do I test this PR?

N/A


tessalt (approved):

src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
```diff
@@ -262,17 +262,17 @@ const ModerateCardContainer: FunctionComponent<Props> = ({
...
-        ? comment.revision.actionCounts.flag.reasons
-            .COMMENT_DETECTED_BANNED_WORD +
-            comment.revision.actionCounts.flag.reasons
-              .COMMENT_DETECTED_SUSPECT_WORD >
-          0
-        : false,
-    [comment]
-  );
+  const highlight = useMemo(() => {
+    if (!comment.revision || !comment.revision.actionCounts) {
```
wyattjoh: `actionCounts` as per the schema cannot be null. If null is being passed, it could mean there's an issue with the resolver because the value for `actionCounts` is computed in the `CommentRevision` resolver. Please take a look there to see if there's an issue!

wyattjoh: Check the subscriptions are including the fragment!

wyattjoh (approved):

PR #3310, opened 2020-11-27
[CORL-1527] Default to pending queue if global premod is enabled

tessalt:
## What does this PR do?

If `settings.moderation` is set to `PRE`: 
- redirect /admin/moderate to /admin/moderate/pending instead of /reported
- hide reported queue

none

## How do I test this PR?

- toggle `settings.moderation` 
- if `PRE`: click "moderate" in navigation, should go to pending queue and hide reported queue
- if `POST`: click "moderate" in navigation, should go to reported queue


src/core/client/admin/routes/Moderate/ModerateNavigation/Navigation.tsx
```diff
@@ -20,6 +20,7 @@ interface Props {
   section?: SectionFilter | null;
   router: Router;
   match: Match;
+  mode?: "PRE" | "POST" | "%future added value" | null;
```
wyattjoh, line 23: Not a comment to change anything here, but we probably need a rule/pattern to handle these in the future.

Any suggestions @cvle?

src/core/client/admin/routes/Moderate/ModerateNavigation/Navigation.tsx
```diff
@@ -69,22 +71,24 @@ const Navigation: FunctionComponent<Props> = ({
...
-            <Localized
-              id="moderate-navigation-comment-count"
-              $count={reportedCount}
-            >
-              {reportedCount}
-            </Localized>
-          </Counter>
-        )}
-      </NavigationLink>
+      {mode && mode === "POST" && (
```
wyattjoh: Technically you can just do `mode === "POST"` instead of checking `mode && mode === "POST"`

src/core/client/admin/routes/Moderate/Queue/ModerationQueue.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+const ModerationQueue: FunctionComponent<Props> = ({ mode }) => {
+  const { router } = useRouter();
+  useEffect(() => {
+    if (mode === "PRE") {
+      router.replace("/admin/moderate/pending");
+    } else if (mode === "POST") {
+      router.replace("/admin/moderate/reported");
+    }
```
wyattjoh: You need to handle when the `storyID`, `siteID`, or `commentID` is passed. Check the possible invocations in `src/core/client/framework/helpers/getModerationLink.ts#getModerationLink` and adjust routes in the router to use this component.

wyattjoh (changes requested):

src/core/client/admin/routes/Moderate/Queue/ModerationQueue.tsx
```diff
@@ -3,17 +3,36 @@ import { FunctionComponent, useEffect } from "react";
...
+
+    if (siteID) {
+      suffix = `sites/${siteID}`;
+    } else if (storyID) {
+      suffix = `stories/${storyID}`;
+    }
+
+    if (path) {
+      router.replace(`/admin/moderate/${path}/${suffix}`);
     }
```
wyattjoh: You can use the helper `getModerationLink` to generate this! Would simplify this instead to:

```ts
router.replace(getModerationLink({ queue: mode === "PRE" ? "pending" : "reported", storyID, siteID }));
```

alinhle:
@wyattjoh @tessalt I'm still unsure if we should implement it this way. We are making the load time twice as slow to access the admin page (when assets are already cached). You won't notice it if you develop locally, but if you have a slow internet connection or a high latency, it can hit the user experience hard. Consider developing countries with slow internet and high latency. I'm just wondering if the cost is really worth for a simple feature that only redirects to a correct page, when there is the alternative approach using the static config. I believe the loading speed is actually a critical criteria from a product perspective and I feel like if we do it this way it needs to be communicated to @kgardnr. I'm very sorry @tessalt for flip flopping on this issue 😢

src/core/client/admin/routes/Moderate/ModerateNavigation/Navigation.tsx
alinhle: What do you mean exactly?

src/core/client/admin/routes/Moderate/Queue/ModerationQueueQuery.tsx
```diff
@@ -0,0 +1,47 @@
...
+    <QueryRenderer<QueryTypes>
+      query={graphql`
+        query ModerationQueueQuery {
+          settings {
+            moderation
+          }
+        }
+      `}
+      variables={{}}
+      fetchPolicy="store-and-network"
```
alinhle, line 31: This should me `network-only` otherwise it would always try to render twice and no worries our network layer has a caching mechanism.

alinhle (changes requested):

src/core/client/admin/routes/Moderate/Queue/ModerationQueueQuery.tsx
alinhle: Never mind! I think `store-and-network` is actually a good compromise.

alinhle (commented):
Changes looks good to me, only haven't tested it yet.

src/core/client/admin/routes/Moderate/ModerateNavigation/Navigation.tsx
wyattjoh: It's common for us to re-use enum's for components, and having to copy the `"%future added value" | null` everywhere feels complex. Do we have a more universal way to generate enum types that include this `"%future added value"` already?

src/core/client/admin/routes/Moderate/ModerateNavigation/Navigation.tsx
wyattjoh: We should actually modify this logic to instead be:

```suggestion
      {(mode === "POST" || (isNumber(reportedCount) && reportedCount > 0) && (
```

Because even if the moderation mode is set to "PRE", we still could have comments that are reported. With this logic, it'll hide itself only if there is no reported comments and is not in "POST" moderation.

src/core/client/admin/routes/Moderate/Queue/ModerationQueue.tsx
```diff
@@ -0,0 +1,32 @@
...
+}) => {
+  const { router } = useRouter();
+  useEffect(() => {
+    if (mode) {
+      router.replace(
+        getModerationLink({
+          queue: mode === "PRE" ? "pending" : "reported",
+          storyID,
+          siteID,
+        })
```
wyattjoh: You should also handle the `section` as well if you can. See `parseModerationOptions` to see how to extract that from the `match` parameter.

wyattjoh (changes requested):

wyattjoh (approved):

alinhle (approved):

PR #3311, opened 2020-11-27
[CORL-1442] add better error message when edit window expires on server

tessalt:
## What does this PR do?

If for some reason the frontend allows a comment edit to be submitted after the edit window has expired, send a better error message and do not report it. 

none

## How do I test this PR?

Easiest way is probably to comment out the code that disables the submit button in CommentForm and then attempt to submit an edit after the window has expired.


src/core/server/models/comment/comment.ts
```diff
@@ -263,8 +266,7 @@ export function validateEditable(
 
   // Check to see if the edit window expired.
```
tessalt, line 267: the other errors have `todo: better error message` as well, but not sure what this means as the other ones are real errors that should be investigated. Do we just want to send more info along on the other ones?

wyattjoh (approved):

PR #3312, opened 2020-11-27
[CORL-1533] Ratings and Reviews Enhancements

wyattjoh:
## What does this PR do?

Adds support for filtering based on rating.

None.

## How do I test this PR?

Try to filter reviews by rating!


src/locales/en-US/stream.ftl
```diff
@@ -803,3 +795,37 @@ profile-changeEmail-saveChanges = Save changes
 profile-changeEmail-email = Email
 profile-changeEmail-title = Email address
 profile-changeEmail-success = Your email has been successfully updated
+
+## Ratings and Reviews
+
+rr-reviewsTab = Reviews
```
tessalt: I don't love `rr` as a prefix here. We've generally been fairly explicit in our localization keys.

wyattjoh: Fair! I'll see what I can do!

tessalt (approved):

PR #3320, opened 2020-12-04
add prometheus buckets for 750ms and 1000ms

cristiandean:
## What does this PR do?
It's very common that we have requests that take more than 500ms, sometimes 1s, especially when we have the perspective API on or in cases that we're overwhelmed. What do you think about adding two more buckets, 750 and 1000ms in prometheus metrics? 

None

## How do I test this PR?
Check new buckets in /metrics route.

Thanks


wyattjoh (approved):

wyattjoh:
This change looks great @cristiandean! Can you rebase this PR against `next` please?

cristiandean:
Done! :)

PR #3321, opened 2020-12-07
[CORL-1539] Generate introspection json for v6 API docs

tessalt:
## What does this PR do?

adds graphql-codegen to generate introspection json from our graphql schema, and adds a task to push up to gcloud. There's a lot of different ways I could have done this, this seemed the most straightforward but take a look!

none


.circleci/config.yml
```diff
@@ -158,6 +158,10 @@ jobs:
           name: Deploy static files to Google Cloud Storage Bucket
           command: |
             gsutil -m cp -r -z "js,css,map,txt" ./static/** "gs://${GOOGLE_CLOUD_BUCKET}/${CIRCLE_TAG/#v}/"
+      - run:
+          name: Deploy introspection json to Google Cloud Storage Bucket
+          command: |
+            gsutil cp "introspection.json" "gs://${GOOGLE_CLOUD_BUCKET}"
```
wyattjoh: I think it makes sense to output this file in both the root where you've done it here, and the versioned folder.

```suggestion
            gsutil cp "introspection.json" "gs://${GOOGLE_CLOUD_BUCKET}"
            gsutil cp "introspection.json" "gs://${GOOGLE_CLOUD_BUCKET}/${CIRCLE_TAG/#v}/"
```

This ensures we persist the files for each release by themselves.

wyattjoh: Also in the `Copy static assets from docker container` step ensure you copy the introspection out of the container.

codegen.yml
```diff
@@ -0,0 +1,7 @@
+schema: ./src/core/server/graph/schema/schema.graphql
+generates:
+  introspection.json:
```
wyattjoh: I would suggest we store the `introspection.json` file instead:

```
src/core/server/graph/schema/__generated__/introspection.json
```

The introspection is then tied closer to the schema available in that folder.

wyattjoh (changes requested):

.circleci/config.yml
```diff
@@ -158,6 +158,11 @@ jobs:
           name: Deploy static files to Google Cloud Storage Bucket
           command: |
             gsutil -m cp -r -z "js,css,map,txt" ./static/** "gs://${GOOGLE_CLOUD_BUCKET}/${CIRCLE_TAG/#v}/"
+      - run:
+          name: Deploy introspection json to Google Cloud Storage Bucket
+          command: |
+            gsutil cp "src/core/server/graph/schema/__generated__/introspection.json" "gs://${GOOGLE_CLOUD_BUCKET}"
+            gsutil cp "src/core/server/graph/schema/__generated__/introspection.json" "gs://${GOOGLE_CLOUD_BUCKET}/${CIRCLE_TAG/#v}/"
```
wyattjoh: These files are not available here without copying them out of the container in the `Copy static assets from docker container` step. Ensure you copy the files in that step.

wyattjoh (approved):

PR #3323, opened 2020-12-07
Directives Fix

wyattjoh:
## What does this PR do?

When using `@graphql-tools`, they do not modify the schema in place, and instead return a mutated copy. This restores those directives.

None.

## How do I test this PR?

Try to do something you're not supposed to via the API, notice it stops you!


tessalt (approved):
huh!

PR #3327, opened 2020-12-09
[CORL-1434] Flatten Replies

alinhle:
## What does this PR do?
Supersedes #3327

- Implements flatten replies (behind feature flag)
- Refactor ReplyListContainers
- Refactor StaticConfig
- Refactor local state
  - StaticConfig is now passed through `initLocalState`
  - Common local state that is managed by the framework is now defined *inside* the framework at `./src/core/client/framework/lib/relay/local.graphql`
  - framework can now contain fragments, queries, and mutations that is shared

## How do I test this PR?
- Enable `FLATTEN_REPLIES` Feature Flag and post comments including testing subscriptions and single conversation view. (There is no need to stub a StaticConfig in dev)


src/core/client/stream/local/local.graphql
```diff
@@ -1,3 +1,6 @@
+# This extends ./src/core/client/framework/lib/relay/local.graphql
```
wyattjoh, line 1: Maybe we can look into imports in the future!

src/core/client/stream/tabs/Comments/PermalinkView/ConversationThreadContainer.tsx
```diff
@@ -129,7 +129,7 @@ const ConversationThreadContainer: FunctionComponent<Props> = ({
                 <IgnoredTombstoneOrHideContainer
                   viewer={viewer}
                   comment={parent}
-                  singleConversationView={true}
+                  allowTombstoneReveal={true}
```
wyattjoh: Should use the more compact boolean version for consistency:

```suggestion
                  allowTombstoneReveal
```

src/core/server/models/user/validFeatureFlagsFilter.ts
```diff
@@ -0,0 +1,45 @@
+import { GQLFEATURE_FLAG } from "coral-server/graph/schema/__generated__/types";
```
wyattjoh: The `context` of this is the settings/feature flags, and not the user. This should instead be located in the settings folder.

src/core/server/models/comment/comment.ts
```diff
@@ -438,6 +438,31 @@ function cursorGetterFactory(
...
+    filter.parentID = parentID;
+  }
+
+  return retrievePublishedCommentConnection(mongo, tenantID, {
+    ...input,
+    filter,
+  });
+};
+
+export const retrieveFlattenedRepliesConnection = (
```
wyattjoh: This function isn't used anymore, please remove!

src/core/server/graph/loaders/Comments.ts
```diff
@@ -343,4 +344,20 @@ export default (ctx: GraphContext) => ({
...
+      ctx.mongo,
+      ctx.tenant.id,
+      storyID,
+      parentID,
+      {
+        first: defaultTo(first, 10),
+        orderBy: defaultTo(orderBy, GQLCOMMENT_SORT.CREATED_AT_DESC),
+        after,
+      }
+    ).then(primeCommentsFromConnection(ctx)),
```
wyattjoh: This loader isn't used, please remove!

src/core/server/graph/loaders/Comments.ts
```diff
@@ -307,7 +307,8 @@ export default (ctx: GraphContext) => ({
         first: defaultTo(first, 10),
         orderBy: defaultTo(orderBy, GQLCOMMENT_SORT.CREATED_AT_DESC),
         after,
-      }
+      },
+      flatten
```
wyattjoh: Keeping with existing flows, I'd suggest moving the `flatten` into a filter spread as you can see above in the `forStory` loader with the `...tagFilter`.

src/core/server/app/router/client.ts
```diff
@@ -114,6 +115,10 @@ const clientHandler = ({
     locale = req.coral.tenant.locale;
   }
 
+  config.featureFlags =
+    req.coral.tenant?.featureFlags?.filter(validFeatureFlagsFilter(req.user)) ||
+    [];
+
```
wyattjoh: This assignment should be avoided, instead use a object spread in the following `res.render` call.

src/core/common/config.ts
```diff
@@ -28,4 +28,9 @@ export interface StaticConfig {
    * available.
    */
   reporter?: ReporterConfig;
+
+  /**
+   * featureFlags are all the feature flags currently enabled on the tenant.
+   */
+  featureFlags: any[];
```
wyattjoh: Maybe makes sense to use `string` instead?

nick-funk (approved):

wyattjoh (changes requested):

src/core/client/stream/local/local.graphql
alinhle: Interesting! Wonder if this would cause a problem like loading it twice.

src/core/server/models/user/validFeatureFlagsFilter.ts
alinhle: yeah wasn't sure, can do.

src/core/server/app/router/client.ts
alinhle: Of course!

src/core/client/stream/local/local.graphql
alinhle: Note: The import order is currently defined by how the react compiler args are provided.

alinhle: imports from `graphql-tools` https://www.graphql-tools.com/docs/schema-loading/ is non standard, and it seems to be a different use case. It's used to stich together a full schema to be used by the server, while this is about client schema extensions.

wyattjoh (approved):

PR #3347, opened 2020-12-17
[CORL-1571] WordList Fixes

wyattjoh:
## What does this PR do?

- Uses the correct lowercase function `toLower` from `lodash`
- Turn off sorting of lists on save

None.

## How do I test this PR?

Submit two different saves, one for `*badword`. Then after a reload, add `badword` to the list and save. Reload to see both of those in the list!

Turns out `lowerCase` does not do what was expected, and acts more like the `camelCase` or `snakeCase` functions.


tessalt (approved):
tested and works as described.

PR #3351, opened 2020-12-18
[CORL-1526] Allow sorting moderation queues

nick-funk:
## What does this PR do?

Adds a sort filter to the moderation queues.

- Adds `orderBy` props to comments in various places on the moderation queue schema queries

## How do I test this PR?

- Head to admin
- Go to mod queues
- Change the sort order

## TODO
- [x] Use `matchContext` from `found` to get sort order from relay cache instead of `localStorage`
- [x] Implement `initialOrderBy` for approved and rejected queue
- [x] Live updates during sort by oldest should follow stream implementation


src/core/server/graph/resolvers/ModerationQueue.ts
```diff
@@ -25,12 +25,17 @@ export const ModerationQueue: GQLModerationQueueTypeResolver<ModerationQueueInpu
 
     return selector;
   },
-  comments: ({ connection }, { first, after }, { mongo, tenant }) => {
+  comments: (
+    { connection },
+    { first, after, orderBy },
+    { mongo, tenant, logger }
+  ) => {
+    logger.info({ connection, first, after, orderBy }, "WYATT");
```
alinhle: Is this solved? :-D

src/core/client/admin/routes/Moderate/Queue/Queue.tsx
```diff
@@ -154,6 +165,20 @@ const Queue: FunctionComponent<Props> = ({
...
+        <Flex justifyContent="center" className={styles.viewNewButtonContainer}>
+          <Localized id="moderate-queue-viewNew" $count={viewNewCount}>
+            <Button onClick={onViewNew} className={styles.viewNewButton}>
+              View {viewNewCount} new comments
+            </Button>
+          </Localized>
+        </Flex>
+      )}
+
       {hasMore && (
```
alinhle: Live updates for `oldest` sort should be implemented like on the comment stream. Instead of showing a button here, we should just tell relay that there is more data to fetch und the ui handles the rest.

See: https://github.com/coralproject/talk/blob/f070cbc0e3ac59f4b3cbafc1474c8a3713b59b5e/src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/CommentEnteredSubscription.tsx#L56

Then we can stop the `QueueCommentEnteredSubscription` completely when there is more data. See implementation in stream:
https://github.com/coralproject/talk/blob/f070cbc0e3ac59f4b3cbafc1474c8a3713b59b5e/src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabContainer.tsx#L104

alinhle (changes requested):

src/core/server/graph/loaders/Comments.ts
```diff
@@ -217,11 +217,12 @@ export default (ctx: GraphContext) => ({
     status,
     tag,
     query,
+    orderBy,
   }: QueryToCommentsArgs) =>
     retrieveCommentConnection(ctx.mongo, ctx.tenant.id, {
       first: defaultTo(first, 10),
       after,
-      orderBy: GQLCOMMENT_SORT.CREATED_AT_DESC,
+      orderBy: orderBy ? orderBy : GQLCOMMENT_SORT.CREATED_AT_DESC,
```
alinhle: Maybe use `defaultTo` as well?

alinhle (changes requested):
Wow, works really nice!

Can you also stop the subscription when there is a next page like on the stream? That would be perfect!

https://github.com/coralproject/talk/blob/f070cbc0e3ac59f4b3cbafc1474c8a3713b59b5e/src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabContainer.tsx#L104

Also while you're at it, don't forget the 
`defaultTo` thing: https://github.com/coralproject/talk/pull/3351/files/999f1d4377066adcd0d605c682de6eb166a72795#diff-fa2f72f0134a75b6ab87c8bde733d43241a2718af0de2b8b101f29c39b7aff51

Thanks!

src/core/client/admin/routes/Moderate/Queue/QueueRoute.tsx
```diff
@@ -130,6 +151,7 @@ export const QueueRoute: FunctionComponent<Props> = ({
     subscribeToQueueCommentEntered,
     subscribeToQueueCommentLeft,
```
alinhle: Almost! While this behavior works for `subscribeToQueueCommentEntered `, `subscribeToQueueCommentLeft` must be always on.

alinhle (changes requested):

alinhle (approved):

PR #3353, opened 2020-12-21
[CORL-1586] My Comments Fix

wyattjoh:
## What does this PR do?

This fixes a bug where comments without reactions would not have a footer.

None.

## How do I test this PR?

Create a new comment, see that it appears with the footer buttons (at least to view the conversation) even though it doesn't have a reaction.


immber (commented):
👍

PR #3361, opened 2021-01-06
[CORL-1452] add fallback for when Domparser.parseFromString fails to return body

tessalt:
## What does this PR do?

I was unable to reproduce[ this bug ](https://sentry.io/organizations/voxmedia/issues/1913993149/?project=1452040&referrer=jira_integration) so this is something of a stab in the dark, but it does sound like it is likely the same bug as reported here: https://github.com/DylanPiercey/set-dom/issues/23 (ours appears in chrome on iOS as well as UIWebView), where`DomParser.parseFromString` sometimes does not return the document body synchronously. Fix is fairly straightforward fallback to innerHTML, not sure if we should just replace with innerHTML always or if there was a reason we wanted to use DOMParser instead. 

none

### DIFF
--- a/src/core/client/framework/lib/bootstrap/CoralContext.tsx
+++ b/src/core/client/framework/lib/bootstrap/CoralContext.tsx
@@ -95,11 +95,23 @@ export const CoralContextConsumer = CoralReactContext.Consumer;
 
 const parser = new DOMParser();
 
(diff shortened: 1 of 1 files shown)


alinhle (approved):
Great job @tessalt! I would keep it as a fallback for legacy browsers and allow modern browsers use more modern api which could be more optimized.

PR #3369, opened 2021-01-08
[CORL-1564] Add link to reply for flattened replies

nick-funk:
## What does this PR do?

Shows a confirmation with a "Go to reply" when you make a reply and the reply level is flattened. This allows commenters to easily jump to their reply in a very long/nested thread of replies.

None

## How do I test this PR?

- Enable flattened replies
- Make a comment that would appear at indent level 4 or higher (reply to a 3rd level indented coment)
- See the "Go to reply" confirmation and button link appear where the reply form used to be
- Click the "Go to reply"
- Stream iframe will scroll to the comment


src/core/client/stream/tabs/Comments/Comment/ReplyCommentForm/ReplyCommentFormContainer.tsx
```diff
@@ -186,10 +202,58 @@ export class ReplyCommentFormContainer extends Component<Props, State> {
...
+  private jumpToComment = (commentID: string | null, onClose?: () => void) => {
+    if (!commentID) {
+      return;
+    }
+
+    if (onClose) {
+      onClose();
+    }
+
+    const elementID = `comment-${commentID}`;
```
wyattjoh: This ID construction should be constructed. I'd suggest creating a function that would return the ID of the element, and suggest modifying code in the `stream` that uses the `comment-${comment.id}` pattern.

src/core/client/stream/tabs/Comments/Comment/ReplyCommentForm/ReplyCommentFormContainer.tsx
```diff
@@ -186,10 +202,58 @@ export class ReplyCommentFormContainer extends Component<Props, State> {
...
+    setTimeout(() => {
+      const element = document.getElementById(elementID);
+      if (element) {
+        element.scrollIntoView({ behavior: "smooth", block: "start" });
```
wyattjoh: This is in an iframe still, so in order to support scrolling for some browsers, we need to use `pym.js`'s `scrollParentToChildEl` function. It takes the `id` of the element in the frame, so you can use the same ID.

src/core/client/stream/tabs/Comments/Comment/ReplyCommentForm/ReplyCommentFormContainer.tsx
```diff
@@ -186,10 +202,58 @@ export class ReplyCommentFormContainer extends Component<Props, State> {
...
+          color="primary"
+        >
+          <Localized id="comments-jumpToComment-GoToReply">
+            <Button
+              onClick={() => {
+                this.jumpToComment(
+                  this.state.jumpToCommentID,
+                  this.props.onClose
+                );
+              }}
```
wyattjoh: As `jumpToComment` is already inside the component, you don't need to pass these parameters to it. If you refactored it to instead receive nothing, you can rewrite this as:

```suggestion
              onClick={() => {
                this.jumpToComment();
              }}
```

wyattjoh (changes requested):

src/core/client/stream/tabs/Comments/Comment/ReplyCommentForm/ReplyCommentFormContainer.tsx
```diff
@@ -17,20 +24,23 @@ import {
...
-  getSubmitStatus,
   shouldTriggerSettingsRefresh,
   shouldTriggerViewerRefresh,
   SubmitStatus,
 } from "../../helpers";
+import { getSubmissionResponse } from "../../helpers/getSubmitStatus";
 import RefreshSettingsFetch from "../../RefreshSettingsFetch";
 import RefreshViewerFetch from "../../RefreshViewerFetch";
 import { RTE_RESET_VALUE } from "../../RTE/RTE";
+import computeCommentElementID from "../commentElementID";
```
wyattjoh:
```suggestion
import commentElementID from "../commentElementID";
```

src/core/client/stream/tabs/Comments/Comment/ReplyCommentForm/ReplyCommentFormContainer.tsx
```diff
@@ -49,194 +59,241 @@ interface Props {
...
-          onDismiss={this.handleOnCancelOrDismiss}
-          buttonClassName={CLASSES.createReplyComment.dismiss}
-          inReviewClassName={CLASSES.createReplyComment.inReview}
-        />
-      );
+    if (props.onClose) {
+      props.onClose();
     }
+
+    const elementID = computeCommentElementID(commentID);
```
wyattjoh:
```suggestion
    const elementID = commentElementID(commentID);
```

wyattjoh (changes requested):

wyattjoh (approved):

PR #3371, opened 2021-01-11
[CORL-1518] Show rejected comments as removed for staff in single conversation view

nick-funk:
## What does this PR do?

Fixes the bug where staff could still see rejected comments in the single conversation view.

Shows the callout indicated the comment is missing or removed if the comment is rejected, which is the same behaviour we see for regular users in the single conversation view.

None

## How do I test this PR?

- Copy the share link to a reply comment 
- Reject the comment on the stream
- Log out
- Log in again
- Paste the link to the replied comment into your browser
- See if the comment still displays

### DIFF
--- a/src/core/client/stream/tabs/Comments/PermalinkView/PermalinkViewContainer.tsx
+++ b/src/core/client/stream/tabs/Comments/PermalinkView/PermalinkViewContainer.tsx
@@ -16,6 +16,7 @@ import {
   useSubscription,
   withFragmentContainer,
 } from "coral-framework/lib/relay";
+import { GQLCOMMENT_STATUS } from "coral-framework/schema";
 import CLASSES from "coral-stream/classes";
 import UserBoxContainer from "coral-stream/common/UserBox";
 import { ViewFullDiscussionEvent } from "coral-stream/events";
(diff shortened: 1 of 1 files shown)


src/core/client/stream/tabs/Comments/PermalinkView/PermalinkViewContainer.tsx
```diff
@@ -86,6 +87,10 @@ const PermalinkViewContainer: FunctionComponent<Props> = (props) => {
     return getURLWithCommentID(url, undefined);
   }, [pym]);
 
+  const commentVisible = !!(
+    comment && comment.status !== GQLCOMMENT_STATUS.REJECTED
```
tessalt: hmm do we not have a concept of which statuses should or should not be visible? this seems overly specific

wyattjoh: We do!

Right now, we have a `PUBLISHED_STATUSES` exported from `coral-server/models/comment/constants`, but we could extract those to `coral-common`.

nick-funk: Unfortunately, it can't be exported to `coral-common` because that area of the site has no access to the generated scheme types of either the server or client. To move this to `coral-common` would mean you'd need it to pull the `GQLCOMMENT_STATUS` type in from one of those areas, polluting either the server code that uses it with client types or the client code that uses it with server types. 

I've settled on just using the `isPublished` already defined `src/core/client/stream/tabs/Comments/helpers/isPublished.ts`.

tessalt (approved):

wyattjoh (approved):

PR #3373, opened 2021-01-13
Fix code around `useRefetch` and pagination `count` variable.

alinhle:
## What does this PR do?
- `useRefetch` did not allow setting `count` to refetch which was always set to 10. This PR fixes this.
- `useRefetch` `variables` was typed optional but is required in fact.
- remove illegitimate uses of `useRefetch`
- resolve mod queues `count` not passing through to pagination at initial render. 

## How do I test this PR?
- Nothing should be broken.
- `pending` queue etc should have `5` items at first render.

- [x] Merge after https://github.com/coralproject/talk/pull/3351 and a rebase


src/core/client/framework/lib/relay/useRefetch.ts
```diff
@@ -14,14 +14,15 @@ type IsRefetching = boolean;
  */
 export default function useRefetch<V = Variables>(
   relay: RelayPaginationProp,
-  variables: V = {} as any
+  totalCount: number,
```
wyattjoh, line 17: `totalCount` here is confusing, as it always seems to be hard set to `5` in a lot of places. Using `count` might be more appropriate?

wyattjoh (changes requested):

src/core/client/framework/lib/relay/useRefetch.ts
wyattjoh: I noticed that the underlying relay API uses `totalCount`... So approving!

wyattjoh (approved):

PR #3407, opened 2021-01-28
[CORL-1603] On small screens, bump the select sort text size

nick-funk:
## What does this PR do?

Makes iOS and Android stop zooming in/shifting the view in weird ways when you click a select field and have a phone with anything that can be considered a "small screen".

Note: @cvle, you were right, it's font size based and I tested as much on various devices.

None

## How do I test this PR?

- Set up your coral dev instance so you can access it from your phone (tenant domain, allowed external IP, etc)
- Grab a tiny phone like an iPhone 5 or an iPhone SE and navigate to the site
- Click on the Sort By field in stream
- No janky scrolling, jumping, zooming, etc should happen


src/core/client/stream/tabs/Comments/Stream/SortMenu.css
```diff
@@ -14,6 +14,15 @@
   line-height: 1.14;
 }
 
+@media screen and (max-width: 500px) {
```
tessalt: would it work if you used the `sm` breakpoint variable (640px)?

nick-funk: D'oh! Yes, it likely can. Let me fix that up.

tessalt (approved):

PR #3408, opened 2021-01-28
[Post-Hotfix] 6.7.2 French Translations

nick-funk:


nick-funk:
Not needed, main is still in a bad state to merge hotfixes back into develop

PR #3409, opened 2021-01-28
[CHORE] Provide a hook version of withStyles called useStyles

alinhle:
## What does this PR do?
- Provides a hooked version of HOC `withStyles`: `useStyles`
- Includes some example implementations where I changed from using the HOC to the hook version.
- Removed incomplete input assertions, and let typescript typing take over this job
## How do I test this PR?
- Everything should work as before.


tessalt (approved):
i love it!

PR #3426, opened 2021-02-03
[CORL-1599] Font descenders cutting off in ff

alinhle:
## What does this PR do?
- Fix select field descenders cutting off in Firefox

image


nick-funk (approved):

nick-funk:
I have approved, but looks like some snapshots need to be updated. Think it's whining about modifying the SelectField component.

PR #3430, opened 2021-02-04
[CORL-1643] Subscribe to and insert new questions on unanswered tab

nick-funk:
## What does this PR do?

Allow "View new comments" and newly posted comments to show when using the "Unanswered" tab in a Q&A conversation.

None

## How do I test this PR?

- Enable Q&A
- Set stream to Q&A mode
- Set yourself as an expert
- Switch to Unanswered stream tab with expert user
- Open another browser tab as commenter
- Go to Unanswered with commenter
- Post a question while viewing the Unanswered tab in commenter browser tab
- See that the comment is inserted into the Unanswered stream on commenter browser tab
- Switch to the expert's browser tab
- See that there is a view new comments button available, click it and ensure the new unanswered comment shows up in the Unanswered tab

### DIFF
--- a/src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
+++ b/src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
@@ -148,17 +148,6 @@ function tagUnansweredQuestions(
       featuredTag.setValue(GQLTAG.UNANSWERED, "code");
       node.setLinkedRecords(tags.concat(featuredTag), "tags");
     }
-    const commentCountsRecord = story.getLinkedRecord("commentCounts");
-    if (!commentCountsRecord) {
-      return;
-    }
-    const tagsRecord = commentCountsRecord.getLinkedRecord("tags");
-    if (tagsRecord) {
-      tagsRecord.setValue(
-        (tagsRecord.getValue("UNANSWERED") as number) + 1,
-        "UNANSWERED"
-      );
-    }
   }
 }
 
--- a/src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
+++ b/src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
@@ -11,6 +11,7 @@ import { CoralContext } from "coral-framework/lib/bootstrap";
 import {
   commitMutationPromiseNormalized,
   createMutation,
+  LOCAL_ID,
   lookup,
   MutationInput,
 } from "coral-framework/lib/relay";
@@ -147,6 +148,23 @@ function tagUnansweredQuestions(
       const featuredTag = store.create(uuidGenerator(), "Tag");
       featuredTag.setValue(GQLTAG.UNANSWERED, "code");
       node.setLinkedRecords(tags.concat(featuredTag), "tags");
+    }
+  }
+
+  // Only manually increment Unanswered count when we're not actively on
+  // the Unanswered questions tab.
+  const local = lookup(environment, LOCAL_ID);
+  if (local.commentsTab !== "UNANSWERED_COMMENTS") {
+    const commentCountsRecord = story.getLinkedRecord("commentCounts");
+    if (!commentCountsRecord) {
+      return;
+    }
+    const tagsRecord = commentCountsRecord.getLinkedRecord("tags");
+    if (tagsRecord) {
+      tagsRecord.setValue(
+        (tagsRecord.getValue("UNANSWERED") as number) + 1,
+        "UNANSWERED"
+      );
     }
   }
 }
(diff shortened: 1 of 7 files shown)


nick-funk:
~~Just realized the unanswered tab count is double incrementing, one sec, will get a quick fix in for that.~~

Fixed.

alinhle (changes requested):
I need to investigate a bit more, but I already got the following:
- Two parts of the code is doing count incrementing, there should be only one.
- Some of the tag handling should take into account what tags the server responses back.
- We should probably utilise optimistic response more in some cases.

I need to take a look at the whole tag handling. Will report next week.

src/core/client/stream/tabs/Comments/Stream/Subscriptions/CommentEnteredSubscription.ts
```diff
@@ -209,6 +213,7 @@ const CommentEnteredSubscription = createSubscription(
         const isTopLevelComent = !parent;
         if (
           variables.tag &&
+          variables.tag !== GQLTAG.UNANSWERED &&
```
alinhle, line 216: What's this line for?

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/PostCommentFormContainer.tsx
```diff
@@ -117,14 +117,21 @@ export const PostCommentFormContainer: FunctionComponent<Props> = ({
...
         body: input.body,
         rating: input.rating,
         media: input.media,
-      });
+      };
+
+      if (tab === "UNANSWERED_COMMENTS") {
+        params.tag = GQLTAG.UNANSWERED;
+        params.connectionKey = "UnansweredStream_comments";
+      }
```
alinhle, line 132: This is incomplete, and would only work for this particular tab.
Look at how R&R is determining its tag inside the `CreateCommentMutation`. I think we should follow the same pattern.

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
```diff
@@ -36,6 +37,7 @@ export type CreateCommentInput = Omit<
 > & {
   commentsOrderBy?: COMMENT_SORT;
   tag?: GQLTAG;
+  connectionKey?: string;
```
alinhle, line 40: I'd remove both of these. `tag` is never used except by the code in this PR. We can infer both from current state.

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
```diff
@@ -91,7 +93,11 @@ function addCommentToStory(
 ) {
   // Get stream proxy.
   const streamProxy = store.get(input.storyID)!;
-  const connectionKey = "Stream_comments";
+
+  let connectionKey = "Stream_comments";
+  if (input.connectionKey) {
+    connectionKey = input.connectionKey;
+  }
```
alinhle, line 100: Instead of taking `connectionKey` here we could infer from current story mode.

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
```diff
@@ -143,6 +149,12 @@ function tagUnansweredQuestions(
       featuredTag.setValue(GQLTAG.UNANSWERED, "code");
       node.setLinkedRecords(tags.concat(featuredTag), "tags");
     }
+  }
+
+  // Only manually increment Unanswered count when we're not actively on
+  // the Unanswered questions tab.
+  const local = lookup(environment, LOCAL_ID);
+  if (local.commentsTab !== "UNANSWERED_COMMENTS") {
```
alinhle, line 157: We don't need the whole count logic here. There is already a routine that manages counts in `incrementTagCommentCounts`, it just needs the right input.

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
```diff
@@ -406,11 +418,31 @@ export const CreateCommentMutation = createMutation(
...
+            if (tag) {
+              sharedUpdater(
+                environment,
+                store,
+                { ...input, tag },
+                uuidGenerator
+              );
+            } else {
+              sharedUpdater(environment, store, input, uuidGenerator);
+            }
```
alinhle, line 445: If we remove `tag` from `input` then we don't need this change here anymore.

alinhle (changes requested):
Finished the Code Review. Let's work this out together next week. Also take a look at  `UnansweredCommentsTabViewNewMutation` and `AllCommentsTabViewNewMutation`. They are responsible for handling counts of incoming comments.

src/core/client/stream/tabs/Comments/Stream/PostCommentForm/CreateCommentMutation.ts
alinhle: To be correct, the `sharedUpdater` should actually use the `tags` that comes from the `response` instead.

alinhle:
Superseded by https://github.com/coralproject/talk/pull/3567

PR #3431, opened 2021-02-04
[CORL-1649] Bugfix: Comment draft deleted on login or registration

tessalt:
## What does this PR do?

- creates new hooks wrapping `usePersistedState` allowing use of either local or session storage
- ~switch `PostCommentForm:draft` from session to local storage so it isn't removed when session is cleared via login/signup~
- Prevent sessionStorage from being cleared after login

none

## How do I test this PR?

1. logout
2. type text into the input box
3. click "register" and register a new user
4. ensure text is still in input box


alinhle:
Looks good. I've changed the fix slightly: 
- Instead of saving state to local storage – which is never cleared – the data is kept inside session storage
- Changed so that a login does not clear the session storage, only a logout does

alinhle (approved):

PR #3432, opened 2021-02-04
Update docs to reflect v6 admin layout

mandy6720:
Note that this is just for the permitted domains. There seem to be a few other changes between the v5 and v6 admin layout

## What does this PR do?
Updates the docs to reflect changes in v6

none


tessalt:
Hi there! Thanks for your contribution. We're actually in the process of re-launching our documentation, but we'll be sure to check that these instructions are correct when we do!

PR #3434, opened 2021-02-04
[CORL-1583] <StreamModal />

alinhle:
## What does this PR do?
- Move `window` into `CoralContext`
- Introduce `renderWindow` into `CoralContext` which is always set to the  `window` we are rendering too.
- Introduce `renderWindow` into `UIContext`.
- In case of rendering to another frame through a portal, `window` and `renderWindow` should differ, otherwise they are the same
- Introduce `<RenderTargetPortal />` that allows you to render to other render targets in other iframes.
- Introduce `<StreamModal />` that renders to a stream modal.

## How to test this PR?
- Write some code that uses `<StreamModal />`. It should behave exactly the same as a normal `<Modal />`


nick-funk (approved):

alinhle:
Merged into https://github.com/coralproject/talk/pull/3697

PR #3438, opened 2021-02-08
[CORL-1641/42] Disable rating edit for R&R, tab count fixes

nick-funk:
## What does this PR do?

- Disable editing a rating only comment for R&R, can still edit if it's a review comment
- Tab counts will update for everything except the Review tab
  - Why? Cause the review tab improperly re-uses the `AllCommentsTab` and simply sets a `tag` to filter the comments. This works for loading the comments but makes for a bunch of extra logic that will need to be added to manually update counts in Relay for both the `CreateComment` and `CommentEnteredSubscription`. If we created another comment stream, it would work better, but alas, it is what it is now.

None

## How do I test this PR?

- Create a R&R stream
- Go to it

Edit:
- Create a rating only comment
- Quickly go to comment history
- Find rating only comment
- Single conversation view on it
- Edit button should be missing

Counts:
- Post as one commenter while watching with another commenter
- Switch the tabs around and make sure the counts update when another commenter posts to different comment tab than the one you are on


tessalt (dismissed):
LGTM but lets make sure to test this behaviour before release

alinhle (changes requested):
The `AllCommentsTab` count update mechanism is implement in  `AllCommentsTabViewNewMutation`.

This introduces another count update mechanism. Now the counts are being updated twice.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/CommentEnteredSubscription.tsx
```diff
@@ -175,6 +175,17 @@ const CommentEnteredSubscription = createSubscription(
...
+                commentCounts {
+                  totalPublished
+                  tags {
+                    FEATURED
+                    UNANSWERED
+                    REVIEW
+                    QUESTION
+                  }
+                }
+              }
```
alinhle: This won't fix the count issues, rather it conflicts with other count logic. You should look to update `AllCommentsTabViewNewMutation` instead.

alinhle (dismissed):

nick-funk:
@cvle , I saw your comment. All I did was add a query to pull in the counts. There is no extra incrementing or decrementing going on from the changes I can see? How does querying from the source of truth double increment it?

alinhle:
@nick-funk The counts are incremented at `AllCommentsTabViewNewMutation`, so it would bite with having them updated during the subscription.

alinhle (approved):

PR #3439, opened 2021-02-08
Auto focus first input field on Sign In and Sign Up popups

nick-funk:
## What does this PR do?

Auto focuses the first input field (the email field) on Sign In and Sign Up popups.

It was really starting to bug me that this was not the first focused element.

None

## How do I test this PR?

- Go to stream
- Click `Register` or `Sign In`
- Notice that the first field (email field) is focused and you can immediately start typing into it

### DIFF
--- a/src/core/client/auth/components/EmailField.tsx
+++ b/src/core/client/auth/components/EmailField.tsx
@@ -1,5 +1,5 @@
 import { Localized } from "@fluent/react/compat";
-import React, { FunctionComponent } from "react";
+import React, { FunctionComponent, useEffect } from "react";
 import { Field } from "react-final-form";
 
 import { streamColorFromMeta } from "coral-framework/lib/form";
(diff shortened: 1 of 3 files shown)


tessalt (approved):

src/core/client/auth/components/EmailField.tsx
```diff
@@ -13,33 +13,49 @@ import { ValidationMessage } from "coral-ui/components/v3";
...
-            fullWidth
-          />
-        </Localized>
-        <ValidationMessage meta={meta} />
-      </FormField>
-    )}
-  </Field>
-);
+const EmailField: FunctionComponent<Props> = (props) => {
+  let inputRef: HTMLInputElement | null = null;
```
wyattjoh: This should use `useRef` instead:

```suggestion
const inputRef = useRef<HTMLInputElement>(null);
```

wyattjoh (dismissed):

alinhle (approved):

PR #3466, opened 2021-02-24
[CORL-1658] AMP Configuration

alinhle:
## What does this PR do?
- Allow AMP support to be enabled inside of `Admin` -> `Config` -> `Advanced`

## How do I test this PR?
- `npm run start:development`
- goto `http://localhost:3000/embed/stream/amp` 
- Check if AMP is really turned on/off
  - It's on when:  `Content-Security-Policy: frame-ancestors` is not sent in the response headers
  - It's off when:  `Content-Security-Policy: frame-ancestors` is sent 

# TODO
- Actual URL to new documentation


nick-funk:
I ran this and tried it with and without AMP enabled. Not seeing `Content-Security-Policy: frame-ancestors` be sent back in the response headers for either option.

nick-funk (changes requested):
I ran this and tried it with and without AMP enabled. Not seeing `Content-Security-Policy: frame-ancestors` be sent back in the response headers whether I enable AMP or not.

nick-funk (approved):

src/core/server/models/tenant/tenant.ts
```diff
@@ -93,7 +100,7 @@ export interface TenantSettings
   /**
    * featureFlags is the set of flags enabled on this Tenant.
    */
-  featureFlags?: GQLFEATURE_FLAG[];
+  featureFlags?: (GQLFEATURE_FLAG | LEGACY_FEATURE_FLAGS)[];
```
wyattjoh: Maybe this might read cleaner?

```suggestion
  featureFlags?: Array<GQLFEATURE_FLAG | LEGACY_FEATURE_FLAGS>;
```

src/core/server/models/tenant/helpers.ts
```diff
@@ -98,3 +98,10 @@ export function supportsMediaType(
       return !!tenant.media?.giphy.enabled && !!tenant.media.giphy.key;
   }
 }
+
+export function isAMPEnabled(tenant: Pick<Tenant, "featureFlags" | "amp">) {
+  if (tenant.amp !== undefined) {
+    return !!tenant.amp;
+  }
```
wyattjoh: Maybe this reads a bit cleaner?

```suggestion
  if (typeof tenant.amp === "boolean") {
    return tenant.amp;
  }
```

PR #3493, opened 2021-04-05
[CORL-1709] adding af-za translations

immber:
## What does this PR do?

Starts to add some translations for Africans


immber:
@tessalt This should be good to go now with all the english strings removed. I'm not able to run the branch locally because my local env is not set up for that, so I'm not able to test the addition of the language to the list, but assuming that looks good this should be ready.

tessalt (approved):

PR #3505, opened 2021-04-15
Fix linting errors in live chat form containers

nick-funk:
## What does this PR do?

Use `usePersistedSessionState` instead of `usePersistedState` so that it stops whining about a missing `localStorage` prop since the `usePersistedSessionState` injects it for us.

Updates this to bring it in line with what we do in the regular comment forms.

None

## How do I test this PR?

Check code, build it, run it, no linting errors.


nick-funk:
Closing as I already pushed these changes in another branch

PR #3506, opened 2021-04-15
Remove `latestCursor` from Local relay cache

nick-funk:
## What does this PR do?

Removes the `latestCursor` value from the Local relay cache.

Instead uses two helper functions to retrieve and persist this constantly changing value.

This same retrieval is also used for the retrieval during `initLocalState` for the `currentCursor`.

None on main schema.
Removes `latestCursor` from `local.graphql` stream side.

## How do I test this PR?

Load up the stream and make sure it is properly remembering your cursor position.

### DIFF
--- a/src/core/client/stream/local/initLocalState.ts
+++ b/src/core/client/stream/local/initLocalState.ts
@@ -12,7 +12,7 @@ import {
   initLocalBaseState,
 } from "coral-framework/lib/relay";
 import { GQLFEATURE_FLAG, GQLSTORY_MODE } from "coral-framework/schema";
-import CursorState from "coral-stream/tabs/Live/cursorState";
+import { getLatestCursor } from "coral-stream/tabs/Live/cursorState";
 
 import { initLocalStateQuery } from "coral-stream/__generated__/initLocalStateQuery.graphql";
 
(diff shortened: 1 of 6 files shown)


src/core/client/stream/tabs/Live/LivePostCommentFormContainer.tsx
```diff
@@ -89,10 +89,10 @@ export const LivePostCommentFormContainer: FunctionComponent<Props> = ({
   const [nudge, setNudge] = useState(true);
   const [submitStatus, setSubmitStatus] = useState<SubmitStatus | null>(null);
 
-  const [draft = "", setDraft, initialDraft] = usePersistedState<string>(
+  const [draft = "", setDraft, initialDraft] = usePersistedSessionState<string>(
```
nick-funk, line 92: Doing these here as well as in the other PR because otherwise it won't compile and we won't be able to merge.

src/core/client/stream/local/initLocalState.ts
```diff
@@ -90,13 +90,11 @@ const initLocalState: InitLocalState = async ({
...
   const query = parseQuery(location.search);
 
-  const key = `liveCursor:${query.storyID}:${query.storyURL}`;
-
-  const rawValue = await context.localStorage.getItem(key);
-  let currentCursor: CursorState | null = null;
-  if (rawValue) {
-    currentCursor = JSON.parse(rawValue);
-  }
+  const currentCursor = await getLatestCursor(
```
alinhle: Just a small suggestion: What do you think about calling it `const latestCursorState = await getLatestCursorState`, that way`latestCursorState.cursor` makes more sense than `currentCursor.cursor`

alinhle (changes requested):

alinhle (approved):

PR #3507, opened 2021-04-15
Live Chat: Mutation fixes and cleanup

alinhle:
## What does this PR do?

- Remove non-live-chat related code from mutations
- fix optimistic responses and data dependencies


nick-funk (approved):

nick-funk:
This looks good! Approving! Tests look to be complaining about a snapshot, should be easy to update so we can get this green and merge it.

PR #3516, opened 2021-04-22
[CORL-1748] Refactor Live Conversation and Mutations

alinhle:
## What does this PR do?

- Refactor Live Conversation
- Refactor out hasNextPageFromMutation


src/core/client/stream/tabs/Live/LiveChatContainer.tsx
```diff
@@ -64,7 +63,10 @@ import styles from "./LiveChatContainer.css";
 
 interface ConversationViewState {
   visible: boolean;
-  comment?: PropTypesOf<typeof LiveConversationContainer>["comment"] | null;
+  comment?:
+    | LiveCommentContainer_comment
+    | NonNullable<LiveCommentContainer_comment["parent"]>
```
nick-funk, line 68: Hah! Nice typing here

src/core/client/stream/tabs/Live/helpers/handleNewCommentInStory.ts
```diff
@@ -7,31 +7,33 @@ import {
...
-      beforeConnection,
-      Boolean(options.fromMutation)
-    );
+    if (!hasNextPage) {
+      insertComment(store, comment, afterConnection);
+    }
+    return;
+  }
+  if (!hasNextPage) {
+    markHasNextPage(afterConnection, beforeConnection);
```
nick-funk, line 37: Ooof, that `return` complicates things a bit, eh? I started to try and write a cleaner version of this, only to realize I'd end up with 3 flat if's instead of a clean `if...else if`...

I might have done this, that way you can comment each specific flow with flat if's:

```
// live insertion, with no next page, we can simply insert the 
// comment and move along
if (options.liveInsertion && !hasNextPage) {
   insertComment(store, comment, afterConnection);
   return;
}
// not live inserting, not tailing, hasNextPage is true, we
// will just let the existing pagination flag handle it
else if (options.liveInsertion && hasNextPage) {
   return;
}

// not tailing, not live inserting, so we can just mark the
// pagination to handle this
if (!hasNextPage) {
   markHasNextPage(afterConnection, beforeConnection);
}
```

nick-funk (approved):

Issue #3357, opened 2020-12-30
Unhelpful validation errors on initial configuration wizard


Sentynel:
**Expected behavior:**
Validation errors on initial configuration tell you what you did wrong.

**Actual behavior:**
![image](https://user-images.githubusercontent.com/2052340/103358709-2aeef380-4aae-11eb-90c0-bb60e4b7caf5.png)
Obvious question: Why is this failing validation? Answer: it isn't; I missed an "https://" off the organisation URL on the previous screen and *that* one has failed validation.
The error message in the dev console reads `"message": "\"tenant.organization.url\" must be a valid uri"` so it should be possible to improve this.

**Related Issues:**
![image](https://user-images.githubusercontent.com/2052340/103358881-9cc73d00-4aae-11eb-93ed-9fd02724ed6f.png)
I don't think the initial configuration wizard particularly needs rate limiting, and it was annoying when trying to fix the previous issue!

**Versions:**

- Coral: git develop dc656e2
- NodeJS: v14.15.3
- NPM: 6.14.9
- MongoDB:
- Redis:
- Browser: Chrome/87.0.4280.107
- OS: Ubuntu 20.04

wyattjoh:
Hi @Sentynel! Thank's for the info. We definitely know that the installation wizard could use some love! Thanks for bringing it to our attention!

PR #3517, opened 2021-04-23
URL validation during installation

ghost:
## What does this PR do?

Addresses #3357 (unhelpful validation errors on initial configuration wizard)

On the installation wizard's add organization step, the site url requires that it begins with either 'https://' or 'http://' -- however, the 'Invalid URL' notice does not show up if someone types in something like 'www.google.com'. Then, when they proceed to the permitted domains page and click finish, a validation error occurs with no details. 

This PR gives a "Be sure to include http:// or https:// in your URL" error message below the input, so users can't proceed to the next installation step, until 

None

## How do I test this PR?

1. Go to the installation wizard
2. Complete installation steps until 'Add Organization and site details'
3. Under organization url try a url without 'http', like 'www.google.com'. It should give an error message to include http://
4. When you try a url with an invalid url like 'http:/google.com', the original 'Invalid URL' error will show up
5. Continue to the 'Site permitted Domains' page, and repeat steps 3 and 4 but with a list of URLs

### DIFF
--- a/src/core/client/framework/lib/messages.tsx
+++ b/src/core/client/framework/lib/messages.tsx
@@ -84,6 +84,12 @@ export const INVALID_URL = () => (
   </Localized>
 );
 
+export const INVALID_URL_START = () => (
+  <Localized id="framework-validation-invalidURLStart">
+    <span>Be sure to include http:// or https:// in your URL</span>
+  </Localized>
+);
+
 export const NOT_A_WHOLE_NUMBER = () => (
   <Localized id="framework-validation-notAWholeNumber">
     <span>Please enter a valid whole number</span>
--- a/src/core/client/framework/lib/validation.tsx
+++ b/src/core/client/framework/lib/validation.tsx
@@ -264,7 +264,7 @@ export const validateDeleteConfirmation = (phrase: string) =>
 /**
  * validateURLListStart is a validator that checks if a list of URLs all begin with 'http'.
  */
- export const validateURLListStart = createValidator((v) => {
+export const validateURLListStart = createValidator((v) => {
   for (const url of v) {
     if (!startsWith(url, "http")) {
       return false;
@@ -277,7 +277,7 @@ export const validateDeleteConfirmation = (phrase: string) =>
 /*
  * validateWeakURLList is validator that Checks if a list of URLs are all valid URLs.
  */
- export const validateWeakURLList = createValidator((v) => {
+export const validateWeakURLList = createValidator((v) => {
   if (!Array.isArray(v)) {
     return false;
   }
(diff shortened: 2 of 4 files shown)


src/core/client/framework/lib/messages.tsx
```diff
@@ -84,6 +84,12 @@ export const INVALID_URL = () => (
   </Localized>
 );
 
+export const INVALID_URL_START = () => (
+  <Localized id="framework-validation-invalidURLStart">
+    <span>Be sure to include http:// or https:// in your URL</span>
+  </Localized>
+);
+
```
alinhle, line 92: Maybe call this `INVALID_URL_NOT_HTTP` because in essence we check if the http(s) protocol is used or no.

src/core/client/framework/lib/validation.tsx
```diff
@@ -106,6 +107,17 @@ export const validateURL = createValidator(
...
+ * validateURLStart is a Validator that checks that the URL starts with 'http'.
+ */
+export const validateURLStart = createValidator((v) => {
+  if (!startsWith(v, "http")) {
+    return false;
+  }
+
+  return true;
+}, INVALID_URL_START());
+
```
alinhle, line 120: It should check `startsWith(v, "http://") || startsWith(v, "https://")` instead and the method should maybe named `validateURLHTTP`.

src/core/client/framework/lib/validation.tsx
```diff
@@ -249,7 +261,23 @@ export const validatePercentage = (min: number, max: number) =>
...
+ */
+export const validateURLListStart = createValidator((v) => {
+  for (const url of v) {
+    if (!startsWith(url, "http")) {
+      return false;
+    }
+  }
+
+  return true;
+}, INVALID_URL_START());
```
alinhle, line 275: See above

src/core/client/framework/lib/validation.tsx
```diff
@@ -249,7 +261,23 @@ export const validatePercentage = (min: number, max: number) =>
...
+
+/*
+ * validateWeakURLList is validator that Checks if a list of URLs are all valid URLs.
+ */
+export const validateWeakURLList = createValidator((v) => {
```
alinhle, line 280: This should still keep the name `validateStrictURLList`, the strictness comes from the `URL_REGEX` that requires the protocol to be specified.

src/core/client/install/App/steps/AddOrganizationStep.tsx
```diff
@@ -141,7 +142,11 @@ class AddOrganizationStep extends React.Component<Props> {
 
               <Field
                 name="siteURL"
-                validate={composeValidators(required, validateURL)}
+                validate={composeValidators(
+                  required,
+                  validateURL,
+                  validateURLStart
```
alinhle, line 148: Switching them the other way round would produce a more helpful error message.

alinhle (changes requested):
Thanks for your PR @parth364! Here are a couple of remarks about the changes.

alinhle:
Closed due to inactivity

Issue #3358, opened 2021-01-01
Unicode story URLs not normalised


Sentynel:
**Expected behavior:**
A story with Unicode in its URL should be normalised by Coral so that all browsers, regardless of behaviour, treat it as the same story.

**Actual behavior:**
Example URL: `https://www.angrymetalguy.com/medico-peste-%D7%91-the-black-bile-review/` (Note: the production instance of this page is running Coral 4, but I have verified that the same issue exists on the latest release of Coral 6.) The title and canonical URL for this page include a unicode character. If you inspect the source, you'll see that WordPress generates a canonical link element `<link rel="canonical" href="https://www.angrymetalguy.com/medico-peste-ב-the-black-bile-review/" />` which contains the actual Unicode character. Chrome and most browsers transparently handle Unicode in URLs and translate it to the percent-encoded UTF8 representation. This means that when the Coral comments embed attempts to fetch the comments stream, the URL that ends up in the Coral database is `"url" : "https://www.angrymetalguy.com/medico-peste-%D7%91-the-black-bile-review/"`. However, this is not true of all browsers. I have three different variations of this URL in my database:
`https://www.angrymetalguy.com/medico-peste-%D7%91-the-black-bile-review/` - current Chrome, Firefox, most browsers
`https://www.angrymetalguy.com/medico-peste-ב-the-black-bile-review/` - pre-Edgium MS Edge, others?
`https://www.angrymetalguy.com/medico-peste-%d7%91-the-black-bile-review/` - (note lowercase percent escape) unknown browser(s)?

Each of these is treated as a separate comments thread, and the applicable browser sees only that thread of comments when viewing the page. I unfortunately do not have a complete list of badly behaved browsers. This was discovered in historical data and my access logs have long since aged off. For testing purposes I've been able to reproduce using Edge, as noted above.

There's definitely an argument to be made that this isn't a Coral problem - maybe WordPress should be generating percent-escaped URLs, maybe browsers should be submitting percent-escaped URLs, etc. (And I realise the misbehaving browsers are not likely to be on the "supported browser" list.) However, in the spirit of being tolerant to weird inputs but consistent about outputs, I do think Coral should be normalising URLs here.

**Related Issues:**
n/a

**Versions:**

- NodeJS: v14.15.3
- NPM: 6.14.9
- MongoDB: v3.6.8
- Redis: v=5.0.7
- Browser: (discussed above)
- OS: (as above)

alinhle:
[...]

PR #3518, opened 2021-04-23
Normalization of urls 

ghost:
## What does this PR do?

This PR addresses #3358. When a comment stream is embedded into a page with no canonical link tag, the comment stream takes the url from the window. If the url in the window has different capitalization or encoding then multiple separate threads are created for the same page. This PR ensures the url from the window is normalized.

None
## How do I test this PR?

1. Embed a comment stream into a page and make a comment
2. Change the capitalization of the URL in the browser and go to that same page (with different capitalization)
3. You should see the same comment already there, because it's a single comment stream

### DIFF
--- a/src/core/client/framework/helpers/resolveStoryURL.ts
+++ b/src/core/client/framework/helpers/resolveStoryURL.ts
@@ -14,15 +14,15 @@ function resolveStoryURL() {
   );
 
   // ensure that urls are encoded properly and lowercase
-  let pathname = window.location.pathname
+  let pathname = window.location.pathname;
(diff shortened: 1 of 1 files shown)


cvle:
This is a bit tricky and we have to be very careful to do this right. There should be a test that proves that all kinds of urls are encoded to the way modern standard browsers do.

alinhle (changes requested):
This is a bit tricky and we have to be very careful to do this right. There should be a test that proves that all kinds of urls are encoded the way modern standard browsers do.

alinhle:
Closed due to inactivity

PR #3519, opened 2021-04-23
[CORL-1739] Live Chat: Tombstone rejected comments with subscription

nick-funk:
## What does this PR do?

Tombstones rejected comments as they happen live within the live chat.

Since users don't refresh or reload the page as often when using live chat, a bad comment that is rejected can be visible for a long time. As such, we use a subscription to tombstone to clear them out.

Adds a new `commentStatusChanged` with payload to the subscriptions.

## How do I test this PR?

- Go to live chat
- Create a comment
- Use another account in another tab to reject it
- Watch the newly created comment get tombstone-ed in real time


src/core/server/graph/schema/schema.graphql
```diff
@@ -8564,6 +8564,29 @@ type CommentEditedPayload {
...
+
+  """
+  oldStatus is the previous status assigned to the Comment.
+  """
+  oldStatus: COMMENT_STATUS!
+
+  """
+  commentID is the comment id that was changed on the story.
+  """
+  commentID: ID!
```
alinhle: Should return whole comment here instead.

src/core/server/graph/schema/schema.graphql
```diff
@@ -8637,6 +8660,11 @@ type Subscription {
   """
   commentEdited(storyID: ID!): CommentEditedPayload!
 
+  """
+  commentRejected returns when a Comment has been rejected anywhere in the Story.
+  """
+  commentStatusChanged(storyID: ID!): CommentStatusChangedPayload!
```
alinhle: We want to expose minimal information about moderation to the public. Because we only care about rejected comments, this should be `commentRejected` instead.

alinhle (changes requested):
This looks good in general! Here are some comments:

alinhle (approved):

PR #3525, opened 2021-04-26
[CORL-1749] Live Chat: stop navigation buttons from blocking other stream controls

nick-funk:
## What does this PR do?

Forces flex alignment `div`'s in the live chat navigation buttons from having a height so they don't block click events behind them on the stream.

None

## How do I test this PR?

- Go to live chat
- View stream when not tailing
- Post a comment with a different account + user
- See the "new comments" or "jump to live" show up
- Line up the button with the "rec", "reply", and "show conversation" buttons of a comment
- Click on the "rec", "reply", or "show conversation" and it will work, no longer blocked by the invisible Flex div that aligns the button center at the bottom of the stream

### DIFF
--- a/src/core/client/stream/tabs/Live/JumpToButton.css
+++ b/src/core/client/stream/tabs/Live/JumpToButton.css
@@ -1,3 +1,7 @@
+.root {
+  position: relative;
+}
+
 .jumpToContainer {
   position: absolute;
 
@@ -6,8 +10,7 @@
   z-index: 998;
 
   transform: translate(-50%, 0%);
-
-  background: transparent;
+  white-space: nowrap;
 }
 
 .jumpButton {
--- a/src/core/client/stream/tabs/Live/JumpToButton.css
+++ b/src/core/client/stream/tabs/Live/JumpToButton.css
@@ -1,22 +1,13 @@
 .jumpToContainer {
-  position: relative;
+  position: absolute;
 
-  top: -20px;
-  left: 0;
-  width: 100%;
-  height: 0px;
+  bottom: var(--spacing-1);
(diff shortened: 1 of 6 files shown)


alinhle (changes requested):
It looks like it is working from the outside. Looking into the implementation raises a lot of question marks. 

For the main view I see 3 flex container elements, which seems excessive and redundant, and makes the `noHeight` unnecessary as well. For the conversation view there is still 2 flex container elements. The use of `overflow: visible` and `height: 0px` is also questionable in general. 

image

I recommend combining `position: relative`, `position: absolute`, `pointer-events: none`, `pointer-events: initial` for a clean implementation, removing the additional flex container inside `LiveChat` and refactor all jump buttons to use `JumpToButton` inside of the conversation view.

nick-funk:
> It looks like it is working from the outside. Looking into the implementation raises a lot of question marks.
> 
> For the main view I see 3 flex container elements, which seems excessive and redundant, and makes the `noHeight` unnecessary as well. For the conversation view there is still 2 flex container elements. The use of `overflow: visible` and `height: 0px` is also questionable in general.
> 
> image
> 
> I recommend combining `position: relative`, `position: absolute`, `pointer-events: none`, `pointer-events: initial` for a clean implementation, removing the additional flex container inside `LiveChat` and refactor all jump buttons to use `JumpToButton` inside of the conversation view.

I didn't need to do `pointer-events` to make this work, but I was able to remove the Flex's and make this work properly using `position: relative` and `position: absolute` with a `transform` to help center the element dynamically. Should be a lot cleaner now.

alinhle (approved):

PR #3531, opened 2021-04-29
[CORL-1747] Live Chat: Merge the root and conversation "comment entered" subscriptions

nick-funk:
## What does this PR do?

Merges the live chat root and conversation "comment entered" subscriptions together to prevent race conditions

None

## How do I test this PR?

- Use live chat
- Create comments
- See that it updates the "jump to new", live insertion, etc on both the main chat and also the conversation views


src/core/client/stream/tabs/Live/LiveChatContainer.tsx
```diff
@@ -133,7 +134,7 @@ const LiveChatContainer: FunctionComponent<Props> = ({
     conversationState,
     showConversation,
     hideConversation,
-  ] = useConversation(eventEmitter);
+  ] = useConversation(eventEmitter, setLocal);
```
alinhle: Instead of passing `setLocal` here which is initialized with a lot more than just conversation related data, do a `useLocal` inside of `useConversation` and return the `conversationRootCommentID`.

src/core/client/stream/tabs/Live/LiveCommentEnteredSubscription.tsx
```diff
@@ -13,13 +13,26 @@ import { LiveCommentEnteredSubscription } from "coral-stream/__generated__/LiveC
...
+
+  return liveChat.tailingConversation;
+}
+
+function conversationRootCommentID(environment: Environment): string | null {
+  const liveChat = lookup(environment, LOCAL_ID).liveChat;
+
+  return liveChat.conversationRootCommentID;
+}
+
```
alinhle: What do you think about calling these `isLiveMainStreamInsertionEnabled`, `isLiveConversationInsertionEnabled`, `getConversationRootCommentID` instead? I mistook them for variables on first glance instead of functions.

alinhle (changes requested):
Pretty awesome! Thanks. I have only minor requests.

alinhle (approved):
Just removed an unused field in a commit, everything else looks awesome!

PR #3537, opened 2021-05-04
[CORL-1756] Live Chat: Fix empty state callout spacing

nick-funk:
## What does this PR do?

Adds some spacing between the top of the scrollable live chat stream and the "There are no comments yet. Why don't you write one?" callout when the stream is empty.

None

## How do I test this PR?

- Create empty live chat stream
- Visit it
- See nicely spaced callout asking you to write a comment


alinhle (approved):
There seem to be a conflict, and CI is weird.

Otherwise this looks good!

PR #3547, opened 2021-05-10
[CORL-1759] Live Chat: Scroll report dialog into view when shown

nick-funk:
## What does this PR do?

Scrolls report dialog into view when shown.

Prevents it from being hidden below the scrolled state of Virtuoso when clicking `Report` on comments near bottom of stream scroll state.

None

## How do I test this PR?

- Go to live chat stream
- Scroll to bottom / jump to live
- Click report on bottom most comment
- See that stream scrolls the report dialog into view under the comment

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26084,7 +26084,7 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+          "resolved": false,
           "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
           "dev": true,
           "optional": true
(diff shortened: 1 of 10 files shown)


alinhle (changes requested):
This can be done in a more efficient way with better ux. I added a commit https://github.com/coralproject/talk/pull/3547/commits/ed1d4a3403bd9067f9929da79914322b4086288c to give you an `ensureInView` function, which automatically finds the first scrollable parent for an element and applies the scrolling if out of view. You should use this for the implementation. 

Additionally instead of using `getElementById` and `setTimeout` it's much more efficient to use `ref` instead, which gives you an instantaneous experience.

Here is some pseudo-code:
```ts
// Could move this in a helpers / utils folder.
const ensureRefInView = (el: HTMLElement | null) => {
  if (el) {
    ensureInView(el);
  }
}; 

const Component = () => {
  return showReportFlow && (
    <div ref={ensureRefInView}>
      <ReportFlowContainer
        viewer={viewer}
        comment={comment}
        settings={settings}
        onClose={toggleShowReportFlow}
      />
    </div>
  );
}
```

alinhle (approved):

PR #3561, opened 2021-05-13
[CORL-1587] Rejected Tombstone in Conversation Thread

alinhle:
## What does this PR do?
- Fix pagination issues in the Conversation Thread of the Single Conversation View.
  - Fix load more showing when there is only one parent which is already rendered
  - Continue rendering root parent when fully expanded

- Refactor/Fix `IgnoredTombstoneOrHideContainer`
  - Move deleted tombstone behavior out into a different Container `DeletedTombstoneContainer`
  - Split `allowTombstoneReveal` into  `allowTombstoneReveal` and `disableHide` to reflect what it is actually doing
  - Fix styling issues with `Show Comment` button

- Refactor  `CommentContainer`
  -  Move out part of deleted tombstone behavior into `DeletedTombstoneContainer`

- Refactor/Fix  `EditCommentMutation`, `RejectCommentMutation`, `ApproveCommentMutation`
  - Some optimistic response fixes
  - Prevent double rendering  

- Rename `RejectedTombstoneContainer` to `ModerationRejectedTombstoneContainer`
- Implement `RejectedTombstoneContainer` for the  `ConversationThread`

## How do I test this PR?
- Test `ConversationThread` pagination
- Test all tombstone related behavior
  - DeletedTombstone in Stream, Single Conversation View
  - RejectedTombstone in Single Conversation View - Convesration Thread
  - ModerationRejectedTombstone
  - IgnoredUserOrHideTombstone in Stream and Single Convo View


src/core/client/stream/classes.ts
```diff
@@ -434,16 +434,22 @@ const CLASSES = {
...
-   * rejectedTombstone is shown when a comment got rejected.
+   * moderationRejectedTombstone is shown to the moderator when a comment got rejected.
    */
-  rejectedTombstone: {
-    $root: "coral coral-rejectedtombstone",
-    goToModerateButton: "coral coral-rejectedtombstone-goToModerateButton",
+  moderationRejectedTombstone: {
+    $root: "coral coral-moderationRejectedTombstone",
+    goToModerateButton:
+      "coral coral-moderationRejectedTombstone-goToModerateButton",
```
nick-funk, line 442: just a note, that if we change this, we need to make sure our client CSS that might be using it for custom styling is updated as well.

src/core/client/stream/classes.ts
```diff
@@ -434,16 +434,22 @@ const CLASSES = {
...
+  /**
+   * rejectedTombstone is shown when a comment has been rejected.
+   */
+  rejectedTombstone: "coral coral-rejectedTombstone",
+
   /**
    * deletedTombstone is shown when a comment has been deleted.
    */
-  deletedTombstone: "coral coral-rejectedtombstone",
+  deletedTombstone: "coral coral-deletedTombstone",
```
nick-funk, line 452: If we're going to be changing the static CSS names for these, we need to make sure we let clients know and update any custom CSS we are in charge of writing for SBN

nick-funk (approved):

src/core/client/stream/classes.ts
alinhle: This is a bug though, as you can see by the naming. I hope we didn't rely on this 😱

PR #3564, opened 2021-05-18
[CORL-1765] Fix: change abusive translation in report

alinhle:
## What does this PR do?
- The spelling of 'behaviour' is British spelling currently, we are changing it to something that works across different versions of English.

### DIFF
--- a/src/locales/en-US/stream.ftl
+++ b/src/locales/en-US/stream.ftl
@@ -566,7 +566,7 @@ comments-reportPopover-reportThisComment = Report This Comment
 comments-reportPopover-whyAreYouReporting = Why are you reporting this comment?
 
 comments-reportPopover-reasonOffensive = This comment is offensive
-comments-reportPopover-reasonAbusive = This is abusive behaviour
+comments-reportPopover-reasonAbusive = This commenter is being abusive
 comments-reportPopover-reasonIDisagree = I disagree with this comment
 comments-reportPopover-reasonSpam = This looks like an ad or marketing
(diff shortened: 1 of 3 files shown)


src/locales/en-US/stream.ftl
```diff
@@ -566,7 +566,7 @@ comments-reportPopover-reportThisComment = Report This Comment
 comments-reportPopover-whyAreYouReporting = Why are you reporting this comment?
 
 comments-reportPopover-reasonOffensive = This comment is offensive
-comments-reportPopover-reasonAbusive = This is abusive behaviour
+comments-reportPopover-reasonAbusive = This commenter is being abusive
```
nick-funk, line 554: I think @wyattjoh indicated that we have to create a new key and not to just update the value of the old key

nick-funk (changes requested):

src/locales/en-US/stream.ftl
wyattjoh: Depends! If we can confirm that the previous english translation was used as the base for other languages, then we need to change the key. If not, then we can keep it the same.

alinhle: We don't need to introduce a new key, because the meaning of the translation has not changed at its core. The sole purpose of this translation change was to solve a specific English problem which does not apply to other languages.

nick-funk: Works for me

nick-funk (approved):

PR #3566, opened 2021-05-19
[CORL-1667] For Review Queue

nick-funk:
## What does this PR do?

Adds a for review queue to the moderation queues.

Adds a `flags(...)` query and also a `markFlagReviewed(...)` mutation to the schema.

## How do I test this PR?

- Enable `FOR_REVIEW` feature flag
- Create some comments
- Report them
- Head to Admin > Moderation > For Review
- See comments
- Mark them reviewed


alinhle (dismissed):

src/core/server/graph/resolvers/Flag.ts
```diff
@@ -25,4 +25,21 @@ export const Flag: GQLFlagTypeResolver<actions.CommentAction> = {
...
+    if (!comment) {
+      return null;
+    }
+
+    const revision = comment.revisions.find((r) => r.id === commentRevisionID);
+
+    // Returning a wrapped revision because this is the format
+    // that the `CommentRevision.ts` resolver definition is
+    // going to expect.
+    const wrappedRevision = { revision, comment };
```
alinhle: `const wrappedRevision: WrappedCommentRevision = ...`

src/core/server/models/action/comment.ts
```diff
@@ -278,20 +284,41 @@ export async function createActions(
...
+      break;
   }
 
-  // Return a connection.
-  return resolveConnection(query, input, (action) => action.createdAt);
+  if (input.filter) {
+    where = { ...tenantFilter, ...where, ...input.filter };
+  }
+
+  query.where(where);
```
wyattjoh: Subsequent calls to `.where(/* */)` on the `query` will actually do this merging for you:

https://github.com/coralproject/talk/blob/6986c4d5cd48766db79f5b6d4536349818e67796/src/core/server/models/helpers/query.ts#L38

src/core/server/models/action/comment.ts
```diff
@@ -300,14 +327,10 @@ export async function retrieveCommentActionConnection(
   input: CommentActionConnectionInput
 ): Promise<Readonly<Connection<Readonly<CommentAction>>>> {
   // Create the query.
-  const query = new Query(collection(mongo)).where({ tenantID });
+  let query = new Query(collection(mongo));
```
wyattjoh: Because subsequent calls to `.where` merge the query conditions, this particular change isn't needed.

src/core/server/stacks/reviewCommentAction.ts
```diff
@@ -0,0 +1,31 @@
+import { Db } from "mongodb";
+
+import { Tenant } from "coral-server/models/tenant";
+import { User } from "coral-server/models/user";
+import { commentActions } from "coral-server/services/mongodb/collections";
+
+const reviewCommentAction = async (
+  mongo: Db,
+  tenant: Tenant,
+  viewer: User | undefined,
```
wyattjoh: The viewer here can't be `undefined`, _someone_ reviewed the comment action! In the mutator (where this is called), because the linked mutation requires a logged in user with an admin/moderator role, you must use `ctx.user!` to cast it to a `User` from a `User | undefined`.

wyattjoh (changes requested):

src/core/server/models/action/comment.ts
nick-funk: Ooooh! Neat!

src/core/server/stacks/reviewCommentAction.ts
nick-funk: 👍

wyattjoh (commented):
Few minor comments! Not blocking!

src/core/client/ui/components/v2/Timestamp/Timestamp.tsx
```diff
@@ -12,31 +10,22 @@ export interface TimestampProps {
...
-    </BaseButton>
+    <TimestampFormatter
+      className={className}
+      rootClassName={styles.root}
+      textClassName={styles.text}
+      toggleAbsolute={toggleAbsolute}
+      onToggleAbsolute={onToggleAbsolute}
+    >
+      {children}
+    </TimestampFormatter>
```
alinhle: What is `TimestampFormatter` doing different than `<Timestamp>`?

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+        <Flex alignItems="center" justifyContent="space-evenly">
+          <CheckBox
+            checked={reviewed}
+            onChange={(event) => {
+              const enabled = !!(event.currentTarget.value === "on");
+              void onReview(id, enabled);
+            }}
+          >
+            {""}
+          </CheckBox>
```
alinhle: Maybe would be nice to move `onChange` to a `useCallback`? 
Why does `CheckBox` has an empty child `{""}`?

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+  return comment.revision.body || "";
+};
+
+const getModerationURL = (comment: ForReviewComment | null) => {
+  if (!comment) {
+    return "";
+  }
+
+  return `${getLocationOrigin()}/admin/moderate/comment/${comment.id}`;
+};
```
alinhle: Would `getModerationLink` already provide this functionality for you?

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+  comment: ForReviewComment | null;
+  username: string | null;
+  reason: COMMENT_FLAG_REASON | null;
+  additionalDetails: string | null;
+  createdAt: string;
+  reviewed: boolean;
+  onReview: (id: string, enabled: boolean) => void;
+}
+
+const getRevision = (comment: ForReviewComment | null) => {
```
alinhle: The name `getRevision` is inaccurate, as we are only returning the `body`. So `getRevisionBody` would be more accurate, but I think it should be just named `getCommentBody` instead.

src/core/client/ui/components/v3/ComponentLink/ComponentLink.tsx
```diff
@@ -0,0 +1,16 @@
...
+  href: string;
+  children: React.Component | Element | React.ReactNode;
+  className?: string;
+}> = ({ href, children, className }) => (
+  <a href={href} className={cn(styles.root, className)}>
+    {children}
+  </a>
+);
+
+export default ComponentLink;
```
alinhle: When do we need this component where `<TextLink>` and regular `<a>` won't suffice?

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+      <TableCell>
+        <TimestampFormatter>{createdAt}</TimestampFormatter>
+      </TableCell>
+      <TableCell>
+        {
+          <ComponentLink href={getModerationURL(comment)}>
+            <CommentParser className={styles.commentLink}>
+              {getRevision(comment)}
+            </CommentParser>
+          </ComponentLink>
```
alinhle: - Needs to be hooked up to `<Link>` from found, otherwise you'll bypass the router entirely.
- Should turn comments body into text before putting it in the table. We don't want links embedded in the comment, or other block elements to show up in the table.

Should be sth like:
```tsx
<Link
  as={TextLink}
  to={props.href}
  onClick={props.onClick}
>
  {getHTMLPlainText(getCommentBody(comment)).trim().substr(0, 50) || "No text content"}
</Link>
```

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
```diff
@@ -0,0 +1,292 @@
...
+>(
+  {
+    query: graphql`
+      fragment ForReviewQueueRoute_query on Query
+        @argumentDefinitions(
+          count: { type: "Int", defaultValue: 5 }
+          cursor: { type: "Cursor" }
+          storyID: { type: "ID" }
+          siteID: { type: "ID" }
+          section: { type: "SectionFilter" }
```
alinhle, line 165: Needs support for sites, stories, and sections.

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
```diff
@@ -0,0 +1,292 @@
...
+                onReview={onReview}
+              />
+            ))}
+          </TableBody>
+        </Table>
+        {relay.isLoading() && (
+          <Flex justifyContent="center">
+            <Spinner />
+          </Flex>
+        )}
```
alinhle: This is not required, `<AutoLoadMore>`  already shows a spinner and I see double spinners with this.

src/core/client/admin/components/Comment/CommentContent.tsx
```diff
@@ -59,48 +19,15 @@ const CommentContent: FunctionComponent<Props> = ({
...
-
-  return <>{children}</>;
+  return (
+    <CommentParser
+      phrases={phrases}
+      highlight={highlight}
+      className={cn(className, styles.root, highlight && styles.highlight)}
+    >
+      {children}
+    </CommentParser>
```
alinhle: I don't think we need `CommentParser` at all. It doesn't do anything different than `CommentContent` except it does not provide any default styles? Default styles can always be overwritten by passing a `className`.

alinhle (dismissed):

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
alinhle: Otherwise I end up here when I use the filters:
grafik

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+}
+
+interface ForReviewComment {
+  id: string;
+  revision: ForReviewRevision;
+}
+
+interface Props {
+  id: string;
+  comment: ForReviewComment | null;
```
alinhle: Comment is never `null` so we don't need to handle this case.

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+  if (!comment) {
+    return "";
+  }
+
+  if (!comment.revision) {
+    return "";
+  }
+
+  return comment.revision.body || "";
+};
```
alinhle: This could be implemented as: 
```ts
const getCommentBody = (comment: ForReviewComment) => comment?.revision?.body || ""; 
```

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
alinhle: Could also ask for what is actually needed instead of nesting an object.
```
commentID: string
commentBody: string
```

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
```diff
@@ -0,0 +1,292 @@
...
+              <ForReviewQueueRow
+                key={a.id}
+                id={a.id}
+                comment={{ ...a.comment, revision: a.revision }}
+                username={a.flagger ? a.flagger.username : ""}
+                reason={a.reason}
+                additionalDetails={a.additionalDetails}
+                createdAt={a.createdAt}
+                reviewed={a.reviewed}
+                onReview={onReview}
```
alinhle: Performancewise, this should actually be implemented as a `Container`. Otherwise it can get really slow the more rows are loaded.

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
alinhle: Would all change anyway if turned into a Container, see: https://github.com/coralproject/talk/pull/3566#pullrequestreview-668244008

src/core/client/ui/components/v2/Timestamp/Timestamp.tsx
nick-funk: Lack of css styling that makes it bolded/off-grey

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
nick-funk: Was not aware this existed, I copy pasted this functionality/usage from another part of admin... but maybe! (probably, lol)

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
nick-funk: Yeah, I have a patch incoming, found this earlier, hooking up now

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
alinhle: `React.memo()` would do this too, but it makes sense to split a container here to keep data requirements close to where it's needed.

src/core/client/ui/components/v2/Timestamp/Timestamp.tsx
alinhle: I see! I think the name is misleading though, because usually the name of the component suggests what will be rendered. In this case I would expect an interactive timestamp formatter UI to be rendered. It's not perfect but maybe we can do `<Timestamp raw />`, use the `variant` prop, or something else.

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+      <TableCell>
+        <ReasonText reason={reason} />
+      </TableCell>
+      <TableCell>{additionalDetails}</TableCell>
+      <TableCell>
+        <Flex alignItems="center" justifyContent="space-evenly">
+          <CheckBox
+            checked={reviewed}
+            onChange={(event) => {
+              const enabled = !!(event.currentTarget.value === "on");
```
alinhle: This seems to be always `true` no matter the state.

alinhle: Should be:
```ts
const enabled = event.currentTarget.checked;
```

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
nick-funk: Unfortunately not, that helper method requires a section, siteID, etc that we don't have and cannot easily know for these comments without adding a lot of checks. I did however simplify this logic.

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
nick-funk: Refactored this away during container refactor you requested elsewhere

src/core/client/ui/components/v3/ComponentLink/ComponentLink.tsx
nick-funk: Replaced and fixed

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRoute.tsx
nick-funk: This is patched and fixed now

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
nick-funk: Correct, fixed, patched, all refactored away with container

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
nick-funk: Refactored away with container and other schema to access revision easier

src/core/client/ui/components/v2/Timestamp/Timestamp.tsx
nick-funk: Fair enough, I think I'll just remove both of these format-less abstractions and override the classname instead

src/core/client/admin/routes/Moderate/Queue/ForReviewQueueRoute/ForReviewQueueRow.tsx
alinhle: Look again. These are all optional.

src/core/server/models/action/comment.ts
```diff
@@ -232,11 +242,19 @@ export async function createAction(
     ...rest,
   };
 
+  // Grab the story that we'll use to check moderation pieces with.
+  const story = await retrieveStory(mongo, tenantID, input.storyID);
+  if (!story) {
+    throw new StoryNotFoundError(input.storyID);
+  }
+
```
wyattjoh: Move this higher please!

wyattjoh (changes requested):

wyattjoh (approved):

src/core/client/admin/routes/Login/AccountCompletionContainer.tsx
```diff
@@ -106,7 +106,15 @@ const AccountCompletionContainer: FunctionComponent<Props> = ({
         if (redirectPath && !redirectPath.startsWith("/admin")) {
           window.location.href = redirectPath;
         } else {
-          router.replace({ pathname: redirectPath || "/admin" });
+          const pathname = redirectPath || "/admin";
```
tessalt, line 109: is this related to the rest of the PR?

tessalt (approved):

alinhle (approved):

PR #3581, opened 2021-06-08
include imported users in new user premod stage

tessalt:
## What does this PR do?

support the new user premod stage for imported users who do not have any approved comments.

### DIFF
--- a/src/core/server/services/comments/pipeline/phases/statusPreModerateNewCommenter.ts
+++ b/src/core/server/services/comments/pipeline/phases/statusPreModerateNewCommenter.ts
@@ -26,11 +26,17 @@ export const statusPreModerateNewCommenter = async ({
     return;
   }
 
+  // If the user is imported and the number of published comments on the user
+  // is greater than or equal to the threshold, OR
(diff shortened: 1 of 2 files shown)


src/core/server/services/comments/pipeline/phases/statusPreModerateNewCommenter.ts
```diff
@@ -26,11 +26,17 @@ export const statusPreModerateNewCommenter = async ({
...
   // If the number of approved comments on the user is greater than or equal to
   // the threshold, then there's nothing to do!
+
   if (
+    (author.importedAt &&
+      author.commentCounts.status.NONE >=
+        tenant.newCommenters.approvedCommentsThreshold) ||
     author.commentCounts.status.APPROVED >=
-    tenant.newCommenters.approvedCommentsThreshold
+      tenant.newCommenters.approvedCommentsThreshold
```
wyattjoh: I would probably suggest placing this on another block after this one as the other blocks are here to reduce the complexity in this `if` block.

wyattjoh (approved):

PR #3585, opened 2021-06-09
[CORL-1627] Show "new commenter" flags even when site wide premod is enabled

nick-funk:
## What does this PR do?

Move up the new-user premod check in pipeline

This allows the new user pre-mod filter to run
before it tries to block via a site wide pre-mod
filter. This lets the "new commenter" tag show
even if site wide pre-mod is turned on.

Otherwise, premod behaves as normally expected.

None

## How do I test this PR?

- Enable site wide premod
- Enable new commenter premod
- Make comments with new commenter
- See that they are marked with 'new commenter' in moderation
- Make comments with non-new commenter
- See that they are pre-modded because of site wide premod
- Turn site wide pre-mod off
- Make more comments with non-new commenter
- See that they are not pre-modded
- Approve the new commenter's comments so they are no longer a new commenter
- Make more comments with the formerly "new commenter"
- See that their comments are no longer pre-modded
- Turn off the new commenter pre-mod option
- Make yet another new commenter
- Make comments with this new commenter
- Make sure the comments are not pre-modded

### DIFF
--- a/src/core/server/services/comments/pipeline/phases/index.ts
+++ b/src/core/server/services/comments/pipeline/phases/index.ts
@@ -61,9 +61,9 @@ export const moderationPhases: IntermediateModerationPhase[] = [
   detectLinks,
 
   // Apply any pre-existing conditions to these comments.
+  statusPreModerateNewCommenter,
   statusPreModerate,
   statusPreModerateUser,
-  statusPreModerateNewCommenter,
 
   // Run any external moderation phase that missed other filters.
   external,


src/core/server/services/comments/pipeline/phases/index.ts
```diff
@@ -61,9 +61,9 @@ export const moderationPhases: IntermediateModerationPhase[] = [
   detectLinks,
 
   // Apply any pre-existing conditions to these comments.
+  statusPreModerateNewCommenter,
```
nick-funk, line 64: Moved up so that it will tag "new commenter" before it gets to the "site wide premod" check.

If site wide pre-mod is enabled, it will fall through to the site-wide pre-mod anyways, so this order doesn't hurt anything, it just allows for an earlier pre-mod check based on new commenter state prior to any site-wide pre-mod checks.

If you have any doubts about this, feel free to test using my testing steps, it should be good. Also, can just check these phases code to see the proof. :)

alinhle (approved):

tessalt (approved):

PR #3589, opened 2021-06-21
[CORL-1761] Show external moderation phases on the moderate card

nick-funk:
## What does this PR do?

Adds markers, "moderate by" state, and "automated actions" details to the comment moderation cards in the mod queues.

image

- Adds external moderation phase metadata to the comment's metadata

## How do I test this PR?

- Create an external moderation phase in Admin > Config > Moderation
  - Will require making a server to call an endpoint. I suggest using a quick express app with a POST endpoint to return some status like: 
    ```
      app.post("/mod", (req, res) => {
        res.send({
          actions: [
            {
              actionType: "FLAG",
              reason: "COMMENT_DETECTED_TOXIC",
            },
            {
              actionType: "FLAG",
              reason: "COMMENT_DETECTED_SPAM",
            },
          ],
        });
      });
    ```
- Create some comments as a non-admin/mod user
- See the comments get flagged with whatever you put into your mod phase and that the comments are marked with tags of the mod phases in the moderate card
- You can also force them into specific queues by making the mod result return `status: "REJECTED"` or similar


src/core/server/models/comment/revision.ts
```diff
@@ -45,6 +54,21 @@ export interface RevisionMetadata {
...
+   */
+  externalModeration?: {
+    name: string;
+    analyzedAt: Date;
+    status?: GQLCOMMENT_STATUS;
+    tags?: GQLTAG[];
+    actions?: {
+      type?: ACTION_TYPE;
+      reason?: FLAG_REASON;
+    }[];
```
wyattjoh: May I suggest nesting this one level deeper under `result` or something similar to distinguish it from the top level moderation properties?

nick-funk: Unsure what you mean here. The other results also do this same pattern? There's:

- perspective
- wordList
- akismet
- externalModeration

You want a results inside the externalModeration type? Or a results with an externalModeration inside?

If the former, then we should add a results subtype to perspective, wordList, akismet, etc.
If the latter, then we need to move akismet, wordList, perspective, etc inside results too.

Also, on the topic of "distinguish it from the top level moderation properties", that is kind of the point of this whole PR after talking with Sam/Andrew. 

The clients that use this want to treat the external moderation phases as 1st class mod phases. We add a lot of visibility in the mod-details and even merge them all under the same "Details" in the mod cards. Even going as far as to mark the mod card "moderated by" to show <external_mod_phase> instead of "System".

I don't think treating these as less than "top level" is accurate anymore.

nick-funk: Ah, after discussion, you meant `add a results subtype` to `externalModeration`. All good, it will be done!

src/core/client/stream/test/comments/permalink/permalinkView.spec.tsx
```diff
@@ -91,8 +91,9 @@ beforeEach(() => {
...
 /* TODO (cvle) - flaky */
 it("renders permalink view", async () => {
-  const tabPane = await waitForElement(() =>
-    within(testRenderer.root).getByTestID("current-tab-pane")
+  const tabPane = await waitForElement(
+    () => within(testRenderer.root).getByTestID("current-tab-pane"),
+    { timeout: 30000 }
   );
   expect(within(tabPane).toJSON()).toMatchSnapshot();
   expect(await within(testRenderer.root).axe()).toHaveNoViolations();
```
alinhle: You can undo this change. The flakyness is due to `axe()` for some reason. you can comment out the line with axe, or try updating axe packages.

src/core/client/admin/components/ModerateCard/ExternalModerationSummaryContainer.tsx
```diff
@@ -0,0 +1,249 @@
...
+        <Localized id="moderate-marker-toxic">
+          <Marker color="reported" key={keyCounter++}>
+            Toxic
+          </Marker>
+        </Localized>
+      );
+    default:
+      return null;
+  }
+};
```
alinhle: Could be a component :-)

src/core/client/admin/components/ModerateCard/ExternalModerationSummaryContainer.tsx
```diff
@@ -0,0 +1,249 @@
...
+                  {em.result.tags?.map(
+                    (
+                      t:
+                        | boolean
+                        | React.ReactChild
+                        | React.ReactFragment
+                        | React.ReactPortal
+                        | null
+                        | undefined
+                    ) => (
```
alinhle: Couldn't that just be:
```ts
                  {em.result.tags?.map(
                    (t) =>

```
?

src/core/client/admin/components/ModerateCard/externalModeration.ts
```diff
@@ -0,0 +1,110 @@
...
+          readonly actions: ReadonlyArray<{
+            readonly reason: GQLCOMMENT_FLAG_REASON_RL | null;
+          }> | null;
+        };
+      }> | null;
+    };
+  } | null;
+  readonly status: GQLCOMMENT_STATUS_RL;
+  tags: ReadonlyArray<CODED_TAG>;
+}
```
alinhle: Should be a fragment, also gives you types :-)

src/core/client/admin/components/ModerateCard/ModeratedByContainer.tsx
```diff
@@ -67,23 +67,49 @@ const ModeratedByContainer: React.FunctionComponent<Props> = ({
     onUsernameClicked(moderatedBy.id);
   }, [onUsernameClicked, moderatedBy]);
 
-  if (!moderatedBy) {
+  const externallyModeratedBy = comment.revision?.metadata?.externalModeration?.filter(
```
alinhle: This should be only set when `statusHistory.edges.length` is zero.

src/core/client/admin/components/ModerateCard/externalModeration.ts
```diff
@@ -0,0 +1,110 @@
...
+    (m: {
+      name: string;
+      result: {
+        actions: ReadonlyArray<{
+          readonly reason: GQLCOMMENT_FLAG_REASON_RL | null;
+        }> | null;
+        status: GQLCOMMENT_STATUS_RL | null;
+        tags: GQLTAG_RL[];
+      };
+    }) => {
```
alinhle: `m` is already typed

src/core/client/admin/components/ModerateCard/externalModeration.ts
```diff
@@ -0,0 +1,110 @@
...
+        spam &&
+        revision.actionCounts.flag.reasons.COMMENT_DETECTED_SPAM >= spam.length
+      ) {
+        return true;
+      }
+
+      // Check if the status matches the external moderation phase status
+      if (
+        status === m.result.status &&
+        m.result.status !== GQLCOMMENT_STATUS.NONE
```
alinhle: That would only make sense when `statusHistory` array is empty

src/core/client/admin/components/ModerateCard/externalModeration.ts
```diff
@@ -0,0 +1,110 @@
...
+      ) {
+        return true;
+      }
+
+      // Check if the tags match the external moderation phase tags
+      if (
+        tags &&
+        m.result.tags?.every((t: GQLTAG_RL) =>
```
alinhle: `t` is already typed?

src/core/client/admin/components/ModerateCard/ExternalModerationSummaryContainer.tsx
alinhle: Also would make `keyCounter` unnecessary.

src/core/client/admin/components/ModerateCard/externalModeration.ts
```diff
@@ -0,0 +1,110 @@
...
+          }> | null;
+        };
+      }> | null;
+    };
+  } | null;
+  readonly status: GQLCOMMENT_STATUS_RL;
+  tags: ReadonlyArray<CODED_TAG>;
+}
+
+const filterValidExternalModItems = (comment?: Comment | null) => {
```
alinhle: Why do we need to filter for valid external mod phases?

alinhle (changes requested):

alinhle (approved):

PR #3593, opened 2021-06-22
[CORL-1655] Show trace ID's to resolve client error's easier

nick-funk:
## What does this PR do?

Creates a unique trace ID on every server request that can be attached to error messages, logged in the server logs, and shown to the user. The trace ID is oblique in that it tells the user nothing of what caused the error, but is an identifier they can provide to support staff that can easily be correlated to server logs to assist in resolving/debugging the issue.

None

## How do I test this PR?

- Throw an error somewhere in the code base
  - An easy one is to make the `body` resolver throw an error in `src/core/server/graph/resolvers/Comment.ts` (Ln: 56)
- Visit the site and trigger this error
- Should see a call out with the trace ID
- Check console running Coral and find matching `contextID` for the error
- Test this in both development and production builds

Example:
image


src/core/server/app/middleware/graphql/apolloServer.ts
```diff
@@ -31,6 +32,7 @@ function contextProvider(options: ContextProviderOptions) {
       tenant,
       logger,
       site,
+      traceID: uuidv4(),
```
wyattjoh: The `id` on this already is sent as the Trace ID in the `X-Trace-ID` header of the response.

src/core/server/app/middleware/graphql/apolloServer.ts
```diff
@@ -45,6 +47,9 @@ function contextProvider(options: ContextProviderOptions) {
       opts.clientID = clientID.slice(0, 36);
     }
 
+    // Set trace ID for debug events if we error
+    req.traceID = opts.traceID;
+
```
wyattjoh: This shouldn't be duplicated as it's already available on the request as `req.coral.id`.

src/core/server/graph/context.ts
```diff
@@ -116,6 +118,7 @@ export default class GraphContext {
     this.signingConfig = options.signingConfig;
     this.clientID = options.clientID;
     this.reporter = options.reporter;
+    this.traceID = options.traceID || "";
```
wyattjoh: This is already provided as `options.id` on the GraphContext.

src/core/server/graph/plugins/helpers.ts
```diff
@@ -108,7 +108,13 @@ function hoistCoralErrorExtensions(
 
   // Hoist the message from the original error into the message of the base
   // error.
-  (err as any).message = extensions.message;
+  const rawError = err as any;
+  rawError.message = extensions.message;
+  rawError.traceID = ctx.traceID;
+
+  if (err.extensions) {
+    err.extensions.traceID = ctx.traceID;
```
wyattjoh: Should be able to use `ctx.id` instead.

src/core/server/graph/plugins/logger.ts
```diff
@@ -45,7 +45,10 @@ export function logAndReportError(
   const report = ctx.reporter.report(getOriginalError(err) || err, scope);
 
   // Log that we reported an error.
-  ctx.logger.error({ err, report }, "graphql query error");
+  ctx.logger.error(
+    { err, report, traceID: ctx.traceID },
+    "graphql query error"
+  );
```
wyattjoh: Shouldn't need to add the `traceID` here, as `ctx.logger` already adds a `contextID` which is the same ID.

wyattjoh (changes requested):
Looks good! I'd just suggest that you should use the already existing trace ID rather than creating a second one. Happy to discuss!

src/core/server/types/express.ts
```diff
@@ -18,13 +18,15 @@ export interface CoralRequest {
...
   logger: Logger;
+  traceID?: string;
 }
 
 export type TenantCoralRequest = RequireProperty<CoralRequest, "tenant">;
 
 export interface Request<T = CoralRequest> extends ExpressRequest {
   coral: T;
   user?: User;
+  traceID?: string;
```
wyattjoh: These changes aren't needed now.

src/core/server/app/middleware/error.ts
```diff
@@ -126,6 +134,7 @@ function wrapAndReport(
     context: {
       error,
       report,
+      traceID: req.traceID,
```
wyattjoh: These should use `req.coral.id` instead of `req.traceID`

wyattjoh (changes requested):

wyattjoh (approved):
Server integration and client side integration looks nice! Excited to see Vinh's comment as well!

alinhle (changes requested):
Exact review is coming.

src/core/client/framework/lib/errors/unknownServerError.ts
```diff
@@ -12,6 +12,7 @@ interface UnknownErrorExtension {
 export default class UnknownServerError extends Error {
```
alinhle: `UnknownServerError`, `InvalidRequestError`, etc should inherit from a new class `TraceableError`.

```ts
class TraceableError extends error {
  public readonly traceID: string;
}
```

src/core/client/framework/lib/network/extractGraphQLError.ts
```diff
@@ -12,19 +12,48 @@ function isGraphQLErrorWithExtension(
...
-    return null;
+    return new MultiError(errors);
   }
   const err = errors[0];
   if (!isGraphQLErrorWithExtension(err)) {
     // No extensions == GraphQL error.
     // TODO: (cvle) harmonize with server.
-    return null;
+    return new MultiError(errors);
   }
```
alinhle: I just ran a thorough analysis and these lines actually should never happen.
When GraphQL Standard error occurs they always return a non 200 status code and as such will never hit this routine.

So we can replace line 47-57 with:
```ts
if (errors.length > 1 || !isGraphQLErrorWithExtension(errors[0])) {
  return new Error(`Unexpected errors: ${JSON.stringify(errors}`);
}
```

`MultiError` can then be removed.

src/core/server/graph/plugins/helpers.ts
```diff
@@ -108,7 +108,13 @@ function hoistCoralErrorExtensions(
...
   // Hoist the message from the original error into the message of the base
   // error.
-  (err as any).message = extensions.message;
+  const rawError = err as any;
+  rawError.message = extensions.message;
+  rawError.traceID = ctx.id;
+
+  if (err.extensions) {
+    err.extensions.traceID = ctx.id;
+  }
```
alinhle: We don't need to attach the traceID to the error directly. Extensions are always available so we can rely on that. This will be more spec compilant.

Should put implementation in `serializeExtensions` e.g. in line 107 it should say 
```ts
const extensions = originalError.serializeExtensions(bundle, traceID);
```

src/core/server/services/comments/pipeline/phases/index.ts
```diff
@@ -61,9 +61,9 @@ export const moderationPhases: IntermediateModerationPhase[] = [
   detectLinks,
 
   // Apply any pre-existing conditions to these comments.
+  statusPreModerateNewCommenter,
```
alinhle: This seems to be from another PR that I just reviewed?

src/core/client/framework/lib/errors/unknownServerError.ts
```diff
@@ -21,6 +22,17 @@ export default class UnknownServerError extends Error {
       Error.captureStackTrace(this, UnknownServerError);
     }
 
+    const traceID: string | null | undefined = (error as any).traceID as string;
+    const traceIDs: string[] | null | undefined = (error as any)
+      .traceIDs as string[];
```
alinhle: As suggested `traceIDs` doesn't need to be handled, as we can remove `MultiError` class.

src/core/client/framework/lib/errors/unknownServerError.ts
```diff
@@ -21,6 +22,17 @@ export default class UnknownServerError extends Error {
       Error.captureStackTrace(this, UnknownServerError);
     }
 
+    const traceID: string | null | undefined = (error as any).traceID as string;
```
alinhle: Instead of casting `error as any` add `traceID` to  type `UnknownErrorExtension` on line 4.

src/core/client/ui/components/v3/QueryError/QueryError.tsx
```diff
@@ -0,0 +1,58 @@
...
+    if (traceIDs && traceIDs.length > 0) {
+      return render(traceIDs.join(", "), error.message);
+    } else if (traceID) {
+      return render(traceID, error.message);
+    } else {
+      return <div>{error.message}</div>;
+    }
+  } catch {
+    return <div>{error.message}</div>;
+  }
```
alinhle: When following the other suggesions, we can use the common `TraceableError` class and also provide the same nice ui for errors without the traceID:
```suggestion
  let traceID = error instanceof TraceableError ? error.traceID : "";
  return (
    <CallOut
      color="error"
      title={
        <Localized id="common-error-title">
          <span>An error has occurred</span>
        </Localized>
      }
    >
      {traceID && (
        <>
          <Localized id="common-error-traceID">
            <div className={styles.heading}>Trace ID</div>
          </Localized>
          <div className={styles.section}>{traceID}</div>
        </>
      )}
      <Localized id="common-error-message">
        <div className={styles.heading}>Message</div>
      </Localized>
      <div className={styles.section}>{message}</div>
    </CallOut>
  );
```

alinhle (changes requested):

src/core/client/framework/lib/errors/invalidRequestError.ts
```diff
@@ -35,9 +36,11 @@ export default class InvalidRequestError extends Error
...
       Error.captureStackTrace(this, InvalidRequestError);
     }
+
     this.extension = extension;
     this.code = extension.code;
     this.id = extension.id;
+    this.traceID = (extension as any).traceID;
     this.param = extension.param;
     this.message = extension.message || extension.code;
   }
```
alinhle, line 46: See review of UnknownServerError

src/core/server/graph/plugins/helpers.ts
wyattjoh: That actually makes a lot of sense!

src/core/client/framework/lib/network/extractGraphQLError.ts
nick-funk: That's odd, I was easily able to make these happen with a simple error on a GraphQL resolver server side. Particularly common in dev builds.

src/core/server/graph/plugins/helpers.ts
nick-funk: This does not work. Also tried modifying the `serializeExtensions` method to receive the traceID and set it in the returned object, but that also does not work. Am fairly certain I need to set this here.

src/core/client/framework/lib/errors/unknownServerError.ts
nick-funk: Please see previous comment for why I created MultiError.

src/core/client/framework/lib/network/extractGraphQLError.ts
nick-funk: I removed it and things keep on chugging, going to keep it this way until I find a reason to re-add a `MultiError` or similar.

src/core/client/ui/components/v3/QueryError/QueryError.tsx
nick-funk: Done!

src/core/client/framework/lib/network/extractGraphQLError.ts
alinhle: > That's odd, I was easily able to make these happen with a simple error on a GraphQL resolver server side. Particularly common in dev builds.

Oh if you can provide an example, it would be super helpful!

src/core/server/graph/plugins/helpers.ts
alinhle: Hmm I just tried it and it worked, let's chat about this.

src/core/client/framework/lib/errors/invalidRequestError.ts
```diff
@@ -12,18 +14,20 @@ interface InvalidRequestExtension {
   message?: string;
   id?: string;
   param?: string;
+  traceID?: string;
```
alinhle: This is a required field and is always provided.

src/core/client/framework/lib/errors/invalidRequestError.ts
```diff
@@ -12,18 +14,20 @@ interface InvalidRequestExtension {
...
  * server.
  */
-export default class InvalidRequestError extends Error
+export default class InvalidRequestError extends TraceableError
   implements InvalidRequestExtension {
   // Keep extension of original server response.
   public readonly extension: InvalidRequestExtension;
   public readonly code: ERROR_CODES;
   public readonly id?: string;
+  public readonly traceID?: string;
```
alinhle: This is a required field and is always provided.

src/core/client/framework/lib/errors/networkOfflineError.ts
```diff
@@ -1,8 +1,10 @@
+import TraceableError from "./traceableError";
+
 /**
  * NetworkOfflineError thrown when an action that required internet connection
  * but network was offline.
  */
-export default class NetworkOfflineError extends Error {
+export default class NetworkOfflineError extends TraceableError {
```
alinhle: This is not a TraceableError.

src/core/client/framework/lib/errors/relayNetworkRequestError.ts
```diff
@@ -1,14 +1,17 @@
 import { RRNLRequestError } from "react-relay-network-modern/es";
 
+import TraceableError from "./traceableError";
+
 /**
  * RelayNetworkRequestError wraps Request errors thrown by Relay Network Layer.
  */
-export default class RelayNetworkRequestError extends Error {
+export default class RelayNetworkRequestError extends TraceableError {
```
alinhle: This is also not a TraceableError.

src/core/client/framework/lib/errors/unknownServerError.ts
```diff
@@ -1,17 +1,21 @@
...
 
 /**
  * UnknownServerError wraps any error returned from the
  * server that we don't know of.
  */
-export default class UnknownServerError extends Error {
+export default class UnknownServerError extends TraceableError {
   // Keep origin of original server response.
   public origin: UnknownErrorExtension;
+  public traceID?: string;
```
alinhle: This is a required field and is always provided.

src/core/client/framework/lib/errors/unknownServerError.ts
```diff
@@ -21,6 +25,12 @@ export default class UnknownServerError extends Error {
       Error.captureStackTrace(this, UnknownServerError);
     }
 
+    const traceID: string | null | undefined = error.traceID as string;
+
+    if (traceID) {
+      this.traceID = traceID;
+    }
+
```
alinhle: `this.traceID = error.traceID` should suffice

src/core/client/framework/lib/errors/unknownServerError.ts
```diff
@@ -1,17 +1,21 @@
+import TraceableError from "./traceableError";
+
 /**
  * Shape of the `UnknownError` extension.
  */
 interface UnknownErrorExtension {
   code: string;
+  traceID?: string;
```
alinhle: This is a required field and is always provided.

src/core/client/framework/lib/errors/traceableError.ts
```diff
@@ -0,0 +1,5 @@
+class TraceableError extends Error {
+  public readonly traceID?: string;
```
alinhle: This should be a required field.

src/core/client/ui/components/v3/QueryError/QueryError.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+    if (traceID) {
+      return render(traceID, error.message);
+    } else {
+      return <div>{error.message}</div>;
+    }
+  } catch {
+    return <div>{error.message}</div>;
+  }
+};
```
alinhle: I suggest rendering the same or similar nice UI no matter if traceID is available or not, see: https://github.com/coralproject/talk/pull/3593#discussion_r657894869.

src/core/client/framework/lib/network/extractGraphQLError.ts
alinhle: You were right, `MultiError` does exist 😱 I missed that. We can sync about this later.

alinhle (approved):

PR #3594, opened 2021-06-23
[CORL-2147] Track queue sort change events into metrics

nick-funk:
## What does this PR do?

When the user changes the sortedBy in the moderation queues, we fire an event with the eventEmitter that can be collected into our metrics.

None

## How do I test this PR?

- Start up app
- Go to mod queues
- Change sort by back and forth
- See event emitter logs in the Coral CLI output

### DIFF
--- a/src/core/client/stream/events.ts
+++ b/src/core/client/stream/events.ts
@@ -627,7 +627,7 @@ export const CancelAccountDeletionEvent = createViewerNetworkEvent<{
  * This event is emitted when the viewer changes the moderation
  * queue sort.
  */
-export const QueueSortChangedEvent = createViewerNetworkEvent<{
+export const QueueSortChangedEvent = createViewerEvent<{
   success: {
     sortOrder: string;
   };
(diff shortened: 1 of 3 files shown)


src/core/client/stream/events.ts
```diff
@@ -622,3 +622,17 @@ export const CancelAccountDeletionEvent = createViewerNetworkEvent<{
...
+ */
+export const QueueSortChangedEvent = createViewerNetworkEvent<{
+  success: {
+    sortOrder: string;
+  };
+  error: {
+    message: string;
+    code?: string;
+  };
+}>("queueSortChangedEvent");
```
alinhle: use `createViewerEvent` instead of `createViewerNetworkEvent` as this action does not cause a network request.

alinhle (changes requested):

alinhle (changes requested):
`createViewerEvent` and `createViewerNetworkEvent` has a slightly different API, after adapting the code, this PR is good to go!

alinhle (commented):
We don't have a way to collect metrics on the admin yet.

nick-funk:
Closing this as we don't have metrics Admin side and we don't have a need to remove the sorting. Instead, we need to make the DB handle this gracefully.

PR #3599, opened 2021-06-28
Fix bug leading to e-mail leak

okybr:
Talk version 4 exposes private information such as e-mail addresses to unauthenticated queries because of a simple bug. This pull-request fixes the bug and thus closes the e-mail leak.

### DIFF
--- a/graph/resolvers/util.js
+++ b/graph/resolvers/util.js
@@ -70,7 +70,7 @@ const wrapCheck = (
  * @param {Array<String>} permissions permissions that the user must have
  */
 const checkPermissions = (ctx, permissions) =>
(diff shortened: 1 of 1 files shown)


wyattjoh (approved):

munishsinghal (commented):
Even after these changes, it seems user with role ADMIN & MODERATOR can still see the email address in postman with below graphql query
query Q1 { user(id: "XXXX") { email }}
query Q2 { users(query: {}) { nodes { email }}

but user with role Staff, Commentor cannot see email address.

okybr:
@munishsinghal, that's how it's supposed to work.

PR #3602, opened 2021-06-29
[CORL-2077] Nonexistent ignoredUsers

marcushaddon:
## What does this PR do?

This PR addresses Jira ticket CORL-2077, in which users deleted via migration are present in existing users' `ignoredUsers` array.

None.

## How do I test this PR?

1. As user A, ignore existing user B.
2. Delete user B (manually, not through the API)
3. As user A, view your profile (there should be no error, and the deleted user should no longer


marcushaddon:
This PR fixes the issue breaking the "Blocking User"'s profile after the "Blocked User" has been deleted, but it also makes comments by the Blocked User reappear to the Blocking User.

The two fixes I can think of (with my limited understanding of the problem/Coral) are:
1. Verify whether or not there exist cases in which users deleted in migration exist in other users' `ignoredUsers` array AND comments by that deleted user still exist / made it through migration. If there aren't, then the fix in this PR should work.
2. Run a script to remove all non-existent users from all users' `ignoredUsers` array, and then patch migration/import tool to do this as well (more work, more permanent)
3. Rather than filtering out users that don't exist, always include at least their id field so that it can still be used to hide comments (much quicker, but the type of `User.ignoredUsers` would then be `({ id: string } | User)[]`)

src/core/server/graph/resolvers/User.ts
```diff
@@ -26,15 +24,15 @@ const maybeLoadOnlyIgnoredUserID = (
     return [];
   }
 
-  // Get the field names of the fields being requested, if it's only the ID,
-  // we have that, so no need to make a database request.
-  const fields = getRequestedFields<GQLUser>(info);
-  if (fields.length === 1 && fields[0] === "id") {
-    return users.map(({ id }) => ({ id }));
-  }
```
wyattjoh: This is an important optimization that should not be removed!

src/core/server/graph/resolvers/User.ts
```diff
@@ -7,16 +7,14 @@ import { roleIsStaff } from "coral-server/models/user/helpers";
...
   GQLUSER_ROLE,
   GQLUserTypeResolver,
 } from "coral-server/graph/schema/__generated__/types";
 
 import { RecentCommentHistoryInput } from "./RecentCommentHistory";
 import { UserStatusInput } from "./UserStatus";
-import { getRequestedFields } from "./util";
 
-const maybeLoadOnlyIgnoredUserID = (
+const maybeLoadOnlyIgnoredUserID = async (
```
nick-funk: What do you think of us renaming this to `loadAndFilterForExistingUsers`?

I know it sounds a little absurd, but we are indeed "loading the users" and then "filtering out null and errored users". Maybe I only like this since I come from agglutinated wording Java/C# land.

That's my only comment, other than that, this looks merge-able. Approved!

nick-funk (approved):

src/core/server/graph/resolvers/User.ts
marcushaddon: @wyattjoh I realized that we actually don't have to remove this optimization (the error doesn't arise from returning id's of nonexistent users, just from literally returning `null`) and have re-added it

PR #3624, opened 2021-07-12
[CORL-2177] Link to legacy docs site

tessalt:
## What does this PR do?

Adds link to legacy documentation on the new documentation site.


wyattjoh:
Ideally we can use the versioned documentation dropdown.. But that's for later I think.

alinhle (approved):

PR #3625, opened 2021-07-12
[CORL-1637] Autohide scrollbar

marcushaddon:
## What does this PR do?

This PR hides the scroll bar in the Giphy widget when there are no results.

None.

## How do I test this PR?

1. Enable always showing scroll bars on OSX (https://heresthethingblog.com/2015/03/10/mac-tip-macs-scroll-bars/)
2. visit a comment thread
3. Click the `+ GIF` button
4. Before entering a gif search, observe that there is no scroll bar between the search box and the Giphy logo
5. Note that the Giphy logo is no longer stretched vertically in Safari or Firefox


alinhle (changes requested):
We use css-modules https://github.com/css-modules/css-modules that turns classnames into hashes to enforce uniqueness on classnames. (Read more https://css-tricks.com/css-modules-part-1-need/). 

During development, the name of the original file and the untransformed className is prefixed to the hash.

So by seeing `GiphyInput-grid-hash` it tells you that this className comes from a file `GiphyInput` with classname `.grid`. As we name the css files after the Component in which it is used (we don't reuse css files for other components, except by using `compose`) we know that it comes from a component called `GiphyInput`.

You find the definition of the style for it on line 43 and line 45 is the `overflow` property you want to modify.

alinhle (approved):

PR #3633, opened 2021-07-16
fix: pass requests with expired jwt

marcushaddon:
## What does this PR do?
This PR updates our JWTStrategy so that when a token is expired but otherwise valid, the request will be allowed to proceed as if it is an unauthenticated request (it may still later be rejected if the specific request required authorization. Essentially our auth layer punts on the issue).

None.

## How do I test this PR?
Let your JWT expire (I temporarily changed my tenants ttl to 10 seconds in the db) and then make a request. You should no longer receive a 401.

### DIFF
--- a/src/core/server/app/middleware/passport/strategies/jwt.ts
+++ b/src/core/server/app/middleware/passport/strategies/jwt.ts
@@ -101,7 +101,7 @@ export async function verifyAndRetrieveUser(
     // request rather than erroring out.
     if (
       err instanceof JWTRevokedError ||
-      err.jse_cause instanceof TokenExpiredError
+      err.cause() instanceof TokenExpiredError
     ) {
       return null;
     }
--- a/src/core/server/app/middleware/passport/strategies/jwt.ts
+++ b/src/core/server/app/middleware/passport/strategies/jwt.ts
@@ -1,4 +1,4 @@
-import jwt from "jsonwebtoken";
+import jwt, { TokenExpiredError } from "jsonwebtoken";
 import { Strategy } from "passport-strategy";
 
 import { AppOptions } from "coral-server/app";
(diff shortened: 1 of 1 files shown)


src/core/server/app/middleware/passport/strategies/jwt.ts
```diff
@@ -99,7 +99,10 @@ export async function verifyAndRetrieveUser(
   } catch (err) {
     // When the JWT was revoked, just indicate that there is no user on the
     // request rather than erroring out.
-    if (err instanceof JWTRevokedError) {
+    if (
+      err instanceof JWTRevokedError ||
+      err.jse_cause instanceof TokenExpiredError
```
nick-funk: Would you mind if we use `.cause()` here instead?

I suggest this because `jse_cause` and its older version `we_cause` is considered a private field and may change over time as we update node versions.

source: https://github.com/joyent/node-verror/issues/37

nick-funk (changes requested):

nick-funk:
Good work @marcushaddon!

@cvle, this will cause the user to see "You are not authorized to access that resource." if they try and write a comment with an expired session token.

image

This is cause we return a `USER_NOT_ENTITLED` error when they post a comment. What would you say to putting the following (or similar relay updates/refresh) in the `handleOnSubmit` for the `PostCommentFormContainer`?

```
// User's session likely expired, refresh page
if (error.code === ERROR_CODES.USER_NOT_ENTITLED) {
     window.location.reload();
}
```

I was thinking we might also do something with the `My Profile` tab query as well, that way we can reload if they click on it with an expired token. Would catch the two most common error points that a user would click on with an expired token.

Alternatively, @wyattjoh, what would you say to throwing a `SESSION_EXPIRED` error when we see this expired token server side and handle that client side instead of returning a `USER_NOT_ENTITLED` downstream somewhere in the server when they click on something?

alinhle:
Oh this is a very old issue and I remember there have been edge cases and disagreements, I need to put my head into this first.

tessalt:
Closing this PR until we can better research and scope out the work.

PR #3638, opened 2021-07-19
[CORL-1596] Move deleted comments into approve/rejected queues

nick-funk:
## What does this PR do?

When a user is deleted, the comments for that user are approved and rejected accordingly depending on whether they have replies or not.

None

## How do I test this PR?

- Create a user
- Create some comments for them
- Get some of the comments to be reported
- Ensure that several of the comments have replies from another commenter too
- Use the following mutation or set the account for deletion and delete the user

```
mutation ($userID: ID!, $clientMutationId: String!) {
  deleteUserAccount(input: {
    userID: $userID,
  }) {
    user {
      id
    }
  }
}
```

- Make sure the deletion mutation returns successfully
- Check the comment streams and queues
- The comments should either be deleted (no replies) or show a tombstone (has replies)
- The comments should be in the rejected queue (no replies) or in the approved queue (has replies)

### DIFF
--- a/src/core/server/services/users/delete.ts
+++ b/src/core/server/services/users/delete.ts
@@ -1,8 +1,16 @@
 import { Collection, Db } from "mongodb";
 
 import { ACTION_TYPE } from "coral-server/models/action/comment";
+import { getLatestRevision } from "coral-server/models/comment";
 import { Story } from "coral-server/models/story";
+import { retrieveTenant } from "coral-server/models/tenant";
 import collections from "coral-server/services/mongodb/collections";
+import { updateAllCommentCounts } from "coral-server/stacks/helpers";
+
+import { GQLCOMMENT_STATUS } from "coral-server/graph/schema/__generated__/types";
+
+import { moderate } from "../comments/moderation";
+import { AugmentedRedis } from "../redis";
 
 const BATCH_SIZE = 500;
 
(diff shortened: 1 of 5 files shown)


src/core/server/services/users/delete.ts
```diff
@@ -101,12 +109,99 @@ async function deleteUserActionCounts(
   });
 }
 
+async function moderateComments(
+  mongo: Db,
+  redis: AugmentedRedis,
+  tenantID: string,
+  filter: any,
```
wyattjoh: This can be typed! Add:

```ts
filter: FilterQuery<Comment>,
```

Where `FilterQuery` is from `coral-server/models/helpers`, and `Comment` is the comment model.

wyattjoh (changes requested):
The rest looks good! Think there's one refinement where we can use proper types for the argument.

src/core/server/services/users/delete.ts
```diff
@@ -101,12 +109,99 @@ async function deleteUserActionCounts(
   });
 }
 
+async function moderateComments(
+  mongo: Db,
+  redis: AugmentedRedis,
+  tenantID: string,
+  filter: FilterQuery<Comment>,
```
wyattjoh, line 116: You need to import the `Comment` interface from `coral-server/models/comment`, this is the global `Comment` interface.

wyattjoh (changes requested):

wyattjoh (approved):

PR #3643, opened 2021-07-21
fix: handle 'truthy value with falsey url' edgecase in media input

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2176 by handling the edge case in which the external media input has a non null Value, but that value has an empty `url` field. This occurs when the user clicks "Insert" before typing anything into the input.

None.

## How do I test this PR?
1. Enable external media
2. Click the "+" button on the comment form
3. click the "Insert" button
4. Observe that the input does not close


alinhle (changes requested):
It's not that final form is behaving wrong, but the validators are configured incorrectly. In `getMediaValidators.ts` the `hasExternalMediaAttached` condition should be `!!values.media && values.media.type === "external"` instead of `!!values.media && values.media.type === "external" && !!values.media.url` so that the validators can kick in. 

Feel free to reach out for a video call if you want to.

alinhle (approved):

PR #3644, opened 2021-07-23
[CORL-1693] send error payload in retry callback abort method

nick-funk:
## What does this PR do?

This will make the retry errors have a message that is
worth reading instead of obfuscating all these errors
with the default message of:

 "Aborted in beforeRetry() callback

This is a bit gnarly, as we're putting a stringified JSON
object into the abort message payload, but it's better than
no context coming through at all.

None

### DIFF
--- a/src/core/client/framework/lib/network/createNetwork.ts
+++ b/src/core/client/framework/lib/network/createNetwork.ts
@@ -81,15 +81,12 @@ export default function createNetwork(
         statusCodes: [500, 503, 504],
         beforeRetry: ({ abort, attempt, lastError }) => {
           if (attempt > 2) {
-            const payload = JSON.stringify({
-              note: "aborted in beforeRetry() callback.",
-              error: {
-                name: lastError?.name,
-                message: lastError?.message,
-                stack: lastError?.stack,
-              },
-            });
-            abort(payload);
+            let message = lastError?.message;
+            if (message && lastError?.name !== "RRNLRetryMiddlewareError") {
+              // Prefix with Error name.
+              message = `(${lastError?.name}) ${message}`;
+            }
+            abort(message);
           }
         },
       }),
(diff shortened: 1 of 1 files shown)


src/core/client/framework/lib/network/createNetwork.ts
```diff
@@ -79,9 +79,17 @@ export default function createNetwork(
...
-            abort();
+            const payload = JSON.stringify({
+              note: "aborted in beforeRetry() callback.",
+              error: {
+                name: lastError?.name,
+                message: lastError?.message,
+                stack: lastError?.stack,
+              },
+            });
+            abort(payload);
```
alinhle: I think `aborted in beforeRetry() callback.` is noise and we don't need to add the note. The abort error is always a `RRNLRetryMiddlewareError` this should already convey the _note_. Stack is also noise imo because we know it is a request problem not related to our code, also sentry can't help with source mapping the stack embedded in this string. 

What do you think of something like this which contains less noise?
```ts
            let message = lastError?.message;
            if (message && lastError?.name !== "RRNLRetryMiddlewareError") {
              // Prefix with Error name.
              message = `(${lastError?.name}) ${message}`;
            }
            abort(message);
          }
```

alinhle (commented):
It looks good! I have a comment about cutting some of the noise. What do you think of my remark?

src/core/client/framework/lib/network/createNetwork.ts
nick-funk: Works for me! I simply go with more verbosity over less when I don't know specifically what errors are being thrown. If you think this is enough, then I'm cool with it.

I just wanted to capture some fragment of `lastError` and pass it through.

I'll get this fixed up quick!

alinhle (approved):

PR #3645, opened 2021-07-26
[CORL-2205] Instead of throwing an error on mailer timeout, log a warning

nick-funk:
## What does this PR do?

Instead of throwing an error on mailer timeout and causing an alert in Slack, log a warning.

This way the mailer can gracefully try and resend the email in its next mailing operation.

None

## How do I test this PR?

It is very hard to test a mailer job timeout locally, but you can still test that this doesn't break mailing operations.

- Use https://docs.coralproject.net/development to set up the mailer testing environment with inbucket
- Create a user
- Request email notifications in the my profile configuration on stream
- Create some comments
- With another user, reply and respect some of the previous user's comments
- Wait for email summaries/notifications to show up in inbucket for the first user with notifications
 
 
## How do we deploy this PR?

Will need to update the incident reporting for mailer errors to match on this message:

```
jsonPayload.err.pvt.reason = "could not send email, not retrying"
```

instead of the previous reason/error name it was using to trigger mail failure incidents.

### DIFF
--- a/src/core/server/queue/tasks/mailer/processor.ts
+++ b/src/core/server/queue/tasks/mailer/processor.ts
@@ -14,7 +14,7 @@ import { JSDOM } from "jsdom";
 import { juiceResources } from "juice";
 import { camelCase, isNil } from "lodash";
 import { Db } from "mongodb";
-import timeoutPromiseAfter from "p-timeout";
+import timeoutPromiseAfter, { TimeoutError } from "p-timeout";
 
 import { LanguageCode } from "coral-common/helpers";
 import { Config } from "coral-server/config";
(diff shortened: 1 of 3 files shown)


src/core/server/queue/tasks/mailer/processor.ts
```diff
@@ -382,10 +382,16 @@ export const createJobProcessor = (
...
 
-      throw new WrappedInternalError(e, "could not send email");
+      if (e instanceof TimeoutError) {
+        // if we timed out, instead of erroring, log a warning and bail out
+        // so that the mailer job can try again later
+        log.warn({ err: e }, "sending email timed out");
+        return;
+      } else {
+        throw new WrappedInternalError(e, "could not send email");
+      }
```
wyattjoh: The mailer _must_ throw an error in order to have the job retried.

wyattjoh (changes requested):

wyattjoh (approved):

PR #3648, opened 2021-07-27
Feat/corl 2164 link to original commenter

marcushaddon:
## What does this PR do?
This PR will address https://vmproduct.atlassian.net/browse/CORL-2164 by causing the "In reply to..." text on replies to link the user back to the original comment. So far, this is successfully happening when the reply's parent is visible on the page (after Vinh's work on unread comments is merged, it will also cause the parent comment to be highlighted). 

If a comment's parent has not been loaded (ie it is "behind" a "Show more comments" button), the InReplyTo element should render as a non-actionable span element.

I do not believe there are any other edgecases that could cause a comment's parent to not be loaded, but if a user clicks this link and it is not found, it will be logged in Sentry.

Personal note: Currently the logic is "assume the comment's parent is loaded unless explicitly indicated otherwise". I might rename the property name to better reflect this.

None.

## How do I test this PR?
Find a comment that is a reply to another comment and click "In Reply to...". The page should snap to the parent comment (currently only if its on the page).
 
 
## How do we deploy this PR?
No special actions should be required.


alinhle:
I think we don't need to provide the jump to functionality inside of ConversationThread in the ancestor tree. The parent comment is right there (or sometimes after clicking on reveal) just above the comment.

alinhle:
There is also some overlap here with the work I'm doing on the Comment Seen feature as you already pointed out.

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
```diff
@@ -2,21 +2,53 @@ import { Localized } from "@fluent/react/compat";
...
-  );
+const InReplyTo: FunctionComponent<Props> = ({ parent, isLink }) => {
+  const { pym } = useCoralContext();
+
+  const navigateToParent = (id: string) => {
+    /* eslint-disable */
+    console.log('navigating to parent');
+    const elemId = computeCommentElementID(id);
+    const elem = document.getElementById(elemId);
+    console.log(elem);
```
nick-funk: Might want to remove the eslint disables and console.log's

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
```diff
@@ -2,21 +2,53 @@ import { Localized } from "@fluent/react/compat";
...
+    if (elem) {
+      void pym?.scrollParentToChildEl(elemId);
+      elem.focus();
+      (elem as any).style["border"] = "5px solid green";
+    } else {
+      /* eslint-disable-next-line */
+      console.log("TODO: its not on the page~");
```
nick-funk: More es-lint and console.log's that should probably be removed

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
```diff
@@ -2,21 +2,53 @@ import { Localized } from "@fluent/react/compat";
...
+  const navigateToParent = (id: string) => {
+    /* eslint-disable */
+    console.log('navigating to parent');
+    const elemId = computeCommentElementID(id);
+    const elem = document.getElementById(elemId);
+    console.log(elem);
+    if (elem) {
+      void pym?.scrollParentToChildEl(elemId);
+      elem.focus();
+      (elem as any).style["border"] = "5px solid green";
```
nick-funk: Thank you for using pym! This is one of the main things I was looking for! 💯

nick-funk (changes requested):

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
nick-funk: Make sure to run this border thing by design to check this is what they want. Super helpful for testing though!

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -80,6 +80,7 @@ import styles from "./CommentContainer.css";
 
 interface Props {
   viewer: ViewerData | null;
+  parentLoaded?: boolean;
```
cvle, line 83: I suggest renaming the property to _what it does_, e.g. `enableJumpToParent`

alinhle: I suggest renaming the property to _what it does_, e.g. `enableJumpToParent`. In your case because the `default` is `true`, I would actually call it `disableJumpToParent`.

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
```diff
@@ -2,21 +2,51 @@ import { Localized } from "@fluent/react/compat";
...
+    const elem = document.getElementById(elemId);
+    if (elem) {
+      void pym?.scrollParentToChildEl(elemId);
+      elem.focus();
+    } else {
+      globalErrorReporter.report(
+        `Assertion Error: Expected to find parent comment with id ${parent?.id} but could not`
+      );
+    }
+  };
```
alinhle: Should use `useCallback` here and it also seems unnecessary to pass in the `id` because you already got this from `parent.id`.

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
```diff
@@ -2,21 +2,51 @@ import { Localized } from "@fluent/react/compat";
...
 import { Flex, Icon } from "coral-ui/components/v2";
 
+import { CommentContainer_comment as CommentData } from "coral-stream/__generated__/CommentContainer_comment.graphql";
+
 import styles from "./InReplyTo.css";
 
 interface Props {
-  username: string;
+  parent: CommentData["parent"];
+  isLink: boolean;
```
alinhle: I suggest here as well to rename the property to what it does like before: e.g. `enableJumpToParent` or `disableJumpToParent` depending on how you want the default to be.

src/core/client/stream/tabs/Comments/PermalinkView/ConversationThreadContainer.tsx
```diff
@@ -213,6 +213,7 @@ const ConversationThreadContainer: FunctionComponent<Props> = ({
               <RejectedTombstoneContainer comment={comment}>
                 <DeletedTombstoneContainer comment={comment}>
                   <CommentContainer
+                    parentLoaded={false}
```
alinhle: this shouldn't be `false`, but in this case with the `parentLoaded` property: `remaining === 0`

alinhle (changes requested):

src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx
```diff
@@ -1,22 +1,55 @@
...
+
+  const navigateToParent = useCallback(() => {
+    if (!parent) {
+      return;
+    }
+    const elemId = computeCommentElementID(parent.id);
+    const elem = document.getElementById(elemId);
+    if (elem) {
+      void pym?.scrollParentToChildEl(elemId);
+      elem.focus();
```
nick-funk, line 34: ah, yes, that's what I forgot in my previous comment, to use pym to facilitate scrolling. Good job here!

nick-funk (approved):
Works well, I think this should be good to merge if we rebase after `commentSeen` stuff is merged into develop!

alinhle (approved):

src/core/client/stream/tabs/Comments/Comment/InReplyTo.css
```diff
@@ -2,6 +2,14 @@ $commentsInReplyToUsernameColor: var(--palette-text-500);
...
 $commentsInReplyToIconColor: var(--palette-text-100);
 
+button {
+  background-color: inherit;
+  border: none;
+  padding: 0;
+  margin: 0;
+  cursor: pointer;
+}
+
```
alinhle: That's global css, which we shouldn't be using here.

alinhle (changes requested):

alinhle (approved):

alinhle (approved):

PR #3649, opened 2021-07-28
Enable webhooks for COMMENT_CREATED events

marcushaddon:
## What does this PR do?
This creates the ability to fire webhooks in response to new comments being created.

`COMMENT_CREATED` was added as a valid value for `WEBHOOK_EVENT_NAME`.

## How do I test this PR?
1. In the admin view, create or configure a webhook. Under "Events to send" you should now see `COMMENT_CREATED` as an option. 
2. Check this and save it. 
3. Create a comment and observe that a server listening for POST requests at the address specified in the webhook receives a POST request.


src/core/server/events/listeners/webhook.ts
```diff
@@ -12,7 +12,10 @@ export type WebhookCoralEventListenerPayloads = StoryCreatedCoralEventPayload;
 export class WebhookCoralEventListener
   implements CoralEventListener<WebhookCoralEventListenerPayloads> {
   public readonly name = "webhook";
-  public readonly events = [CoralEventType.STORY_CREATED];
+  public readonly events = [
+    CoralEventType.STORY_CREATED,
+    CoralEventType.COMMENT_CREATED,
```
nick-funk, line 17: I think you may want to also listen for `COMMENT_REPLY_CREATED`? That or I think you can use `COMMENT_ENTERED` and it will trigger for both root comments and replies?

nick-funk (approved):

alinhle (changes requested):
This looks great! Can you also add docs to `WEBHOOKS.md`? Thanks!

WEBHOOKS.md
```diff
@@ -218,3 +226,64 @@ function when comparing signatures.
...
+     * siteID is the Site that the newly created Story was created on.
+     */
+    siteID: string;
+
+    /**
+     * commentID is the comment to which the new comment was a reply
+     */
+    commentID: string;
+  }
+  createdAt: string;
```
alinhle: As already mentioned, the comments should be adapted and we should probably double check the data that is being sent out.

alinhle (approved):

PR #3650, opened 2021-08-03
CORL-2122: remove point-events: none; from scroll container for modal

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2122 by fixing a bug where modals can't scroll. 

None.

## How do I test this PR?
Create a deep reply thread, then report a comment near the bottom. As an admin, click on the comment in "moderate" tab, then click "SHOW MORE OF THIS CONVERSATION" until the content extends past the height of the screen. The modal should be scrollable, while the rest of the UI remains locked. 

## How do we deploy this PR?
No special actions should need to be taken.


alinhle (approved):
I moved the backdropClick handler up to make that work again. Rest looks great!

PR #3653, opened 2021-08-04
Allow disabling rate limiting in production mode 

alinhle:
## What does this PR do?
- Allow disabling rate limiting in production mode e.g. for load testing on a staging environment


nick-funk (approved):

immber (approved):
👍

PR #3656, opened 2021-08-05
[CORL-2231] Make stream bundle work with old embed versions in cache

alinhle:
## What does this PR do?
- In older versions `embed.js` is put in local browser cache for 7 days. Our stream code might run into an issue where it would use a newer feature from `embed.js` that is not yet available.
- This PR adds backwards compatibility for the newer storage access.

## How do I test this PR?
- Find an old `embed.js`
- Build this branch
- Replace with old `embed.js`
- Run Coral with modified files
- See it working!


tessalt (approved):
clever!

nick-funk (approved):

PR #3657, opened 2021-08-05
feat: embed.js now reports its version to stream

alinhle:
## What does this PR do?
- `embed.js` now sends its version over to the stream bundle
- This lets us detect and handle situations where we encounter an old cached `embed.js`
 
## How do we deploy this PR?
- Maybe just add a `console.log` somewhere?


nick-funk:
I see no problem with just calling `console.log(embedVersion)` during the `initLocalStorage`. Go for it!

nick-funk (approved):

PR #3666, opened 2021-08-12
[CORL-2222/2223/2225] Add story archiving features to Coral

nick-funk:
## What does this PR do?

Adds the ability to archive stories to a secondary mongo instance.

Adds the mutations to archive stories and also adds the auto-archiving job features to let stories archive as they age and pass a certain age.

Adds two mutations to `archiveStories` and `unarchiveStories` along with their input's and payloads.

## How do I test this PR?

- Spin up another mongo instance to be your archiving Db
```
docker run -d -p 27018:27017 --restart always --name mongo-archive mongo:4.2
```
- Set the environment variable for this new archiving URI
```
MONGODB_ARCHIVE_URI=mongodb://127.0.0.1:27018/coral
```
- Find a story to archive and use the following mutation to do so:
```
mutation($storyIDs: [ID!]!, $clientMutationId: String!) {
  archiveStories(input: {
    storyIDs: $storyIDs,
  }) {
    stories {
      id
      isArchiving
      isArchived
    }
  }
}

// Query variables
{
	"storyIDs": [
		"<STORY_ID_YOU_WANT_TO_ARCHIVE>"
	],
	"clientMutationId": "0"
}
```
- Go check the story in stream and see that it is closed, archived, and is completely read-only
 
## How do we deploy this PR?

- If a client is not using archiving, you don't need to do anything.

- If you want to only use archiving and only use manual archive mutations on a client, you need to set the archive URI env variable on their regular and job deploy yaml:

```
MONGODB_ARCHIVE_URI
```
(if this is not set, it will default to the main mongodb URI.)

- If you want to use auto-archiving job features, you need to set the following on the job deploy yaml (or whichever instance will do archiving jobs).

```
ENABLE_AUTO_ARCHIVING <-- set to true to turn archive jobs on
AUTO_ARCHIVE_OLDER_THAN <-- optional, has a default
AUTO_ARCHIVING_BATCH_SIZE <-- optional, has a default
```


src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -643,7 +653,10 @@ export const CommentContainer: FunctionComponent<Props> = ({
                         onClick={toggleShowReplyDialog}
                         active={showReplyDialog}
                         disabled={
-                          settings.disableCommenting.enabled || story.isClosed
+                          settings.disableCommenting.enabled ||
+                          story.isClosed ||
+                          story.isArchived ||
+                          story.isArchiving
```
wyattjoh, line 682: Aren't all stories that are archived (or archiving) also closed?

src/core/client/stream/tabs/Configure/OpenOrCloseStream/OpenOrCloseStreamContainer.tsx
```diff
@@ -48,6 +48,10 @@ const OpenOrCloseStreamContainer: FunctionComponent<Props> = ({
     setShowSuccess(false);
   }, [setShowSuccess]);
 
+  if (story.isArchiving || story.isArchived) {
+    return null;
+  }
+
```
wyattjoh: Maybe instead it would be best to show a readonly button with a warning that "this story is archived"? Clarification from @losowsky obviously.

src/core/server/cron/accountDeletion.ts
```diff
@@ -14,6 +14,7 @@ import {
 
 interface Options {
   mongo: Db;
+  archive: Db;
```
wyattjoh: Why not take the combined Mongo context here?

src/core/server/data/context.ts
```diff
@@ -0,0 +1,28 @@
+import { Db } from "mongodb";
+
+import { AugmentedRedis } from "coral-server/services/redis";
+
+export interface MongoContext {
+  readonly main: Db;
```
wyattjoh: I almost like the `live` used in the options better than `main`? `live` is more meaningful.

src/core/server/data/context.ts
```diff
@@ -0,0 +1,28 @@
...
+  readonly archive: Db;
+}
+
+interface DataContextOptions {
+  mongoLive: Db;
+  mongoArchive: Db;
```
wyattjoh: You could possibly embed this a bit further to simplify the access:

```suggestion
  mongo: MongoContext;
```

src/core/server/graph/loaders/Comments.ts
```diff
@@ -190,21 +191,34 @@ export default (ctx: GraphContext) => ({
...
-        ...tagFilter(tag),
-        ...sectionFilter(ctx.tenant, section),
-        // If these properties are not provided or are null, remove them from
-        // the filter because they do not exist in a nullable state on the
-        // database model.
-        ...requiredPropertyFilter({ storyID, siteID, status }),
+  }: QueryToCommentsArgs) => {
+    let story: Readonly<Story> | null = null;
+    if (storyID) {
+      story = await retrieveStory(ctx.mongo, ctx.tenant.id, storyID);
```
wyattjoh: Should instead use:

```suggestion
      story = await ctx.loaders.Stories.story.load(storyID);
```

src/core/server/graph/mutators/Stories.ts
```diff
@@ -140,4 +149,56 @@ export const Stories = (ctx: GraphContext) => ({
...
+          ctx.tenant.id,
+          storyID
+        );
+      }
+
+      const result = await retrieveStory(ctx.mongo, ctx.tenant.id, storyID);
+      stories.push(result);
+    }
+
+    return stories;
```
wyattjoh: This logic should live in a new `archiving` service folder.

src/core/server/graph/mutators/Stories.ts
```diff
@@ -140,4 +149,56 @@ export const Stories = (ctx: GraphContext) => ({
...
```
wyattjoh: Same here, should live in a new `archiving` service folder.

src/core/server/data/context.ts
```diff
@@ -0,0 +1,28 @@
...
+  redis: AugmentedRedis;
+}
+
+export default class DataContext {
```
wyattjoh: It's common GraphQL lingo to refer to this as a `Connector`, so this should be called `Connectors`.

src/core/server/graph/context.ts
```diff
@@ -86,6 +89,8 @@ export default class GraphContext {
   public readonly signingConfig?: JWTSigningConfig;
   public readonly user?: User;
 
+  public readonly dataContext: DataContext;
```
wyattjoh: Following the feedback around renaming the `DataContext` class to `Connectors`, similarrly this would then become:

```suggestion
  public readonly connectors: Connectors;
```

src/core/server/graph/context.ts
```diff
@@ -104,6 +109,7 @@ export default class GraphContext {
     this.i18n = options.i18n;
     this.pubsub = options.pubsub;
     this.mongo = options.mongo;
+    this.archive = options.archive;
```
wyattjoh: If the connectors are being added to the context directly, you should remove the reference to these specifically.

src/core/server/index.ts
```diff
@@ -221,7 +224,19 @@ class Server {
...
+    // is, then use the same db for both mongo instances
+    if (
+      archiveMongoURI === config.default("mongodbArchive") &&
+      mongoURI !== config.default("mongodb")
+    ) {
+      archiveMongoURI = mongoURI;
+    }
+
+    this.mongo = await createMongoDB(mongoURI);
+    this.archive = await createMongoDB(archiveMongoURI);
```
wyattjoh: I would rather see this be a bit different! Right now, if the same URL is for both the archive and the mongo instance, we actually create 2X the connections to the same instance! Instead, I'd suggest the following logic:

```suggestion
    this.mongo = await createMongoDB(mongoURI);

    const mongoURI = config.get("mongodb");
    let archiveMongoURI = config.get("mongodbArchive");
    // if the archive URI isn't specified, but the regular mongo uri
    // is, then use the same db for both mongo instances
    if (
      archiveMongoURI === config.default("mongodbArchive") &&
      mongoURI !== config.default("mongodb")
    ) {
      this.archive = this.mongo;
    } else {
      this.archive = await createMongoDB(archiveMongoURI);
    }
```

src/core/server/models/comment/comment.ts
```diff
@@ -467,7 +511,7 @@ export const retrieveCommentRepliesConnection = (
  * @param pagination pagination options to paginate the results
  */
 export async function retrieveCommentParentsConnection(
-  mongo: Db,
+  mongo: MongoContext,
```
wyattjoh: I think for this function, it could additionally get a `isArchived: boolean` to help pass down to the `retrieveManyComments`.

src/core/server/models/comment/comment.ts
```diff
@@ -388,24 +396,54 @@ export async function editComment(
...
+    id,
+    tenantID,
+  });
+
+  return archivedComment;
 }
 
 export async function retrieveManyComments(
-  mongo: Db,
+  mongo: MongoContext,
```
wyattjoh: This should instead have a `isArchived` parameter to avoid the double pulling for this scenario.

src/core/server/models/comment/comment.ts
```diff
@@ -388,24 +396,54 @@ export async function editComment(
...
+  const liveCommentsCursor = collection(mongo.main).find({
     id: {
       $in: ids,
     },
     tenantID,
   });
 
-  const comments = await cursor.toArray();
+  const liveComments = await liveCommentsCursor.toArray();
+  if (liveComments && liveComments.length > 0) {
```
wyattjoh:
```suggestion
  if (liveComments.length > 0) {
```

src/core/server/models/comment/comment.ts
```diff
@@ -388,24 +396,54 @@ export async function editComment(
...
 }
 
 export async function retrieveManyComments(
-  mongo: Db,
+  mongo: MongoContext,
   tenantID: string,
   ids: ReadonlyArray<string>
 ) {
-  const cursor = collection(mongo).find({
+  // Try and find it in live comments collection
```
wyattjoh, line 406: Best to add a guard at the top to prevent unneeded database calls:

```ts
if (ids.length === 0) {
  return []
}
```

src/core/server/models/story/story.ts
```diff
@@ -727,3 +742,306 @@ export async function retrieveStorySections(
   // the null entries, the resulting array can not contain null.
   return results.filter((section) => section !== null).sort() as string[];
 }
+
```
wyattjoh, line 747: I would strongly suggest extracting this to a service function, and exposing the MongoDB methods as methods that are imported from each model module. Reaching between collections in a single model file should be avoided at all costs.

src/core/server/models/story/story.ts
```diff
@@ -727,3 +742,306 @@ export async function retrieveStorySections(
...
+  id: string,
+  now: Date
+) {
+  const result = await collection(mongo).findOneAndUpdate(
+    {
+      id,
+      tenantID,
+      $and: [
+        { $or: [{ isArchiving: { $exists: false } }, { isArchiving: false }] },
+        { $or: [{ isArchived: { $exists: false } }, { isArchived: false }] },
```
wyattjoh: This can be more compactly written:

```suggestion
        { isArchiving: { $in: [null, false] } },
        { isArchived: { $in: [null, false] } },
```

src/core/server/models/story/story.ts
```diff
@@ -727,3 +742,306 @@ export async function retrieveStorySections(
...
+  const multiplier = negate ? -1 : 1;
+
+  const result: RelatedCommentCounts = createEmptyRelatedCommentCounts();
+
+  if (commentCounts.action) {
+    for (const key in commentCounts.action) {
+      if (result.action.hasOwnProperty.call(commentCounts.action, key)) {
+        const value = commentCounts.action[key];
+        result.action[key] = value * multiplier;
+      }
```
wyattjoh: Best to invert this for more clarity:

```suggestion
      if (!Object.prototype.hasOwnProperty.call(commentCounts.action, key)) {
        continue;
      }
      
      const value = commentCounts.action[key];
      result.action[key] = value * multiplier;
```

src/core/server/models/story/story.ts
```diff
@@ -727,3 +742,306 @@ export async function retrieveStorySections(
...
+    },
+    {
+      returnOriginal: false,
+    }
+  );
+
+  return result.value;
+}
+
+const getCommentCounts = (options: {
```
wyattjoh: Best to move this into the `src/core/server/models/comment/counts/counts.ts` file as well.. Call it something like `negateCommentCounts`.

src/core/server/models/story/story.ts
wyattjoh: Apply the same changes below!

src/core/server/models/story/story.ts
```diff
@@ -727,3 +742,306 @@ export async function retrieveStorySections(
...
+  redis: AugmentedRedis,
+  tenantID: string,
+  id: string
+) {
+  const targetStory = await collection(mongo.main).findOne({ id, tenantID });
+  if (!targetStory) {
+    throw new StoryNotFoundError(id);
+  }
+
+  if (targetStory.isArchived) {
```
wyattjoh: Maybe also check if the story is currently already being archived?

src/core/server/models/story/story.ts
```diff
@@ -727,3 +742,306 @@ export async function retrieveStorySections(
...
+    );
+  }
+  if (
+    targetCommentModerationActions &&
+    targetCommentModerationActions.length > 0
+  ) {
+    await archivedCommentModerationActions(mongo.archive).insertMany(
+      targetCommentModerationActions
+    );
+  }
```
wyattjoh: I have a bit of an issue with how _all_ this data is loaded into memory and only then dumped out to each collection. I'd rather instead see something where you could batch these into more manageable chunks. This would be why the "comments" model could instead do something like:

```
cursor = < create cursor of comments to archive >
batch = []
for comment in cursor:
  batch.push(comment)
  if batch.length >= BATCH_MAX:
    archivedComments.insertMany(batch);
    batch = []
if batch.length > 0:
  archivedComments.insertMany(batch);
  batch = []
```

That way the documents are handed over in reasonable batches. `toArray` does this batching internally, but it just keeps going until the cursor is consumed.

wyattjoh (changes requested):

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
nick-funk: Yes

src/core/server/data/context.ts
nick-funk: I originally called it `live` and @cvle asked me to change it to `main` to avoid confusion. You two can decide what you prefer.

src/core/server/data/context.ts
nick-funk: Alright, this and your other comment makes me feel like I should go all in and refactor this context through the whole code base. No worries, I hear you and it will be done.

src/core/server/models/story/story.ts
nick-funk: Thanks, will update!

src/core/server/models/story/story.ts
nick-funk: 👍

src/core/server/models/story/story.ts
nick-funk: 💯 👍

src/core/server/models/story/story.ts
nick-funk: It will always be archiving cause of the way the `markStoryForArchiving` and `markStoryForUnarchiving` locks the story.

src/core/client/stream/tabs/Configure/OpenOrCloseStream/OpenOrCloseStreamContainer.tsx
nick-funk: Yeah, we're still waiting for Andrew to get back. This was brought up by myself and others in the "what status do we want to show for archiving?" and then we realized both Sam and Andrew were out, so we're sitting on this for now.

src/core/server/models/story/story.ts
nick-funk: This has been batched

src/core/server/app/middleware/passport/strategies/jwt.ts
```diff
@@ -60,8 +60,8 @@ export function createVerifiers(
 ): Array<Verifier<Token>> {
   return [
     new OIDCVerifier(options),
-    new SSOVerifier(options),
-    new JWTVerifier(options),
+    new SSOVerifier({ ...options, mongo: options.mongo.live }),
+    new JWTVerifier({ ...options, mongo: options.mongo.live }),
```
wyattjoh: Looks like you adjusted the `OIDCVerifier` but not these two.

src/core/server/graph/loaders/Comments.ts
```diff
@@ -190,21 +190,34 @@ export default (ctx: GraphContext) => ({
...
-        ...tagFilter(tag),
-        ...sectionFilter(ctx.tenant, section),
-        // If these properties are not provided or are null, remove them from
-        // the filter because they do not exist in a nullable state on the
-        // database model.
-        ...requiredPropertyFilter({ storyID, siteID, status }),
+  }: QueryToCommentsArgs) => {
+    let story: Readonly<Story> | null = null;
+    if (storyID) {
+      story = await retrieveStory(ctx.mongo.live, ctx.tenant.id, storyID);
```
wyattjoh: Should use the loader here:

```suggestion
      story = await ctx.loaders.Stories.story.load(storyID)
```

src/core/server/graph/mutators/Stories.ts
```diff
@@ -138,6 +151,77 @@ export const Stories = (ctx: GraphContext) => ({
...
+
+      const result = await retrieveStory(
+        ctx.mongo.live,
+        ctx.tenant.id,
+        storyID
+      );
+      if (result) {
+        stories.push(result);
+      }
+    }
```
wyattjoh: This logic should be moved to "coral-server/services/archive".

src/core/server/models/comment/comment.ts
```diff
@@ -353,7 +357,11 @@ export async function editComment(
   );
   if (!result.value) {
     // Try to get the comment.
-    const comment = await retrieveComment(mongo, tenantID, id);
+    const comment = await retrieveComment(
+      { live: mongo, archive: mongo },
+      tenantID,
+      id
+    );
```
wyattjoh: In this case, I'd suggest either making a separate function for `retrieveComment` that only accepts a single `mongo: Db`, as realistically, this is adding an extra DB call for an edge case. Then you could modify it to instead look like:

```ts
function isMongoContext(mongo: Db | MongoContext): mongo is MongoContext {
  return (
    typeof (mongo as MongoContext).live !== "undefined" &&
    (mongo as MongoContext).live instanceof Db &&
    typeof (mongo as MongoContext).archive !== "undefined" &&
    (mongo as MongoContext).archive instanceof Db
  );
}

export async function retrieveComment(
  mongo: MongoContext | Db,
  tenantID: string,
  id: string
) {
  const live = isMongoContext(mongo) ? mongo.live : mongo;
  const archive = isMongoContext(mongo) ? mongo.archive : null;

  const liveComment = await collection(live).findOne({ id, tenantID });
  if (liveComment) {
    return liveComment;
  }

  // If the passed mongo object wasn't a context, we have no archive to check!
  if (!archive) {
    return null;
  }

  const archivedComment = await archivedComments(archive).findOne({
    id,
    tenantID,
  });
  return archivedComment;
}
```

So you end up writing:

```suggestion
    const comment = await retrieveComment(mongo, tenantID, id);
```

Just an idea! I just don't like the idea of always checking the archive (for this use case) which for editing a comment, can't happen anyways.

src/core/server/cron/autoArchiving.ts
```diff
@@ -0,0 +1,109 @@
...
+  tenantCache,
+  config,
+}) => {
+  const enabled = config.get("enable_auto_archiving");
+  if (!enabled) {
+    log.info("cancelling auto archiving operation as it is not enabled");
+    return;
+  }
+
+  const amount = config.get("auto_archiving_amount");
```
wyattjoh: `BATCH_SIZE` is more correct.

src/core/server/cron/autoArchiving.ts
```diff
@@ -0,0 +1,109 @@
...
+  const age = config.get("auto_archiving_age");
```
wyattjoh: `auto_archive_older_than`

src/core/server/index.ts
```diff
@@ -221,14 +222,30 @@ class Server {
...
+
+    // If we have an archive URI, use it, otherwise, default
+    // to using the live database
+    let archive: Db | null = null;
+    const archiveURI = config.get("mongodb_archive");
+    if (
+      archiveURI === config.default("mongodb_archive") &&
+      liveURI !== config.default("mongodb")
+    ) {
+      archive = live;
```
wyattjoh: Upon further reflection, I'd argue that an archive feature that uses the same db connection isn't helpful. While it's useful in development to archive to the same db instance, it provides none of the expected performance benefits.

src/core/server/models/comment/comment.ts
wyattjoh: But actually move `MongoContext` use to _only_ services.

src/core/server/models/story/helpers.ts
```diff
@@ -67,6 +67,20 @@ export function getStoryClosedAt(
...
+) {
+  if (story.isArchived) {
+    return false;
+  }
+
+  return story.isArchiving || false;
+}
+
+export function isStoryArchived(story: Pick<Story, "isArchived">) {
+  return story.isArchived || false;
```
wyattjoh:
```suggestion
  return story.isArchived ?? false;
```

src/core/server/cron/autoArchiving.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  // For each of the tenant's, process their users notifications.
+  for await (const tenant of tenantCache) {
+    log = log.child({ tenantID: tenant.id }, true);
+
+    const now = new Date();
+    const dateFilter = new Date(now.getTime() - age);
+    const stories = await retrieveStoriesToBeArchived(
```
wyattjoh: Risk! By bulk fetching stories and marking them at once means if a failure occurs, you might be in a larger broken state than one at a time!

src/core/server/cron/autoArchiving.ts
```diff
@@ -0,0 +1,109 @@
...
+      mongo.live,
+      tenant.id,
+      dateFilter,
+      amount
```
wyattjoh:
```ts
for (let j = 0; j < amount; j++) {
```

And get one story at a time.

src/core/server/models/story/story.ts
```diff
@@ -727,3 +735,81 @@ export async function retrieveStorySections(
...
+      tenantID,
+      isArchiving: { $in: [null, false] },
+      isArchived: { $in: [null, false] },
+    },
+    {
+      $set: {
+        isArchiving: true,
+        closedAt: now,
+        updatedAt: now,
+        archivedAt: now,
```
wyattjoh: `startedArchivingAt` instead! Should add a `TODO` around handling partially archived stories.

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+  commentModerationActions,
+  comments,
+  stories,
+} from "coral-server/services/mongodb/collections";
+import { AugmentedRedis } from "coral-server/services/redis";
+
+const getCommentCounts = (options: {
+  story: Readonly<Story>;
+  negate: boolean;
+}) => {
```
wyattjoh:
```suggestion
const negateCommentCounts = (commentCounts: Readonly<RelatedCommentCounts>) => {
```

wyattjoh: Also, move to `src/core/server/models/comment/counts/counts.ts` please!

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+    story: { commentCounts },
+    negate,
+  } = options;
+  const multiplier = negate ? -1 : 1;
+
+  const result: RelatedCommentCounts = createEmptyRelatedCommentCounts();
+
+  if (commentCounts.action) {
+    for (const key in commentCounts.action) {
+      if (result.action.hasOwnProperty.call(commentCounts.action, key)) {
```
wyattjoh:
```suggestion
      if (Object.prototype.hasOwnProperty.call(commentCounts.action, key)) {
```

For the rest of this file please!

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+  const targetCommentActions = commentActions(mongo.live).find({
+    tenantID,
+    storyID: id,
+  });
+
+  logger.info("archiving comments");
+  const targetCommentIDs = await moveDocuments({
+    tenantID,
+    source: comments(mongo.live),
+    selectionCursor: targetComments,
```
wyattjoh: Can use https://github.com/coralproject/talk/blob/develop/src/core/server/models/helpers/query.ts#L12 instead!

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+    {
+      $set: {
+        isArchiving: false,
+        isArchived: true,
+      },
+    },
+    {
+      returnOriginal: false,
+    }
+  );
```
wyattjoh: Move to the story model!

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+    {
+      $set: {
+        isArchiving: false,
+        isArchived: false,
+      },
+    },
+    {
+      returnOriginal: false,
+    }
+  );
```
wyattjoh: Move to models!

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+  return result.value;
+}
+
+interface MoveDocumentsOptions {
+  tenantID: string;
+  source: Collection;
+  selectionCursor: Cursor;
+  destination: Collection;
+  returnMovedIDs?: boolean;
+}
```
wyattjoh:
```suggestion
interface MoveDocumentsOptions<T> {
  tenantID: string;
  source: Collection<T>;
  fulter: FilterQuery<T>;
  destination: Collection<T>;
  returnMovedIDs?: boolean;
}
```

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+
+const BATCH_SIZE = 100;
+const moveDocuments = async ({
+  tenantID,
+  source,
+  selectionCursor,
+  destination,
+  returnMovedIDs = false,
+}: MoveDocumentsOptions) => {
+  let insertBatch: any[] = [];
```
wyattjoh:
```suggestion
async function moveDocuments<T extends { id: string }>({
  tenantID,
  source,
  filter,
  destination,
  returnMovedIDs = false,
}: MoveDocumentsOptions<T>) {
  let insertBatch: T[] = [];
```

src/core/server/services/archive/index.ts
```diff
@@ -0,0 +1,328 @@
...
+
+  while (await selectionCursor.hasNext()) {
+    const document = await selectionCursor.next();
+    if (!document) {
+      continue;
+    }
+
+    insertBatch.push(document);
+
+    const hasID = Object.prototype.hasOwnProperty.call(document, "id");
```
wyattjoh: Not required now!

src/core/server/index.ts
```diff
@@ -221,14 +222,30 @@ class Server {
...
+    } else {
+      archive = await createMongoDB(archiveURI);
+    }
+
+    this.mongo = { live, archive };
```
wyattjoh: Move the creation logic inside the context, so you'll have a `createMongoContext(config)`

wyattjoh (dismissed):

src/core/server/cron/autoArchiving.ts
nick-funk: They're not being marked, but they are being queried out as a list. I'll make this walk a cursor instead.

src/core/client/admin/routes/Stories/StoryActions/ArchiveStoriesMutation.ts
```diff
@@ -0,0 +1,54 @@
...
+              isArchived: true,
+              isArchiving: false,
+            },
+          ],
+          clientMutationId: clientMutationId.toString(),
+        },
+      },
+      variables: {
+        input: {
+          ...input,
```
alinhle: I recommend not using the spread operator here. Historically a lot bugs was caused by this. For example someone passes more parameters than necessary into input and Typescript does not complain or a developer enriches the input with other data that is needed for handling the mutation but is not part of the mutation variables.

src/core/client/admin/routes/Stories/StoryActions/StoryActions.tsx
```diff
@@ -56,7 +64,7 @@ const StoryActions: FunctionComponent<Props> = ({
                   </DropdownButton>
                 </Localized>
               )}
-              {canClose && (
+              {canClose && !canUnarchive && (
```
alinhle: Check for !canUnarchive seems unnecessary. canClose means Story is open, so canUnarchive is always false, because story would be closed.

src/core/client/admin/routes/Stories/StoryActions/UnarchiveStoriesMutation.ts
```diff
@@ -0,0 +1,54 @@
...
+              isArchived: true,
+              isArchiving: false,
+            },
+          ],
+          clientMutationId: clientMutationId.toString(),
+        },
+      },
+      variables: {
+        input: {
+          ...input,
```
alinhle: See comment on other mutation on this.

src/core/client/stream/common/ModerateStream/ModerateStreamContainer.tsx
```diff
@@ -46,6 +49,19 @@ const ModerateStreamContainer: FunctionComponent<Props> = ({
...
+  if (isArchived || isArchiving) {
+    return (
+      <Marker color="warning" variant="filled">
+        <Icon size="sm" className={styles.icon}>
+          archive
+        </Icon>
+        <Localized id="general-archived">
+          <span>Archived</span>
+        </Localized>
+      </Marker>
```
alinhle: I think this is a good candidate to refector it out to outer scope.

e.g. 
```ts
const ArchiveMarker = () => (
<Marker color="warning" variant="filled">
        <Icon size="sm" className={styles.icon}>
          archive
        </Icon>
        <Localized id="general-archived">
          <span>Archived</span>
        </Localized>
      </Marker>
);

const ModerateStreamContainer: FunctionComponent<Props> = ({
[...]
  if (isArchived || isArchiving) {
    return <ArchiveMarker />;
  }
```

src/core/client/stream/tabs/Comments/Comment/ArchivedReportFlow/ArchivedReportFlowContainer.tsx
```diff
@@ -0,0 +1,111 @@
...
+      <Localized
+        id="comments-archivedReportPopover-doesThisComment"
+        a={
+          <a
+            href={`mailto:${settings.organization.contactEmail}?subject=${subject}&body=${emailBody}`}
+          >
+            {settings.organization.url}
+          </a>
+        }
+        $orgURL={settings.organization.url}
```
alinhle: grafik

Should that render the organisation email or at least the organisation name instead? Rendering the organisation url here, and then linking to a mail address is very confusing.

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -708,14 +722,17 @@ export const CommentContainer: FunctionComponent<Props> = ({
...
-        {showReportFlow && (
+        {showReportFlow && !story.isArchived && !story.isArchiving && (
           <ReportFlowContainer
             viewer={viewer}
             comment={comment}
             settings={settings}
             onClose={toggleShowReportFlow}
           />
         )}
+        {showReportFlow && story.isArchived && !story.isArchiving && (
```
alinhle: Shouldn't this say `showReportFlow && (story.isArchived || story.isArchiving)`? Genuinely asking.

src/core/client/stream/tabs/Comments/Comment/PermalinkButton/PermalinkCopyButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+  variant = "regular",
+  paddingSize = "small",
+  upperCase = false,
+}) => {
+  const emitCopyEvent = useViewerEvent(CopyPermalinkEvent);
+
+  const [copied, setCopied] = useState(false);
+
+  const handleCopy = useCallback(() => {
+    setCopied(true);
```
alinhle, line 54: I think we should add a timeout here that restores the `non copied` state for the user to click again if needed. 

The timeout should be set to around 800 to avoid issues with the permalink popover auto close timeout.

src/core/client/stream/tabs/Comments/Comment/PermalinkButton/PermalinkCopyButton.tsx
```diff
@@ -0,0 +1,84 @@
...
+    emitCopyEvent({ commentID });
+
+    onCopied();
+  }, [setCopied, emitCopyEvent, commentID, onCopied]);
+
+  return (
+    <CopyToClipboard text={permalinkURL} onCopy={handleCopy}>
+      <Button
+        className={cn(
+          CLASSES.sharePopover.copyButton,
```
alinhle: Props should accept `className` and `CLASSES.sharePopover.copyButton` should be passed by the Popover.

src/core/client/stream/tabs/Comments/Comment/PermalinkButton/PermalinkPopover.tsx
```diff
@@ -46,56 +43,24 @@ const PermalinkPopover: FunctionComponent<Props> = ({
...
 
   return (
     <Flex
       itemGutter="half"
       className={cn(styles.root, CLASSES.sharePopover.$root)}
     >
-      <CopyToClipboard text={permalinkURL} onCopy={handleCopy}>
-        <Button
-          className={cn(
-            CLASSES.sharePopover.copyButton,
```
alinhle, line 69: This className should be passed into the PermalinkcopyButton.

src/core/client/stream/tabs/Comments/Comment/ArchivedReportFlow/ArchivedReportFlowContainer.tsx
```diff
@@ -0,0 +1,111 @@
...
+          or spam? Send an email to our moderation team at org email address
+          with a link to this comment and a brief explanation.
+        </div>
+      </Localized>
+
+      <Localized id="comments-archivedReportPopover-needALink">
+        <div className={styles.heading}>Need a link to this comment?</div>
+      </Localized>
+
+      <PermalinkCopyButton
```
alinhle, line 80: add `className={CLASSES.reportPopover.copyLinkButton}`

src/core/client/stream/tabs/Configure/ArchivedConfigurationContainer.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+  return (
+    <div className={styles.root}>
+      <CallOut
+        color="warning"
+        title={
+          <Localized id="configure-archived-title">
+            This comment stream has been archived
+          </Localized>
+        }
```
alinhle: You could make this accessible by:
```suggestion
        color="warning"
        aria-labelledby="configure-archived-title"
        title={
          <Localized id="configure-archived-title">
            <div id="configure-archived-title">
              This comment stream has been archived
            </div>
          </Localized>
        }
```

src/core/client/stream/tabs/Configure/Configure.tsx
```diff
@@ -21,21 +22,43 @@ export interface Props {
...
+
+  if (isArchived) {
+    return (
+      <div>
+        <UserBoxContainer viewer={viewer} settings={settings} />
+        <ArchivedConfigurationContainer story={story} />
+      </div>
+    );
+  }
+
```
alinhle: This is just picky, but could be more dry if you want:

```ts
const UserBoxElement = <UserBoxContainer viewer={viewer} settings={settings} />;
```

and then use `{UserBoxElement}` for both occasions.

src/core/client/stream/tabs/Configure/OpenOrCloseStream/OpenOrCloseStreamContainer.tsx
```diff
@@ -100,6 +104,8 @@ const enhanced = withFragmentContainer<Props>({
     fragment OpenOrCloseStreamContainer_story on Story {
       id
       isClosed
+      isArchiving
+      isArchived
```
alinhle: Seems like changes in this file is no longer relevant, as we don't show this container anymore?

src/core/client/stream/tabs/Configure/UnarchiveStoriesMutation.ts
```diff
@@ -0,0 +1,54 @@
...
+              isArchived: true,
+              isArchiving: false,
+            },
+          ],
+          clientMutationId: clientMutationId.toString(),
+        },
+      },
+      variables: {
+        input: {
+          ...input,
```
alinhle: This spread is a comment source of bugs, I recommend explicitly specifying the input properties.

src/core/server/app/handlers/api/comment/featured.ts
```diff
@@ -61,20 +61,20 @@ export const featuredHander = ({
...
+      retrieveManyStories(mongo.live, tenant.id, ids)
     );
     const getAuthors = new DataLoader((ids: string[]) =>
-      retrieveManyUsers(mongo, tenant.id, ids)
+      retrieveManyUsers(mongo.live, tenant.id, ids)
     );
 
     const comments = await retrieveFeaturedComments(
-      mongo,
+      mongo.live,
```
alinhle: So featured comments are no longer available once the story has been archived?

src/core/server/app/middleware/passport/index.ts
```diff
@@ -38,7 +38,7 @@ export function createPassport(options: Options): passport.Authenticator {
   const auth = new Authenticator();
 
   // Use the LocalStrategy.
-  auth.use(createLocalStrategy(options));
+  auth.use(createLocalStrategy({ ...options, mongo: options.mongo.live }));
```
alinhle: Looks like all the other authors accept `MongoContext` instead, maybe we should adapt `createLocalStrategy` to behave similiarly as well?

src/core/server/app/handlers/api/story/ratings.ts
```diff
@@ -91,7 +91,7 @@ export const ratingsJSONPHandler = ({
       return respond(res, null);
     }
 
-    const ratings = await retrieveStoryRatings(mongo, tenant.id, story.id);
+    const ratings = await retrieveStoryRatings(mongo.live, tenant.id, story.id);
```
alinhle: I think anything related to ratings will break when the story is archieved.

src/core/server/app/handlers/api/user/download.ts
```diff
@@ -42,7 +42,7 @@ export const userDownloadHandler = ({
       const latestContentDate = new Date(iat * 1000);
 
       // Send the export down the response.
-      await sendUserDownload(res, mongo, tenant, user, latestContentDate);
+      await sendUserDownload(res, mongo.live, tenant, user, latestContentDate);
```
alinhle: Download GDPR data is broken as it does not include archived comments.

src/core/server/services/users/delete.ts
```diff
@@ -232,19 +253,27 @@ export async function deleteUser(
...
+    .findOne({ id: tenantID });
   if (!tenant) {
     throw new Error("could not find tenant by ID");
   }
 
   // Delete the user's action counts.
-  await deleteUserActionCounts(mongo, userID, tenantID);
+  await deleteUserActionCounts(mongo.live, userID, tenantID);
+  if (mongo.archive) {
+    await deleteUserActionCounts(mongo.archive, userID, tenantID);
```
alinhle: This should fail, as it would target the wrong collection.

src/core/server/services/users/delete.ts
```diff
@@ -262,5 +291,15 @@ export async function deleteUser(
...
 
+  // Delete the user's archived action counts.
+  if (mongo.archive) {
+    await deleteUserActionCounts(mongo.archive, userID, tenantID);
+  }
+
+  // Delete the user's archived comments.
+  if (mongo.archive) {
+    await deleteUserComments(mongo, redis, userID, tenantID, now, true);
+  }
```
alinhle: Looks like duplicate?

src/core/server/services/users/users.ts
```diff
@@ -1640,7 +1641,7 @@ export async function updateUserLastCommentID(
  * @param user the User that we're looking up the limit for
  */
 export async function retrieveUserLastComment(
```
alinhle: Mhh maybe call this `retrieveUserLastCommentNotArchived`, long name but clear...

src/core/server/stacks/createComment.ts
```diff
@@ -187,17 +188,21 @@ export default async function create(
...
 
   // Grab the story that we'll use to check moderation pieces with.
-  const story = await retrieveStory(mongo, tenant.id, input.storyID);
+  const story = await retrieveStory(mongo.live, tenant.id, input.storyID);
   if (!story) {
     throw new StoryNotFoundError(input.storyID);
   }
 
+  if (story.isArchiving || story.isArchived) {
+    throw new Error("Cannot create comments on an archive story");
```
alinhle: Would be great, to make this a standardized error in  `src/core/server/errors/index.ts`

src/core/server/stacks/editComment.ts
```diff
@@ -134,14 +134,18 @@ export default async function edit(
...
+    mongo.live,
     tenant.id,
     originalStaleComment.storyID
   );
   if (!story) {
     throw new StoryNotFoundError(originalStaleComment.storyID);
   }
 
+  if (story.isArchiving || story.isArchived) {
+    throw new Error("Cannot create comments on an archive story");
```
alinhle: Would be great, to make this a standardized error in  `src/core/server/errors/index.ts`

src/core/server/cron/autoArchiving.ts
```diff
@@ -0,0 +1,115 @@
...
+  const enabled = config.get("enable_auto_archiving");
+  if (!enabled) {
+    log.info("cancelling auto archiving operation as it is not enabled");
+    return;
+  }
+
+  const amount = config.get("auto_archiving_batch_size");
+  const age = config.get("auto_archive_older_than");
+
+  // For each of the tenant's, process their users notifications.
```
alinhle: Looks like a c&p comment

src/core/server/graph/loaders/Comments.ts
```diff
@@ -190,21 +192,34 @@ export default (ctx: GraphContext) => ({
...
-        // If these properties are not provided or are null, remove them from
-        // the filter because they do not exist in a nullable state on the
-        // database model.
-        ...requiredPropertyFilter({ storyID, siteID, status }),
+  }: QueryToCommentsArgs) => {
+    let story: Readonly<Story> | null = null;
+    if (storyID) {
+      story = await ctx.loaders.Stories.story.load(storyID);
+    }
+
```
alinhle, line 201: Maybe we should return an empty connection here when `story.isArchiving` is true? Because the data will be inconsistent.

alinhle: The frontend could also show a nice ui saying `apologies, this story is currently being archived`.

src/core/server/graph/loaders/Comments.ts
```diff
@@ -243,20 +258,33 @@ export default (ctx: GraphContext) => ({
...
-      after,
-      filter: {
-        // Filter optionally for comments with a specific tag.
-        "tags.type": tag,
+  ) => {
+    const story = await ctx.loaders.Stories.story.load(storyID);
+    if (!story) {
+      throw new Error("cannot get comments for a story that doesn't exist");
+    }
+
```
alinhle, line 279: Same here and the other occassions on this file

src/core/server/models/comment/comment.ts
```diff
@@ -389,24 +389,28 @@ export async function editComment(
...
   });
 
-  const comments = await cursor.toArray();
+  const items = await cursor.toArray();
+  if (items.length > 0) {
+    return ids.map((id) => items.find((comment) => comment.id === id) || null);
+  }
 
-  return ids.map((id) => comments.find((comment) => comment.id === id) || null);
+  return [];
```
alinhle: Why was this change made? I see logic that depends on the nulls in the array in other parts of the codebase.

src/core/server/models/comment/counts/counts.ts
```diff
@@ -226,3 +227,52 @@ export async function updateRelatedCommentCounts<
...
+  negate: boolean;
+}) => {
+  const { commentCounts, negate } = options;
+  const multiplier = negate ? -1 : 1;
+
+  const result: RelatedCommentCounts = createEmptyRelatedCommentCounts();
+
+  if (commentCounts.action) {
+    for (const key in commentCounts.action) {
+      if (Object.prototype.hasOwnProperty.call(commentCounts.action, key)) {
```
alinhle, line 242: Early return: I would do `if (!Object.prototype.hasOwnProperty.call(commentCounts.action, key)) { continue; })`. But this is fine too.

src/core/server/models/comment/counts/counts.ts
```diff
@@ -226,3 +227,52 @@ export async function updateRelatedCommentCounts<
 
   return result.value || null;
 }
+
+export const negateCommentCounts = (options: {
```
alinhle, line 231: I wonder if lodash has something for doing the negate easier.

src/core/server/models/story/helpers.ts
```diff
@@ -67,6 +67,20 @@ export function getStoryClosedAt(
...
 }
 
+export function isStoryArchiving(
+  story: Pick<Story, "isArchiving" | "isArchived">
+) {
+  if (story.isArchived) {
+    return false;
+  }
+
+  return story.isArchiving ?? false;
```
alinhle, line 77: Intersting , thanks for showing me the`??`. This is essentially the same as `Boolean(story.isArchiving)` in this case right?

src/core/server/models/story/story.ts
```diff
@@ -727,3 +735,124 @@ export async function retrieveStorySections(
...
+  storyID: string,
+  now: Date
+) {
+  const result = await collection(mongo).findOneAndUpdate(
+    { id: storyID, tenantID },
+    {
+      $set: {
+        isArchiving: false,
+        isArchived: true,
+        archivedAt: now,
```
alinhle, line 858: I wonder if we should set 

```ts
{
  unarchivedAt: null,
  startedUnarchivingAt: null,
}
```

src/core/server/models/story/story.ts
```diff
@@ -727,3 +735,124 @@ export async function retrieveStorySections(
...
+  storyID: string,
+  now: Date
+) {
+  const result = await collection(mongo).findOneAndUpdate(
+    { id: storyID, tenantID },
+    {
+      $set: {
+        isArchiving: false,
+        isArchived: false,
+        unarchivedAt: now,
```
alinhle, line 881: I wonder if we should set 

```ts
{
  archivedAt: null,
  startedArchivingAt: null,
}
```

src/core/server/services/comments/comments.ts
```diff
@@ -79,3 +110,174 @@ export async function removeTag(
...
+
+  if (liveComment) {
+    return liveComment;
+  }
+
+  if (mongo.archive && !skipArchive) {
+    const archivedComment = await retrieveCommentModel(
+      mongo.archive,
+      tenantID,
+      id
```
alinhle: This should not work, as it would access the wrong collection.

src/core/server/services/comments/comments.ts
```diff
@@ -79,3 +110,174 @@ export async function removeTag(
...
+    isArchived && mongo.archive
+      ? archivedComments(mongo.archive)
+      : comments(mongo.live);
+  return retrieveCommentParentsConnectionModel(
+    collection,
+    tenantID,
+    comment,
+    paginationParameters
+  );
+}
```
alinhle, line 404: I kinda like to see comments here for the new methods introduced.

src/core/server/services/mongodb/collections.ts
```diff
@@ -18,16 +18,26 @@ export const tenants = createCollection<Tenant>("tenants");
...
 export const sites = createCollection<Site>("sites");
 
 export const commentModerationActions = createCollection<
   CommentModerationAction
 >("commentModerationActions");
 
+export const archivedCommentModerationActions = createCollection<
+  CommentModerationAction
+>("archivedCommentModerationActions");
+
```
alinhle, line 40: It seems passing mongo around in the code now with two instances is still very awkward and error prone.

Here is a crazy idea: 
I would throughout the whole application only pass `MongoContext` instead of `Db`, everywhere. 

For this I would adapt `createCollection` that would accept also `MongoContext` instead of `Db` and then select the appropriate database automatically.

So the api would look like:
 ```ts
export const commentModerationActions = createCollection<
  CommentModerationAction
>("commentModerationActions");

export const archivedCommentModerationActions = createCollection<
  CommentModerationAction
>("archivedCommentModerationActions", { archive: true });
```

Now we essentially tie the collection to the specific db instance, which also reflects reality, and we don't ever have to care about knowing which mongo db instance to choose from throughout the entire codebase.

File: src/core/client/admin/routes/Stories/StoryActions/ArchiveStoriesMutation.ts [hunk 78]
```diff
@@ -0,0 +1,54 @@
...
+  (environment: Environment, input: MutationInput<MutationTypes>) =>
+    commitMutationPromiseNormalized<MutationTypes>(environment, {
+      mutation: graphql`
+        mutation ArchiveStoriesMutation($input: ArchiveStoriesInput!) {
+          archiveStories(input: $input) {
+            stories {
+              id
+              status
+              isArchived
+              isArchiving
```
alinhle, line 26: `isArchived` and `isArchiving` should be also added to the `CloseStoryMutation` and `OpenStoryMutation` so that in case they fail because of archiving, the status is immediately updated.

alinhle (changes requested):
This touched quite a bit of stuff, still, looking good so far. 

Before fixing other things, you might want to look at this comment here: https://github.com/coralproject/talk/pull/3666#discussion_r713517312
This could make things hell lot of less confusing and easier.

Some things I have noted while doing the review.
- Rating is probably broken
- Reactions are gone for achived story?
- Feature comments are gone for achived story?

src/core/client/stream/tabs/Configure/OpenOrCloseStream/OpenOrCloseStreamContainer.tsx
```diff
@@ -48,6 +48,10 @@ const OpenOrCloseStreamContainer: FunctionComponent<Props> = ({
     setShowSuccess(false);
   }, [setShowSuccess]);
 
+  if (story.isArchiving || story.isArchived) {
+    return null;
+  }
```
alinhle: This check seems now to be obsolete as this Component is longer rendered when story isArchived or in progress of archiving. If you wanna a check here, I'd recommend you to render an error instead.

nick-funk:
> This touched quite a bit of stuff, still, looking good so far.
> 
> Before fixing other things, you might want to look at this comment here: [#3666 (comment)](https://github.com/coralproject/talk/pull/3666#discussion_r713517312)
> This could make things hell lot of less confusing and easier.
> 
> Some things I have noted while doing the review.
> 
>     * Rating is probably broken
> 
>     * Reactions are gone for achived story?
> 
>     * Feature comments are gone for achived story?

Addressing some of your comments here:

"Rating is probably broken"
Try setting the `ENABLE_RATINGS_AND_REVIEWS` feature flag, R&R appears to be working fine on this branch. ~~I think the submission flow might be bugged though, so I'll have to fix that up and make it aware of archiving state.~~ Did some testing and archiving seems to work with ratings and reviews.

"Reactions are gone for achived story?"
Correct, you cannot react to an archived story.

"Feature comments are gone for achived story?"
~~Hmmm, these do seem to be missing. I'll take a quick peek, should be able to get this fixed.~~
This has been fixed, featured comments now show.

src/core/server/app/handlers/api/comment/featured.ts
nick-funk: This is a REST handler for the dashboard. It does not use archive because the dashboard does not use archived data.

src/core/server/app/handlers/api/story/ratings.ts
nick-funk: That is not the case and I have tested it and it works.

src/core/server/models/story/helpers.ts
nick-funk: Correct, it is the same in this case.

src/core/server/models/story/story.ts
nick-funk: It would seem like that's a good idea at first glance. However, we use the unarchivedAt as a filter to prevent auto-archiving from running on stories that admin's specifically decide they want to unarchive. Otherwise, whenever they unarchive a story, the auto-archiver would just archive it again in an hour or two.

src/core/server/models/story/story.ts
nick-funk: I'm not sure I like removing this metadata. It would prevent me from seeing when a story was archived and when it is un-archived. @wyattjoh largely asked me to add these in to aid in recover-ability and debugging should archiving issues crop up. Removing this data doesn't make the document more confusing in my mind. The reason being that if we have an `archivedAt` date that is less than the `unarchivedAt` date, we have all the data we need to grab both the archiving and un-archiving logs from Sentry/Cloud-Console. This seems like a good thing to keep.

src/core/client/stream/tabs/Configure/ArchivedConfigurationContainer.tsx
```diff
@@ -0,0 +1,84 @@
...
+          <Localized id="configure-archived-title">
+            <div id="configure-archived-title">
+              This comment stream has been archived
+            </div>
+          </Localized>
+        }
+        icon={
+          <Icon size="sm" className={styles.icon}>
+            archive
+          </Icon>
```
alinhle, line 44: Should add `container="section"`

src/core/server/app/handlers/api/story/ratings.ts
alinhle: Actually it is, the ratings become zero:
Screenshot 2021-10-05 at 21 44 33

src/core/server/services/users/delete.ts
```diff
@@ -36,25 +37,29 @@ interface Batch {
...
+    if (!isArchived) {
+      await executeBulkOperations<Story>(
+        collections.stories(mongo),
+        batch.stories
+      );
+      batch.stories = [];
+    }
   }
 
   const cursor = collections
```
alinhle: I think the collection needs to be switched here for archived

alinhle (changes requested):

src/core/server/services/users/delete.ts
```diff
@@ -36,25 +37,29 @@ interface Batch {
...
+
+    await executeBulkOperations<Comment>(comments(mongo), batch.comments);
     batch.comments = [];
 
-    await executeBulkOperations<Story>(
-      collections.stories(mongo),
-      batch.stories
-    );
-    batch.stories = [];
+    if (!isArchived) {
```
alinhle, line 53: Why don't we run this for `isArchived`?

src/core/server/services/comments/comments.ts
```diff
@@ -79,3 +109,175 @@ export async function removeTag(
...
+  if (ids.length === 0) {
+    return [];
+  }
+
+  const liveComments = await retrieveManyCommentModels(
+    mongo.comments(),
+    tenantID,
+    ids
+  );
+  if (liveComments.length > 0) {
```
alinhle: Should filter out null elements, before this check. (I know it's confusing as hell)

alinhle: Sth like `liveComments.filter(c => Boolean(c)).length > 0`

src/core/server/services/users/delete.ts
nick-funk: We don't process story count/state modifications on archived stories. We also don't apply `moderate(...)` to approve/reject comments that are deleted when archived cause it will screw up the mod queue counts. Andrew and Wyatt agreed that if a story is archived, it's done and no one is likely to go back and re-review a story that is older than the archiving limit.

src/core/server/services/comments/comments.ts
nick-funk: Yeah, found that when testing the permalink view, it's fixed up now.

src/core/server/services/users/delete.ts
nick-funk: Yeah, I fixed this up.

src/core/server/graph/loaders/Comments.ts
nick-funk: It now returns an empty connection

src/core/server/app/handlers/api/story/ratings.ts
nick-funk: Check now, should be fixed after recent changes.

src/core/client/stream/tabs/Comments/Comment/ArchivedReportFlow/ArchivedReportFlowContainer.tsx
```diff
@@ -84,6 +85,7 @@ const ArchivedReportFlowContainer: FunctionComponent<Props> = ({
         variant="outlined"
         paddingSize="extraSmall"
         upperCase
+        className={CLASSES.sharePopover.copyButton}
```
alinhle: This one needs its own CLASSES name.

alinhle: Like `CLASSES.reportPopover.copyLinkButton`

alinhle:
`Open Story` should not show, when Story is archived.

Screenshot 2021-10-05 at 21 37 25

Resolved

src/core/server/models/comment/comment.ts
```diff
@@ -939,9 +966,49 @@ export async function removeCommentTag(
...
+  if (mongo.archive && archivedStories.length > 0) {
+    archivedCounts = await retrieveStoryCommentTagCountsFromDb(
+      mongo,
+      tenantID,
+      archivedStories,
+      true
+    );
+  }
+
+  return [...liveCounts, ...archivedCounts];
```
alinhle: Counts must be returned back in original order, or counts will be switched.

src/core/server/queue/tasks/rejector.ts
```diff
@@ -31,16 +28,22 @@ export interface RejectorData {
...
+    mongo,
+    tenantID,
+    authorID,
+    {
+      orderBy: GQLCOMMENT_SORT.CREATED_AT_DESC,
+      first: 100,
+      after: connection ? connection.pageInfo.endCursor : undefined,
+    },
+    false
+  );
```
alinhle: Banning users should reject their archived comments as well. We can reuse the method we used for `deleteUser`. (Or can be done later)

alinhle (changes requested):

src/core/server/queue/tasks/rejector.ts
nick-funk: this is fixed up now :)

src/core/server/models/comment/comment.ts
nick-funk: This should be properly ordered and sorted now :D

alinhle (approved):

PR #3669, opened 2021-08-16
corl-2216-dont-submit-invalid-external-image

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2216 by creating a nested Form + Field (without rendering a nested <form> element) around the ExternalImageInput in order to verify the validity of the pasted image url before alerting the actual parent Form of it's presence. This addresses an edge case in which a user pastes an invalid image url, putting the MediaField form into an invalid state from which there is no exit.

None.

## How do I test this PR?
In a new comment, paste a Twitter link (or any value that is either not a valid URL or not a URL of an image) into the external media widget (accessed via the '+' button). It should alert you that this is an invalid image link. Then paste the same link in the body of the comment, at which point you should be presented with a prompt to attach the Tweet to your comment.
 
 
## How do we deploy this PR?
There shouldn't be any special considerations for deploying this PR.


alinhle (changes requested):
That change doesn't really fix the underlying issue. The lines that have been removed are actually necessary because we don't want an existing media to be replaced during a paste.

The issue is a bit more complex though.

We actually don't want the inner media form to influence the post comment form as much as it currently does. An invalid value in the media form would block the submission of the comment form entirely and disables the media buttons. 

What we want is to separate the two better.

The media form should be handled by it's own FinalForm instance (without using another `<form>` tag because nested forms are invalid HTML), and only when everything was validated, it should set the `media` value on the outer `FinalForm` instance.

Ping me whenever you want me to go over the code with you.

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -55,39 +53,72 @@ const ExternalImageInput: FunctionComponent<Props> = ({ onSelect }) => {
...
-          </Flex>
-        </HorizontalGutter>
-      </HorizontalGutter>
-    </div>
+    <Form
+      onSubmit={() => {
+        onSelect(url);
+      }}
+    >
+      {({ handleSubmit }) => (
```
alinhle: I think you should consider using more of the capabilities of the Form library:
```ts
({
  handleSubmit,
  submitting,
  pristine,
  invalid
})
```

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -24,25 +31,16 @@ interface Props {
...
     }
   }, []);
 
-  const onClick = useCallback(() => {
-    onSelect(url);
-
-    // TODO: we should handle this state better...
-    setURL("");
-  }, [url, setURL, onSelect]);
+  const [url, setURL] = useState("");
```
alinhle: FinalForm does state handling already, that's what the library is for. No need to burden yourself with syncing the state.

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -55,39 +53,72 @@ const ExternalImageInput: FunctionComponent<Props> = ({ onSelect }) => {
...
-              >
-                Insert
-              </Button>
-            </Localized>
-          </Flex>
-        </HorizontalGutter>
-      </HorizontalGutter>
-    </div>
+    <Form
+      onSubmit={() => {
```
alinhle: You can access formdata through their passed parameter:

```
onSubmit={(formData) => {
```

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -55,39 +53,72 @@ const ExternalImageInput: FunctionComponent<Props> = ({ onSelect }) => {
...
+                          className={styles.input}
+                          onChange={(evt) => {
+                            setURL(evt.currentTarget.value);
+                            input.onChange(evt);
+                          }}
+                          onKeyPress={onKeyPress}
+                          fullWidth
+                          variant="seamlessAdornment"
+                          color="streamBlue"
+                          ref={ref}
```
alinhle: Remove `name` and `onChange` and spread the `input` value into `TextField`:

```
<TextField
  {...input}
```

That's how you integrate with FinalForm.

alinhle: The current code produces an uncontrolled input that is manually synced, which is not desired.

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -55,39 +53,72 @@ const ExternalImageInput: FunctionComponent<Props> = ({ onSelect }) => {
...
+                    </HorizontalGutter>
+                    {meta.dirty && meta.error && (
+                      <CallOut
+                        color="error"
+                        title={meta.error}
+                        titleWeight="semiBold"
+                        icon={<Icon>error</Icon>}
+                        role="alert"
+                      />
+                    )}
```
alinhle: Use `ValidationMessage` from `framework` which is fully integrated into FinalForm

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -55,39 +53,72 @@ const ExternalImageInput: FunctionComponent<Props> = ({ onSelect }) => {
...
+                        />
+                        <Localized id="comments-postComment-insertImage">
+                          <Button
+                            color="stream"
+                            disabled={!!meta.error}
+                            onClick={(e) => {
+                              if (!url || meta.error) return;
+
+                              handleSubmit();
+                            }}
```
alinhle: This suggestion will integrate FinalForm.

```suggestion
                            disabled={submitting || invalid || pristine}
                            onClick={handleSubmit}
```

src/core/client/stream/tabs/Comments/ExternalImageInput/ExternalImageInput.tsx
```diff
@@ -55,39 +53,72 @@ const ExternalImageInput: FunctionComponent<Props> = ({ onSelect }) => {
...
+                      <Flex>
+                        <TextField
+                          name="externalImg"
+                          id="coral-comments-postComment-pasteImage"
+                          className={styles.input}
+                          onChange={(evt) => {
+                            setURL(evt.currentTarget.value);
+                            input.onChange(evt);
+                          }}
+                          onKeyPress={onKeyPress}
```
alinhle: This suggestion would redirect Enter to submitting this form instead of cancelling the enter entirely for a better user experience.

```suggestion
                          onKeyPress={(e: KeyboardEvent<HTMLInputElement>) => {
                            if (e.key !== "Enter") {
                              return;
                            }
                            handleSubmit();
                            e.preventDefault();
                          }}
```

alinhle (changes requested):
It seems that the code only partly integrated FinalForm.
For example state syncing is done manually and currently this also produces an uncontrolled field instead of a controlled one.

The changes I suggested are for fully integrating into FinalForm and let it handle the form and state handling.

Still it kinda works and is much better already!

alinhle (approved):

PR #3670, opened 2021-08-17
Feat/CORL-2213 remove story live update config

marcushaddon:
## What does this PR do?
This PR removes the option to toggle live updates via the config panel in the stream.

None.

## How do I test this PR?
While logged in as an admin, navigate to a stream's Configure tab, and see that there is no longer the option to toggle live updates for that particular story (they can still be toggled site wide).

## How do we deploy this PR?
This shouldn't require any special considerations.


marcushaddon:
@nick-funk @tessalt I just removed LiveUpdateConfig component from the stream config component, without deleting source for for LiveUpdateConfig. Let me know if I should delete all of it or if we have some way of identifying dead code.

src/core/client/stream/tabs/Configure/Configure.tsx
```diff
@@ -33,8 +33,6 @@ const Configure: FunctionComponent<Props> = (props) => {
       <AddMessageContainer story={props.story} />
       <QAConfigContainer story={props.story} settings={props.settings} />
       <HorizontalRule />
-      <LiveUpdatesConfigContainer story={props.story} />
-      <HorizontalRule />
```
nick-funk: Are there translations that might need to be removed from `stream.ftl` as well?

nick-funk (changes requested):

alinhle (changes requested):
Hey @marcushaddon, you can remove the files entirely. We can always take it back from this PR. Also there are quite a few unused translations. Can you find and remove them? Unfortunately we have do this all by hand at the moment.

nick-funk (approved):

src/locales/en-US/stream.ftl
```diff
@@ -694,26 +694,11 @@ configure-premodLink-commentsContainingLinks =
...
-configure-liveUpdates-description =
-configure-enableLiveUpdates-description =
-  When enabled, the comments will be updated instantly as new comments and
-  replies are submitted, instead of requiring a page refresh. You can
-  disable this in the unusual situation of an article getting so much
-  traffic that the comments are loading slowly.
-configure-enableLiveUpdates-enable = Enable
-
-configure-disableLiveUpdates-title = Disable live updates
 configure-disableLiveUpdates-description =
```
alinhle: Here is one left to be removed :-)

src/core/client/stream/tabs/Configure/LiveUpdatesConfig/LiveUpdatesConfigContainer.tsx
```diff
@@ -1,161 +0,0 @@
...
-  story: graphql`
-    fragment LiveUpdatesConfigContainer_story on Story {
-      id
-      settings {
-        live {
-          configurable
-          enabled
-        }
-      }
-    }
```
alinhle, line 157: The fragment `LiveUpdatesConfigContainer_story` is still being used, that's why we got all the build errors.

alinhle (approved):

nick-funk (approved):
Approving again

PR #3682, opened 2021-08-25
Fix Circle CI dependency failures

nick-funk:
## What does this PR do?

~~Update image to node:14 LTS and~~ disable sudo npm update -g
(Note: we will update to Node 14 later when we have updated Coral to run on Node 14)

We can't be using sudo to do npm updates as it modifies
the ~/.npm and node_modules/ ownership. Instead we need to
use the CircleCI images as-is and regularly update to the
latest LTS as they come out.

This will give us the security of a stable and secure npm
command as well as prevent permissions issues around the
dependencies breaking our builds and causing `npm ci` to
hang.

None

## How do I test this PR?

See working builds in CI from CircleCI.
 
## How do we deploy this PR?

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ job_node_defaults: &job_node_defaults
   <<: *job_defaults
   working_directory: ~/coralproject/talk
   docker:
-    - image: circleci/node:12
+    - image: circleci/node:14
   environment:
     <<: *job_node_environment
 
@@ -37,9 +37,10 @@ jobs:
             - v1-dependency-cache-{{ checksum "package-lock.json" }}
             # Find the most recently generated cache used from any branch
             - v1-dependency-cache-
-      - run:
-          name: Update NPM
-          command: sudo npm update -g npm
+      # Disabled as we don't want to use sudo to update packages
+      # - run:
+      #     name: Update NPM
+      #     command: sudo npm update -g npm
       # Disabled until there's capabilities to ignore a specific vun. Related:
       # https://npm.community/t/please-provide-option-to-ignore-packages-in-npm-audit/403/4
       # https://github.com/npm/cli/pull/10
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -37,9 +37,10 @@ jobs:
             - v1-dependency-cache-{{ checksum "package-lock.json" }}
             # Find the most recently generated cache used from any branch
             - v1-dependency-cache-
-      - run:
-          name: Update NPM
-          command: sudo npm update -g npm
+      # Disabled as we don't want to use sudo to update packages
+      # - run:
+      #     name: Update NPM
+      #     command: sudo npm update -g npm
```
nick-funk, line 43: Why are we disabling this?

We're not supposed to be updating npm globally in the CircleCI images. They build them specifically to have an updated LTS version of npm. If we want to update npm, we need to be changing the docker image to a newer image (which I've also done for this config to keep us up to date since we were still running node:12).

The next image will be `circleci/node:16` when that becomes the LTS for Node.js.

As well, this update has another bad side effect. It causes the `~/.npm` directory to become owned by the sudo user. This makes the pulling from the cache tied to `~/.npm` into our local `node_modules/` impossible and causes the command to hang.

So, we really don't want to be using this update command with sudo.

Lastly, why not just run the update without sudo? Cause the update fails cause the CircleCI does some funny permissions stuff in their particular install of Node.js on this docker image. Fun stuff...

wyattjoh: This does mean though that the image is running Node 14 though instead of 12.

tessalt (approved):

.circleci/config.yml
nick-funk: I have reverted the node:14 change to node:12 as requested by @wyattjoh because Coral is still running on Node 12 for the time being. We will update this in near future though.

wyattjoh (approved):

PR #3689, opened 2021-09-02
[CORL-2181] MobileToolbar and Z-Key

alinhle:
## What does this PR do?
- Introduce MobileToolbar and Z-Key functionality
- Prevent capturing unnecessary Keypresses
- Added class `coral coral-mobileToolbar`

## How do I test this PR?
- Use `Z` or `Shift+Z` to go to next unread.
- Use `Shift+A` to mark all visible comments as read
- Reduce screensize to trigger the mobile toolbar
- Keypresses should not trigger when the comment stream is not in view or you're currently on an inputfield even on the parent page.


src/core/client/ui/theme/breakpoints.ts
```diff
@@ -1,6 +1,7 @@
 const breakpoints = {
   xs: 320,
   mobile: 400,
+  mobileMax: 460,
```
nick-funk, line 4: iPhone XS Max naming has made it into the code-base (lol)

src/core/server/graph/schema/schema.graphql
```diff
@@ -564,6 +564,11 @@ enum FEATURE_FLAG {
   """
   COMMENT_SEEN
 
+  """
+  Z_KEY will allow Z key traversal and the mobile toolbar.
+  """
+  Z_KEY
```
nick-funk, line 570: Since this key may change in the future, it's a bit funny we're bringing SBN-isms into Coral. Would have maybe called this `ENABLE_KEYBOARD_COMMENT_TRAVERSAL` or something that describes what `Z_KEY` is doing. But the comment will suffice for now. We can fix this later.

nick-funk (approved):
This looks good. ✅

nick-funk (approved):

PR #3696, opened 2021-09-07
Fix dependencies timing out on Circle CI

nick-funk:
## What does this PR do?

image

Just kidding, there's more to it than green check marks. This counters the issue with CircleCI proxying calls to npm through their network and how that makes npm installs (or `npm ci`) slow. I've boosted the timeout for the installation to 20 minutes. Our builds are taking 13 minutes to install dependencies, so this should give some wiggle room in case something slows down even more.

In case you were wondering, "can we make the npm installs faster?"

Yes, we can, but that's a lot of work right now, and we need builds working today. We have a ticket logged to look into this a bit more robustly or even moving to GitHub Actions to resolve these slow install issues. `npm ci` only takes 5 minutes locally or 6 minutes via a Docker container build, so this is squarely a Circle CI -> NPM network issue.

If we want to continue to use CircleCI long-term, we might want to consider hosting our own package cache that we can pull from which we can ensure we won't get network slow-downs through.

None

## How do I test this PR?

Check the CircleCI builds and see that this branch isn't failing when everything else is failing.
 
## How do we deploy this PR?

Merge it into develop and let it run free!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,6 +52,7 @@ jobs:
       - run:
           name: Install dependencies
           command: sh .circleci/npmci.sh
+          no_output_timeout: 20m
       - save_cache:
           key: v1-dependency-cache-{{ checksum "package-lock.json" }}
           paths:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,9 +34,9 @@ jobs:
           keys:
             # Find a cache corresponding to this specific package-lock.json
             # checksum when this file is changed, this key will fail.
-            - v2-dependency-cache-{{ checksum "package-lock.json" }}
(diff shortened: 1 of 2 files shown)


alinhle (approved):

marcushaddon (approved):

tessalt (approved):

.circleci/config.yml
```diff
@@ -52,6 +52,7 @@ jobs:
       - run:
           name: Install dependencies
           command: npm ci
+          no_output_timeout: 20m
```
nick-funk, line 55: The default for this is 10 minutes. Our install on Circle CI takes ~13m30s, so this default would always cause our builds to fail due to a timeout. I have increased this to let the packages take the time they need because Circle's proxy connection to NPM is slow sometimes.

PR #3704, opened 2021-09-08
Feat/corl 2230 configurable flatten replies

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2230 and makes "Flatten Replies" configurable via the admin UI, as opposed to a feature flag.

FLATTEN_REPLIES is no longer a valid feature flag.

## How do I test this PR?
1. Create a reply chain that is long enough to be hidden behind a "Show more of this conversation" button. 
2. Under admin > configure > general, navigate to Flatten Replies and check "Yes"
3. back in the stream, refresh and confirm that all replies are shown.
 
## How do we deploy this PR?
No special actions should be necessary to deploy this change NOTE: Need to confirm this wont break tenants/instances where flatten replies are enabled as a feature flag / if we need some sort of migration.


src/core/client/admin/routes/Configure/sections/General/FlattenRepliesConfig.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <Localized id="configure-general-flattenReplies-title">
+        <Header container="h2">Flatten replies</Header>
+      </Localized>
+    }
+  >
+    <Localized
+      id="configure-general-flattenReplies-explanation"
+      strong={<strong />}
+    >
+      <FormFieldDescription> TODO: Get copy!</FormFieldDescription>
```
alinhle: It still says, TODO here.

src/core/server/graph/schema/schema.graphql
```diff
@@ -1916,6 +1909,11 @@ type Settings @cacheControl(maxAge: 30) {
   """
   closeCommenting: CloseCommenting!
 
+  """
+  flattenReplies contains settings related to hiding/showing all replies to a comment
+  """
+  flattenReplies: Boolean
```
alinhle: Maybe reuse the old comment?

``` 
"""
  flattenReplies when enabled, will cause all replies that would normally be
  hidden by a "Show more of this conversation" button to be displayed linearly
  sorted by date.
"""
```

src/core/client/stream/local/initLocalState.ts
```diff
@@ -75,7 +83,16 @@ const initLocalState: InitLocalState = async ({
...
 
-  const featureFlags = await determineFeatureFlags(environment, staticConfig);
+  /* eslint-disable */
+  console.log('INITTING LOCAL STATE', {
+    environment,
+    context,
+    auth,
+    staticConfig,
+    rest
+  })
```
alinhle: Leftover from development?

src/core/server/graph/schema/schema.graphql
alinhle: I think the type should be `Boolean!`, there is no need to keep the `null` state.

alinhle (changes requested):
Looks good overall. I've added some review comments.

There is also one missing change:
- The `settings.flattenReplies` resolver should default to the former feature flag `FLATTEN_REPLIES`
  - Look at the `amp` resolver in `src/core/server/graph/resolvers/Settings.ts` for an example

nick-funk (dismissed):
This feature works rather well. Nicely done. Fix up the small comments @cvle mentioned and this will be good to go!

src/core/server/models/tenant/tenant.ts
```diff
@@ -102,6 +103,12 @@ export interface TenantSettings
    */
   featureFlags?: Array<GQLFEATURE_FLAG | LEGACY_FEATURE_FLAGS>;
 
+  /**
+   * flattenReplies is whether the tenant wants replies to be hidden behind
+   * a "Show more of this conversation" link.
+   */
+  flattenReplies?: boolean;
+
```
alinhle: I think we should add to this file instead: `src/core/server/models/settings/settings.ts` where the `Settings` type is exported. There you find all the other settings as well.

src/core/server/graph/schema/schema.graphql
```diff
@@ -1910,7 +1910,9 @@ type Settings @cacheControl(maxAge: 30) {
   closeCommenting: CloseCommenting!
 
   """
-  flattenReplies contains settings related to hiding/showing all replies to a comment
+  flattenReplies when enabled, will cause all replies that would normally be
+  hidden by a "Show more of this conversation" button to be displayed linearly
+  sorted by date.
   """
   flattenReplies: Boolean
```
alinhle: https://github.com/coralproject/talk/pull/3704#discussion_r709537504

alinhle (changes requested):
Nearly done! Just one comment, and a remaining comment from the previous review.

src/core/client/admin/routes/Configure/sections/General/FlattenRepliesConfig.tsx
```diff
@@ -0,0 +1,55 @@
...
+  <ConfigBox
+    title={
+      <Localized id="configure-general-flattenReplies-title">
+        <Header container="h2">Flatten replies</Header>
+      </Localized>
+    }
+  >
+    <Localized
+      id="configure-general-flattenReplies-explanation"
+      strong={<strong />}
```
alinhle: Just remove this and we are done!

This one makes the `strong` tag available in the translation string, but it is not being used.
```suggestion
```

alinhle (changes requested):
Last thing and we're done!

alinhle (approved):

alinhle:
Just linting errors left.

PR #3711, opened 2021-09-09
[CORL-2289] Fix slow dependency install

alinhle:
## What does this PR do?
- replace material-design-icons with smaller package

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25854,21 +25854,21 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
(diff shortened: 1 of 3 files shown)


nick-funk (approved):

nick-funk (commented):
This is causing a lot of build failures on my local, investigating if it's my environment or something actually bad.

nick-funk (commented):
Nevermind, I got my enviro fixed, this is working good. Ready to merge.

Issue #3718, opened 2021-09-11
Duplicate email notifications for replies being generated when pre-moderation of comments is off


Sentynel:
**Expected behavior:**

Me again, sorry..! A user should be notified once, and only once, about a given reply to one of their comments. This notification should be sent, or queued for digests, at the point the comment becomes visible.

**Actual behavior:**

If comment pre-moderation is disabled, notifications for normal users are generated and sent twice: once when the comment is originally posted (and becomes visible), and once when a moderator marks the comment as approved, _even though this action does not change the visibility of the comment_.

Here is an example of a comment (irrelevant bits elided):
```
{
        "id": "38135553-f402-482f-8abe-bca76af23ec1",
        "createdAt": ISODate("2021-09-11T14:05:57.095Z"),
        "status" : "APPROVED",
}
```
Here is the corresponding commentModerationActions entry for that comment (trimmed again):
```
{
        "createdAt" : ISODate("2021-09-11T14:12:10.120Z"),
        "commentID" : "38135553-f402-482f-8abe-bca76af23ec1",
        "status" : "APPROVED"
}
```
Here are the relevant bits of the user whose comment this is a reply to:
```
{
        "notifications" : {                                                                                                 
                "onReply" : true,                                                                                           
                "onFeatured" : false,                                                                                       
                "onModeration" : false,                                                                                     
                "onStaffReplies" : false,                                                                                   
                "digestFrequency" : "HOURLY"                                                                                
        },
        "digests" : [
                {
                        "template" : {
                                "name" : "notification/on-reply",
                                "context" : {
                                        "commentPermalink" : "https://www.angrymetalguy.com/aborted-maniacult-review/?commen
tID=38135553-f402-482f-8abe-bca76af23ec1",
                                }
                        },
                        "createdAt" : ISODate("2021-09-11T14:05:57.171Z")
                },
                {
                        "template" : {
                                "name" : "notification/on-reply",
                                "context" : {
                                        "commentPermalink" : "https://www.angrymetalguy.com/aborted-maniacult-review/?commen
tID=38135553-f402-482f-8abe-bca76af23ec1",
                                }
                        },
                        "createdAt" : ISODate("2021-09-11T14:12:10.136Z")
                }
        ],
}
```
Two notifications, one generated at comment posting time, the other generated at comment moderation time.

**Related Issues:**

n/a

**Versions:**

- NodeJS: 14.17.6
- NPM: 6.14.15
- MongoDB: 3.6.8
[...]

PR #3719, opened 2021-09-11
Fix duplicate reply notification emails

Sentynel:
## What does this PR do?

Fixes receiving duplicate notifications of comment replies when a moderator approves a comment in post-moderation mode. See #3718.

None

## How do I test this PR?

In an instance configured with post moderation, create a comment, and enable onReply notifications for that account. From a different (non-staff) account, reply to that comment. Observe that a notification is generated (either via having an SMTP test server configured, or by enabling digests for the replied-to user and checking the user's digests field in the database). In the moderation panel, accept the reply. Observe that an additional notification is _not_ generated.
 
## How do we deploy this PR?

No special actions required.


alinhle:
@Sentynel This is awesome, thanks for your contribution!

alinhle (approved):

PR #3720, opened 2021-09-13
[CORL-2300] Stop highlighting comments longer than 500 characters

nick-funk:
## What does this PR do?

For really really long comments, this prevents the UI from hanging up trying to parse both a huge word list and a huge comment body for suspect/banned words.

None

## How do I test this PR?

- See CORL-2300 case for notes on the comment that prompted this bug fix. Make one.
- Report the comment
- Copy a huge banned/suspect word list from a big client
- Head to mod queue and see the mod queue load even with the giant comment body
    - No slowdowns or hangs
- Note that there is no highlighting on those long comments due to resource constraints on our RegEx
 
## How do we deploy this PR?

No special notes, will deploy same as any PR.

### DIFF
--- a/src/core/client/admin/helpers/markHTMLNode.ts
+++ b/src/core/client/admin/helpers/markHTMLNode.ts
@@ -35,6 +35,8 @@ function markPhrasesHTML(text: string, expression: RegExp) {
 // markHTMLNode manipulates the node by looking for #text nodes and adding
 // markers.
 export default function markHTMLNode(parentNode: Node, expression: RegExp) {
+  let exceededLength = false;
+
   parentNode.childNodes.forEach((node) => {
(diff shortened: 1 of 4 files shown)


src/core/client/admin/helpers/markHTMLNode.ts
```diff
@@ -1,7 +1,13 @@
+const LENGTH_CUTOFF = 500;
```
nick-funk, line 1: If you go any higher than this with a specific high profile client's banned/suspect word list, you will get massive slowdowns when running the regex powered `text.split(...)` over it...

nick-funk:
Closing as I have a better PR coming that will dramatically improve the performance here

PR #3721, opened 2021-09-14
[CORL-2251] Fix sign in bug

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2251 by displaying validation errors only when a field has actually been "dirtied" or if the form as been unsuccessfully submitted (previously they would display as soon as a field lost focus)

None.

## How do I test this PR?
Navigate to the stream while not logged in. Click the "Sign in" button, and without typing anything into the email field, click the top half of any of Facebook/Google sign in buttons, and observe that the modal begins the OAuth flow, rather than sliding out from under your pointer.
 
## How do we deploy this PR?
No special considerations should have to be taken.


src/core/client/framework/lib/form/helpers.tsx
```diff
@@ -97,11 +97,15 @@ export const formatStringList = (v: string[] | null) => {
...
 >;
 
-export const hasError = ({ touched, error, submitError }: FieldMeta) =>
-  touched && (error || submitError);
+export const hasError = ({
+  error,
+  submitError,
+  dirty,
+  submitFailed,
+}: FieldMeta) => (dirty || submitFailed) && (error || submitError);
```
alinhle: That's a nice improvement! Even better I think is leaving `dirty` out and only use `submitFailed && (error || submitError);`

alinhle (changes requested):

src/core/client/framework/lib/form/helpers.tsx
nick-funk: Yeah, this makes sense, no need to check for dirty. That seems like an overzealous "error" state haha.

Also, seconding that this is a nice fix @marcushaddon!

src/core/client/auth/views/SignIn/SignInWithEmail.tsx
```diff
@@ -50,7 +50,7 @@ const SignInWithEmail: FunctionComponent<SignInWithEmailForm> = (props) => {
               />
             )}
             <div className={cn(CLASSES.login.field, styles.field)}>
-              <EmailField disabled={submitting} autofocus />
+              <EmailField disabled={submitting} />
```
nick-funk: If you follow @cvle's suggestion and remove `dirty` from the `hasError` check, you can put this `autofocus` back in which improves keyboard usability by allowing folks to immediately start typing in the form when logging in with email.

nick-funk (dismissed):

src/core/client/auth/views/SignIn/SignInWithEmail.tsx
alinhle: Thanks @nick-funk  I thought I put out a similar comment, but I guess I didn't click on submit :-P

marcushaddon: ah i forgot to re-add that

src/core/client/framework/lib/form/helpers.tsx
marcushaddon: this works perfecty!

alinhle (approved):

nick-funk (approved):

PR #3724, opened 2021-09-14
[CORL-2300] Store and re-use detected words for mod card highlighting

nick-funk:
## What does this PR do?

Persists the detected words to the wordList metadata so that we can use it to highlight the moderate cards instead of re-doing the regex detection we already did.

Added banned and suspect words to the wordList metadata on comments.

## How do I test this PR?

- Add some suspect and banned words to your word lists
- Create comments with these words
- Check mod queue and see the words highlighted on these comments
 
## How do we deploy this PR?

Same as any feature, just deploy with the release from develop to main.


src/core/server/services/comments/pipeline/wordList.ts
```diff
@@ -1,19 +1,26 @@
...
   suspect: TestWithTimeout | false;
 }
 
 export type Options = Pick<Tenant, "id" | "locale" | "wordList">;
 
+// interface TestResult {
+//   isMatch: boolean | null;
+//   matches: string[];
+// }
+
```
nick-funk: remove

src/core/client/admin/components/Comment/CommentContent.tsx
```diff
@@ -34,9 +31,7 @@ const getSanitize: (highlight: boolean) => Sanitize = (() => {
           // Allow all RTE features to be displayed.
           features: ALL_FEATURES,
           config: {
-            FORBID_TAGS: highlight
-              ? ["b", "strong", "i", "em", "s", "span"]
-              : [],
+            FORBID_TAGS: [],
```
alinhle: Reminder to self: Review this change.

src/core/server/helpers/createTesterWithTimeout.ts
```diff
@@ -27,25 +44,58 @@ export default function createTesterWithTimeout(
...
+        // found a matched word
+        if (t % 4 === 2) {
+          const index = sumPriorLengths(tokens, t);
+          matches.push({
+            value: token,
+            index,
+            length: token.length,
+          });
+        }
+      }
```
alinhle, line 71: What do you think about moving this into a separate helper and write a test?

src/core/client/admin/helpers/markHTMLNode.ts
```diff
@@ -1,58 +1,101 @@
...
+  end: number;
+  replace?: boolean;
+}
+
 // markPhrasesHTML looks for `suspect` and `banned` words inside `text` given
 // the settings applied for the locale and highlights them by returning an HTML
 // string.
-function markPhrasesHTML(text: string, expression: RegExp) {
-  const tokens = text.split(expression);
+export default function markPhrasesHTML(
```
alinhle, line 12: This would be a good candidate for a unit test.

src/core/client/admin/components/ModerateCard/ModerateCard.tsx
```diff
@@ -13,6 +13,7 @@ import React, {
 import { MediaContainer } from "coral-admin/components/MediaContainer";
 import { HOTKEYS } from "coral-admin/constants";
 import { GetPhrasesRegExpOptions } from "coral-admin/helpers";
```
alinhle: Phrases are no longer used and should be removed.

alinhle: As well as from the props

src/core/server/services/comments/pipeline/createServerWordListRegEx.ts
```diff
@@ -0,0 +1,51 @@
...
+    )
+    // For each of these words, wrap a `|` or OR.
+    .join("|");
+
+  // Wrap the pattern in split rules. We want to match any word that either is
+  // at the start of a string, or a word boundary. The word must also either be
+  // at the end of the string or at another word boundary.
+  const pattern = `(^|${rule.boundary})(${words})($|${rule.boundary})`;
+
+  return new RE2(pattern);
```
alinhle: `coral-common/utils/createWordListRegExp.ts` should be moved to `src/core/server/services/comments/pipeline/helpers/createServerWordListReg.ts` and incorporate this change, as the original `createWordListRegExp` is not used anymore.

The existing tests should be also brought over and should be passing.
(They will fail, try and you'll see why)

Likewise `src/core/client/admin/helpers/getPhrasesRegExp.ts` is no longer used and should be removed.

alinhle (changes requested):

wyattjoh:
What happens if the wordlist changes? Do you re-process any of the comments?

alinhle:
@wyattjoh nah, we won't reprocess comments.

src/core/client/admin/helpers/markHTMLNode.ts
nick-funk: It is now tested. Have a look at `markHTMLNode.spec.ts` 🙂

src/core/client/admin/components/ModerateCard/ModerateCard.tsx
nick-funk: Removed and cleaned up

src/core/server/services/comments/pipeline/createServerWordListRegEx.ts
nick-funk: Tests have been moved over, are now passing, and the proper flags are set on the RE2 initialization to make this regex work properly.

src/core/server/helpers/createTesterWithTimeout.ts
nick-funk: I vote we leave it, the majority of this code is completely removed/refactored in the upcoming worker thread implementation for this.

src/core/client/admin/components/Comment/CommentContent.tsx
nick-funk: Yeah, it should work fine with replacement now cause all it does is a simple marking instead of a bunch of regex client side. Please let me know if you find a bug, I think we can always re-introduce this if we need to.

alinhle: Revert this change as a prerequisite to support catching words that are hidden like this:
```
<strong>s</strong>uspe<strong>ct</strong>
```

alinhle (changes requested):
As we just talked about in our meeting @nick-funk: The new wordlist system needs to detect suspect/banned words that are hidden in tags.

```
<em>s<strong>uspect</strong></em>
```

src/core/client/admin/components/Comment/CommentContent.tsx
nick-funk: This has been reverted, as well, I have made it so the server sanitizes the comment body prior to scanning for suspect/banned words the same way the client was previously doing here.

src/core/server/services/comments/pipeline/wordList.ts
```diff
@@ -97,19 +115,25 @@ export class WordList {
...
     }
 
     const timer = createTimer();
 
+    const sanitized = this.sanitizer(testString);
+
     // Test the string against the list and timeout if it takes too long.
-    const result = test(testString);
+    const result = test(sanitized.innerHTML);
     if (result === null) {
```
alinhle: This should not test for `result === null` but for `result.timedOut`

src/core/server/services/comments/pipeline/wordList.ts
```diff
@@ -97,19 +115,25 @@ export class WordList {
...
       logger.info(
-        { tenantID: options.id, listName, took: timer(), testString },
+        { tenantID: options.id, listName, took: timer(), sanitized },
```
alinhle: Should not log the whole HTML Node, but the `innerHTML`.

src/core/server/services/comments/pipeline/wordList.ts
```diff
@@ -97,19 +115,25 @@ export class WordList {
...
 
     const timer = createTimer();
 
+    const sanitized = this.sanitizer(testString);
+
     // Test the string against the list and timeout if it takes too long.
-    const result = test(testString);
+    const result = test(sanitized.innerHTML);
     if (result === null) {
       logger.info(
```
alinhle: While at it: I think this should be a `warning` instead.

alinhle (changes requested):

src/core/server/services/comments/pipeline/wordList.ts
nick-funk: Fixed, good catch from the legacy `null` return type for timed out results.

src/core/server/services/comments/pipeline/wordList.ts
nick-funk: Now logs the sanitized string only.

src/core/server/services/comments/pipeline/wordList.ts
nick-funk: Is now a warning.

alinhle (approved):

PR #3733, opened 2021-09-20
[CORL-604] Force admin local auth via feature flag

nick-funk:
## What does this PR do?

Adds a feature flag called `FORCE_ADMIN_LOCAL_AUTH` that forces the local authentication for the admin side of Coral on and disables the ability to turn it off.

It also shows a callout explaining why this is happening in the config settings so users know why it is greyed out.

Adds the feature flag `FORCE_ADMIN_LOCAL_AUTH`.

## How do I test this PR?

- Enable the feature flag `FORCE_ADMIN_LOCAL_AUTH`
- Go to Admin > Config > Authentication > Login with email authentication
- See that you can no longer toggle the Coral Admin option and it is enabled
- See that you cannot turn off Local Auth either 
 
## How do we deploy this PR?

- Enable the feature flag `FORCE_ADMIN_LOCAL_AUTH` for clients that need it


alinhle (changes requested):
To follow the existing patterns: The `LocalAuthConfig.tsx` should be turned into a FragmentContainer and ask for the necessary data itself. This will also fix the error that currently is happening when you try to save the form, saying that `featureFlags` is not part of the input.

There is a helper in `src/core/server/models/tenant/helpers.ts` that calls `hasEnabledAuthIntegration` that needs to handle the new feature flag too, otherwise passwort reset, linking, and so on won't work.

Other than that, looks good!

src/core/client/admin/routes/Configure/sections/Auth/LocalAuthConfigContainer.tsx
```diff
@@ -0,0 +1,99 @@
...
+          stream
+        }
+      }
+    }
+  }
+`;
+
+interface Props {
+  disabled?: boolean;
+  settings: any;
```
alinhle: Change this to `settings: LocalAuthConfigContainer_settings`, import the types and we can merge this! :-)

alinhle (approved):

PR #3737, opened 2021-09-22
[CORL-604] Make `FORCE_ADMIN_LOCAL_AUTH` into an env var config option

nick-funk:
## What does this PR do?

Make `FORCE_ADMIN_LOCAL_AUTH` into an env var config option

Was previously a feature flag, can now more easily be set across tenants via an environment variable.

Removes the previously added `FORCE_ADMIN_LOCAL_AUTH` feature flag in favour of a same named env var instead.

## How do I test this PR?

- Set `FORCE_ADMIN_LOCAL_AUTH=true` in your `.env`
- Go to Admin > Config > Authentication and see that the `Login with email authentication` is forced on and the local auth for Admin is forced on
 
 
## How do we deploy this PR?

- When deploying on clients that need local admin auth forced on, set `FORCE_ADMIN_LOCAL_AUTH` in the deployment yaml.

### DIFF
--- a/src/core/client/admin/local/initLocalState.ts
+++ b/src/core/client/admin/local/initLocalState.ts
@@ -48,7 +48,13 @@ const initLocalState: InitLocalState = async ({
     }
   }
 
-  await initLocalBaseState({ environment, context, auth, ...rest });
+  await initLocalBaseState({
+    environment,
+    context,
+    auth,
+    staticConfig,
+    ...rest,
+  });
 
   const modQueueSortOrder = await context.localStorage.getItem(
     MOD_QUEUE_SORT_ORDER
--- a/src/core/client/admin/local/initLocalState.ts
+++ b/src/core/client/admin/local/initLocalState.ts
@@ -17,6 +17,7 @@ const initLocalState: InitLocalState = async ({
   environment,
   context,
   auth = null,
+  staticConfig,
   ...rest
 }) => {
   let redirectPath = await context.localStorage.getItem(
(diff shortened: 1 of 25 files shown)


src/core/client/admin/local/initLocalState.ts
```diff
@@ -17,6 +17,7 @@ const initLocalState: InitLocalState = async ({
   environment,
   context,
   auth = null,
+  staticConfig,
```
wyattjoh, line 20: This `staticConfig` _must_ be passed to the later call to `initLocalBaseState`. TS is happy now because the type of the parameter is `StaticConfig | null | undefined`.

wyattjoh (changes requested):

src/core/client/admin/local/initLocalState.ts
nick-funk: No problem, is fixed, just updating some tests and this will be pushed up.

src/core/server/graph/resolvers/AuthenticationTargetFilter.ts
```diff
@@ -1,14 +1,11 @@
...
 import {
   GQLAuthenticationTargetFilter,
   GQLAuthenticationTargetFilterTypeResolver,
-  GQLFEATURE_FLAG,
 } from "coral-server/graph/schema/__generated__/types";
 
 export const AuthenticationTargetFilter: GQLAuthenticationTargetFilterTypeResolver<GQLAuthenticationTargetFilter> = {
-  admin: ({ admin }, _, { tenant }) => {
-    if (hasFeatureFlag(tenant, GQLFEATURE_FLAG.FORCE_ADMIN_LOCAL_AUTH)) {
+  admin: ({ admin }, _, { tenant, config }) => {
```
wyattjoh: `tenant` here isn't used!

wyattjoh (approved):

PR #3739, opened 2021-09-27
Fix/af za translations

HenkoR:
## What does this PR do?
Add missing translation on reply comment box placeholder for the af-ZA locale


tessalt (approved):

tessalt:
Thanks!

PR #3741, opened 2021-09-28
[CORL-2270] Throw better error when removing story with comments

nick-funk:
## What does this PR do?

When removing a story with comments and `includeComments = false/not set`, throws a better error message to indicate our preference that they use `mergeStory` instead.

None

## How do I test this PR?

- Create a story
- Create comments
- Call `removeStory` mutation
- See that it throws a descriptive error message
 
## How do we deploy this PR?

No special deploy actions necessary.

### DIFF
--- a/src/core/server/services/stories/index.ts
+++ b/src/core/server/services/stories/index.ts
@@ -234,8 +234,9 @@ export async function remove(
       "attempted to remove story that has linked comments without consent for deleting comments"
     );
 
-    // TODO: (wyattjoh) improve error
-    throw new Error("asset has comments, cannot remove");
+    throw new Error(
+      "Failed to delete story. Attempted to remove a story that still has comments. It is preferable that you merge a story instead of trying to delete it. If you are sure about what you're doing and want to delete this story with its comments. Use the option `includeComments = true`. For more details check the schema suggestions for this mutation input."
+    );
   }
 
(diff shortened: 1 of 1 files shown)


alinhle (changes requested):
Let's talk about why this safeguard is here and whether it makes sense to remove it or instead provide a better error message.

src/core/server/services/stories/index.ts
```diff
@@ -234,8 +234,9 @@ export async function remove(
       "attempted to remove story that has linked comments without consent for deleting comments"
     );
 
-    // TODO: (wyattjoh) improve error
-    throw new Error("asset has comments, cannot remove");
+    throw new Error(
+      "Failed to delete story. Attempted to remove a story that still has comments. It is preferable that you merge a story instead of trying to delete it. If you are sure about what you're doing and want to delete this story with its comments. Use the option `includeComments = true`. For more details check the schema suggestions for this mutation input."
```
alinhle:
```suggestion
      "Failed to delete story. Attempted to remove a story that still has comments. It is preferable that you merge a story or update its Story URL instead of trying to delete it. If you are sure about what you're doing and want to delete this story with its comments. Use the option `includeComments = true`. For more details check the schema suggestions for this mutation input."
```

alinhle (approved):

PR #3745, opened 2021-10-05
Fix nb-NO translation

ingriddraagen:
## What does this PR do?
Fix a translation-isuue


alinhle (changes requested):
The original English translation is:
```
comments-userPopover-ignore = Ignore
```
Which is only the verb ignore. 
I think line 255-256 should be removed and
line 254 should be something like: `comments-userPopover-ignore = Ignorer`

Wdyt?

ingriddraagen:
> The original English translation is:
> 
> ```
> comments-userPopover-ignore = Ignore
> ```
> 
> Which is only the verb ignore. I think line 255-256 should be removed and line 254 should be something like: `comments-userPopover-ignore = Ignorer`
> 
> Wdyt?

Great catch! I was sick last week, so sorry for the late reply. I'll push the changes asap

tessalt (approved):

alinhle (approved):

PR #3746, opened 2021-10-06
[CORL-2356] Decision history should work when comment revisions get deleted

alinhle:
## What does this PR do?
- When a comment was moderated and the author gets deleted, the decision history would break.
- This will fix it.
Screenshot 2021-10-07 at 00 08 42

## How do I test this PR?
- Moderate a comment of a user
- Delete user
- Open decision history


nick-funk (approved):

alinhle:
Will be part of https://github.com/coralproject/talk/pull/3666

PR #3750, opened 2021-10-14
[CORL-2331] Searchable Sites List

kabeaty:
## What does this PR do?
This PR addresses [CORL-2331](https://vmproduct.atlassian.net/browse/CORL-2331) by adding a site search by name. This is to support cases, such as SBNation, that include many sites, to make it easier for someone to go through the list and find a specific site they're looking for. These changes add site search components to the admin that are then used in the stories route. These include a `SiteSearchListContainer` that displays search results, the `SiteSearchTextField`, and the `SiteFilterOption`, which shows each search option. The `SiteFilterOption` component was already being used by the stories route before.   The idea is that in the future, these components could also be used in the moderation route in place of its current `SiteSelector` dropdown, as well as potentially in other places where a site search would be useful.

This adds a query to the text index for sites for the text search by name.

## How do I test this PR?
You’ll need to have multiple sites (at least 10 is ideal) added. A few of the sites should have at least one story. Then, navigate to the stories route. Search within the new “Search by site name” input. 

Test selecting different sites changes the input value. Test selecting different sites correctly updates the stories shown . Test that selecting “All sites” shows all stories. Test that clicking outside of the site search closes the site search result list if it’s open. Test that the search result list shows both “All sites” and “No sites were found with that search” when no site is found (any feedback on what messaging is shown in this case is welcomed!).
 
## How do we deploy this PR?
Upon deploy, I believe that we will need to add a text index for the sites query.
```
use coral;
db.sites.createIndex({ tenantID: 1, "$**": "text", createdAt: -1 });
```


nick-funk (approved):
This is incredible work, heartily approved!

src/core/client/admin/test/fixtures.ts
```diff
@@ -214,6 +214,11 @@ export const settings = createFixture<GQLSettings>({
   amp: false,
 });
 
+export const settingsWithMultisite = createFixture<GQLSettings>({
+  ...settings,
+  multisite: true,
+});
```
alinhle: `createFixture` accepts the base fixture and does recursive merging for you :-)

```suggestion
export const settingsWithMultisite = createFixture<GQLSettings>({
  multisite: true,
}, settings);
```

src/core/client/admin/components/SiteSearch/SiteSearchList.css
```diff
@@ -0,0 +1,28 @@
+.list {
+  position: absolute;
+  z-index: 99;
```
alinhle:
```suggestion
  z-index: $zindex-popover;
```

alinhle (changes requested):
As we talked in our meeting:
- Searchable list should be implemented with its own query and not attach to the main Query.

alinhle (changes requested):
Excellent work! 

Here is a summary of our meeting today for a next round of improvements:

tl:dr 
- make components as dumb as possible with a clear responsibility, move logic up as much as possible, try avoiding dependencies between components, props should make sense if you look at a component in isolation, rely on props that has an “on” prefix, like “onSelect”, “onSearch”, “onSubmit”.
- Rename CSS file

src/core/client/admin/components/SiteSearch/SiteSearch.tsx
```diff
@@ -0,0 +1,95 @@
...
+import SiteSearchListQuery from "./SiteSearchListQuery";
+import SiteSearchTextField from "./SiteSearchTextField";
+
+interface Props {
+  onSelect: (id: string | null) => void;
+}
+
+const SiteSearch: FunctionComponent<Props> = ({ onSelect }) => {
+  const [searchFilter, setSearchFilter] = useState<string>("");
+  const [tempSearchFilter, setTempSearchFilter] = useState<string>("");
```
cvle, line 27: `searchTextFieldValue` could be a better naming.

src/core/client/admin/components/SiteSearch/SiteFilterOption.tsx
```diff
@@ -1,38 +1,35 @@
...
 
-import { SiteFilterOption_site } from "coral-admin/__generated__/SiteFilterOption_site.graphql";
+import { SiteSearchListContainer_query } from "coral-admin/__generated__/SiteSearchListContainer_query.graphql";
 
 import styles from "./SiteFilterOption.css";
 
 interface Props {
-  site: SiteFilterOption_site | null;
-  onSelect: (id: string | null) => void;
+  site: SiteSearchListContainer_query["sites"]["edges"][0]["node"] | null;
```
alinhle: Instead of creating an indirect dependency between this an SiteSearchListContainer, use `{ name: string; id: string}` or make it into a fragment container.

alinhle (approved):
Excellent job!

PR #3752, opened 2021-10-18
[CORL-2312]: Add option to provide another domain for comment downloads

kabeaty:
## What does this PR do?
This PR adds an optional server config env var for the domain to be used in the download GDPR comments link that's sent out in emails. If set to its default empty string, the tenant domain will still be used. If set to a domain, then this domain will be used in the download link sent out.

I am interested in any feedback on what to do in `constructDownloadLinkUrl` regarding the domain in the case of working in `development`. I opted to include the link domain override in that url even though it means the url won't work in that case, so that the link could be inspected for testing that the override is working.

None

## How do I test this PR?
You can test this PR by
1. Setting the new env var (`DOWNLOAD_GDPR_COMMENTS_LINK_DOMAIN`) to a domain. (Can also test with it not set to make sure still working as expected in that case.)
1. Configuring Coral to allow users to download their comment history.
1. You'll need your local dev set up to mock email sending.
1. Request your comment history from your profile in the stream.
1. View the email you receive at localhost:9000. You should see when you hover over `Download my comment archive` that the link is using the domain you set.
 
## How do we deploy this PR?
This adds a new env var to the server config, but it doesn't need to be set to anything.

### DIFF
--- a/src/core/server/config.ts
+++ b/src/core/server/config.ts
@@ -455,6 +455,13 @@ const config = convict({
     default: ms("30s"),
     env: "MAILER_JOB_TIMEOUT",
   },
+  download_gdpr_comments_link_domain: {
+    doc:
+      "Specifies an alternative domain to be used for the download GDPR comments link sent out in emails. If set to default empty string, will use the tenant domain. Example: yourdomain.com",
+    format: String,
+    default: "",
+    env: "DOWNLOAD_GDPR_COMMENTS_LINK_DOMAIN",
+  },
   non_fingerprinted_cache_max_age: {
     doc: "Max age for the ",
     format: "ms",
--- a/src/core/server/config.ts
+++ b/src/core/server/config.ts
@@ -58,6 +58,16 @@ convict.addFormat({
   },
 });
 
+// Add a custom format for domain.
+convict.addFormat({
+  name: "domain",
+  validate: (domain: string) => {
+    if (domain) {
+      Joi.assert(domain, Joi.string().domain());
+    }
+  },
+});
+
 // Add a custom format for a list of comma seperated strings.
 convict.addFormat({
   name: "list",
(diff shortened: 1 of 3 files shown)


src/core/server/config.ts
```diff
@@ -58,6 +58,16 @@ convict.addFormat({
...
 
+// Add a custom format for domain.
+convict.addFormat({
+  name: "domain",
+  validate: (domain: string) => {
+    if (domain) {
+      Joi.assert(domain, Joi.string().domain());
+    }
+  },
+});
```
nick-funk, line 69: This created a cool learning moment for me, I thought we already had a url type for our config validation, but this is for a doman and not an `"optional-trailing-url"` like the `static-uri`. Nice add!

src/core/server/services/users/download/token.ts
```diff
@@ -67,10 +70,15 @@ export async function generateDownloadLink(
...
+  const downloadLinkDomainOverride = config.get(
+    "download_gdpr_comments_link_domain"
+  );
+
+  return constructDownloadLinkURL(
     config,
     tenant,
-    `/account/download#downloadToken=${token}`
+    `/account/download#downloadToken=${token}`,
+    downloadLinkDomainOverride
```
nick-funk, line 81: How is the `downloadLinkDomainOverride` bringing in the download token here? I see it being added for the regular `path` var, maybe I'm missing something, because I think it might be missing for the override var?

nick-funk (changes requested):
Just a couple small things to fix here, but otherwise, this PR is great!

src/core/server/services/users/download/token.ts
kabeaty: You got it--the download token is still being sent through via the token in the path and then combined with the override (which is just for the domain) when the url is constructed.

nick-funk: ah, nevermind, I missed that, this is good to go then!

nick-funk (approved):
This looks good to go!

alinhle (approved):
Very well done! Good work @kabeaty!

PR #3756, opened 2021-10-21
[CORL-2375] Use worker threads to process regex to keep regex operations async friendly

nick-funk:
NOTE: this PR is branched off of CORL-2300's PR, merge that first and retarget this to develop before merging this PR.

## What does this PR do?

Create an internal wordlist service that uses worker threads

None

## How do I test this PR?

- set banned and suspect words in the config
- create comments with commenter account using those words
- see them get detected in mod queues and/or rejected from stream
 
## How do we deploy this PR?

Nothing special here.


src/core/server/services/wordList/serviceThread.ts
```diff
@@ -0,0 +1,92 @@
...
+const run = async (options: WordlistServiceOptions) => {
+  if (isMainThread) {
+    return;
+  }
+  if (!options) {
+    return;
+  }
+  if (!parentPort) {
+    return;
+  }
```
wyattjoh, line 30: If you did something like:

```ts
const port = parentPort;
if (!port) {
  return;
}
```

Then later, you wouldn't need to do `parentPort!` because the type of `port` can't be undefined.

src/core/server/services/wordList/wordlistService.ts
```diff
@@ -0,0 +1,130 @@
...
+    const timeout = this.timeout;
+
+    const promise = new Promise<WordlistThreadResult>((resolve, reject) => {
+      requests.set(id, {
+        id,
+        resolve,
+        reject,
+      });
+
+      setTimeout(() => {
```
wyattjoh: An alternatve! You can do something like this:

```ts
const handle = setTimeout(() => { /* ... */ }, timeout);
requests.set(id, { id, resolve, reject, handle });
```

And then later, when the request is processed, instead do:

```ts
clearTimeout(handle);
```

To avoid having all those dangling timeouts that need processing.

src/core/server/services/wordList/wordlistService.ts
```diff
@@ -0,0 +1,130 @@
...
+    if (!this.requests.has(msg.id)) {
+      return;
+    }
+
+    const request = this.requests.get(msg.id);
+    if (!request) {
+      return;
+    }
+
+    request.resolve(msg);
```
wyattjoh, line 158: The implmentation of this worker pool is leaking (the ID for example)! Perhaps, only `resolve` with the contents of `msg.result` and `reject` with an error instead!

src/core/server/services/comments/pipeline/wordList.ts
```diff
@@ -31,12 +28,10 @@ export class WordList {
...
-    // Generate the regular expression for this list.
-    const regexp = createServerWordListRegEx(locale, list);
-
-    // Create a managed regular expression from the provided regular expression
-    // so we can time it out if it takes too long!
-    return createTesterWithTimeout(regexp, timeout);
+    return new WordlistService(logger, timeout, {
+      lang: locale,
+      phrases: list,
+    });
```
wyattjoh, line 34: This will create one `WordlistService` (read, one `worker_thread`) PER Tenant! I'd suggest a refactor that lets you send the wordlist configuration (the RegExp and Tenant ID) over to the SINGLE worker.

wyattjoh: Note that this is also being tracked as a `WeakMap`! The underlying `Worker` is never terminated even though it's left the scope!

wyattjoh: It does provide a `.terminate()` or better, `.unref()` to shutdown the worker thread if you need to shut it down. I'd expect though that changes to tenant patterns could be handled with a `type` field in the message sent over the `postMessage` connection to the worker.

src/core/server/services/wordList/serviceThread.ts
nick-funk: I did, there is a guard clause, but TS linter kept whining

Weird that it needs it to be dumped into a const var, I'll try that, thanks

src/core/server/services/wordList/wordlistService.ts
```diff
@@ -0,0 +1,130 @@
...
+    this.script = "./dist/core/server/services/wordList/serviceThread.js";
+    this.worker = new Worker(this.script, {
+      workerData: options,
+    });
+
+    this.onMessageDelegate = this.onMessage.bind(this);
+    this.onErrorDelegate = this.onError.bind(this);
+
+    this.worker.on("message", this.onMessageDelegate);
+    this.worker.on("error", this.onErrorDelegate);
```
wyattjoh, line 67: If this error is encountered too, it means that the worker has crashed! You should remove it to ensure that the process crashes.

nick-funk: It now logs an error and crashes

src/core/server/services/comments/pipeline/wordList.ts
nick-funk: I'm trying to de-couple the word list service from the tenant like you have suggested here. The idea being that we only want one worker thread for the whole Coral instance.

I have run into a bit of a pickle though. The problem is that the wordlists are cached and generated based on a tenant's org's banned and suspect word lists. If I create a singular thread for the entire Coral pod, then I have to have separate regex with different banned/suspect words for each cached within the service thread.

Then I also have to then tell the worker thread, "I am this tenant, please retrieve my specific regex for banned/suspect words" and "use that to test this string I have asked you to test for banned words" as opposed to how it currently is where it goes "I have a service for my banned words, please test this string that you already immediately know how to test cause you know my banned words cause you're my service for my tenant for my banned words".

This also breaks the enclosed Tenant scope we previously created throughout all of Coral. Instead of the Tenant/GraphContext being consolidated and scoped to a request for a specific Tenant, we will now have this service living outside of any Tenant that also holds part of the state of every Tenant on a Coral instance. It feels very much like it would turn this service into a Singleton pattern with disparate entitities that are not within its scope of ownership.

In my experience, this is not a good pattern to create. Singleton's aren't bad per se, but they are bad when they hold resources from many others trying to centralize some work because we try and force it to do the work for n-many things when scoping it to do 1 thing for 1 scope makes for a cleaner abstraction of the service logic.

tessalt:
Can't merge until we update node to v16

PR #3757, opened 2021-10-22
[CORL-2316]: Add moderator messaging option

kabeaty:
## What does this PR do?

This PR adds in the option to send a commenter a message if you are a moderator without changing their status (i.e. without a warning attached). This is for https://vmproduct.atlassian.net/browse/CORL-2316. 

Mod messaging is very similar to sending a warning message except that it doesn't change the user's status in any way.

To the GraphQL schema, these changes add `ModMessageStatus` and `ModMessageStatusHistory`, as well as adds these under `modMessage` to the `UserStatus`. These include the current status of moderation messages (is there currently an active mod message?) as well as the history of past moderation messages and their acknowledgment. In addition, these changes add `ModMessageUserInput`, `ModMessageUserPayload`, `AcknowledgeModMessageInput`, and `AcknowledgeModMessagePayload`. They also add `modMessageUser` and `acknowledgeModMessage`.

To the database schema, these changes add `ModMessageStatus` and `ModMessageStatusHistory`, as well as adds these under `modMessage` to the `UserStatus`. These changes add `modMessageUser` and `acknowledgeModMessage`. They also add `ConsolidatedModMessageStatus` and `consolidateUserModMessageStatus` to grab the message from the latest active moderation message.

## How do I test this PR?

 To test this, you can create at least one user that is a commenter. Then
 1. Logged in as admin/moderator, navigate to `Community` in the admin. 
 1. Find the user to whom you want to send a moderation message in the user table and click their status. Select `Message` at the bottom of the options.
 1. You should see a modal open with an area for entering a message to the user. Enter a message and click `Message User`. You should see a success modal confirming that a message was sent to the user.
 1. Now log in as the user to whom you sent the message. At the top of the stream, you should see the moderation message that you sent displayed with an option to acknowledge the message. See that you are still able to comment on the story.
 
 Other things to test with this flow include:
 - As the admin/moderator, visit the user's account history throughout the process and confirm that the message and message acknowledgment history is visible there.
 - If multiple messages have been sent by the moderator and unacknowledged, only the latest active moderation message is displayed at the top of the stream for acknowledgment.
 - A user sees both a warning and a moderation message to acknowledge if they have both.
 
## How do we deploy this PR?

I don't think there's anything that needs to be done on deploy.


src/core/client/admin/components/UserStatus/ModMessageForm.tsx
```diff
@@ -0,0 +1,79 @@
...
+                    <Label className={styles.label}>Message</Label>
+                  </Localized>
+                  <Localized id="community-modMessageModal-message-required">
+                    <span className={styles.required}>Required</span>
+                  </Localized>
+                </Flex>
+              </HorizontalGutter>
+              <Field component="textarea" name="message" validate={required}>
+                {({ input }) => (
+                  <Textarea id="modMessageModal-message" {...input} fullwidth />
```
alinhle: As this is only used for testing, we usually use `data-testid="modMessageModal-message" and `getByTestID` instead. But you can leave it as is.

src/core/client/admin/components/UserStatus/ModMessageUserMutation.ts
```diff
@@ -0,0 +1,77 @@
...
+                }
+              }
+            }
+            clientMutationId
+          }
+        }
+      `,
+      variables: {
+        input: {
+          ...input,
```
alinhle: We do have a lot of bugs coming from spreading input like this in mutations, because Typescript allows passing in an `input` that contains more props than the type defines. That's why I recommend picking or passing the props individually here instead.

src/core/client/admin/components/UserStatus/UserStatusChange.tsx
```diff
@@ -194,6 +196,19 @@ const UserStatusChange: FunctionComponent<Props> = ({
...
             )}
+            <Localized id="community-userStatus-message">
+              <DropdownButton
+                className={styles.dropdownButton}
+                onClick={() => {
+                  if (onModMessage) {
+                    onModMessage();
+                    toggleVisibility();
+                  }
+                }}
```
alinhle, line 207: As a good practice I usually recommend using `useCallback` which memoizes the callback instead of generating a new callback every render. Keeping the same reference when the data/object has not changed is a prerequisite for any performance optimisation in React.

This is not not important, just a reminder for a good practise.

src/core/server/models/user/user.ts
```diff
@@ -2329,6 +2367,125 @@ export async function acknowledgeOwnWarning(
 
   return result.value;
 }
+
+export async function modMessageUser(
```
alinhle: I would start a method with a verb like `sendModMessage`

src/core/server/services/users/users.ts
```diff
@@ -1316,6 +1319,61 @@ export async function acknowledgeWarning(
...
+/**
+ * modMessage will send a moderation message to a specific user.
+ *
+ * @param mongo mongo database to interact with
+ * @param tenant Tenant where the User will be messaged on
+ * @param moderator the User that is messaging the User
+ * @param userID the ID of the User being messaged
+ * @param now the current time that the message was sent
+ */
+export async function modMessage(
```
alinhle: Same here, would start method with a verb.

src/core/server/models/user/user.ts
```diff
@@ -2390,6 +2553,23 @@ export function consolidateUserWarningStatus(
   };
 }
 
+export function consolidateUserModMessageStatus(
```
alinhle, line 2542: How do you think about putting a comment here to explain what `consolidate` means?

src/core/server/models/user/user.ts
```diff
@@ -2329,6 +2367,125 @@ export async function acknowledgeOwnWarning(
...
+        {
+          "status.modMessage.active": {
+            $ne: false,
+          },
+        },
+        {
+          "status.modMessage.history": {
+            $size: 0,
+          },
+        },
```
alinhle: Interesting, what's this second condition for?

src/core/server/models/user/user.ts
```diff
@@ -369,6 +400,12 @@ export interface UserStatus {
    * warnings
    */
   warning?: WarningStatus;
+
+  /**
+   * modMessage stores whether a user has an unacknowledged moderation message and
+   * a history of moderation messages
+   */
+  modMessage: ModMessageStatus;
```
alinhle: Should be `modMessage?: ModMessageStatus;` I believe, because `modMessage` is undefined for users that are already in the db before this lands.

src/core/client/stream/tabs/Comments/Stream/StreamContainer.tsx
```diff
@@ -300,6 +302,7 @@ export const StreamContainer: FunctionComponent<Props> = (props) => {
                 />
               )}
               {warned && <WarningContainer viewer={props.viewer} />}
+              {modMessaged && <ModMessageContainer viewer={props.viewer} />}
             </section>
```
alinhle: Can you make the section into a `HoriziontalGutter`? Like this:

```tsx
            <HorizontalGutter
              id={VIEWER_STATUS_CONTAINER_ID}
              aria-label="Account Status"
              container="section"
            >
```

src/core/server/graph/schema/schema.graphql
```diff
@@ -8339,12 +8444,26 @@ type Mutation {
...
+  acknowledgeWarning will remove a warning
   """
   acknowledgeWarning(
     input: AcknowledgeWarningInput!
   ): AcknowledgeWarningPayload @auth(permit: [WARNED])
 
+  """
+  modMessageUser will send a user a moderation message that they need to acknowledge
+  """
+  modMessageUser(input: ModMessageUserInput!): ModMessageUserPayload!
```
alinhle: Is `modMessage` a verb 🤔 otherwise let's find a verb to start the name of this mutation.

src/core/server/graph/schema/schema.graphql
```diff
@@ -8339,12 +8444,26 @@ type Mutation {
     @auth(roles: [ADMIN, MODERATOR])
 
   """
-  acknowledgWarning will remove a warning
+  acknowledgeWarning will remove a warning
   """
   acknowledgeWarning(
     input: AcknowledgeWarningInput!
   ): AcknowledgeWarningPayload @auth(permit: [WARNED])
```
alinhle: Amazing work! Can you also adapt this line to have the same permission as your `acknowledgeModMessage` 
`@auth(permit: [SUSPENDED, BANNED, PENDING_DELETION, WARNED])`. 

I know it's not part of your story, but while you're here.. :-D

alinhle (changes requested):
Some remarks, but this is excellent work @kabeaty!

src/core/server/models/user/user.ts
```diff
@@ -2329,6 +2367,125 @@ export async function acknowledgeOwnWarning(
...
+  };
+
+  const result = await mongo.users().findOneAndUpdate(
+    {
+      id,
+      tenantID,
+      $or: [
+        {
+          "status.modMessage.active": {
+            $ne: false,
```
alinhle: Maybe true?

src/core/server/graph/schema/schema.graphql
kabeaty: As discussed, was thinking of `modMessage` as a verb, but will update to `sendModMessage` throughout for clarity 👍

src/core/server/models/user/user.ts
kabeaty: Updated to `$eq: true` here and in `acknowledgeOwnWarning` 👍

src/core/server/models/user/user.ts
kabeaty: Removed this second condition both here and in `acknowledgeOwnWarning` 👍

src/core/client/stream/test/comments/stream/modMessaged.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+    within(modMessageCallout).queryByText(
+      "This is a friendly reminder about our community guidelines."
+    )
+  ).toBeDefined();
+
+  // can still see comment stream and interact with comments
+  expect(within(tabPane).getByTestID("comments-allComments-log")).toBeDefined();
+  expect(within(tabPane).queryByTestID("comment-reply-button")).toBeDefined();
+  expect(within(tabPane).queryByTestID("comment-report-button")).toBeDefined();
+  expect(within(tabPane).queryByTestID("comment-edit-button")).toBeDefined();
```
alinhle: `getByX` commands will automatically fail when element was not found, while `queryBy` does not and return either null or the element it found.

alinhle (approved):

PR #3769, opened 2021-11-01
[CORL-2373]: Enable single site moderation by default

kabeaty:
## What does this PR do?

This PR removes the `SITE_MODERATOR` feature flag as part of https://vmproduct.atlassian.net/browse/CORL-2373 and enables single site moderation by default.

Where the feature flag was previously checked, either only whether the tenant is multisite is checked (front end) or nothing needs to be checked (back end).

These changes remove the `SITE_MODERATOR` feature flag.

## How do we deploy this PR?

Is there anything we need to do for the removal of the `SITE_MODERATOR` feature flag?

- No @cvle


src/core/server/config.ts
```diff
@@ -513,6 +513,13 @@ const config = convict({
     default: false,
     env: "FORCE_ADMIN_LOCAL_AUTH",
   },
+  enable_site_moderator: {
+    doc:
+      "Will enable the Site Moderator role for individual sites on a tenant by assigning a new moderation scope to a user.",
```
nick-funk: It might make more sense for this to read something like this:

```
Will allow the creation of site moderators whose scope of moderation permissions are limited to the sites in which they are assigned. Also requires your organization to have at least two sites.
```

The reason I say this and don't want to specify `tenant` here is this will affect all tenants that are on the pod deployment since this is an environment variable. Let me know if any of this doesn't jive with the copy from design, not trying to be confusing.

src/core/server/graph/schema/schema.graphql
```diff
@@ -2070,6 +2064,11 @@ type Settings @cacheControl(maxAge: 30) {
   amp is whether Accelerated Mobile Pages Support is activated.
   """
   amp: Boolean!
+  """
+  siteModeratorEnabled is whether the config var is enabled for
+  the single site moderator role
+  """
+  siteModeratorEnabled: Boolean!
```
nick-funk: @cvle and @wyattjoh, I am fine with this existing here on the settings object, but could also see this being populated via the nunjucks injected config and existing as a prop stored in the React Relay local cache on `Local`. What are your thoughts?

nick-funk (dismissed):
I'm putting request changes, but it's mostly me requesting feedback on opinions from @cvle and @wyattjoh. Overall, this PR looks good!

src/core/server/graph/schema/schema.graphql
alinhle: I will bring this up in scrum. I think if the goal is to eventually get rid of this switch all together, then I would probably avoid adding this to the schema and either use the StaticConfig as you said @nick-funk or removing the switch and the env var entirely.

nick-funk: Sounds good, hopefully we can get this resolved with minimal changes to this PR. The rest of the code looks good to me.

src/core/client/stream/tabs/Comments/Comment/UserBanPopover/UserBanPopoverContainer.tsx
```diff
@@ -38,14 +34,6 @@ const UserBanPopoverContainer: FunctionComponent<Props> = ({
   const banUser = useMutation(BanUserMutation);
   const { localeBundles } = useCoralContext();
 
-  // Not checking for multisite here due to permission issues.
-  // We shouldn't have this enabled unless we're already multisite
-  // anyways. Also, if we send site ID's and multisite is off, the
-  // backend will handle this.
-  const moderationScopesEnabled = settings.featureFlags.includes(
```
kabeaty, line 45: We don't need any checks here because site moderators do not see this `UserBanPopoverContainer` in the stream; only admins or organization moderators do. This is also why the alternative description below has been removed.

alinhle (approved):

PR #3773, opened 2021-11-03
CORL-2371 Fix Staff Badges

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2371 by fixing a bug causing staff badges to not be added to the appropriate comments. It also begins a fixture collection for server side testing

None.

## How do I test this PR?
Create 2 sites under the same tenants, and a story for each site, as well as a member with the role of Staff, 
Observe that as a Staff user, you receive Staff badges on both stories.

Additionally: 
Create an Admin user, an Organization moderator, a Site moderator for the first site, a Site moderator for the second site and a general commenter.

Observe that:
1. Admin and Organization moderator users are given badges on both stories.
2. Site Moderator users receive badges on the story for which the site is in their moderation scopes, and not on the other story.
3. General Commenters do not receive badges anywhere.
 
## How do we deploy this PR?
This PR should not require any special actions to deploy.


src/core/server/services/comments/pipeline/phases/tagStaff.spec.ts
```diff
@@ -0,0 +1,221 @@
...
+    throw new Error("tagStaff unexpectadly applied more than one tag");
+  }
+
+  return result?.tags?.[0];
+};
+
+const tenant = createTenantFixture();
+const tenantID = tenant.id;
+
+// TODO: this will need to be updated once https://github.com/coralproject/talk/pull/3769 is merged
```
marcushaddon: This will be a very fast fix, but the way this feature is indicated will need to be updated once @kabeaty 's PR is in.

nick-funk (approved):

src/core/server/services/comments/pipeline/phases/tagStaff.spec.ts
nick-funk: I've approved, but put a `don't merge` tag on this cause it sounds like you want this other PR to get merged first.

alinhle (commented):
I agree with everything that has been said :-)

PR #3775, opened 2021-11-05
CORL-2358 gracefully close connections

marcushaddon:
## What does this PR do?
This PR addresses https://vmproduct.atlassian.net/browse/CORL-2348 by closing connections on the express server on SIGTERM, as well as setting up a method in which any other resources can be cleaned up if need be.

None.

## How do I test this PR?
TODO
 
## How do we deploy this PR?

### DIFF
--- a/src/core/server/index.ts
+++ b/src/core/server/index.ts
@@ -279,6 +279,16 @@ class Server {
   }
 
   /**
+   * stop gracefully closes connections and cleans up any resources.
+   */
+  public async stop() {
+    if (this.httpServer) {
+      this.httpServer.close(() => logger.info("Http server closed"));
+    }
+    // Do we need additional cleanup here?
+  }
+
+  /**
    * process will start the job processors and ancillary operations.
    */
   public async process() {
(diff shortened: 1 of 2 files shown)


src/core/server/index.ts
```diff
@@ -278,6 +278,16 @@ class Server {
...
   }
 
+  /**
+   * stop gracefully closes connections and cleans up any resources.
+   */
+  public async stop() {
+    if (this.httpServer) {
+      this.httpServer.close(() => logger.info("Http server closed"));
+    }
+    // Do we need additional cleanup here?
```
marcushaddon, line 288: ie mongo/redis connections?

alinhle (changes requested):
Screenshot 2021-11-10 at 14 34 25

Unfortunately only closing the `httpServer` is not enough, because all of our connections are long living (keep-alive) tcp connections. 

Try finding a library that helps closing these connections as well.

src/core/server/index.ts
alinhle: That won't be necessary.

tessalt:
closing until we can re-evaluate options.

PR #3780, opened 2021-11-10
[CORL-2221]: Replace anchor tags with their text when copy/pasting into RTE

kabeaty:
## What does this PR do?

These changes update the RTE sanitization logic to replace anchor tags with their text when copy/pasting into the RTE. Currently, anchor tags' text is updated to their href when copy/pasted into the RTE. This is for https://vmproduct.atlassian.net/browse/CORL-2221. When replacing the anchor tags with text, it was adding nbsp, so this adds a hook to make sure those aren't included/don't change how the comment is displayed, too.

None

## How do I test this PR?

You can copy/paste text that includes linked text into the RTE for comments in the comment stream. See that the text is pasted into the RTE but it is no longer a link.


tessalt (approved):
validated this works as intended locally.

PR #3785, opened 2021-11-15
[CORL-2399] Do not archive ratings and reviews stories

nick-funk:
## What does this PR do?

- Block auto-archiving and standard archiving for stories whose mode is set to `RATINGS_AND_REVIEWS`.
- Filter out R&R stories when searching for new stories to auto-archive

None

## How do I test this PR?

- Create a ratings and reviews story
- Attempt to archive it
- See that it does not succeed
- Edit its `createdAt` date so that it will be auto-archived
- Wait for or trigger the auto archiving job
    - Set `ENABLE_AUTO_ARCHIVING=true`
    - Modify the `AUTO_ARCHIVING_INTERVAL` to something that runs more often than every 15 minutes
    - For more info on these env vars, see `src/core/server/config.ts`
- See that this story is not auto archived
    - It can be helpful to have fewer than 500 stories so that auto archiving will definitely try and pick up the story if it can
 
## How do we deploy this PR?

N/A

### DIFF
--- a/src/core/client/admin/routes/Stories/StoryActions/StoryActionsContainer.tsx
+++ b/src/core/client/admin/routes/Stories/StoryActions/StoryActionsContainer.tsx
@@ -3,7 +3,7 @@ import { graphql } from "react-relay";
 
 import { Ability, can } from "coral-admin/permissions";
 import { useMutation, withFragmentContainer } from "coral-framework/lib/relay";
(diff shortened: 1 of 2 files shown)


tessalt (approved):

kabeaty (approved):
Looks good! Thanks for making the update to not showing the `Archive story` option in the `Actions` dropdown if the story mode is `RATINGS_AND_REVIEWS`

PR #3786, opened 2021-11-16
[CORL-2349] Use translation keys for relayNetworkRequestError strings

nick-funk:
## What does this PR do?

Use translation keys to get error message text for the `RelayNetworkRequestError` defined in `src/core/client/framework/lib/errors/relayNetworkRequestError.ts`.

None

## How do I test this PR?

Put the following after line 14 in your `customErrorMiddleware`:

```
const msg = getMessage(
    localeBundles,
    "common-error-relayNetworkRequestError-anUnexpectedNetworkError",
    "THIS FAILED AND USED DEFAULT"
  );
  // eslint-disable-next-line no-console
  console.log("LOOK HERE", msg);
```

Import `getMessage` from `i18n` in `customErrorMiddleware` via:

```
import { getMessage } from "../i18n";
```

Have any error throw client side and see that the console shows the `LOOK HERE` translation as translated correctly from `common.ftl` with `An unexpected network error occured, please try again later.` and is not using the text `THIS FAILED AND USED DEFAULT`.
 
## How do we deploy this PR?

N/A


src/locales/en-US/common.ftl
```diff
@@ -20,3 +20,6 @@ common-experimentalTag-tooltip-title = Experimental feature
 common-error-title = An error has occurred
 common-error-message = Message
 common-error-traceID = Trace ID
+common-error-relayNetworkRequestError-anUnexpectedNetworkError =
```
kabeaty: Would these fit better in `framework.ftl`? Also, I know you are using the message that was there, but I think it should be `occurred` not `occured` in the message.

kabeaty (approved):
Looks good and works as expected! Left one small note on the translations file, but approved.

src/locales/en-US/common.ftl
nick-funk: Let me have a quick look at `framework.ftl`. I always get confused which ftl does what in this layer of the front-end.

I will also fix up the typo! 😄

nick-funk: Ah, it is indeed in the framework sub-dir... into `framework.ftl` it goes!

PR #3787, opened 2021-11-17
[CORL-2284]: fix for Go to moderate view link does not pass through SSO token correctly

kabeaty:
## What does this PR do?

These changes ensure that the SSO access token is added to the `Go to moderate to review this decision` link after a comment is rejected in the stream, if the token exists. These changes are for https://vmproduct.atlassian.net/browse/CORL-2284.

None

## How do I test this PR?

You will need to be logged in as an SSO moderator/admin. You can use the `multi-site-test` to do this easily. Then go to the stream. Reject a comment. Check the url for the link `Go to moderate to review this decision` and see that it includes the access token. Click `Go to moderate to review this decision` and see that you remain logged in and are taken to the expected place to review the comment moderation decision. 

You can also test that everything works as expected still for non-SSO users clicking this link.


nick-funk (approved):
Looks and works good! Approved!

PR #3788, opened 2021-11-18
[CORL-2279]: Add list of positive reactions to rejected comments to comment cards

kabeaty:
## What does this PR do?

These changes add a `Reactions` tab to the `Details` dropdown on comment cards when a comment has been reacted to. In this tab, the usernames of the users who have reacted to the comment are shown. When their usernames are clicked on, their user history drawer opens up to show more information about them. These changes are for https://vmproduct.atlassian.net/browse/CORL-2279.

None

## How do I test this PR?

Make some users react to a comment. Go to that comment's card in `Moderate` and click `Details`. See that there is a `Reactions` tab and that it is the second tab after `Info`. Click on the `Reactions` tab. See that this is when the query is run to get the reaction data. See that the usernames for the reactions are displayed. If you click on a username, the user history drawer for that user opens. If there are more users than 20 who have reacted to a comment, a `Load More` button will appear to load more.

If a comment has no reactions, if you click `Details` on its comment card, it should not have a `Reactions` tab option.


src/core/client/admin/components/ModerateCard/ReactionDetailsContainer.tsx
```diff
@@ -0,0 +1,136 @@
...
+            </Flex>
+          )
+      )}
+      {hasMore && (
+        <Localized id="moderateCardDetails-tab-reactions-loadMore">
+          <Button
+            variant="outlined"
+            color="secondary"
+            fontSize="extraSmall"
+            onClick={loadMore}
```
nick-funk, line 66: For safety's sake and to prevent duplicate clicks on the load more button, would you mind encapsulating this in a callback similar to the following?

```
const onLoadMore = useCallback(() => {
    if (!relay.hasMore() || relay.isLoading()) {
      return;
    }

    loadMore();
  }, [loadMore]);
```

I don't think duplicate clicks ever caused any harm, but I've seen this done elsewhere in places.

nick-funk (changes requested):
This all works and looks good, I have one request, is there something we can do to add a scroll or wrapped lay out the details list so it doesn't grow so tall?

image

It starts to look a touch silly when I have more than ~10 likes haha

src/core/client/admin/components/ModerateCard/ReactionDetailsContainer.tsx
```diff
@@ -0,0 +1,136 @@
...
+      return {
+        ...fragmentVariables,
+        count,
+        cursor,
+        commentID: props.comment.id,
+      };
+    },
+    query: graphql`
+      # Pagination query to be fetched upon calling 'loadMore'.
+      # Notice that we re-use our fragment, and the shape of this query matches our fragment spec.
```
nick-funk, line 121: Nice! I didn't realize we could put comments in here, but it makes sense now that I see it! 👍

src/core/client/admin/components/ModerateCard/ReactionDetailsContainer.tsx
kabeaty: I think this is already taken care of by disabling the button with `disabled={isLoadingMore}` if it's loading more and only showing the button if it has more?

kabeaty:
> This all works and looks good, I have one request, is there something we can do to add a scroll or wrapped lay out the details list so it doesn't grow so tall?
> 
> It starts to look a touch silly when I have more than ~10 likes haha

Yes, good point. I added a max-height and overflow of auto to scroll if needed. I also updated the default number of reactions that are loaded in at first to just 10, so that the `Load more` should be visible on first load, plus this is also consistent with the number of flags that we include in the `INFO` tab. Let me know what you think.

nick-funk:
> > This all works and looks good, I have one request, is there something we can do to add a scroll or wrapped lay out the details list so it doesn't grow so tall?
> > It starts to look a touch silly when I have more than ~10 likes haha
> 
> Yes, good point. I added a max-height and overflow of auto to scroll if needed. I also updated the default number of reactions that are loaded in at first to just 10, so that the `Load more` should be visible on first load, plus this is also consistent with the number of flags that we include in the `INFO` tab. Let me know what you think.

This is much better! I'm sure there could be whole design passes on this, but I'm alright with this for now.

nick-funk (approved):

PR #3792, opened 2021-11-23
[CORL-2275]: Static stream representation before loads in

kabeaty:
## What does this PR do?

These changes add a static representation of the eventual stream UI in the DOM, before React loads in, so that readers know where the comments will appear. These changes are for https://vmproduct.atlassian.net/browse/CORL-2275.

I originally was thinking in-line styles for everything so that it would all be replaced when React loads in and to keep it simple, but then switched to classes with a `load` namespace to be able to do a media query and for readability/maintainability. Open to feedback on what we think the best approach is here.

None

## How do I test this PR?

Navigate to the stream and see the static representation before the actual stream loads in. Check it on different sizes.


src/core/client/stream/index.html
```diff
@@ -5,20 +5,193 @@
...
+      .load-lMargin-s {
+        margin-left: 0.5rem;
+      }
+      .load-lMargin-auto {
+        margin-left: auto;
+      }
+      .load-hr {
+        border: 1px solid #eaeff0;
+      }
+      @media (max-width: 639px) {
```
nick-funk, line 79: I've seen a few of these show up lately. I've done them myself.

I've seen them in `ConversationModalQuery.css` and `UserHistoryDrawer.css`.

What do you think about adding this one to `sharedVariables.ts` as something like `mediaStops` and call this whichever size you think makes sense (maybe `lg` or similar)? Maybe it would be `625px` and we add to this or something with `calc`. I don't know how we should be consolidating media queries to make the UI bits re-composable. Thoughts?

I was mostly just trying to clean up the magic numbers in our CSS a bit. Maybe it's a non-issue.

nick-funk (approved):
This works great! If you have some time, check my one comment and see if that makes sense. Approving as my suggestion is minor at best.

PR #3795, opened 2021-11-23
Chore/typescript upgrade 4.1.0

marcushaddon:
## What does this PR do?
This PR upgrades us to Typescript ^4.1.0 to support the use of `react-hook-forms` (and to keep us up to date in general).

None.

## How do I test this PR?
Build and/or test this branch.
 
 
## How do we deploy this PR?


kabeaty:
Closing for now, as further investigation into upgrade is needed.

PR #3796, opened 2021-11-24
[CORL-2299] Iframeless (Shadow DOM) Epic 

alinhle:
## What does this PR do?
- Take Coral out of the iframe

## API Deprecation
- `bodyClassName` is now deprecated. Use: `containerClassName`

## Breaking Changes 
- Raised min browser requirement, needs Shadow DOM Support, dropped support for IE11
- Font definition needs to be in a separate file (Custom Font CSS) and is configurable in admin
- `body {}`, `html {}`, `:root {}` in Custom CSS must be replaced with `#coral {}`.

## Conflict
- This PR removes feature: https://github.com/coralproject/talk/pull/3792.
  - Must be reimplemented in an iframeless way.

## Deliverables
- [x] First working version of Coral Stream outside of iframe
- [x] Bring back functionality that depended on iframe
  - [x] iframeless local, session, indexstorage
  - [x] Login / Logout
  - [x] events
  - [x] Localization
  - [x] StaticConfig
  - [x] ExternalConfig
  - [x] global CSS
  - [x] registerClickFarAway
  - [x] AMP
  - [x] Modal  / Toolbar
  - [x] Refresh Access Token
  - [x] liveCommentCount
  - [x] setCommentID
  - [x] AutoRender
  - [x] BodyClassName
  - [x] Set public path / static url
  - [x] Enable restricted Cors
  - [x] CSS Support
  - [x] Async Chunk CSS Support
  - [x] Custom CSS Support
  - [x] Custom Font Support
  - [x] Find solutions for media queries
  - [x] Error Reporter
  - [x] Isolate MobileToolbar within its own shadow dom
- [x] Expose bootstrap data at `/embed/bootstrap` 
- [x] Utilize client mounted routes in development instead of separate web dev server html files
- [x] Reduce global polluting babel Polyfills
- [x] Integrate Shadow-Dom
- [x] Integrate emotion into Shadow-Dom because of @giphy/react-components
- [x] Prevent flash of unstyled elements
- [x] Smarter preloading
- [x] Allow disabling Coral typography
- [x] Stream not as a separate library?
- [x] Deprecate/Rename BodyClassName
- [x] ~~Allow disabling individual polyfills~~
- [x] Set minimum browser requirements
- [x] Remove IE11 baggage
- [x] ~~IE11 Support~~
- [x] IframeEncapsulation
- [x] Frame bundle
- [x] Webpack isolation
- [x] Support custom Websocket ports during dev
## TODO
- [x] Resolve site
- [x] Refactor: Replace `CoralContext.renderWindow` with `CoralContext.window` (Add to backlog)
- [x] Make eventemitter universal that replaces previous pym- and postmessages between iframe and embed
- [x] Reduce global namespace pollution from polyfills
- [x] Remove global CSS
- [x] Replace StaticConfig
- [x] Replace @media queries previously relying on iframe width
- [x] Replace code that depends on `areWeInIframe()`
- [x] Replace localStorage with iframeless version
- [x] Replace sessionStorage with iframeless version
- [x] Replace indexedDBStorage with iframeless version
- [x] Replace postMessage that was used to communicate between iframes
- [x] Replace `registerClickFarAway`
- [x] Replace `pym` except on external medias
- [x] Replace `withAutoHeight` decorator
	- Support amp dynamic height
- [x] Replace `withClickEvent` decorator
- [x] Replace `withConfig` decorator
- [x] Replace `withEventEmitter` decorator
- [x] Replace `withIndexedDBStorage` decorator
- [x] Replace `withKeyPressEvent` decorator
- [x] Remove `withIOSSafariWidthWorkaround` decorator
- [x] Replace `withLiveCommentCount` decorator
- [x] Replace `withPostMessageStorage` decorator
- [x] Replace `withRefreshAccessToken` decorator
- [x] Replace `withRenderTargets` 
- [x] Remove RenderTarget functionality
- [x] Replace `withSetCommentID` decorator
- [x] Replace `OnPymLogin` listener
- [x] Replace `OnPymLogout` listener
- [x] Replace `OnEvents` listener
  - Support rudderstack events propagation
- [x] Replace `onPymSetCommentID`
- [x] Replace pym refresh token getter in `RefreshTokenHandler`
- [x] MobileToolbar / Modals / Portals must create a derived version of the ShadowRoot that also syncs CSS files into the shadow tree. MobileToolbar should be rendered using a Portal to make it independent of stacking contexes and put into its own ShadowRoot.
- [x] Merge with develop


yogeshbeniwal:
@cvle Will this require code change in WordPress plugin also?

alinhle:
@yogeshbeniwal The api has no breaking changes and should work without changes to the Wordpress plugin.

marcushaddon:
Still early in testing this, but I'm unable to delete pasted text in the comment form (even after refreshing and seeing the persisted text).

src/core/client/embed/StreamEmbed.ts
```diff
@@ -1,203 +1,318 @@
...
   }
 
+  /** Queue calledback to be called after bootstrap config has loaded */
+  private runAfterBootstrapConfigLoaded(cb: () => void) {
+    this.onBootstrapConfigLoaded.push(cb);
+  }
+
+  /** This is called when BootstrapConfig ist Loaded */
+  private onBootstrapConfigLoad(bootstrapConfig: EmbedBootstrapConfig) {
+    this.boostrapConfig = bootstrapConfig;
```
kabeaty: typo, `boostrapConfig` --> `bootstrapConfig`

src/core/client/framework/hooks/useModerationLink.ts
```diff
@@ -0,0 +1,13 @@
+import getModerationLink, {
+  Options,
+} from "coral-framework/helpers/getModerationLink";
+import { useCoralContext } from "coral-framework/lib/bootstrap";
+
+/**
+ * usePrevious is a react hook that will return the
```
kabeaty: Can you please update this comment to match the `useModerationLink` function?

src/core/client/framework/lib/errors/reporter/createReporter.ts
```diff
@@ -19,41 +24,28 @@ function createDevelopmentReporter() {
 
 function createReporter(
   window: Window,
```
kabeaty: It looks like we are no longer using `window` here. Can we remove?

src/core/client/stream/tabs/Profile/ProfileQuery.tsx
```diff
@@ -21,13 +20,9 @@ const loadProfileContainer = () =>
   import("./ProfileContainer" /* webpackChunkName: "profile" */);
 
 // (cvle) For some reason without `setTimeout` this request will block other requests.
-const preloadAndPolyfill = once((window: Window) =>
+const proload = once((window: Window) =>
```
kabeaty: Do we still need `window` here? Looks like it isn't used now. Also, typo of `proload` --> `preload`

src/core/client/stream/tabs/Discussions/DiscussionsQuery.tsx
```diff
@@ -21,13 +20,9 @@ const loadDiscussionsContainer = () =>
   import("./DiscussionsContainer" /* webpackChunkName: "profile" */);
 
 // (cvle) For some reason without `setTimeout` this request will block other requests.
-const preloadAndPolyfill = once((window: Window) =>
+const preload = once((window: Window) =>
```
kabeaty: Do we still need `window` here? Looks like it isn't used now.

src/core/client/stream/tabs/Configure/ConfigureQuery.tsx
```diff
@@ -20,13 +19,9 @@ const loadConfigureContainer = () =>
   import("./ConfigureContainer" /* webpackChunkName: "configure" */);
 
 // (cvle) For some reason without `setTimeout` this request will block other requests.
-const preloadAndPolyfill = once((window: Window) =>
+const preload = once((window: Window) =>
```
kabeaty: Do we still need `window` here? Looks like it isn't used now.

src/core/client/ui/components/v2/MatchMedia/MatchMedia.tsx
```diff
@@ -46,6 +58,86 @@ interface Props {
...
+      if (resultRef.current !== result) {
+        resultRef.current = result;
+        rerender();
+      }
+    };
+    return onShadowRootWidthChange(shadowRoot, callback);
+  }, [shadowRoot, minWidth, maxWidth]);
+
+  const minWidthChanged = prevMinWidth !== null && prevMinWidth !== minWidth;
+  const maxWidthChanged = prevMaxWidth !== null && prevMaxWidth !== minWidth;
```
kabeaty: Should this be checking for `prevMaxWidth !== maxWidth` instead?

src/core/client/ui/shadow/ReactShadowRoot.tsx
```diff
@@ -0,0 +1,122 @@
...
+interface Props {
+  /** Root prop allows you to pass a different instance of `react-shadow` */
+  Root?: typeof DefaultReactShadow;
+  /** containerClassName is applied to the container inside of the shadow dom */
+  containerClassName?: string;
+  /** CSS assets to be added */
+  cssAssets?: CSSAsset[];
+  /** CSS assets to be added at the end */
+  customCSSAssets?: CSSAsset[];
+  style?: React.CSSProperties;
```
kabeaty: I think it could be nice to add a comment here for `style` too about how it's used for hiding unstyled content before css is loaded.

src/core/client/ui/shadow/ReactShadowRoot.tsx
```diff
@@ -0,0 +1,122 @@
...
+  return <div {...props} className={cn(props.className, className)} />;
+};
+
+/**
+ * `ReactShadowRootDerived` allows you to have multiple shadow root containers.
+ * In the React Tree it must be a descendant of the main `ReactShadowRoot` element and
+ * it derives all props that are set on `ReactShadowRoot` essentially giving
+ * you the same shadow dom environment with everything setup like CSS assets loading.
+ */
+export const ReactShadowRootDerived: FunctionComponent<DerivedProps> = (
```
kabeaty: I think this name works fine, but other ideas would be `ReactShadowRootDerivedChild` (longer but more explicit about how must be a descendant of main shadow root element) or just `ReactShadowRootChild` (although this might lose some of the clarity that it inherits from `ReactShadowRoot`).

src/core/server/app/helpers/manifestLoader.ts
```diff
@@ -0,0 +1,168 @@
...
+
+import { waitFor } from "coral-common/helpers";
+import logger from "coral-server/logger";
+
+import Entrypoints, { Entrypoint, RawEntrypoint } from "./entrypoints";
+
+/**
+ * Sometimes webpack produces an invalid manifest with empty entrypoints.
+ * We retry whenever we encounter this and usually webpack fixes this.
+ * This is a hardlimit on how many retires we should do.
```
kabeaty: typo, `retires` --> `retries`

src/core/server/app/middleware/graphql/apolloServer.ts
```diff
@@ -132,14 +133,17 @@ export const apolloGraphQLMiddleware = ({
     debug: false,
   });
 
-  // Get the GraphQL middleware.
+  // Check request coming from known origin.
+  const corsOptionsDelegate = createCorsOptionsDelegate(options.mongo);
+
+  // Get the GraphQL miqddleware.
```
kabeaty: typo, `miqddleware` --> `middleware`

src/locales/en-US/admin.ftl
```diff
@@ -857,6 +857,11 @@ configure-wordList-suspect-wordListDetailInstructions =
 configure-advanced-customCSS = Custom CSS
 configure-advanced-customCSS-override =
   URL of a CSS stylesheet that will override default Embed Stream styles.
+configure-advanced-customCSS-stylesheetURL = Custom CSS Stylesheet URL
+configure-advanced-customCSS-fontsStylesheetURL = Custom CSS Style Sheet URL for Font Faces
```
kabeaty: Can we change `Style Sheet` --> `Stylesheet` for consistency here? Also, for `configure-advanced-customCSS-containsFontFace` below, I think `stylesheets` should just be singular `stylesheet` (either capitalized like the others, or all lowercased).

docs/docs/cms.md
```diff
@@ -37,8 +37,9 @@ Options object for `createStreamEmbed`:
 - `storyURL` - [string] _(optional)_: URL for the story. May alternately specify via `storyID` or allow Coral to scrape and determine automatically.
 - `accessToken` - [string] _(optional)_: Access token to log in a user via [SSO](/sso).
 - `refreshAccessToken` - [function] _(optional)_: Callback to obtain a new access token when the current one has expired. A parameter `nextAccessToken` is passed as the first argument that should be called with the next access token.
-- `bodyClassName` - [string] _(optional)_: HTML class name to add to the body of the stream embed iframe for CSS targeting.
+- `containerClassName` - [string] _(optional)_: HTML class name to add to the container of the stream embed iframe for CSS targeting.
```
kabeaty: Should this comment just say `container of the stream embed for CSS...` without `iframe`?

docs/docs/css.md
```diff
@@ -35,11 +35,11 @@ You can also navigate to https://github.com/coralproject/talk/blob/main/src/core
 
 ### Custom body class for theming
 
-You can set the class name of the `<body>` tag inside the embed by using the `bodyClassName` parameter when calling `Coral.createStreamEmbed`:
+You can set the class name of the `<body>` tag inside the embed by using the `containerClassName` parameter when calling `Coral.createStreamEmbed`:
```
kabeaty: Should this doc and its header be updated to something more like `You can set the class name of the container inside of the embed by using...` and `Custom container class for theming` without a reference to the `<body>` tag?

src/core/build/createWebpackConfig.ts
```diff
@@ -471,15 +476,56 @@ export default function createWebpackConfig(
...
+            // typography.css must be included in the light dom whereas the other css
+            // could be in the shadow dom. That's why we are treating it here separately.
+            {
+              test: /\/typography\.css$/,
+              use: [
+                {
+                  loader: require.resolve("style-loader"),
+                  options: {
+                    injectType: "linkTag",
+                    /* , insert: insertLinkTag */
```
kabeaty: Why is this commented out?

src/core/client/embed/Coral.ts
```diff
@@ -30,35 +30,51 @@ export interface Config {
   enableDeprecatedEvents?: boolean;
 
   /** Allow setting className of body tag inside iframe */
```
kabeaty: Can this comment be updated to match `containerClassName`?

src/core/client/embed/test/basic.spec.ts
```diff
@@ -1,56 +1,204 @@
...
+        attach: attachStub,
+      };
+      scriptElement!.onload!({} as any);
+      expect(attachStub.called).toBe(true);
+      expect(args && args.storyURL).toBe("http://localhost/canonical");
+    } finally {
+      server.restore();
+    }
   });
   it("should remove iframe", () => {
```
kabeaty: Can the name of this test be updated?

src/core/client/embed/test/basic.spec.ts
```diff
@@ -1,56 +1,204 @@
...
-    link.rel = "canonical";
-    link.href = "http://localhost/canonical";
-    document.head.appendChild(link);
-    const CoralEmbedStream = Coral.createStreamEmbed({
-      id: "basic-integration-test-id",
-    });
-    CoralEmbedStream.render();
-    expect(container.innerHTML).toMatchSnapshot();
-    document.head.removeChild(link);
-    expect(console.warn).not.toHaveBeenCalled();
```
kabeaty, line 41: Should this check still be included in this test? Looks like it is calling a warning about canonical link here when previously we expected it not to?

src/core/client/framework/components/loadables/MarkdownEditor/MarkdownEditor.css
```diff
@@ -561,11 +561,14 @@ $fullscreenZIndex: 10;
       text-shadow: inherit;
     }
   }
+
+  /* @cvle: This part does not seem to be used? */
```
kabeaty: Do you think we can safely remove this then, or should we test further first to confirm?

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -529,10 +518,13 @@ const KeyboardShortcuts: FunctionComponent<Props> = ({ loggedIn }) => {
         return;
       }
 
-      if (loadMoreEvents.includes(e) && pym) {
+      if (loadMoreEvents.includes(e)) {
         // Need to send new height to pym after more comments/replies load
         // in instead of waiting for polling to update it
-        pym.sendHeight();
+
+        // TODO: (@cvle) pym no longer exist, we usually don't need this. However we might need a mechanism like this for amp?
```
kabeaty: Do you recommend we just do some amp testing to confirm this?

src/core/client/stream/App/listeners/OnEmbedSetCommentID.ts
```diff
@@ -1,26 +1,24 @@
...
     super(props);
 
-    // Sets comment id through pym.
-    props.pym.onMessage("setCommentID", (raw) => {
+    props.eventEmitter.on("embed.setCommentID", (value: string) => {
       commitLocalUpdate(this.props.relayEnvironment, (s) => {
-        const id = raw || null;
+        const id = value || null;
         if (s.get(LOCAL_ID)!.getValue("commentID") !== id) {
           s.get(LOCAL_ID)!.setValue(id, "commentID");
```
kabeaty: Do we still need the below, or can it be removed?

src/core/client/embed/StreamEmbed.ts
```diff
@@ -1,203 +1,318 @@
...
+      if (
+        config.commentID ||
+        !getBrowserInfo(window).supports.intersectionObserver
+      ) {
         // If we are going to auto-render this comment stream and we're on a
         // specific comment, render the stream now so we can auto-scroll the
         // user to the embed.
         this.render();
       } else {
+        // Preload assets when nearing embed.
```
kabeaty, line 148: 🎉

kabeaty (changes requested):
Such exciting changes! Looking good overall to me. I left some comments and questions throughout.   

Through manual testing, I also noticed some things that I added to the QA doc. 
Outside of those, I’m also seeing locally: 
- some warnings in the console about “couldn’t fulfill desired order of chunk groups”. Do you know what’s causing this?
- On the “Story with button” page, clicking “Show comments” doesn’t load the comments. I’m seeing `http://undefined/embed/bootstrap` in the console on this page, so looks like it’s not getting the rootURL from config.  
- I also did some searches for “pym” and “iframe” in the code. I’m still seeing a few mentions of both in the code that I think can be removed/updated. For example, in `retrieveSiteFromQuery`, `CallOut.tsx`, `getAriaPoliteMacOSWorkaround`, and a few more files. Also some docs such as in `seo.md`.

package.json
```diff
@@ -466,12 +468,11 @@
   "browsers": [
     ">1%",
     "last 4 versions",
-    "IE 11",
-    "iOS >= 9",
-    "Android >= 4.4.4",
-    "Edge >= 17",
-    "Firefox >= 68",
-    "Chrome >= 49",
+    "iOS >= 10",
```
nick-funk, line 474: Just a suggestion, but as of XCode 12.1, Apple recommends apps to support iOS 12 - 15 presently. Are there clients we know of that are using iOS 10, 11?

src/core/build/createWebpackConfig.ts
nick-funk: They moved it up into the `MiniCssExtractPlugin` 😃

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
alinhle: I readded it.

src/core/build/createWebpackConfig.ts
alinhle: Will be removed

src/core/client/framework/components/loadables/MarkdownEditor/MarkdownEditor.css
alinhle: I think it's safe

src/core/client/stream/App/listeners/OnEmbedSetCommentID.ts
alinhle: Should be removed

src/core/client/embed/test/basic.spec.ts
alinhle: readded!

src/core/common/utils/getHost.ts
```diff
@@ -0,0 +1,4 @@
+export default function getHost(url: string) {
+  const split = url.split("/");
+  return split[2];
+}
```
nick-funk, line 4: can we do something like the following here?

```
const path = new URL(url);
const host = path.hostname;
```

nick-funk (changes requested):
This is an incredible amount of work, thank you so much for all your efforts @cvle !

I am only putting `Request changes` because I found some bugs. I pulled recent changes and the RTE text input still seems a bit weird and after I post a comment, the RTE freezes up and I can't type into it again.

The second issue I ran into is that the live updates websockets can't seem to connect. I'm digging into this to see if I can find out why.

image

I'm sure we'll get these issues resolved, amazing work and thank you again for all your effort into setting this up for all of us!

package.json
alinhle: iOS 10 and iOS11 work just fine without special treatment, do we currently gain anything from removing support of the two versions?

src/core/common/utils/getHost.ts
alinhle: Hmm I guess, though this implementation can be used server side as well, so as long as the test passes, I think it's fine?

alinhle:
> This is an incredible amount of work, thank you so much for all your efforts @cvle !
> 
> I am only putting `Request changes` because I found some bugs. I pulled recent changes and the RTE text input still seems a bit weird and after I post a comment, the RTE freezes up and I can't type into it again.
> 
> The second issue I ran into is that the live updates websockets can't seem to connect. I'm digging into this to see if I can find out why.
> 
> image
> 
> I'm sure we'll get these issues resolved, amazing work and thank you again for all your effort into setting this up for all of us!

This should be addressed now at least during development. Using custom ports confuses some code we introduced for the nginx proxy support.

image

alinhle:
> Such exciting changes! Looking good overall to me. I left some comments and questions throughout.   
> 
> Through manual testing, I also noticed some things that I added to the QA doc. Outside of those, I’m also seeing locally: 
> 
> * some warnings in the console about “couldn’t fulfill desired order of chunk groups”. Do you know what’s causing this?

This means that in one bundle/chunk we asked for one ordering while in another we asked for the opposite order to load the CSS. This should do no harm though, you can try to track down this and fix the ordering.

> * On the “Story with button” page, clicking “Show comments” doesn’t load the comments. I’m seeing `http://undefined/embed/bootstrap` in the console on this page, so looks like it’s not getting the rootURL from config.

Fixed, thanks!

> * I also did some searches for “pym” and “iframe” in the code. I’m still seeing a few mentions of both in the code that I think can be removed/updated. For example, in `retrieveSiteFromQuery`, `CallOut.tsx`, `getAriaPoliteMacOSWorkaround`, and a few more files. Also some docs such as in `seo.md`.

Thanks! I removed all of the `pym` reference and most of the `iframe` as well. We still support AMP which puts Coral in an iframe so sometimes it still is valid.

The `seo.md` must probably adapted, or the section about iframe fully dropped. Probably @losowsky & @tessalt know what to do with it.

src/core/client/ui/shadow/ReactShadowRoot.tsx
alinhle: There is no longer a `ReactShadowRootDerived` component and the ShadowRoot has been refactored into a higher level `Encapsulation` concept as we now also have an `IframeEncapsulation` component.

package.json
nick-funk: Ah, just trying to cut down our testing. The less devices we support, the less we have to test. No worries, we can leave it as-is.

src/core/common/utils/getHost.ts
nick-funk: You can use `URL` class server side as well.

I ran the following in node server side:
```
const rawURL = "https://www.theverge.com/2022/2/2/22914970/facebook-app-loses-daily-users-first-time-earnings";

const url = new URL(rawURL);
console.log("url based hostname:", url.hostname);

const split = rawURL.split("/");
console.log("split based hostname:", split[2]);
```

Got this result:
```
url based hostname: www.theverge.com
split based hostname: www.theverge.com
```

The only reason I suggest it is that URL will validate the link with all edge cases. For instance, web browsers will accept `http:/` as well as `http://`. The `new URL(...)` library does this as well, but the split based method will fail and return 2022 for the above example `rawURL`.

Also, `URL` will throw other errors if the URL is malformed in other weird ways and give an informative message.

alinhle: Ah didn't know it works serverside. I guess I can use  `new URL(url).host` directly then, thanks!

alinhle: Actually no, it doesn't support omitting the protocol like `//domain.com` which browsers do accept and is standardized. `coral-party` repository e.g. is depending on it.

nick-funk: Browsers only accept `//domain.com` to point to `file:////domain.com` though? We use that in `coral-party`?

nick-funk: Anyways, this is fine for now. Not holding a review over a two line change.

kabeaty (approved):

nick-funk (approved):

PR #3798, opened 2021-11-29
Fix/corl 2327 user ban bug

marcushaddon:
## What does this PR do?
This PR should address https://vmproduct.atlassian.net/browse/CORL-2327 by making user ban management more fine grained.

This introduces the `UpdateUserBanInput` input and `UpdateUserBanPayload` type, as well as the `updateUserBan(...)` mutation

## How do I test this PR?
- a "Manage User Ban" modal should now be accessible regardless of a user's status (unblocking site mods from adding additional bans to a user already banned on other sites)
- unscoped moderators should be able to add and remove bans to individual sites for a user (within the same form submission), as well as ban a user across the entire organization, and unban a user across the entire organization
- scoped moderators should be able to add additional single-site bans to users even if that user is already banned on other sites
- scoped mods will see an "All sites" option for banning, which should ban the user from all sites within that moderators scope
- scoped mods will not see a "No Sites" option (as they are not able to remove bans)
- when managing user bans on individual sites, scoped mods should be able to see all sites that a user is banned on, but not be able to unban users on any sites
 
 
## How do we deploy this PR?
I don't think any special considerations should be needed.


src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
```diff
@@ -275,16 +275,18 @@ const ModerateCardContainer: FunctionComponent<Props> = ({
 
   const handleBanConfirm = useCallback(
     async (
+      updateType: string,
```
kabeaty: It doesn't look like `updateType` and `unbanSiteIDs` are used anywhere here. Can they be removed?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -24,12 +26,15 @@ import styles from "./BanModal.css";
 
 interface Props {
   username: string | null;
+  banStatus?: UserStatusChangeContainer_user["status"]["ban"];
```
kabeaty: I think ban sites are already coming through with `userScopes` so it should be possible to just get the sites that way instead of passing through again?

src/core/client/admin/components/UserStatus/UserStatusSitesList.tsx
```diff
@@ -48,44 +52,78 @@ const UserStatusSitesList: FunctionComponent<Props> = ({ viewerScopes }) => {
...
+        unbanSiteIDs.onChange(unbanSiteIDs.value.filter((id) => id !== siteID));
+      }
+    },
+    [banSiteIDs, unbanSiteIDs]
+  );
 
-      selectedIDsInput.onChange(changed);
+  const onToggleSite = useCallback(
+    (id: string, on: boolean) => {
+      if (on) {
```
kabeaty: What do you think about using a ternary operator here instead for readability? Also, what do you think about using `siteID` here and below as well for consistency?

src/core/client/admin/components/UserStatus/UserStatusSitesList.tsx
```diff
@@ -103,31 +141,57 @@ const UserStatusSitesList: FunctionComponent<Props> = ({ viewerScopes }) => {
             (viewerIsScoped && !viewerIsSingleSiteMod)) && (
             <Flex className={styles.sitesToggle} spacing={5}>
               <FormField>
+                {/* MARCUS: can we disable this is user is top level banned? */}
                 <Localized id="community-banModal-allSites">
-                  <RadioButton checked={!showSites} onChange={onHideSites}>
+                  <RadioButton
+                    checked={updateType.value === "ALL_SITES"}
```
kabeaty: What do you think about making these updateTypes into an enum somewhere?

src/core/client/admin/test/community/community.spec.tsx
```diff
@@ -809,20 +827,147 @@ it("remove user ban", async () => {
...
+
+  act(() => {
+    within(modal).getByType("form").props.onSubmit();
+  });
+
   within(userRow).getByText("Active");
   expect(resolvers.Mutation!.removeUserBan!.called).toBe(true);
 });
 
+it.only("manages bans on individual sites", async () => {
```
kabeaty: Looks like you still need to remove the `only` here :)

src/core/client/admin/test/community/community.spec.tsx
```diff
@@ -923,7 +1068,7 @@ it("invites user", async () => {
   });
 });
 
-it("ban user across specific sites", async () => {
+it.skip("ban user across specific sites", async () => {
```
kabeaty: Did you mean to keep this test skipped? Also, this seems very similar to the test for individual sites above. Could they be combined into one test instead?

src/core/server/graph/schema/schema.graphql
```diff
@@ -7466,6 +7466,52 @@ type BanUserPayload {
   clientMutationId: String!
 }
 
+input UpdateUserBanInput {
+  """
+  userID is the id of the user whos ban status should be updated
```
kabeaty: typo: `whos` --> `whose`

src/core/server/graph/schema/schema.graphql
```diff
@@ -8500,7 +8546,14 @@ type Mutation {
     @auth(roles: [ADMIN, MODERATOR])
 
   """
-  removeUserBan will remove an active ban from a User if they have one.
+  updateUserBan will ban a specified user from specified sites (if
```
kabeaty, line 8561: `updateUserBan` is adding and removing bans, right? Can you update the comment here to clarify?

src/core/server/services/users/users.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+          mailer,
+          rejector,
+          tenant,
+          commentor,
+          badUser.id,
+          "Test message",
+          false,
+          [siteA.id]
+        )
+    ).rejects.toThrow(); // TODO: assert correct error
```
kabeaty: Are these still TODOs you're looking into to assert correct errors?

src/core/server/services/users/users.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+  const admin = createUserFixture({
+    tenantID,
+    role: GQLUSER_ROLE.ADMIN,
+  });
+
+  /* eslint-disable-next-line */
+  const userService = require("coral-server/models/user");
+
+  it("rejects updates by non mod users", async () => {
+    const commentor = createUserFixture({
```
kabeaty: nit: can you make this `commenter`?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -67,27 +73,26 @@ const BanModal: FunctionComponent<Props> = ({
...
-        }
-
-        if (isSiteMod && (!selectedIDs || selectedIDs.length === 0)) {
-          return { [FORM_ERROR]: "At least one site must be selected" };
-        }
+        const { banSiteIDs, unbanSiteIDs, updateType } = input;
+
+        const inScope = (siteID: string) =>
+          !moderationScopesEnabled ||
+          !viewerScopes?.sites?.length ||
```
kabeaty: Would it work to just check for `!isSiteMod` in place of lines 79 & 80 here?

src/core/client/admin/test/community/community.spec.tsx
```diff
@@ -672,12 +676,14 @@ it("ban user", async () => {
     within(userRow).getByLabelText("Change user status").props.onClick();
   });
 
-  const popup = within(userRow).getByLabelText(
+  const statusMenu = within(userRow).getByLabelText(
```
kabeaty: Can you update this to be `dropdown` as well, as you did in the other tests, for consistency?

src/core/server/services/users/users.ts
```diff
@@ -1175,6 +1175,140 @@ export async function ban(
   return user;
 }
 
+/**
+ * ban will ban a specific user from interacting with Coral.
```
kabeaty: I think this comment should say `updateUserBan`. Also, it's possible to ban and unban with `updateUserBan`, right? I think it would be good to update to clarify that as well.

src/core/server/services/users/users.ts
```diff
@@ -1175,6 +1175,140 @@ export async function ban(
...
+  banSiteIDs?: string[] | null,
+  unbanSiteIDs?: string[] | null,
+  now = new Date()
+) {
+  // Ensure valid role
+  if (
+    banner.role !== GQLUSER_ROLE.ADMIN &&
+    banner.role !== GQLUSER_ROLE.MODERATOR
+  ) {
+    throw new Error("User not authorized to peform UpdateUserBan"); // Look for specific error
```
kabeaty: It looks like there's still a TODO here around the specific error?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -67,27 +73,26 @@ const BanModal: FunctionComponent<Props> = ({
...
+          viewerScopes?.sites?.some(({ id }) => id === siteID);
+
+        const filteredBans = banSiteIDs?.filter((siteID: string) =>
```
kabeaty: To me, something like `inScopeFilteredBans` would be clearer than just `filteredBans` regarding how the bans are filtered when I'm reading through. What do you think?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -205,9 +213,13 @@ const BanModal: FunctionComponent<Props> = ({
                         Cancel
                       </Button>
                     </Localized>
-                    <Localized id="community-banModal-banUser">
-                      <Button type="submit" ref={lastFocusableRef}>
-                        Ban User
+                    <Localized id="community-banModal-updateBan">
+                      <Button
+                        data-testid="save-ban-status"
```
kabeaty: Is this `data-testid` used anywhere?

src/core/client/admin/components/UserStatus/UserStatusChange.tsx
```diff
@@ -72,37 +61,21 @@ const UserStatusChange: FunctionComponent<Props> = ({
...
-              </Localized>
-            ) : (
-              <Localized id="community-userStatus-ban">
+            {
+              <Localized id="community-userStatus-manageBan">
                 <DropdownButton
+                  data-testid="manage-ban"
                   className={styles.dropdownButton}
-                  disabled={!onBan}
+                  // disabled={banned && viewerIsScoped} MARCUS: When can banned be true?
```
kabeaty: This can just be removed now, right, since the `Manage Ban` is always an option?

src/core/client/admin/components/UserStatus/UserStatusChange.tsx
```diff
@@ -72,37 +61,21 @@ const UserStatusChange: FunctionComponent<Props> = ({
...
-                >
-                  Remove ban
-                </DropdownButton>
-              </Localized>
-            ) : (
-              <Localized id="community-userStatus-ban">
+            {
+              <Localized id="community-userStatus-manageBan">
                 <DropdownButton
+                  data-testid="manage-ban"
```
kabeaty: Is this `data-testid` used anywhere?

src/core/server/services/users/users.ts
```diff
@@ -1175,6 +1175,140 @@ export async function ban(
...
+        newUnbans
+      );
+    }
+  }
+  // if any new bans and rejectExistingCommments, reject existing comments
+  if (newBans && rejectExistingComments) {
+    await rejector.add({
+      tenantID: tenant.id,
+      authorID: targetUser.id,
+      moderatorID: banner.id, // MARCUS: is this going to reject them across all sites?
```
kabeaty: Is this resolved?

src/core/client/admin/components/UserStatus/UserStatus.tsx
```diff
@@ -25,9 +25,9 @@ const render = (className: string, content: React.ReactNode) => (
 );
 
 const UserStatus: FunctionComponent<Props> = (props) => {
+  // MARCUS: possibly create bug ticket
```
kabeaty: Is this resolved?

kabeaty (changes requested):
Looks good overall! One issue that I noticed during manual testing was that I got an error when a user was banned on specific sites and then I tried to ban them on all sites. I got an error in this case that the user was already banned.

I'm also wondering if the text and email message should change in the `No sites` case? It still asks if the viewer is sure that they want to ban the user instead of remove bans for the user. Since it's possible to remove bans with the Specific sites option too, maybe the text should also be updated in this case too to include adding and removing bans?

src/core/client/admin/components/UserStatus/BanModal.tsx
marcushaddon: Doesn't the `Scope` type of `userScopes` refer to a moderation scope for a moderator, and `banStatus` refer to the ban status of a commenter?

src/core/client/admin/components/UserStatus/BanModal.tsx
marcushaddon: oh thats much better!

src/core/client/admin/components/UserStatus/BanModal.tsx
kabeaty: Ah, so in`UserStatusChangeContainer`, what is passed in for `userScopes.sites` is `user.status.ban.sites`. I believe only the viewer has `moderationScopes`, as they are the ones viewing the modal and banning/unbanning in this case, and then the user is the commenter who does/doesn't have a ban status and ban sites.

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -1,16 +1,25 @@
...
-import React, { FunctionComponent, useCallback, useMemo } from "react";
+import React, {
+  FunctionComponent,
+  useCallback,
+  useMemo,
+  useState,
+} from "react";
 import { Field, Form } from "react-final-form";
 
+import { UserStatusChangeContainer_user } from "coral-admin/__generated__/UserStatusChangeContainer_user.graphql";
```
kabeaty: Can you move this import down below all the coral imports but above the `ModalHeader` for consistency with where it's imported in other files?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -187,10 +248,62 @@ const BanModal: FunctionComponent<Props> = ({
...
+                    </Flex>
                   )}
 
+                  <UserStatusSitesList
+                    bannedSites={userBanStatus?.sites || []}
+                    viewerScopes={viewerScopes}
+                    banActive={userBanStatus?.active}
+                    banState={[banSiteIDs, setBanSiteIDs]}
+                    unbanState={[unbanSiteIDs, setUnbanSiteIDs]}
+                    visible={
```
kabeaty: What do you think about just doing a check here in `BanModal` for whether to show `UserStatusSitesList` or null instead of passing it through here to check? I think that will be more readable to understand when it is/isn't shown, and more consistent with the other checks we're doing in this modal for when to show things.

src/core/client/admin/components/UserStatus/UserStatusSitesList.tsx
```diff
@@ -31,64 +22,91 @@ export interface Scopes {
 }
 
 interface Props {
+  visible: boolean;
+  readonly banActive?: Readonly<boolean>;
```
kabeaty: What do you think about just passing `banActive` and `bannedSites` through here as `userBanStatus`, too? I think it would be more readable to keep this consistent.

src/core/server/services/users/users.ts
```diff
@@ -1175,6 +1177,157 @@ export async function ban(
...
+  unbanSiteIDs?: string[] | null,
+  now = new Date()
+) {
+  // Ensure valid role
+  if (
+    banner.role !== GQLUSER_ROLE.ADMIN &&
+    banner.role !== GQLUSER_ROLE.MODERATOR
+  ) {
+    throw new UserForbiddenError(
+      "User not authorized to peform UpdateUserBan",
```
kabeaty: typo `peform` --> `perform`. Also wondering if it would be friendlier to make the error something more like `User not authorized to manage user bans.`

src/core/client/admin/components/UserStatus/UpdateUserBanMutation.ts
```diff
@@ -0,0 +1,43 @@
...
+                  }
+                }
+              }
+            }
+          }
+        }
+      `,
+      variables: {
+        input: {
+          ...input,
```
kabeaty: Can you please update to explicitly define the expected `input` here, so that we know exactly what this mutation expects?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -58,53 +77,96 @@ const BanModal: FunctionComponent<Props> = ({
...
+          inScopeFilteredBans.length + inScopeFilteredUnbans.length === 0
         ) {
-          selectedIDs = [viewerScopes.sites[0].id];
-        }
-
-        if (isSiteMod && (!selectedIDs || selectedIDs.length === 0)) {
-          return { [FORM_ERROR]: "At least one site must be selected" };
+          return {
+            [FORM_ERROR]: (
+              <Localized id="specificSitesSelect-validation">
```
kabeaty: The error message here doesn't seem to match what's in the localization file. Do we want to update the error message to this there too? 

Also, I'm seeing this error when `All sites` and `No sites` are selected if the error was already triggered on `Specific sites` and then you navigate to them. For this reason, I think it makes sense to include the error only when `Specific sites` is selected instead of as a form error here (when you merge in my changes from PR 3804, you should be able to use something similar to how I set it up there, and it should hopefully just resolve this issue).

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -58,53 +77,99 @@ const BanModal: FunctionComponent<Props> = ({
...
     (input) => {
       try {
-        let selectedIDs = input.selectedIDs || [];
+        const inScope = (siteID: string) =>
+          !isSiteMod || viewerScopes?.sites?.some(({ id }) => id === siteID);
+
+        const [banSiteIDs] = banIDsState;
+        const [unbanSiteIDs] = unbanIDsState;
+
+        const inScopeFilteredBans = banSiteIDs?.filter((siteID: string) =>
```
kabeaty: Are these checks just to be sure that ban/unban siteIDs aren't included that are out of a moderator's scope? Since they shouldn't be able to select anything out of their scope. I think that a comment could be helpful here.

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -58,53 +77,99 @@ const BanModal: FunctionComponent<Props> = ({
...
+  const viewerIsAdmin = viewerScopes.role === GQLUSER_ROLE.ADMIN;
+  const viewerIsOrgAdmin =
+    viewerScopes.role === GQLUSER_ROLE.MODERATOR &&
+    !!(!viewerScopes.sites || viewerScopes.sites?.length === 0);
+
+  const [updateType, setUpdateType] = useState<UpdateType>(
+    userBanStatus?.active ? UpdateType.NO_SITES : UpdateType.ALL_SITES
+  );
+
+  const banIDsState = useState<string[]>([]);
```
kabeaty: I'm wondering if it would work better to use `useField` in `UserStatusSitesList` in this case instead of `useState` here. That's kind of how it's being used, right? Then we could just check the input for `banIDs`/`unbanIDs` on form submit, and it wouldn't be needed to pass this state down through to `UserStatusSitesList`.

src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
```diff
@@ -375,26 +395,24 @@ const ModerateCardContainer: FunctionComponent<Props> = ({
...
+          }
+          open={showBanModal}
+          onClose={handleBanModalClose}
+          onConfirm={handleBanConfirm}
+          viewerScopes={{
+            role: viewer.role,
+            sites: viewer.moderationScopes?.sites?.map((s) => s),
+          }}
+          moderationScopesEnabled={settings.multisite}
+          userBanStatus={comment.author.status.ban}
```
kabeaty, line 413: Nice renaming to better clarify what's getting passed through here!

src/core/server/models/user/user.ts
```diff
@@ -1848,6 +1848,9 @@ export async function siteBanUser(
       $addToSet: {
         "status.ban.siteIDs": { $each: siteIDs },
       },
+      $set: {
```
kabeaty, line 1851: Why are we setting this to false here if they are being banned?

src/locales/en-US/admin.ftl
```diff
@@ -1235,12 +1233,18 @@ community-userStatus-changeButton =
 community-userStatus-premodUser = Always pre-moderate
 community-userStatus-removePremod = Remove pre-moderate
 
-community-banModal-areYouSure = Are you sure you want to ban <username></username>?
-community-banModal-consequence =
+community-banModal-allSites-title = Are you sure you want to ban <username></username>?
+community-banModal-specificSites-title = Are you sure you want to manage ban status of <username></username>?
```
kabeaty: Maybe adding a `the` to make this `...to manage the ban status of...` would flow better?

kabeaty (changes requested):
Thanks for making the updates to this! Looking good. I just have a few comments/questions on some of the changes.

Also, it looks like the ban modal is no longer preserving the selection of rejecting existing comments / custom ban message when switching between radio buttons. Can you take a look at this?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -58,169 +72,236 @@ const BanModal: FunctionComponent<Props> = ({
...
+            sendCustomMessage: false,
+            rejectExistingComments: false,
+            customMessage: getDefaultMessage,
+            banSiteIDs: [],
+            unbanSiteIDs: [],
+          }}
+        >
+          {({
+            handleSubmit,
+            submitError,
```
kabeaty: Is this being used anywhere, or can it be removed?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -58,169 +72,236 @@ const BanModal: FunctionComponent<Props> = ({
...
+  const viewerIsAdmin = viewerScopes.role === GQLUSER_ROLE.ADMIN;
+  const viewerIsOrgAdmin =
+    viewerScopes.role === GQLUSER_ROLE.MODERATOR &&
+    !!(!viewerScopes.sites || viewerScopes.sites?.length === 0);
+
   const onFormSubmit = useCallback(
-    (input) => {
+    ({
+      updateType,
+      sendCustomMessage,
```
kabeaty: Is this being used anywhere, or can it be removed?

src/core/client/admin/components/UserStatus/UserStatusChangeContainer.tsx
```diff
@@ -172,17 +164,36 @@ const UserStatusChangeContainer: FunctionComponent<Props> = ({
...
-      void banUser({
-        userID: user.id,
-        message,
-        rejectExistingComments,
-        siteIDs,
-      });
+  const handleUpdateBan = useCallback(
+    (updateType, rejectExistingComments, banSiteIDs, unbanSiteIDs, message) => {
+      switch (updateType) {
+        case "ALL_SITES":
```
kabeaty: What about using the enum here as well?

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -58,169 +72,236 @@ const BanModal: FunctionComponent<Props> = ({
...
-                </HorizontalGutter>
-              </form>
-            )}
-          </Form>
-        </HorizontalGutter>
+                      <Localized id="community-banModal-updateBan">
+                        <Button
+                          type="submit"
+                          ref={lastFocusableRef}
+                          disabled={
```
kabeaty: With the `Save` button disabled this way, if a viewer bans a user across specific sites, then all sites, and then wants to ban across the same specific sites again, they won't be able to do so, as the `Save` button will be disabled in this case. I'm not sure how often this would happen, but it would create an odd state if it does...

src/core/client/admin/components/UserStatus/UserStatusSitesList.css
```diff
@@ -2,10 +2,6 @@
   margin-bottom: var(--spacing-3);
 }
 
-.sitesToggle {
-  margin-bottom: var(--spacing-3);
-}
-
 .validationMessage {
```
kabeaty: Are you still using this, or should it be removed?

kabeaty (commented):
Looking good, and thanks for making changes! I have just a few more small comments on the updates.

marcushaddon:
@kabeaty re: https://github.com/coralproject/talk/pull/3798#discussion_r767861411
To enable submit in this case would require that every checked site in the site ban list be submitted, which they are not. only sites  actively added by the user after selecting Specific Sites are sent with the mutation to allow for showing what sites a user is banned on, without having a site moderator submit updates to site bans for sites that are out of their scope. It's not ideal but I think it would be wiser to see how actual users interact with this component before doing more non-trivial reworking based on speculation.

kabeaty:
re: https://github.com/coralproject/talk/pull/3798#issuecomment-992982273 I'm replying here for future reference that this was resolved by ensuring general bans remove site bans.

kabeaty (approved):
Looks and works great! Thanks for making the changes.

PR #3799, opened 2021-11-30
[CPSRE-205] Enable image push to GCR on commit to develop

dkolkena:
Testing addition of image push to container registry

### DIFF
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -100,7 +100,7 @@ fi
 
 if [ "$1" = "deploy-commit" ]
 then
-  if [ -n "${CIRCLE_SHA1}" ]
+  if [[ -n "${CIRCLE_SHA1}" && -n "${GCR_IMAGE_NAME}" ]]
   then
     deploy_commit
   fi
--- a/scripts/docker.sh
+++ b/scripts/docker.sh
@@ -55,6 +55,15 @@ deploy_branch() {
   docker push "coralproject/talk:${CIRCLE_BRANCH}"
 }
 
+deploy_commit() {
+  SHORT_GIT_HASH=$(echo $CIRCLE_SHA1 | cut -c -6)
+  echo "==> tagging commit ${SHORT_GIT_HASH}"
+  docker tag "coralproject/talk:latest" "${GCR_IMAGE_NAME}:${SHORT_GIT_HASH}"
+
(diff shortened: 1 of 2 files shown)


dkolkena:
Test to verify CI pipeline - do not merge. (Not sure how to add the don't merge tag yet)

scripts/docker.sh
```diff
@@ -55,6 +55,15 @@ deploy_branch() {
   docker push "coralproject/talk:${CIRCLE_BRANCH}"
 }
 
+deploy_commit() {
+  SHORT_GIT_HASH=$(echo $CIRCLE_SHA1 | cut -c -6)
+  echo "==> tagging commit ${SHORT_GIT_HASH}"
+  docker tag "coralproject/talk:latest" "${GCR_IMAGE_NAME}:${SHORT_GIT_HASH}"
```
nick-funk, line 61: What do you think about passing the tag in directly to `docker.sh` as an argument when we call `docker.sh` so that we don't rely on the defaults of `coralproject/talk:latest` here?

I'm sure this works! It could be worthwhile to be specific in the tagging when we build the image instead of relying on an indirect re-tag though.

Let me know your thoughts 😄

nick-funk (approved):
I'm approving as this looks good and CI is showing it works properly. ✅ 

I made a comment, it's more of a discussion/possible cleanup, might be unnecessary though. Have a look, see if that seems reasonable. Feel free to merge regardless.

nick-funk:
> Test to verify CI pipeline - do not merge. (Not sure how to add the don't merge tag yet)

There are A LOT of menus and widgets in this GitHub UI. Not surprised that you missed it. It's on the right side named `Labels`. Click on the gear icon when in the `Conversation` tab of a PR.

image

scripts/docker.sh
dkolkena: I think that's a good idea, but because this is modeled after `deploy_latest()` and `deploy_branch()`, both of which include that string hard-coded already, I think it's a good idea to push this as is and abstract all of them together as a separate improvement if it becomes necessary.

wyattjoh: Thee `coralproject/talk:latest` was chosen intentially as it's the tag that's created when the `docker build ...` happens elsewhere in this script. This script creates many different tags, and it's not in this case linked to the branch name or tag name.

PR #3803, opened 2021-12-03
[CORL-1697]: Enable pre-moderation across single sites

kabeaty:
## What does this PR do?

This adds the ability to configure pre-moderation for all comments across single sites for multi-site tenants. When this single site configuration is selected with one or more sites, those specific sites will have all of their comments pre-moderated. All other sites will have their comments post-moderated. This is for https://vmproduct.atlassian.net/browse/CORL-1697.

This adds `premoderateAllCommentsSites` to the settings. It also adds a `SINGLE_SITES_PRE` option to the `MODERATION_MODE`s.

I thought about adding `moderation` to `Site`s, but ultimately this seemed to make how we saved and retrieved the moderation information more complex. I was also thinking that it's starting to make sense to consolidate premoderation settings into a `PremoderationConfiguration`, with things like `premodLinksEnable`, `premoderateSuspectWords`, and `premoderateAllCommentsSites`. Or maybe a `ModerationConfiguration` that also includes the moderation mode? I didn't, however, make this change in the scope of this PR, but I'd be interested in others' thoughts on if I should, or if perhaps a new issue should be opened to address this.

## How do I test this PR?

- Test that the `SINGLE_SITES` option only shows up for multi-site tenants (`ON`/`OFF` should still be the options for single-site tenants)
- Test that `All sites` and `Off` options still pre-moderate/post-moderate all comments as expected
- Test that `Single sites` pre-moderates all comments for selected sites and not for others. Go to stream and add comments; see that they don't get added to the stream right away but go to pending moderation queue.
- Test that if `Single sites` is selected and no site is added, there is an error state that says at least one site must be selected
- Test that if a story is configured in the stream to pre-moderate all comments, this still works as expected, no matter which tenant option is selected


src/core/client/admin/routes/Configure/sections/Moderation/PreModerateAllCommentsConfig.tsx
```diff
@@ -0,0 +1,142 @@
...
+                You must select at least one site.
+              </ValidationMessage>
+            </Localized>
+          ) : null}
+        </div>
+      )}
+      <Field name="moderation" type="radio" value="POST">
+        {({ input }) => (
+          <RadioButton {...input} id={`${input.name}-POST`} disabled={disabled}>
+            <Localized id="configure-onOffField-off">
```
kabeaty, line 133: Is there any reason not to also use this localization here for the `Off` option, from the on/off field?

src/core/client/admin/routes/Configure/ConfigureRoute.tsx
```diff
@@ -29,6 +30,11 @@ class ConfigureRoute extends React.Component<Props, State> {
     data: Parameters<Props["updateSettings"]>[0]["settings"],
     form: FormApi
   ) => {
+    // This ensures sites aren't saved to premoderateAllCommentsSites
+    // if the SINGLE_SITES_PRE moderation mode isn't selected
+    if (data.moderation !== GQLMODERATION_MODE.SINGLE_SITES_PRE) {
```
nick-funk: Not a big deal, but for consistency sake, would mind if we renamed this  to `SINGLE_SITES_PREMOD`?

src/core/client/admin/routes/Configure/sections/Moderation/PreModerateAllCommentsConfig.tsx
nick-funk: Nah, this is fine. `configure-onOffField-off` is a generic component localization for the entirety of the configure form.

nick-funk (changes requested):
The code here is good, I only have one rename request for the enum.

I can't comment on it cause its empty, but it looks like `src/core/client/admin/routes/Configure/UpdatePremodSiteMutation.ts` isn't used and can be deleted?

src/core/client/admin/routes/Configure/ConfigureRoute.tsx
kabeaty: I was going for consistency here with the `PRE` moderation mode. Are you thinking for consistency with the `PREMOD` status for comments?

nick-funk: Oh, I see. My mistake! Yes, this is good. I see the `PRE` and `POST` in there. We can keep this the way you have it!

nick-funk (approved):

nick-funk (approved):
Re-reviewed, re-tested, approved.

PR #3806, opened 2021-12-07
[CORL-2413]: Add site search to site moderation modal

kabeaty:
## What does this PR do?

This PR updates the site moderator modal for admins to use the site search. In addition, this fixes a bug where if the user selected no sites to assign, the org moderator role would be assigned; instead, the `Assign` button is now disabled if no sites are selected in this modal. These changes are for https://vmproduct.atlassian.net/browse/CORL-2413. 

None

## How do I test this PR?

Log in as an admin. Click on the `Role` change dropdown. Select `Site moderator`. See that if a user already has sites that they can moderate, these are displayed as selected checkboxes. See the site search. If you search and add sites, they should be added to the selected checkboxes. You should be able to select/deselect these checkboxes once you've added a site. Click assign. See that the selected sites are assigned for site moderation for that user.

Also check that if you have no sites selected in the site moderator modal, the `Assign` button is disabled (greyed out) until you have selected at least one site to assign permissions for.

See that if you are logged in as a site moderator and you select the `Role` change dropdown, you still see the site moderator's list of scoped sites to apply to the user. (There's a separate issue open to address making these checkboxes instead.)


src/core/client/admin/components/UserRole/SiteModeratorModal.tsx
```diff
@@ -71,50 +73,56 @@ const SiteModeratorModal: FunctionComponent<Props> = ({
...
+                    <Flex justifyContent="flex-end" itemGutter="half">
+                      <Localized id="community-siteModeratorModal-cancel">
+                        <Button variant="flat" onClick={onCancel}>
+                          Cancel
+                        </Button>
+                      </Localized>
+                      <Localized id="community-siteModeratorModal-assign">
+                        <Button
+                          type="submit"
+                          disabled={submitting || values.siteIDs.length === 0}
```
kabeaty, line 114: The main changes to this file are adding this `disabled` check that at least one site has been selected, as well as adding `data-testids` for the modal and submit button. And removing the `query` prop.

nick-funk (commented):
> See that if you are logged in as a site moderator and you select the Role change dropdown, you still see the site moderator's list of scoped sites to apply to the user. (There's a separate issue open to address making these checkboxes instead.)

I was unable to follow this flow. Are we letting Site Mods promote users to Site Mod? I vaguely remember that once being a feature, but I think we removed it?

nick-funk:
> 



> > See that if you are logged in as a site moderator and you select the Role change dropdown, you still see the site moderator's list of scoped sites to apply to the user. (There's a separate issue open to address making these checkboxes instead.)
> 
> I was unable to follow this flow. Are we letting Site Mods promote users to Site Mod? I vaguely remember that once being a feature, but I think we removed it?

Further testing looks like the Org Mods can't change roles either. I think we broke the role drop down flow somewhere.

nick-funk (changes requested):
Code looks good and the modal flows seem fine. However, something is up with the moderator role management. Only Admins seem to be able to change roles now.

nick-funk:
> Code looks good and the modal flows seem fine. However, something is up with the moderator role management. Only Admins seem to be able to change roles now.

My mistake, I retested and this is working perfectly. Approving!

nick-funk (approved):
This works great! Approved!

PR #3809, opened 2021-12-10
Use Redis as a queue to lock stories for archiving

nick-funk:
## What does this PR do?

Instead of searching for a story sequentially in batches which default to 500 at a time. Use Redis + Bull job system to queue up the batches prior to performing the archiving step. This allows us to query all the stories we want to archive in 1 query instead of 500 separate queries. This prevents repeat queries and eases the load on mongo to find new stories to archive which is an expensive query.

None

## How do I test this PR?

- set these vars in your `.env` file:

   ```
   ENABLE_AUTO_ARCHIVING=true
   AUTO_ARCHIVING_INTERVAL=0,5,10,15,20,25,30,35,40,45,50,55 * * * *
   AUTO_ARCHIVE_OLDER_THAN=60 days
   ```
- create a bunch of comments that are older than the `AUTO_ARCHIVE_OLDER_THAN`
- build and start up Coral
- watch your logs as it queues and archives the stories you created every 5 minutes on the clock
- check your db and see that stories are being archived and comments are moving over
 
## How do we deploy this PR?

Any previous or future archiving pods with `AUTO_ARCHIVING_ENABLED` set to `true` need to update their intervals as such:

```
- name: AUTO_ARCHIVING_INTERVAL
   value: "0,15,30,45 * * * *"
```


src/core/server/cron/autoArchiving.ts
```diff
@@ -35,38 +36,62 @@ export function registerAutoArchiving(
...
+    const log = logger.child({ tenantID: tenant.id }, true);
 
+    const redisKey = `${REDIS_ARCHIVING_QUEUE_KEY}:${tenant.id}`;
     const now = new Date();
-    const dateFilter = new Date(now.getTime() - age);
 
     let count = 0;
     do {
-      const story = await retrieveLockedStoryToBeArchived(
+      const redisRecords = await redis.spop(redisKey, 1);
```
wyattjoh: Rather than using a custom integration with Redis, instead you could introduce a new Bull task that will do this immediatly instead. I suggest this because the bull tasks have a full lifecycle management. In the case that there's a failure, it can try again.

With this current implementation, if there's an error during the archiving process, it will fail without ever trying to repeat it (unless it's batched again I suppose later).

You could further cap the max number of stories that are processed through use of the limter options if that appeals to you too: https://github.com/OptimalBits/bull/tree/develop/docs#rate-limiter

nick-funk: Sweet! I will look into this!

nick-funk: I have updated this to use `bull`, works like a charm. Thanks for the suggestion!

src/core/server/queue/Task.ts
```diff
@@ -37,6 +38,7 @@ export default class Task<T extends TenantResource, U = any> {
     jobIdGenerator = null,
     queue,
     timeout = 30000,
+    attempts = MAX_JOB_ATTEMPTS,
```
wyattjoh: Maybe rename this to `DEFAULT_MAX_JOB_ATTEMPTS`?

wyattjoh: Also, replace the above function that reads `isLastAttempt` with:

```
  return job.attemptsMade >= (job.opts.attempts ?? MAX_JOB_ATTEMPTS) - 1;
```

src/core/server/queue/tasks/autoArchive.ts
```diff
@@ -0,0 +1,101 @@
...
+    log.error({ storyID }, "unable to archive story");
+    throw new Error("unable to archive story");
+  }
+
+  log.debug({ took: timer() }, "attempted archive operation ended");
+};
+
+export type ArchiverQueue = Task<ArchiverData>;
+
+export function createArchiverTask(
```
wyattjoh: This `JOB_NAME` is `archiver`, and all the types in this file say `Archiver`, maybe the file should be named `archiver.ts`?

wyattjoh (approved):

src/core/server/queue/tasks/autoArchive.ts
nick-funk: Renamed 👍

src/core/server/queue/Task.ts
nick-funk: Updated 👍

src/core/server/models/story/story.ts
```diff
@@ -765,6 +765,26 @@ function getMarkStoryForArchivingSetParam(now: Date) {
   };
 }
 
+export async function forceMarkStoryForArchiving(
```
kabeaty, line 768: I think it would be helpful to add comments and/or otherwise refactor/rename for clarity regarding the difference between `forceMarkStoryForArchiving` and `markStoryForArchiving`.

kabeaty (approved):
Looks good and works as expected! I left one small comment.

PR #3811, opened 2021-12-14
[CORL-2381] Persist seen comments in the back-end

nick-funk:
## What does this PR do?

Persists seen comment state in the back-end so it can be a shared state between devices/logins for the same user.

- Adds a `seen` flag to the `Comment` type
- Adds a `markCommentAsSeen` mutation to the `Mutators` list

## How do I test this PR?

- Create the indexes listed in the deploy notes below
- Visit the stream as a logged in user and see that you can key through or refresh the stream and have comments go yellow to white
- Load up on your phone, another tab/browser and sign in as the same user, see that seen state is persisted to your user account
- Check your mongo for records in the `seenComments` collection
- set the following index, the `30` is 30 seconds (source: https://docs.mongodb.com/manual/core/index-ttl/)
    ```
    db.seenComments.createIndex({ lastSeenAt: 1 }, { expireAfterSeconds: 30 });
    ```
- wait 30 seconds
- watch your seenComments records get cleaned up after they're 30 seconds old
 
## How do we deploy this PR?

Create the following indexes in Mongo:

```
db.seenComments.createIndex({ tenantID: 1, storyID: 1, userID: 1 });
db.seenComments.createIndex({ lastSeenAt: 1 }, { expireAfterSeconds: 604800 });
```

Where `604800` is `7 days * 24 hours * 60 minutes * 60 seconds` for a TTL in seconds of how long to keep the seenComments records before deleting them gracefully within Mongo.


src/core/server/models/comment/comment.ts
```diff
@@ -143,6 +143,8 @@ export interface Comment extends TenantResource {
    * undefined, this Comment is not deleted.
    */
   deletedAt?: Date;
+
+  seen?: boolean;
```
kabeaty, line 152: Can you add a comment here for `seen`?

src/core/server/models/seenComments/seenComments.ts
```diff
@@ -0,0 +1,126 @@
+import { v4 as uuid } from "uuid";
+
+import { MongoContext } from "coral-server/data/context";
+import { FindSeenCommentsInput } from "coral-server/graph/loaders/SeenComments";
+
+import { TenantResource } from "../tenant";
+
+export interface SeenComments extends TenantResource {
```
kabeaty, line 16: What do you think about adding some comments here for `lastSeenAt`, `isDeleting`, `comments` etc.?

src/core/server/cron/seenCommentsCleanup.ts
```diff
@@ -0,0 +1,74 @@
...
+  const job = new ScheduledJob(options, {
+    name: `Daily ${NAME}`,
+    cronTime,
+    command: cleanupSeenComments,
+  });
+
+  return { name: NAME, schedulers: [job] };
+}
+
+const cleanupSeenComments: ScheduledJobCommand<Options> = async ({
```
wyattjoh: Why not instead utilize a TTL index in MongoDB?

https://docs.mongodb.com/manual/core/index-ttl/

It's designed to do this automatically.

src/core/server/models/seenComments/seenComments.ts
```diff
@@ -0,0 +1,120 @@
...
+      tenantID,
+      storyID,
+      userID,
+      comments,
+      lastSeenAt: now,
+    });
+  }
+}
+
+export async function retrieveSeenCommentsForDeletion(
```
kabeaty: Can this be removed now?

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
```diff
@@ -419,7 +432,7 @@ export const CommentContainer: FunctionComponent<Props> = ({
       data-testid={commentElementID}
       // Added for keyboard shortcut support.
       data-key-stop
-      data-not-seen={seen || !commentSeenEnabled ? undefined : true}
+      data-not-seen={commentSeenEnabled ? !comment.seen : false}
```
kabeaty: I think this needs to be something like `commentSeenEnabled && comment.seen ? undefined : true`. 

In `KeyboardShortcuts`, in `toKeyStop`, it's just checking whether `data-not-seen` is there or not. So, setting it to `false` here is making it pass the check as not seen, and the `Z` key shortcut therefore isn't working as expected.

kabeaty (changes requested):
Looks good overall! I left a few comments. Also, it looks like the `CommentContainer` styles for the logged-out state are missing some padding now. Can you take a look at this?

src/core/server/models/seenComments/seenComments.ts
nick-funk: Yup! Removed it!

src/core/server/models/seenComments/seenComments.ts
nick-funk: Comments have been added!

src/core/server/models/comment/comment.ts
nick-funk: Yup! Added!

src/core/client/stream/tabs/Comments/Comment/CommentContainer.tsx
nick-funk: I inverted it to be `commentSeenEnabled && !comment.seen ? true : undefined` to make it more in-line with the fact the tag is `data-not-seen`. Other than that, updated and good catch!

nick-funk:
> Looks good overall! I left a few comments. Also, it looks like the `CommentContainer` styles for the logged-out state are missing some padding now. Can you take a look at this?

I have no idea what's going on with the padding, but it's happening on develop. Logging a bug ticket.

nick-funk:
> > Looks good overall! I left a few comments. Also, it looks like the `CommentContainer` styles for the logged-out state are missing some padding now. Can you take a look at this?
> 
> I have no idea what's going on with the padding, but it's happening on develop. Logging a bug ticket.

Found it, also found more bugs in traversal. All should be fixed up now. I noticed it seems to not want to let Z key do anything unless signed in. We might want to revisit this behaviour, but at least this no longer breaks existing padding/traversal flow now!

kabeaty (approved):
Looks great! Thanks for making the updates.

PR #3816, opened 2021-12-15
[CORL-1697]: add premoderateAllSitesComments to settings resolver with default

kabeaty:
## What does this PR do?

This adds `premoderateAllSitesComments` to the settings resolver with a default.


nick-funk (approved):
This works, thanks for quick fix!

PR #3818, opened 2021-12-16
use correct variable for rejectComments

marcushaddon:
## What does this PR do?
This corrects a bug where the BanModal checks a form field instead of a state variable for "rejectComments", and so doesn't submit the correct value.

None.

## How do I test this PR?
Ban a user via the BanModal, selecting "Reject existing comments" and see that existing comments are rejected.
 
 
## How do we deploy this PR?

### DIFF
--- a/src/core/client/admin/components/UserStatus/BanModal.tsx
+++ b/src/core/client/admin/components/UserStatus/BanModal.tsx
@@ -113,7 +113,7 @@ const BanModal: FunctionComponent<Props> = ({
       try {
         onConfirm(
           updateType,
-          input.rejectExistingComments,
+          rejectComments,
           banSiteIDs,
           unbanSiteIDs,
           customizeMessage ? emailMessage : getDefaultMessage
@@ -132,6 +132,7 @@ const BanModal: FunctionComponent<Props> = ({
       emailMessage,
       customizeMessage,
       getDefaultMessage,
+      rejectComments,
     ]
   );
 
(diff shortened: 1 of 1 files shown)


src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -113,7 +113,7 @@ const BanModal: FunctionComponent<Props> = ({
       try {
```
kabeaty: It looks like `input` is no longer being used here. Also, it seems like we are no longer using the form input here at all; can we also just remove lines 178-183 setting initial values for these?

kabeaty (approved):
Looks good and thanks for fixing. I left one small comment about some things that I think can be removed now. Other than that, approved.

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -264,15 +254,16 @@ const BanModal: FunctionComponent<Props> = ({
...
-                  {!!moderationScopesEnabled &&
-                    updateType === UpdateType.SPECIFIC_SITES && (
-                      <UserStatusSitesList
-                        userBanStatus={userBanStatus}
-                        viewerScopes={viewerScopes}
-                        banState={[banSiteIDs, setBanSiteIDs]}
-                        unbanState={[unbanSiteIDs, setUnbanSiteIDs]}
-                      />
-                    )}
+                  {(viewerIsSingleSiteMod ||
```
kabeaty, line 268: I think that the submit button disabled check also needs to be updated for the form now, so that it can't be submitted without the single site mod's site being checked. I think something like this will work:
```
disabled={
  ((updateType === UpdateType.ALL_SITES &&
  viewerIsSingleSiteMod) ||
  updateType === UpdateType.SPECIFIC_SITES) &&
  !pendingSiteBanUpdates
}
```

nick-funk (approved):

PR #3819, opened 2021-12-17
[CORL-2370] Add archiving copy to stream/admin

marcushaddon:
## What does this PR do?

- Injects the `archivingEnabled` and `autoArchiveOlderThanMs` into the `staticConfig` and from there into the `Local` relay store for both admin and stream
- If archiving is enabled, shows the following throughout stream/admin
    - `My Profile > my comments` will indicate that the comments shown are only those that are new enough to not be archived based on the archiving age threshold
    - Dashboard labels will show that they are relative to the archiving threshold and no longer for "all time"
    - User history dashboard will indicate that the comments shown are only those that are new enough to not be archived based on the archiving age threshold

None

## How do I test this PR?

- Enable archiving by setting `MONGODB_ARCHIVE_URI` in your `.env` file
    - I use a second mongo instance using this command: 
     ```
     docker run -d -p 27018:27017 --restart always --name mongo-archive mongo:4.2
     ```
    - I then use the connection string for that second Mongo I spun up for archiving in my `.env` file:
     ```
     MONGODB_ARCHIVE_URI=mongodb://127.0.0.1:27018/coral
     ```
    - The reason I spin up another Mongo is because archiving is meant to be used with archived comments living in a different cluster (mongo instance) than the live comments. This docker setup matches production most closely for local testing.
- Build and run using `npm run build:development` and `npm run start:development`
- Use an external commenting/iframe host (i.e. the `multisite-test-bed`)
    - This is because the `npm run watch` tooling does not properly update the local state and staticConfig injection due to its configuration
- Check that `My Profile > my comments`, the dashboard, and the user history all comments tab show the appropriate archiving specific notifications and labels
    - Further details on what that should be are listed above in `What does this PR do?`
 
## How do we deploy this PR?


src/core/client/stream/test/create.tsx
```diff
@@ -26,6 +27,16 @@ export default function create(params: CreateTestRendererParams) {
       localRecord.setValue(false, "enableCommentSeen");
       localRecord.setValue(false, "enableZKey");
       localRecord.setValue(false, "amp");
+      localRecord.setValue(true, "archivingEnabled");
+      localRecord.setValue(
+        DEFAULT_AUTO_ARCHIVE_OLDER_THAN,
+        "autoArchiveOlderThanMs"
+      );
+      localRecord.setValue(false, "archivingEnabled");
```
kabeaty, line 30: Looks like these values are being set twice?

src/core/client/stream/test/profile/myComments.spec.tsx
```diff
@@ -115,3 +127,27 @@ it("loads more comments", async () => {
...
+    within(testRenderer.root).getByTestID("profile-commentHistory")
+  );
+
+  expect(
+    within(commentHistory).getByTestID("archivedComments-thresholdNotification")
+  ).toBeDefined();
+});
+
+it("doesn't show archived notification when archiving disabled", async () => {
+  const testRenderer = createTestRenderer({ archivingEnabled: false });
```
kabeaty, line 148: We don't need to set `archivingEnabled` to false here, right, since that's the default?

src/core/common/config.ts
```diff
@@ -57,4 +57,16 @@ export interface StaticConfig {
    * teams can access the admin with their Coral local authentication.
    */
   forceAdminLocalAuth: boolean;
+
+  /**
+   * archivingEnabled will be true when the deployment has set a valid Mongo DB
```
kabeaty: nit - Do you mind updating this to `MongoDB` instead of `Mongo DB`? Will help with searching to have correct/consistent.

src/core/common/config.ts
```diff
@@ -57,4 +57,16 @@ export interface StaticConfig {
...
+   * URI for MONGODB_ARCHIVE_URI.
+   */
+  archivingEnabled: boolean;
+
+  /**
+   * autoArchivingIntervalMs is the time in milliseconds that a story will
```
kabeaty: Can you update this comment to say `autoArchiveOlderThanMs` to match what it's for?

src/core/client/admin/routes/Dashboard/sections/Today.tsx
```diff
@@ -30,6 +36,22 @@ interface Props {
...
+  const [{ archivingEnabled, autoArchiveOlderThanMs }] = useLocal<
+    TodayTotalsLocal
+  >(graphql`
+    fragment TodayTotalsLocal on Local {
+      archivingEnabled
+      autoArchiveOlderThanMs
+    }
+  `);
+
+  const seconds = Math.floor(autoArchiveOlderThanMs / 1000);
```
kabeaty, line 48: Since this and the below is reused in a few places, does it make sense to add it to a shared helper somewhere?

src/core/client/stream/test/profile/myComments.spec.tsx
```diff
@@ -115,3 +127,27 @@ it("loads more comments", async () => {
...
+
+it("shows archived notification when archiving enabled", async () => {
+  const testRenderer = createTestRenderer({ archivingEnabled: true });
+  const commentHistory = await waitForElement(() =>
+    within(testRenderer.root).getByTestID("profile-commentHistory")
+  );
+
+  expect(
+    within(commentHistory).getByTestID("archivedComments-thresholdNotification")
+  ).toBeDefined();
```
kabeaty, line 144: What if we get this by the actual text that we expect it to show to check that's as expected? I also think that would be more readable.

src/locales/en-US/admin.ftl
```diff
@@ -1541,10 +1574,103 @@ dashboard-heading-last-updated = Last updated:
...
+    }
+    [year] { $value ->
+      [1] year
+      *[other] years
+    }
+    *[other] unknown unit
+  } average
 dashboard-today-staff-comments = Staff comments
 dashboard-alltime-staff-comments = All time total
+dashboard-alltime-staff-comments-archiveEnabled = { $value } { $unit ->
```
kabeaty, line 1643: Is it possible to reuse the `framework-timeago` that already exists for the places where we are using this in translations? And then a separate translation for the `average`/`total`, etc.?

kabeaty (approved):
Looks good, and works as expected! Approving, with some small comments/questions and potential ideas for refactoring in a few places.

src/core/client/stream/test/create.tsx
nick-funk: Hah! Yup, fixed!

src/core/client/stream/test/profile/myComments.spec.tsx
nick-funk: I like to be explicit when I do tests. Even though it's the default (and I set the default to make future tests easier to setup if you don't need to be explicit). For this test, it is important that it's false, so setting it to false explicitly conveys what the test is testing.

That's just a personal preference of mine. Verbosity is preferable in my opinion when testing, even when it seems redundant.

src/core/common/config.ts
nick-funk: Yup! Fixed!

src/core/common/config.ts
nick-funk: Yup! Fixed!

src/locales/en-US/admin.ftl
nick-funk: I really want to do this! I've tried to nest Localizations in the past, FTL doesn't seem to allow it. It's kind of annoying and I wish I could do it, but I've never successfully managed to do so.

src/core/client/admin/routes/Dashboard/sections/Today.tsx
nick-funk: I generally agree with grouping similar functionality into a shared function. Removes duplicates of the same code. In this case though, I started thinking "what would I name this?" and got `convertMillisecondsToSeconds` and got flashbacks of that time I was 12 and I tried to re-invent DateTime utility functions. Given the specificity for the flooring that makes the rounding not re-usable for accuracy and the simplicity of just dividing by 1000, it seems un-necessary.

I wish I could just put the localization for this into a re-usable component and just use it for both of the notifications, but FTL dislikes it if you try and nest a `<Localized>` in another `<Localized>`... I'll grump about that later.

src/locales/en-US/admin.ftl
kabeaty: Ah...hmm. Would it work to do something like two `span`s beside each other, since that wouldn't be nesting the localizations?

src/core/client/stream/test/profile/myComments.spec.tsx
nick-funk: Can do, but I won't be able to do the whole text, since it's interpolated, but I'll do the first bit that is consistent and searchable. Should get it towards more read-able.

src/locales/en-US/admin.ftl
nick-funk: I agree with your attempt to do what we can to make modular translation components. I really want to do that!

I try not to do multi-span/div translations for when someone localizes based on a language that goes right to left instead of left to right. I want them to be able to localize it seamlessly as one `<Localized>` object that they just update the text + values for.

PR #3822, opened 2021-12-21
[CORL-2416]: Role promotions/demotions for site mods should have checkboxes

kabeaty:
## What does this PR do?

These changes add checkboxes to the role change promotion/demotion modal for site moderators on the Community tab in the admin. This allows site moderators to have more fine-grained control over granting/removing moderation privileges for their site(s). This is for issue https://vmproduct.atlassian.net/browse/CORL-2416.

None

## How do I test this PR?

As a site moderator (both single- and multi-site), promote and demote users on the Community tab in the admin. See that you can check/uncheck to add/remove sites to users' moderation scopes. Test out adding/removing privileges to users that have site moderator privileges outside of the scope of the viewer that you are signed in as.

See that admin user role change experience is still the same, and that it isn't affected by these changes.


src/core/server/services/users/users.ts
```diff
@@ -781,7 +783,7 @@ export async function demoteUser(
     mongo,
     tenant.id,
     userID,
-    viewer.moderationScopes.siteIDs
+    siteIDs
```
nick-funk, line 808: When we demote/promote with these siteIDs are we checking that the `viewer` (user doing the demote/promote) has all of the siteID's in their moderation scopes?

If we haven't done this elsewhere, maybe we want to check similar to the following?
```
if (
    isSiteModerationScoped(viewer.moderationScopes) &&
    !siteIDs.every((siteID) =>
      viewer.moderationScopes?.siteIDs?.includes(siteID)
    )
  ) {
    throw new Error("you are not permitted to promote the user on these sites");
  }
  ```

kabeaty: This is checked in the `SiteModeratorActions` that controls which sites are shown as options for promotion/demotion. So, the `viewer` shouldn't be given the option to promote/demote on sites outside of their scope. But, I agree that it is also a good idea to also check this here just in case. Added...thanks!

nick-funk: Cool, resolving and approving!

nick-funk (approved):
Looks good! Approved!

PR #3833, opened 2022-01-05
Fix/terser webpack memory leak

tessalt:


nick-funk:
Closing as this was to target the old release branch which no longer exists since we decided to hotfix `main` instead. This fix is also already in that hotfix branch.

PR #3846, opened 2022-01-13
[CORL-2305]: Email domain moderation settings for new users

kabeaty:
## What does this PR do?

These changes add the ability to configure email domain moderation settings for new accounts in the admin. They make it possible to add certain email domains that, if new users sign up with an email address that uses these domains, they can be configured to be either always pre-moderated or banned. This is for https://vmproduct.atlassian.net/browse/CORL-2305

To the schema, this adds `emailDomainModeration` to the settings to keep track of the email domain moderation settings for a tenant. For this, it also adds an `EmailDomain` type and `NEW_USER_MODERATION` enum for the `BAN` or `ALWAYS_PREMOD` options. It also adds `emailDomain` to the `Query` type for the configuration form routes. It also adds `createEmailDomain`, `updateEmailDomain`, and `deleteEmailDomain` mutations.

To the models, it adds creating, updating, and deleting email domains.

no

## How do I test this PR?

To test this PR, you can go to the admin --> configure --> moderation tab. Scroll to the bottom of the page, to the new `Email domain` section. See that no table is shown before you have any email domains added, and that there are also no errors. Add a new email domain. See it appear in the table. Check that if you sign up a new user with that email address domain, their account is set to banned/always pre-moderated.

Update the email domain that you added, and see that it's updated as expected. Add/update more email domains. Delete an email domain, and see that it's deleted as expected. See that if you click `Cancel` while adding/updating domains, you are taken back to the `Email domain` configuration section. You should also be taken there after adding/updating an email domain.

Check that the validation works as expected. For the email domain, it should check that there is no `@` included at the beginning, and that there is at least one `.` included in the domain. It should also check that the domain exists, as it's required. You should also see that if you try to add or update to an email domain that is already configured, you will see an error on add/update.

See that if you sign up with an account with an email address domain that's not included in these email domains, everything still works as expected.

Check that you can view the Account History with ban/premod status there, and that it says the ban/premod status was set by the `System`, and there are no errors.
 
## How do we deploy this PR?

n/a


src/core/client/admin/routeConfig.tsx
```diff
@@ -8,27 +8,27 @@ import { Ability } from "./permissions";
 import { createAuthCheckRoute } from "./routes/AuthCheck";
 import CommunityRoute from "./routes/Community";
 import ConfigureRoute from "./routes/Configure";
+import InnerFormLayout from "./routes/Configure/InnerFormLayout";
```
kabeaty, line 11: Refactored to use this new `InnerFormLayout` for all cases similar to adding/editing domains (moderation phases, webhook endpoints, sites adding/editing), since they were all using the same layout already.

src/core/server/services/users/users.ts
```diff
@@ -238,6 +248,23 @@ export async function create(
...
+  // Check the new user's email address against emailDomain configurations
+  // to see if they should be set to banned or always pre-moderated
+  if (tenant.emailDomainModeration) {
+    const newUserEmailDomainModeration = checkForNewUserEmailDomainModeration(
+      user,
+      tenant.emailDomainModeration
+    );
+    if (newUserEmailDomainModeration) {
+      if (newUserEmailDomainModeration === NEW_USER_MODERATION.BAN) {
+        await banUser(mongo, tenant.id, user.id);
```
kabeaty, line 260: Using `banUser` here instead of `ban` since we don't need to check banner roles, existing bans/premod/suspension/warnings, reject existing comments, or send an email in this new account case. (Similar reasoning for why using `premodUser` over `premod` below).

nick-funk:
I'm getting a `DUPLICATE_EMAIL_DOMAIN` when I try and `Edit email domain`, change the Moderation action, and then click `Update`.

image

```
[2022-01-17T17:57:38.840Z] ERROR: coral/42240 on MacBook-Pro.localdomain(/Users/nickfunk/code/talk/src/core/server/graph/plugins/logger.ts:25 in logAndReportError): graphql query error (context=graph, contextID=f585c900-77be-11ec-ba3d-39cadbd3b946, tenantID=41b25510-481d-4498-bc91-a24230615332, report=null)
    err: {
      "message": "DUPLICATE_EMAIL_DOMAIN",
      "name": "DuplicateEmailDomainError",
      "stack": "Object.updateTenantEmailDomain (src/core/server/models/tenant/tenant.ts:527:11)\nrunMicrotasks (<anonymous>)\nObject.updateEmailDomain (src/core/server/services/tenant/tenant.ts:319:19)\nupdateEmailDomain (src/core/server/graph/resolvers/Mutation.ts:343:15)\n",
      "id": "f5870180-77be-11ec-ba3d-39cadbd3b946",
      "context": {
        "pub": {},
        "pvt": {
          "emailDomain": "premod.com"
        }
      }
    }
```

nick-funk:
I noticed that if I use a sub-domain of a banned/pre-modded domain, I can get through the ban/pre-mod check. Do we want to make sure we're filtering by sub-domains as well?

For instance, if I set `nickfunk.com` to be `ban all new commenter accounts` and sign up with `test@test.nickfunk.com`, I will not be banned.

I checked with my email, and if I send an email to `test@test.nickfunk.com` it will send emails to the `nickfunk.com` mailbox. This seems like a convenient way for these domain spammers to sneak through our moderation system. They can set their mailbox rules to allow `*@spam.com` to go to their email. Then they just create a new sub-domain email for every account they sign up i.e.

```
1@1.spam.com
2@2.spam.com
...
```

nick-funk (changes requested):
Found a small bug and a way to bypass the domain filter. Code looks good otherwise. If we can patch these two small items, should be good to go!

kabeaty:
> Found a small bug and a way to bypass the domain filter. Code looks good otherwise. If we can patch these two small items, should be good to go!

Great catches, thank you! I think that we do want to filter by subdomain as well. Updated to address both.

nick-funk (approved):
This all looks good! Great work!

PR #3852, opened 2022-01-21
Use a copy based docker build for develop builds

nick-funk:


docker/dev/Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+# SEE: https://github.com/nodejs/docker-node/blob/a2eb9f80b0fd224503ee2678867096c9e19a51c2/docs/BestPractices.md#non-root-user
+RUN chown -R node /usr/src/app
+USER node
+
+# Node alpine image does not include ssh. This is a workaround for https://github.com/npm/cli/issues/2610.
+RUN git config --global url."https://github.com/".insteadOf ssh://git@github.com/ && \
+    git config --global url."https://".insteadOf ssh://
+
+# Install build static assets and clear caches.
+RUN npm prune --production
```
nick-funk, line 30: On this dev based docker file, we already copied over the build + static assets + node_modules, so we don't need to do anything but prune.

It copies over the node_modules and build files cause I removed them from the corresponding .dockerignore 😄

scripts/docker.sh
```diff
@@ -72,8 +72,17 @@ then
...
+if [ "$1" = "deploy" ]
+then
+  docker build -t coralproject/talk:latest ${ARGS} -f docker/prod/Dockerfile .
+fi
+
+# build the quick dev copy version if this is just a circle CI commit build
+if [ "$1" = "deploy-commit" ]
+then
+  docker build -t coralproject/talk:latest ${ARGS} -f docker/dev/Dockerfile .
+fi
```
nick-funk, line 85: If we're not doing an actual deploy, use the copy based dev dockerfile to not do "yet another build" for the dev image.

nick-funk:
Closing this as this is really just a stop-gap that will never get merged. We can instead replace babel with esbuild/swc instead and all the build slowness will likely go away.

PR #3855, opened 2022-01-21
Enabling DLC and removing workflow deps from dev deploy step

dkolkena:
## What does this PR do?

Adds `docker_layer_caching: true` to the remote docker defaults, enabling layer caching within the CircleCI pipeline.

None.

No ENV variables but it enables a new CircleCI opt-in setting within any job invoking `setup_remote_docker_defaults`.

## How do I test this PR?

Post a commit and check the Circle workflow. `push_to_gcr` should no longer require `lint` or `unit_tests` to begin, and within the job's stdout you should be able to observe the line `---> Using cache` in the first few docker build steps.
 
## How do we deploy this PR?

No deploy necessary, just merge to develop.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,6 +271,8 @@ workflows:
               only:
                 - /^release-.*/
           requires:
+            - lint
+            - unit_tests
             - build
 
   build-test-deploy:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -270,8 +271,6 @@ workflows:
               only:
                 - /^release-.*/
           requires:
-            - lint
-            - unit_tests
```
wyattjoh: Removing these steps could release code that cannot start as an image, linting should only fail for serious issues (including TS related ones).

dkolkena: The idea was to shorten the pipeline time by letting the push_to_gcr step kick off sooner. However, since this step no longer happens on every `develop` commit and only upon a `/^release-.*/`-tagged branch, you're right that the time gains don't justify the added risk. I've reverted the dependency change.

immber (approved):
👍

PR #3858, opened 2022-01-26
Fix/corl 2468 ban modal fixes

marcushaddon:
## What does this PR do?
In the UserBanModal:
- displays "Specific Sites" by default if the user in question has any single site bans
- only displays "No Sites" option if user is currently banned across all sites (and viewer is admin)

None.

No.

## How do I test this PR?
- open the user ban modal for a user with a single site ban and no blanket ban, and observe the the "Specific Sites" tab is selected by default, showing the individual sites on which the user is banned
- Open the UserBanModal as a non-admin user and observe that the "No Sites" option is not present (as currently only admins can remove bans)
 
## How do we deploy this PR?


src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -98,8 +98,16 @@ const BanModal: FunctionComponent<Props> = ({
     viewerScopes.role === GQLUSER_ROLE.MODERATOR &&
     !!(!viewerScopes.sites || viewerScopes.sites?.length === 0);
 
+  const userIsBlanketBanned = !!userBanStatus?.active;
```
kabeaty, line 101: This works fine, but since I don't think we use the term `blanket banned` anywhere else, I would lean toward making this something more like `userIsAllSitesBanned`.

kabeaty (approved):
Looks good and works great! Left one small comment.

PR #3860, opened 2022-02-01
Feat/corl 2469 story drawer updates

marcushaddon:
## What does this PR do?
- Adds a "Moderate Story" button to the info drawer
- Removes meta data display
- Change story mode select to checkbox
- Change checkbox wording to match stream config

None

No.

## How do I test this PR?
Look at https://vmproduct.atlassian.net/browse/CORL-2469 and compare changes.


marcushaddon:
The buttons mentioned in the ticket (Moderate story and Update) are blue in the stream, but using the same color for the button admin side results in this green color, and it seems like the admin is themed differently.

src/core/client/admin/components/StoryInfoDrawer/ModerateStoryButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+  viewer
+}) => {
+  const [{ accessToken }] = useLocal<ModerateStoryButton_local>(graphql`
+    fragment ModerateStoryButton_local on Local {
+      accessToken
+    }
+  `);
+
+  const href = useMemo(() => {
+    let link = getModerationLink({ storyID: id });
```
marcushaddon, line 37: There is a bit of duplicated between this and the stream-side ModerateStreamConainer, but putting that compnent in a shared directory was tricky because of the differing `Local` schema, and so I went with duplicating it for now.

src/core/client/admin/components/StoryInfoDrawer/ModerateStoryButton.tsx
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable */
```
kabeaty: Why is eslint disabled for this file?

src/core/client/admin/components/StoryInfoDrawer/ModerateStoryButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+import { getModerationLink } from "coral-framework/helpers";
+import { withFragmentContainer, useLocal } from "coral-framework/lib/relay";
+import { Ability, can } from "coral-framework/permissions";
+
+import { ModerateStoryButton_local } from "coral-admin/__generated__/ModerateStoryButton_local.graphql";
+import { ModerateStoryButton_settings } from "coral-admin/__generated__/ModerateStoryButton_settings.graphql";
+import { ModerateStoryButton_story } from "coral-admin/__generated__/ModerateStoryButton_story.graphql";
+import { ModerateStoryButton_viewer } from "coral-admin/__generated__/ModerateStoryButton_viewer.graphql";
+
+import * as styles from "./ModerateStoryButton.css";
```
kabeaty: Can we do `import styles from...` here for consistency?

src/core/client/admin/components/StoryInfoDrawer/RescrapeStory.tsx
```diff
@@ -34,7 +34,9 @@ const ScrapeStory: FunctionComponent<Props> = ({ storyID }) => {
             Triggered
           </Localized>
         ) : (
-          <Localized id="storyInfoDrawer-triggerRescrape">Rescrape</Localized>
+          <Localized id="storyInfoDrawer-triggerRescrape">
+            Rescrape Story Metadata
```
kabeaty: The issue looks like it just has `Rescrape Metadata` here, so just wanted to check we want it to say `Rescrape Story Metadata`

src/core/client/admin/components/StoryInfoDrawer/StorySettingsContainer.tsx
```diff
@@ -112,44 +131,19 @@ const StorySettingsContainer: FunctionComponent<Props> = ({
...
-                          ]}
-                          selected={{
-                            value: moderation,
-                            localizationID: localizedModMode(moderation),
-                          }}
-                        />
-                      </Localized>
-                    </Flex>
-
                     {/* SAVE/SUBMIT */}
```
kabeaty: Do we want this button to say `Update` instead of `Save` now to match the stream update button? Or we only wanted to update the styles to match?

src/core/client/admin/components/StoryInfoDrawer/ModerateStoryButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+        className={styles.moderateStory}
+        variant="outlined"
+        color="primary"
+        paddingSize="small"
+        fontSize="extraSmall"
+        href={href}
+        target="_blank"
+        upperCase
+      >
+        Moderate
```
kabeaty, line 69: In the localization file it says `Moderate story`, and here and in the stream it just says `Moderate`. Which do we want for this button?

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerContainer.tsx
```diff
@@ -23,26 +20,16 @@ import StoryStatus from "./StoryStatus";
 export interface Props {
   onClose: () => void;
```
kabeaty: Looks like `onClose` isn't used. Can we remove?

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerContainer.tsx
```diff
@@ -4,17 +4,14 @@ import { graphql } from "relay-runtime";
...
+import { Flex, HorizontalGutter, TextLink } from "coral-ui/components/v2";
 import ArchivedMarker from "coral-ui/components/v3/ArchivedMarker/ArchivedMarker";
 
+import { StoryInfoDrawerContainer_settings } from "coral-admin/__generated__/StoryInfoDrawerContainer_settings.graphql";
 import { StoryInfoDrawerContainer_story } from "coral-admin/__generated__/StoryInfoDrawerContainer_story.graphql";
+import { StoryInfoDrawerContainer_viewer } from "coral-admin/__generated__/StoryInfoDrawerContainer_viewer.graphql";
 
+import ModerateStoryButton from "./ModerateStoryButton";
 import RescrapeStory from "./RescrapeStory";
 import styles from "./StoryInfoDrawerContainer.css";
```
kabeaty: I know this was like this, but do you mind moving the `styles` import down below this section for consistency?

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerQuery.tsx
```diff
@@ -33,6 +39,10 @@ const StoryInfoDrawerQuery: FunctionComponent<Props> = ({
           return <QueryError error={error} />;
         }
 
+        if (!props?.viewer) {
+          return <h1>TODO NO VIEWER?!</h1>;
```
kabeaty: It looks like this case still needs to be handled. Maybe we show a spinner here too?

src/core/client/admin/components/StoryInfoDrawer/StorySettingsContainer.tsx
```diff
@@ -90,13 +88,34 @@ const StorySettingsContainer: FunctionComponent<Props> = ({
...
               submitSucceeded,
               dirtySinceLastSubmit,
             }) => {
               return (
                 <form onSubmit={handleSubmit}>
                   <Flex direction="column">
+                    <Flex direction="row" className={styles.setting}>
+                      {/* PREMOD ALL COMMENTS ENABLED */}
+                      <Field
+                        name="premoderateComments"
```
kabeaty: Can we make the field name here `moderation` and then use a `parse` and `format` like in `PremodConfig.tsx`, for example, to change to/from a boolean? This would be consistent with how we're handling this in other places, and I think also more readable.

kabeaty (changes requested):
Looking good overall! I left a few comments/questions throughout.

I also noticed that if someone clicks `Rescrape Metadata` for an archived story, there is an error. Should this (and other configuration settings) be disabled here for archived stories?

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerQuery.tsx
marcushaddon: I think i can assume that there will be a viewer here since Admin is an authenticated view, so I'm going to just assert not null, but let me know if that seems problematic

src/core/client/admin/components/StoryInfoDrawer/StorySettingsContainer.tsx
marcushaddon: Updated but just want to make sure this is what you were referring to

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerQuery.tsx
marcushaddon: I changed it to that, but are there cases where viewer is null and it't not because the current user is not logged in? @nick-funk

src/core/client/admin/components/StoryInfoDrawer/StorySettingsContainer.tsx
marcushaddon: agreed!

src/core/client/admin/components/StoryInfoDrawer/StorySettingsContainer.tsx
kabeaty: Yep, that's what I was asking about, thanks! One thing I just noticed--it looks like this button is missing localization. Can you please add?

src/core/client/admin/components/StoryInfoDrawer/ModerateStoryButton.tsx
```diff
@@ -0,0 +1,104 @@
...
+    if (
+      accessToken &&
+      settings.auth.integrations.sso.enabled &&
+      settings.auth.integrations.sso.targetFilter.admin
+    ) {
+      link += `#accessToken=${accessToken}`;
+    }
+
+    return link;
+  }, [accessToken, settings, id]);
```
nick-funk: We have this pattern of using `getModerationLink` and then doing the rigamarole of appending an `accessToken` to it.

How do we feel about updating the `getModerationLink` to take an optional `accessToken` parameter so that we can append in this access token more cleanly in one line and not have duplicates of this append logic all over the admin?

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerQuery.tsx
nick-funk: The viewer can be null because you can access one part of the admin (the login page) and have the viewer be null.

If there is no viewer when you're using the StoryInfoDrawerQuery, it is fine to show a spinner or return null. We don't want to be presenting any of the drawer without a valid viewer.

nick-funk (changes requested):
Overall, code looks good! Made some small requests and addressed your TODO question around the null viewer.

src/core/client/admin/components/StoryInfoDrawer/StoryInfoDrawerQuery.tsx
marcushaddon: aha, good to know

src/core/client/admin/components/StoryInfoDrawer/ModerateStoryButton.tsx
marcushaddon: I like that idea. That function does change to a hook in the Shadow DOM PR, and I have a note to update this PR/area of the code once that is merged, maybe it would be easier to do it then?

src/core/client/admin/components/StoryInfoDrawer/StorySettingsContainer.tsx
```diff
@@ -57,9 +48,16 @@ const StorySettingsContainer: FunctionComponent<Props> = ({
   const updateSettings = useMutation(UpdateStorySettingsMutation);
 
   const onSubmit = async (
-    values: Partial<StorySettingsContainer_storySettings>
+    values: Partial<StorySettingsContainer_storySettings> & {
+      premoderateComments?: boolean;
```
kabeaty: Can this be removed now that we're not using it for the field anymore?

kabeaty (approved):
Looks good, thanks for making the updates! I left one more comment, but approved.

nick-funk (approved):
Approving as we can make the refactor I suggested later when shadowDOM changes are merged.

PR #3878, opened 2022-02-16
[CORL-2450]: Refactor HOC to hooks part 1

kabeaty:
## What does this PR do?

These changes refactor the HOCs of `withLocalStateContainer`, `withForm`, `withGetMessage`, `withContext`, `withUIContext`, `withRouter`, and `withFetch`.

They are refactored to hooks as follows:
`withLocalStateContainer` --> `useLocal`
`withForm` --> `useForm` from `react-final-form`
`withGetMessage` --> `useGetMessage`
`withContext` --> `useCoralContext`
`withUIContext` --> `useUIContext`
`withRouter` --> `useRouter` from `found`
`withFetch` --> `useFetch`

Many of the above changes are fairly straightforward. The most significant changes include refactoring class components to function components where needed. In these cases, I erred on the side of changing as little as possible.

Note that I haven't been able to figure out why one test in the `TrapFocus` spec is failing now. When doing manual testing, everything seems to work as expected. I've added a comment to the test and skipped it for now.

None

No

None

## How do I test this PR?

You can test the components that were changed in this PR. The components that were refactored to function components have the most significant changes and warrant the most testing. These include:
`AuthConfigContainer` --> test out making changes to the auth configuration
`MarkdownEditor` --> test out using the markdown editor. It's used in the Configure tab in the stream, as well as in a few places in General configuration in the admin
`Popover` --> test out opening/closing popovers, such as those on the Community tab, for example, for Status and Role
`Popup` --> test out the AuthPop in the stream, as well as the ChangePassword popup
`TrapFocus` --> test out modals and see that when a modal is opened, the focus is trapped inside of it when tabbing around
`NavigationWarningContainer` --> test out that you will get an alert when navigating away from configure with unsaved changes, but not any other time when navigating around in the admin
`AuthCheckRoute` --> test out logging in/out as a user with/without access to the admin. See that you are redirected to login and to the admin as expected
`OIDCConfigContainer` --> test out updating OpenIDConnect
`EditCommentFormContainer` --> test out editing a comment and see that it works as expected. Including testing out that if max characters are changed while you're editing, it will show an error on submit if your comment is over the new max, and it will preserve your comment text. 
`LoginRoute` and `SignInContainer` --> test login


nick-funk:
Something weird is going on with the login window, it pops back up again after I've already successfully logged in.

double-login-prompt

nick-funk (changes requested):
Very thorough and clean code changes! I only found one bug, which was in the auth popup on the stream. Everything else looks and works great!

nick-funk (approved):

PR #3880, opened 2022-02-22
[CORL-2504]: implement virtualized rendering

kabeaty:
## What does this PR do?

These changes integrate `react-virtuoso` into the comments stream and update the comment stream to use virtualized rendering.

As part of that, these changes support the Z key and the alternate oldest view. They also update how media embeds are shown (default size that can then be expanded to be viewed in full).

Also adds a new config option to the admin that controls whether comments are shown in infinite scroll or with one `Load all comments` button after the initial 20 comments.

These changes also update a few test files to use RTL.

This adds `loadAllComments` to the schema, to support the configuration of whether comments should load in infinite scroll or whether a Load all comments button should be shown first.

no

n/a

## How do I test this PR?

Check out and interact with the stream. Add new comments, replies, change sort order, etc.

Test out Z key with Z_KEY and SEEN_COMMENTS enabled. Test Z key to root comments, replies, looping around from the end of the stream back to the beginning, with new comments via subscription, with new replies via subscription, with the Show more replies button when a comment has many replies, from a null current comment with traversal focus.

When you scroll comments up and out of view, they should be marked as seen with Z_KEY enabled. The unmark all button / shift + A should mark all comments on the story as seen, instead of just all of those in view.

Test out alternate oldest view and clicking `Add comment` takes you to the post comment form below the `Load all comments` button (or initial comments if fewer than 20). Test new comments via subscription in alternate oldest view.

Test out comments with gifs, Tweets, Youtube media, and embedded images. Test out the new expansion for this embedded media. Test out hiding/showing it.

Test out the new config in the admin around whether to show comments in infinite scroll or with a Load all comments button.
 
## How do we deploy this PR?

This requires that the story tree backend work is already deployed and story trees generated, for the Z key portion of these changes.


Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM node:14-alpine
 
+ENV NODE_OPTIONS=--max-old-space-size=4096
```
nick-funk: We seem to have this set to 8 GB on CircleCI, do we want to make that update here as well, just to be safe?

`.circleci/config.yml` (Ln 8)
```
NODE_OPTIONS: "--max-old-space-size=8192"
```

nick-funk (approved):

src/core/server/models/seenComments/seenComments.ts
```diff
@@ -84,9 +84,21 @@ export async function markSeenComments(
...
+  now: Date,
+  markAllAsSeen?: boolean
 ) {
-  const comments = reduceCommentIDs(commentIDs, now);
+  let comments;
+  if (markAllAsSeen) {
+    const markAllCommentIDs = (
+      await mongo
+        .comments()
+        .find({ storyID }, { projection: { id: 1 } })
```
nick-funk: This needs to filter by `tenantID` as well. It's a small change, so I made it and pushed it up.

nick-funk (approved):
Approving again, nice work!

I updated the `markSeenComments` action query when `markAllAsSeen` is set to also filter by `tenantID`. Was a one line change, didn't feel necessary to comment and make you update it when I was already in there and is such a minor tweak. It's simply a safety check, the query would still have worked, but we want to always also filter by `tenantID` just to be safe.

marcushaddon (commented):
meant to leave this comment on another PR, disregard

PR #3884, opened 2022-02-28
[CORL-2513]: Integrate React Testing Library with initial test

kabeaty:
## What does this PR do?
This introduces React Testing Library (RTL) into Coral's tests. It also refactors one initial test (`src/core/client/admin/test/configure/organization.spec.tsx`) over to start using React Testing Library.

It also adds a `customRenderWithApp` to be used in tests for RTL rendering for each testing folder. 

It also updates some test helpers so that it's possible to create context separate from creating a test renderer. This is what we'll want to use for our tests as we migrate them over to RTL.

In addition to RTL, these changes bring in:
`eslint-plugin-jest-dom`: should help us catch errors in tests (https://github.com/testing-library/eslint-plugin-jest-dom)
`@testing-library/user-event`: for things like `type` and `click` over `fireEvent` (https://github.com/testing-library/user-event)
`@testing-library/jest-dom`: for custom matchers like `toBeDisabled` (https://github.com/testing-library/jest-dom)

None

No

n/a

## How do I test this PR?

You can test this PR by running the migrated test and seeing that it passes without errors.
Also just looking over other updates.


nick-funk (approved):
Looking great!

marcushaddon (approved):
I'm excited to write front end tests now!

PR #3888, opened 2022-03-10
Fix bad import on NewCommentersConfiguration resolver

nick-funk:
## What does this PR do?

Fixes a bad import on NewCommentersConfiguration resolver.

Was importing from client types instead of the server side
schema types which crashes the server when done with a
production build.

None

No

No indexes added.

## How do I test this PR?

```
npm run build
npm run start
```
 
- See that the build does not fail and that you don't get a bad import error when the server starts up.
 
## How do we deploy this PR?

No special considerations.

### DIFF
--- a/src/core/server/graph/resolvers/NewCommentersConfiguration.ts
+++ b/src/core/server/graph/resolvers/NewCommentersConfiguration.ts
@@ -3,7 +3,7 @@ import * as settings from "coral-server/models/settings";
 import {
   GQLMODERATION_MODE,
   GQLNewCommentersConfigurationTypeResolver,
-} from "core/client/framework/schema/__generated__/types";
+} from "coral-server/graph/schema/__generated__/types";
 
 export const NewCommentersConfiguration: GQLNewCommentersConfigurationTypeResolver<settings.NewCommentersConfiguration> = {
   premodEnabled: ({ premodEnabled }) => premodEnabled,


src/core/server/graph/resolvers/NewCommentersConfiguration.ts
```diff
@@ -3,7 +3,7 @@ import * as settings from "coral-server/models/settings";
 import {
   GQLMODERATION_MODE,
   GQLNewCommentersConfigurationTypeResolver,
-} from "core/client/framework/schema/__generated__/types";
+} from "coral-server/graph/schema/__generated__/types";
```
nick-funk, line 6: This is a gotcha that will get you bad as it is invisible until you do a production build (i.e. `npm run build` and not `npm run build:development` or `npm run watch` and start it up)

This is a reminder to make sure we always use `coral-server/` types when we are in server code. There are types that are nearly (or sometimes exactly) identical for the client side types as well, and if you let VSCode "auto-import" this for you, it will pick the client side types first because `c` in `client` comes before `s` in `server` alphabetically.

This works fine in dev builds cause we have access to both client and server code when building/running. As soon as we do a production build though, the code is running with a bundled js for client and a TS transpiled js output for the server. When the server in a prod build goes looking for client code, it's gone! Cause it's all wrapped up in the client side static asset bundles.

You all probably already know to do this, but double check your import types! VSCode autocomplete will getcha!

That being said, this is an honest mistake and could happen to anyone. Maybe I can get the linter to check this somehow for us.

kabeaty: I agree that we should look into seeing if the linter can catch this for us!

kabeaty (approved):

iambrianchan (approved):

PR #3890, opened 2022-03-16
[CORL-2525]: Replace invalid URL anchor tag with text in comments

kabeaty:
## What does this PR do?
These changes better handle the case where a comment includes an anchor tag with an invalid URL for its href. Instead of throwing an error, it will turn the anchor tag into text. See docs for info about invalid URLs: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

none

no

n/a

## How do I test this PR?
Check out the test and see that it's working as expected.


nick-funk (approved):
Tested this and it even works great with the suspect/banned words highlighting!

PR #3891, opened 2022-03-22
[CORL-2406]: external user profile URL configuration

kabeaty:
## What does this PR do?

These changes add, behind a feature flag of `CONFIGURE_PUBLIC_PROFILE_URL`, the ability to configure an external public profile link that uses either the user's username or id to construct the link. Once configured, it is then shown in the user history drawer beneath the database id. The link text is `External profile URL` and it opens in a new tab.

Within the scope of these changes, all `moderation.spec.tsx` tests have been migrated to React Testing Library. In addition, commented-out tests for the user history drawer have been migrated over to RTL as well and the commented-out tests removed.

Adds `externalProfileURL` configuration to settings and a new feature flag of `CONFIGURE_PUBLIC_PROFILE_URL`.

Yes, `CONFIGURE_PUBLIC_PROFILE_URL`

n/a

## How do I test this PR?

You can see that nothing is shown for this configuration under `Moderation/Users` in the admin or in the user history drawer without the feature flag set.

You can enable the feature flag. See that you can configure the external profile url in the admin under `Moderation/Users`. Try to add invalid strings (without either `$USER_NAME` or `$USER_ID`) for the configuration and see a validation message. With no url configured, even with the feature flag set, you should see nothing in the user history drawer. Save a valid external profile url. Go to the user history drawer again and see that now a link for `External profile URL` is included beneath database id and it links, opening in a new tab, to the url you configured, with either user name or user id substituted in for `$USER_NAME` or `$USER_ID`.
 
## How do we deploy this PR?

Feature flag of `CONFIGURE_PUBLIC_PROFILE_URL` will need to be enabled for any tenant that wants to configure an external profile url


nick-funk (approved):
👍

PR #3894, opened 2022-03-24
Begin refactoring community.spec.tsx to React Testing Library

nick-funk:
## What does this PR do?

Begin refactoring community.spec.tsx to React Testing Library

None

No

No indexes added.

## How do I test this PR?

Run tests, see they pass.
 
## How do we deploy this PR?

No special considerations.


src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+        ),
+      },
+    },
+  });
+});
+
+it("renders the invite button", async () => {
+  await createTestRenderer();
+
+  const inviteUsersButton = await screen.findByTestId("invite-users-button");
```
kabeaty: I think you can use `findByRole("button", { name: "Invite" })` here instead of the `data-testid`. (Same comment for other instances of invite buttons in these tests, too, that are currently using a `data-testid`).

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+  inviteUsersButton.click();
```
kabeaty: Can you update to use `userEvent` here, so `userEvent.click(inviteUsersButton);`? (Also same comment for all other places in these tests where we are clicking).

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+  const modal = await screen.findByTestId("invite-users-modal");
+  expect(modal).toBeDefined();
```
kabeaty: I think it would be better to assert with `toBeVisible()` here to check that the element is in the document and visible. (Not going to leave a comment in each place, but same with other uses of this assertion throughout these tests).

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+  const moderator = users.moderators[0];
+  await createTestRenderer({
+    resolvers: createResolversStub<GQLResolver>({
+      Query: {
+        viewer: () => moderator,
+      },
+    }),
+  });
+
+  // We use `queryByTestId` here because it will throw an error
```
kabeaty: This comment is fine, but since we're including this info in our best practices, I wonder if we also need it in every test here?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+      Query: {
+        users: createQueryResolverStub<QueryToUsersResolver>(
+          () => emptyCommunityUsers
+        ),
+      },
+    }),
+  });
+
+  const selectField = await screen.findByLabelText("Search by role");
+  fireEvent.change(selectField, {
```
kabeaty: We can use `userEvent` here instead. Also, I think this should be selecting a role option, like `Commenters`, as I don't think `Test` is actually an option in this case. So:
```
userEvent.selectOptions(selectField, ["Commenters"]);
```

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+  await screen.findByText("We could not find anyone", { exact: false });
+});
+
+it("can't change viewer role", async () => {
+  await createTestRenderer();
+
+  const viewerRow = await screen.findByTestId(
+    `community-row-${adminViewer.id}`
+  );
+  expect(() => within(viewerRow).getByLabelText("Change role")).toThrow();
```
kabeaty: As discussed regarding best practices, let's go with `expect(within(viewerRow).queryByLabelText("Change role")).not.toBeInTheDocument();` here.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+  const moderator = users.moderators[0];
+  await createTestRenderer({
+    resolvers: createResolversStub<GQLResolver>({
+      Query: {
+        viewer: () => moderator,
+      },
+    }),
+  });
+
+  expect(screen.queryByLabelText("Change user status")).toBeInTheDocument();
```
kabeaty: I think we should update to `expect(screen.getByLabelText("Change user status")).toBeVisible();` here, as we discussed in best practices.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+  const userRow = await screen.findByTestId(`community-row-${user.id}`);
+  const changeRoleButton = within(userRow).getByLabelText("Change role");
+  userEvent.click(changeRoleButton);
+
+  const popup = within(userRow).getByLabelText(
+    "A dropdown to change the user role"
+  );
+
+  const siteModButton = within(popup).getByTestId(
```
kabeaty: I think we can just do `const siteModButton = within(popup).getByRole("button", { name: "Site Moderator" });` here.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+  );
+  siteModButton.click();
+
+  const modal = await screen.findByTestId("site-role-modal");
+
+  // The submit button should be disabled until at least 1 site is selected
+  const submitButton = await screen.findByTestId(
+    "site-role-modal-submitButton"
+  );
+  expect(submitButton.classList.contains("Button-disabled")).toBe(true);
```
kabeaty: We can just check `expect(submitButton).toBeDisabled();` here.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+  const siteSearchField = within(modal).getByTestId("site-search-textField");
```
kabeaty: You should be able to grab this with `const siteSearchField = within(modal).getByRole("textbox", { name: "Search by site name" });` here.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+  userEvent.click(siteSearchField);
```
kabeaty: You shouldn't need to click here, as `userEvent.type` by default clicks the element before typing.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+  const popup = within(userRow).getByLabelText(
+    "A dropdown to promote/demote a user to/from sites"
+  );
+  const siteModButton = within(popup).getByRole("button", {
+    name: "Remove my sites from moderator",
+  });
+  siteModButton.click();
+
+  await waitFor(() => screen.getByTestId("siteModeratorActions-modal"));
```
kabeaty: I think we should update to `await screen.findByTestId("siteModeratorActions-modal");`

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+  const loadMore = screen.getByText("Load More");
+  loadMore.click();
+
+  await waitFor(() =>
+    expect(screen.queryByText("Load More")).not.toBeInTheDocument()
+  );
+
+  // Make sure third user was added.
+  screen.getByText(users.commenters[1].username!);
```
kabeaty: I think it would be good to make this an explicit assertion here, so `expect(screen.getByText(users.commenters[1].username!)).toBeVisible();`, so that it's clear quickly what we are asserting.

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,545 @@
...
+        },
+      },
+    }),
+  });
+
+  const searchField = screen.getByLabelText("Search by username", {
+    exact: false,
+  });
+  userEvent.click(searchField);
+  userEvent.type(searchField, "search{enter}");
```
kabeaty, line 537: Cool!

kabeaty (changes requested):
Tests are looking good! I left a few comments throughout on where I think we could improve the tests.

src/core/client/admin/components/UserRole/UserRoleChangeButton.tsx
```diff
@@ -10,17 +10,20 @@ import styles from "./UserRoleChangeButton.css";
 interface Props extends Omit<PropTypesOf<typeof TranslatedRole>, "container"> {
   active?: boolean;
   onClick: () => void;
+  testID: string;
```
kabeaty: Should this be optional?

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/components/UserRole/UserRoleChangeButton.tsx
nick-funk: Yup! Fixed!

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
nick-funk: 👍

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+
+  const popup = within(userRow).getByLabelText(
+    "A dropdown to change the user role"
+  );
+  const staffButton = await within(popup).findByRole("button", {
+    name: "Staff",
+  });
+  // userEvent.click(...) throws an error here cause the button is set to
+  // pointer-events: none when disabled. So we do the other `.click()` here.
+  staffButton.click();
```
kabeaty: In this case, can we do `fireEvent.click(staffButton);` instead?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+  const popup = within(userRow).getByLabelText(
+    "A dropdown to change the user role"
+  );
+
+  const siteModButton = within(popup).getByRole("button", {
+    name: "Site Moderator",
+  });
+  // userEvent.click(...) throws an error here cause the button is set to
+  // pointer-events: none when disabled. So we do the other `.click()` here.
+  siteModButton.click();
```
kabeaty: In this case, can we do `fireEvent.click(siteModButton);` instead?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+
+  const popup = within(userRow).getByLabelText(
+    "A dropdown to promote/demote a user to/from sites"
+  );
+  const siteModButton = within(popup).getByRole("button", {
+    name: "Remove my sites from moderator",
+  });
+  // userEvent.click(...) throws an error here cause the button is set to
+  // pointer-events: none when disabled. So we do the other `.click()` here.
+  siteModButton.click();
```
kabeaty: In this case, can we do `fireEvent.click(siteModButton);` instead?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+  const popup = within(userRow).getByLabelText(
+    "A dropdown to promote/demote a user to/from sites"
+  );
+
+  const siteModButton = within(popup).getByRole("button", {
+    name: "Site Moderator",
+  });
+  // userEvent.click(...) throws an error here cause the button is set to
+  // pointer-events: none when disabled. So we do the other `.click()` here.
+  siteModButton.click();
```
kabeaty: In this case, can we do `fireEvent.click(siteModButton);` instead?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+  const siteModButton = within(popup).getByRole("button", {
+    name: "Remove my sites from moderator",
+  });
+  // userEvent.click(...) throws an error here cause the button is set to
+  // pointer-events: none when disabled. So we do the other `.click()` here.
+  siteModButton.click();
+
+  await screen.findByTestId("siteModeratorActions-modal");
+  const removeButton = screen.getByRole("button", { name: "Remove" });
+  removeButton.click();
```
kabeaty: Can you update this one too please?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+      },
+    },
+  });
+  await createTestRenderer({
+    resolvers,
+  });
+
+  const userRow = await screen.findByTestId(`community-row-${user.id}`);
+  const changeRoleButton = within(userRow).getByLabelText("Change role");
+  changeRoleButton.click();
```
kabeaty: Can you update this one too please?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -0,0 +1,555 @@
...
+                },
+              };
+          }
+        },
+      },
+    }),
+  });
+
+  const loadMore = screen.getByText("Load More");
+  loadMore.click();
```
kabeaty: Can you update this one too please?

kabeaty (commented):
Thanks for updating! Just left a few more comments around button clicking.

kabeaty (approved):

PR #3910, opened 2022-04-11
[CORL-2523]: propagate jwt verification errors properly

kabeaty:
## What does this PR do?

These changes improve the error thrown if no suitable jwt verifier is found by clarifying that it could be due to a verifier not being enabled or to validation errors. If there are validation errors, those are included along with which verifier they are for. Interested in any input into the error messages--their wording and how they include the errors--to make sure they clarify the potential issues and better help people resolve them.

none

no

n/a

## How do I test this PR?

You can test this by making some errors happen and trying different combinations of SSO/OIDC/local being enabled. You can make some errors happen by just updating the token in the code (removing required fields, etc.). See that if multiple auth integrations are enabled, as long as one has a valid token, there will be no error. However, if none have a valid token, then all token validation errors are included in the error context.


src/core/server/app/middleware/passport/strategies/jwt.ts
```diff
@@ -83,17 +95,29 @@ export async function verifyAndRetrieveUser(
...
+      // First check that verifier is enabled on the tenant
+      if (verifier.enabled(tenant, token)) {
+        // Then check for token validation errors if verifier is enabled
+        const error = verifier.checkForValidationError(token);
+        if (error) {
+          if (validationErrors.length === 0) {
+            validationErrors += error;
+          } else {
+            validationErrors += " " + error;
+          }
```
nick-funk: This is purely my personal preference, I tend to keep lists of error strings and combine them later using `Array.join(", ")` or similar.

```
const validationErrors: string[] = [];
...

if (error) {
   validationErrors.push(error);
}
...

const combinedError = errors.join(", ");
```

nick-funk (approved):
Approved, made a small suggestion, up to you if you think it makes the code cleaner. :)

PR #3928, opened 2022-04-26
[CORL-2401] Update html-to-text so that it parses token URL's properly

nick-funk:
## What does this PR do?

Updates the `html-to-text` library we use for emails so that it no longer fails to encode password reset URL's that contain a token with any of the following characters: `[, '`.

None

No

No indexes were added.

## How do I test this PR?

- Start up the app
- Make sure you have a dev mailer setup as described in the docs: https://docs.coralproject.net/development#email
- Send a forgot password request
- Check your dev mailer (likely on `localhost:9000`)
- See that the email appears properly with a "Click here to reset your password" link in the mail body
- On `Ln 111` of `src/core/server/app/handlers/api/auth/local/forgot.ts` change the `resetURL` to be a link string containing `[`, `,`, `'` characters.
- Send a forgot password request again
- See that the email appears properly with a "Click here to reset your password" link in the mail body
 
## How do we deploy this PR?

No special considerations.


kabeaty (approved):
Looks good and works as expected!

PR #3929, opened 2022-04-27
[CORL-2524] Create a simple import blocker that checks for cross imports

nick-funk:
## What does this PR do?

Prevents you from importint `coral-server` into `src/core/client` source files and vice versa for `coral-client` into `src/core/server`.

Runs as part of our linting.

Configuration is in `importBlocker.json` at the root of the repository.

None

No

No new indexes added.

## How do I test this PR?

- Run `npm run lint`
- Check that the output includes `Blocked imports linter found 0 errors.`
- Add an import from the server to a client source file, the following is a good example:
  ```
  import logger from "coral-server/logger";
  import {
    Cursor,
    createCollection,
  } from "coral-server/models/helpers";
  ```
- Add an import from the client to a server source file, the following is a good example:
  ```
  import { CommentContainer_viewer as ViewerData } from "coral-stream/__generated__/CommentContainer_viewer.graphql";
  ```
- Run `npm run lint` again
- See that the output includes errors from the blocked imports linter and it lists out the files, line numbers, and line context
 
## How do we deploy this PR?

No special considerations needed to deploy.


dennisrjohn:
The import server to client example above works correctly, but the import client to server doesn't fail linting with the example above. In the code, the blockedImports for server is:
"blockedImports": [
        "coral-client/"
      ],

The example above is importing from coral-stream, if I change the blockedImports to:
"blockedImports": [
        "coral-client/",
        "coral-stream/"
      ],
Then linting fails appropriately.

marcushaddon (approved):

PR #3933, opened 2022-05-04
V4 to v7 migration docs

tessalt:
## What does this PR do?

Updates the docs site to include detailed instructions for the v4 to v7+ migration process. Includes moving WEBHOOKS.md and EXTERNAL_MODERATION_PHASES.md to the docs site. Migration docs will be published without sidebar links for beta testing. Main doc is at /migrate-from-talk and links to the other steps.


docs/docs/migrate-auth.md
```diff
@@ -0,0 +1,33 @@
...
+title: "Migrating to v7: Auth strategies"
+---
+
+# Migrating to v7: Auth strategies
+
+Coral v7+ supports more authentication strategies than v4 out of the box with no plugins required. Read more about [social authentication strategies and OIDC (new)](/auth)
+
+## SSO in v7+
+
+In Legacy Coral, SSO authentication requried:
```
nick-funk: typo `requried` -> `required`

docs/docs/migrate-data.md
```diff
@@ -0,0 +1,213 @@
+---
+title: "Migrating to v7: Migrating Data"
+---
+
+## Before you start
+
+Before you start the data migration process, make sure you have the following:
+
+- a tenant ID for the new v7 coral instance you have created (you can find this by inspecting the record in the `tenants` collection in your new mongo instance)
+- a site ID for the new v7 coral instance (inspect the one record in the `sites` collection)
```
nick-funk, line 10: ~~As far as I know, this is not known prior to performing the migration. When Wyatt and myself did the migration steps, we generated new UUID's for these values and used them as inputs to the `coral-importer` and `coral-import-mapper`.~~

~~When you were performing your steps, did you create a v7 instance first and use that tenant/site id instead?~~

docs/docs/migrate-data.md
```diff
@@ -0,0 +1,213 @@
...
+- a mongo instance with a replica of your v4 production data set
+- a local machine or virtual machine with sufficient RAM and processing power to run `mongoexport` and `mongoimport` on your data set. In our experience, large data sets (10+ GB) will require a minimum of 16GB RAM and 8 cores.
+- _before_ you start the migration process, it is a good idea to create several test users (one commenter, one moderator, one admin). Following the data migration, attempt to log in and comment/moderate as these users to verify that account and login related data was migrated correctly, and that authentication has bee configured correctly.
+
+## 1. Obtain JSON files for v4 data
+
+The first step in the data migration process is to get one JSON file for each collection in your v4 data. You will need to use [mongoexport](https://www.mongodb.com/docs/database-tools/mongoexport/) to get these files. The following bash script will create a folder containing the JSON files for each collection. This script may take a long time to run on large data sets.
+
+```bash
+# Set this to the TALK_MONGO_URL used by Coral.
```
nick-funk: I'm sure this was copied from the README, however, it might be helpful to comment this as the "v4 TALK_MONGO_URL" so folks know we're exporting the v4 JSON out during this step and not the v7 instance. I know, this sounds very pedantic, I'm just trying to be almost ridiculously thorough on these steps.

docs/docs/migrate-data.md
```diff
@@ -0,0 +1,213 @@
...
+# make the output directory
+mkdir -p "${CORAL_OUTPUT_DIRECTORY}"
+
+# Set this to the ID of your new v7 Tenant.
+export CORAL_TENANT_ID=""
+
+# Set this to the ID of your v7 Site.
+export CORAL_SITE_ID=""
+
+# This importer tool is designed to work with Coral at thw following migration
```
nick-funk: typo on `thw` -> `the`

docs/docs/migrate-data.md
```diff
@@ -0,0 +1,213 @@
...
+
+coral-importer legacy --dryRun map
+
+```
+
+If this script succeeded, run it a second time, removing the `--dryRun` flag. If it did not succeed, inspect the logs.json file for more information.
+
+If you have multiple values for `provider` in your dataset, you will need to complete this step multiple times, once for each provider. For example, if you have two `provider` values, `auth-1` and `auth-2`, you will first set `CORAL_MAPPER_USERS_SSO_PROVIDER` to `auth-1`, run the `map` command. Once this has succeeded, replace the `users.json` file in the output directory with the `users.json` file created by the `map` command. Then set `CORAL_MAPPER_USERS_SSO_PROVIDER` to `auth-2` and run the command again.
+
+Once you have completed the import mapping process, replace the `users.json` file in your output directory with the `users.json` file created by the mapper and move on to the next step.
```
nick-funk: Since this outputs a new `users.json` every time, how can it be updating the provider on the users? It has no recollection of the previous run, so it doesn't know to "append a provider" to an existing user.

I would run this by Wyatt, if you need clarification, feel free to pull me into the discussion as I believe this is an edge-case we forgot to address.

nick-funk (commented):
I brought up some suggestions and a few areas where I'm not sure if the tools will do what the docs suggest (though the docs are what we _want_ it to do).

@tessalt, let me know what you think!

docs/docs/migrate-data.md
nick-funk: Nevermind, as you pointed out, SRE does this already. Never mind me and carry on, I'm just being overly cautious lol

docs/docs/migrate-data.md
nick-funk: Discussed this with Wyatt. You probably shouldn't validate multiple v4 auth providers into multiple v6 SSO profiles. 

An example to better demonstrate the way you'd port from v4 to v6:

- We have 100 users
- 30 users are set up with an SSO provider `A`.
- 70 remaining users are set up with an SSO provider `B`.
- Ensure that NO users have both provider `A` AND provider `B` in their v4 profiles or you'll get duplicate users in the next step

We need to do the following to make this work:

- Run `coral-import-mapper` targeting provider A
   - set the output dir to `providerA/users.json`
- Run the `coral-import-mapper` target provider B
   - set the output dir to `providerB/users.json`
- Combine the two users.json together

**Now that we have our head wrapped around this example, here are some more weird edge cases that can happen with resolution steps:**

So, what happens if I have users with 2 or more v4 SSO profiles? Don't I want to map them to 2 v6 SSO profiles? 
**A:** No, you don't really want to do this. Pick one SSO provider for v6 and stick with it.

What if I have two providers (`A` and `B`) and the majority are using `A`, but some are using only `B`?
**A:** You'll have to manually handle these users by doing the following:

- Run `coral-import-mapper` targeting provider B
   - set the output dir to `providerB/users.json`
- Import these users into the v6 database
- Tell users to use auth A for this account and have it login to Coral
   - likely involves host side updating of user account to be using auth A instead of auth B for their account
- The user will now be SSO associated to provider A instead of provider B for v6+

nick-funk (approved):

PR #3949, opened 2022-05-12
[CORL-2551]: fix so RTE buttons show as activated in shadow dom

kabeaty:
## What does this PR do?

These changes add an `rteElementID` to support updates to the `isActive` check in this corresponding PR in the `rte`: https://github.com/coralproject/rte/pull/40.

none

no

n/a

## How do I test this PR?

With the changes in this PR (https://github.com/coralproject/rte/pull/40), `npm link` `rte` with this branch (run `npm link path/to/rte` in `talk`) and run. See that if you select text in the editor and bold, italicize, or apply other styles to it, the corresponding button is activated in the RTE. See that if you click to another part of the page, the button is no longer activated. If you select the text again, the button is again activated as expected.

### DIFF
--- a/src/core/client/stream/tabs/Comments/RTE/__snapshots__/RTE.spec.tsx.snap
+++ b/src/core/client/stream/tabs/Comments/RTE/__snapshots__/RTE.spec.tsx.snap
@@ -22,6 +22,7 @@ exports[`renders correctly 1`] = `
     placeholder="Post a comment"
     placeholderClassName="coral coral-rte-placeholder RTE-placeholder"
     placeholderClassNameDisabled=""
+    rteElementID="Coral-RTE"
     sanitizeToDOMFragment={[Function]}
     sanitizeValue={true}
     toolbarClassName="coral coral-rte-toolbar RTE-toolbar RTE-toolbarHidden"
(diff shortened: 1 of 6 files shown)


src/core/client/stream/tabs/Comments/RTE/__snapshots__/RTE.spec.tsx.snap
```diff
@@ -22,6 +22,7 @@ exports[`renders correctly 1`] = `
     placeholder="Post a comment"
     placeholderClassName="coral coral-rte-placeholder RTE-placeholder"
     placeholderClassNameDisabled=""
+    rteElementID="Coral-RTE"
```
nick-funk, line 25: does it make sense to make this a constant and import both here and in `IframeEncapsulation.tsx` to ensure this stays consistent?

nick-funk (approved):
Am approving cause this works and looks good!. Appears that there are some minor test updates necessary now we have a new `rteElement` prop.

Left one small comment to move the element name into a shared const, not necessary, but might make things a tiny bit cleaner.

Let me know if you want me to test again after you release a new RTE and update the version reference here.

Good work!

PR #3958, opened 2022-05-19
add new v4 to v7 migration docs to sidebar

tessalt:
## What does this PR do?

Updates the docs to add the new v4 to v7 docs to the sidebar. Replaces the old https://docs.coralproject.net/migrating-from-4 url which said not to attempt with the new docs, with full instructions. Adds a redirect from the previously unlisted /migrate-from-talk url because Andrew already gave it out to some people. Updates some titles and labels. 

Screen Shot 2022-05-19 at 1 21 14 PM


nick-funk (approved):
Awesome, thank you!

PR #3971, opened 2022-05-30
[CORL-2564] Use review counts for R&R stories in count scripts

nick-funk:
## What does this PR do?

Updates the count scripts to show ratings for Ratings & Reviews stories instead of comment counts.

None

No new feature flags.

No new indexes added.

## How do I test this PR?

- Create a R&R story
- Ensure you embed the count script on the story page and also on another page to reference the story like a main page on a site with links to many stories where you'd want to see comment counts beside the story link
- make some reviews
- See that the count scripts show `Ratings` / `Rating` in their counts
 
## How do we deploy this PR?

No special considerations.


src/core/server/app/handlers/api/story/count.ts
```diff
@@ -35,6 +41,44 @@ interface StoryCountJSONPQuery {
...
+  i18n: I18n,
+  count: number
+) {
+  // Use translated string.
+  const bundle = i18n.getBundle(tenant.locale);
+
+  let html = "";
+
+  if (storyMode === GQLSTORY_MODE.RATINGS_AND_REVIEWS) {
+    html = translate(
```
kabeaty, line 56: What do you think about doing something like:
```
const text =
    storyMode === GQLSTORY_MODE.RATINGS_AND_REVIEWS ? "Ratings" : "Comments";
  const localizationID =
    storyMode === GQLSTORY_MODE.RATINGS_AND_REVIEWS
      ? "comment-count-ratings-and-review"
      : "comment-count";
  const html = translate(
    bundle,
    `<span class="${NUMBER_CLASS_NAME}">${count}</span> <span class="${TEXT_CLASS_NAME}">${text}</span>`,
    `${localizationID}`,
    {
      number: count,
      numberClass: NUMBER_CLASS_NAME,
      textClass: TEXT_CLASS_NAME,
    }
  );
```

This would have less repetition and I think would be clearer the difference between the two.

src/core/server/app/handlers/api/story/count.ts
```diff
@@ -132,12 +163,31 @@ export const countHandler = ({
...
+
+async function calculateStoryCount(
+  mongo: MongoContext,
+  story: Readonly<Story>
+) {
+  if (story.settings.mode === GQLSTORY_MODE.RATINGS_AND_REVIEWS) {
+    const tallies = await retrieveManyStoryRatings(mongo, story.tenantID, [
+      story.id,
+    ]);
+    if (tallies.length === 0) {
```
kabeaty: Maybe just `return tallies.length === 0 ? 0 : tallies[0].count;`?

src/core/server/locales/en-US/common.ftl
```diff
@@ -12,4 +12,11 @@ comment-count =
     *[other] Comments
   }</span>
 
+comment-counts-ratings-and-review =
```
kabeaty: What do you think about naming this `comment-counts-ratings-and-reviews` with plural `reviews` for consistency for how we refer to ratings and reviews elsewhere in the code?

kabeaty (approved):
Looks good and works as expected! Left a few comments for you to update if you agree.

src/core/server/app/handlers/api/story/count.ts
nick-funk: The reason I'm doing this with explicit if/else is that we're likely going to see a need to do this for QA mode as well. Also, I need to do yet another mode for a special SBN count mode. So I want to get it ready for what will be 4 flows instead of just 2.

That being said, I like your idea to mitigate duplication, but in this case, to keep the rendering isolated to what will be separate flows, I'm going this route with a bit of duplication.

src/core/server/app/handlers/api/story/count.ts
nick-funk: Fixed

src/core/server/locales/en-US/common.ftl
nick-funk: Renamed

PR #3972, opened 2022-06-02
[CORL-2529] Show unread counts in comment count scripts

nick-funk:


nick-funk:
Closing as we can't do this for both performance and security reasons.

PR #3973, opened 2022-06-03
[CORL-2545] Show singular site options for banning on single site organizations

nick-funk:
## What does this PR do?

Update ban modal dialog options for single site organizations so that it shows a single ban/unban option when there is only one site in the organization.

None

No

No new indexes.

## How do I test this PR?

- Setup Coral so there is only one site listed under `Admin > Configure > Organization > Sites`
- Create a mod user and log in as them
- Go to `Community` and attempt to ban a user
- See that the flow is simply to ban the user outright, no `All sites` or `Specific sites` options
- Ban them
- Go to `Manage ban` on same user
- See that the flow is simply to unban the user outright
- Go to `Admin > Configure > Organization > Sites` and add a new site
- Go back to `Community` and attempt to ban user again
- See that the flow now has `All sites` and `Specific sites` options, with save button for ban
- Ban them
- Go to `Manage ban` on the same user
- See that the flow now has `All sites` and `Specific sites` options, with save button for unban
- Unban them
 
## How do we deploy this PR?

No special considerations for deploy.


src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
```diff
@@ -410,6 +410,7 @@ const ModerateCardContainer: FunctionComponent<Props> = ({
           moderationScopesEnabled={settings.multisite}
           userBanStatus={comment.author.status.ban}
           userRole={comment.author.role}
+          isMultisite={settings.multisite}
```
kabeaty, line 412: Since we are already passing through `settings.multisite` for `moderationScopesEnabled`, it would be nice to combine these and just pass whether it is multisite through once but use it for both purposes.

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -237,7 +262,12 @@ const BanModal: FunctionComponent<Props> = ({
                   {(viewerIsAdmin ||
                     viewerIsOrgAdmin ||
                     (viewerIsScoped && !viewerIsSingleSiteMod)) && (
-                    <Flex className={styles.sitesToggle} spacing={5}>
+                    <Flex
```
kabeaty: What about just adding `isMultisite` to the conditional check that starts on line 262 for showing the All / Specific / No sites options? Then we wouldn't need to add these `hidden` styles.

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -307,15 +337,23 @@ const BanModal: FunctionComponent<Props> = ({
...
-                    <Localized id="community-banModal-updateBan">
-                      <Button
-                        type="submit"
-                        ref={lastFocusableRef}
-                        disabled={disableForm}
-                      >
-                        Save
-                      </Button>
-                    </Localized>
+                    {isMultisite && (
```
kabeaty: This is somewhat of a personal preference, but I find it clearer to do 
```
isMultisite ? (<Button></Button>) : (<Button></Button>)
```

I also think it could make sense and be cleaner here to just create a separate component called something like `BanModalSubmitButton` here and then determine what text and localization id to use in that component based on whether it's multisite and what the user's ban status is, since those are the main differences. Then the `createSingleSiteSubmitButton` logic could move into that component too.

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -55,8 +54,33 @@ interface Props {
...
+      <Localized id="community-banModal-unban">
+        <Button type="submit" ref={lastFocusableRef} disabled={disabled}>
+          Unban
+        </Button>
+      </Localized>
+    );
+  } else {
+    return (
+      <Localized id="community-banModal-ban">
+        <Button type="submit" ref={lastFocusableRef} disabled={disabled}>
```
kabeaty: Looking at the issue, this button should be changed to use the `alert` color option and be red in this case. Can you please confirm and update?

kabeaty (changes requested):
Works as expected and looks good! I left a few comments throughout.

src/core/client/admin/components/UserStatus/BanModal.tsx
nick-funk: You're correct, I will fix this up!

src/core/client/admin/components/UserStatus/BanModal.tsx
nick-funk: Great suggestion, refactoring now

src/core/client/admin/components/UserStatus/BanModal.tsx
nick-funk: Fixed and incoming shortly

src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
nick-funk: Agreed, will fix

src/core/client/admin/components/UserStatus/BanModal.tsx
```diff
@@ -52,20 +50,75 @@ interface Props {
...
+  // Otherwise, we're doing a single-site ban flow, show appropriate
+  // ban/unban accordingly
+  if (isBanned) {
+    return (
+      <Localized id="community-banModal-unban">
+        <Button
+          type="submit"
+          ref={lastFocusableRef}
+          disabled={disabled}
+          color="alert"
```
kabeaty: Looking at the issue, I think the unban is still supposed to be green and not use the alert color.

src/core/client/admin/components/UserStatus/UserStatusChangeContainer.tsx
```diff
@@ -289,7 +289,7 @@ const UserStatusChangeContainer: FunctionComponent<Props> = ({
           open
           onClose={handleBanModalClose}
           onConfirm={handleUpdateBan}
-          moderationScopesEnabled={moderationScopesEnabled}
+          isMultisite={settings.multisite && moderationScopesEnabled}
```
kabeaty: I think that `moderationScopesEnabled` is just equal to `settings.multisite`, so I don't think we need to check both here.

kabeaty (changes requested):
Looking good! Thanks for the updates. Just left two small notes.

src/core/client/admin/components/UserStatus/BanModal.tsx
nick-funk: Fixed

src/core/client/admin/components/UserStatus/UserStatusChangeContainer.tsx
nick-funk: Fixed

kabeaty (approved):

PR #3983, opened 2022-06-13
[CORL-2562] During sign up or change username, make sure username is unique

nick-funk:
## What does this PR do?

When signing up or changing your username, check that the proposed username is a unique username and not used by anyone else.

Notifies the user with an error if the username is already taken.

None

No

No new indexes

## How do I test this PR?

* try and sign up (local auth) and choose a username that is already taken
   * see that it blocks you from creating the account and notifies you the username is taken
* try and sign up (local auth) and choose a new username that is NOT taken
   * see that you are able to successfully create the new unique user
* enable username changes in Admin via `Configure > Authentication > Commenter account management features > Allow users to > Change their usernames`
* attempt to change username to a name that is already taken
   * see that it blocks you from changing username and notifies you the username is taken
* change the username to something that is NOT taken
   * see that it allows you change to a valid, non-taken username
 
## How do we deploy this PR?

No special considerations


src/core/server/services/users/users.ts
```diff
@@ -277,6 +279,19 @@ export async function processAutomaticPremodForUser(
   }
 }
 
+export async function usernameAlreadyExists(
+  mongo: MongoContext,
+  tenant: Tenant,
+  username: string
+) {
+  const existingUser = await retrieveUserByUsername(mongo, tenant.id, username);
+  if (existingUser) {
```
kabeaty: My preference would be refactoring here to `return !!existingUser;` so it's just one line.

kabeaty (changes requested):
Looks good overall! I left one comment on a potential refactor you could do. 

I also have a question about the functionality of changing the username. After trying to change to a username that already exists, if I update the username that I want to change to, `Save changes` is still disabled. It's necessary to hit `Cancel` to be able to then click `Change` again and update to a different username. What do you think about updating so that `Save changes` doesn't stay disabled if you update the username you want to change to after receiving the error about a username already existing?

src/core/server/services/users/users.ts
nick-funk: Fixed

nick-funk:
> Looks good overall! I left one comment on a potential refactor you could do.

> I also have a question about the functionality of changing the username. After trying to change to a username that already exists, if I update the username that I want to change to, Save changes is still disabled. It's necessary to hit Cancel to be able to then click Change again and update to a different username. What do you think about updating so that Save changes doesn't stay disabled if you update the username you want to change to after receiving the error about a username already existing?


Fixed up the issue with not being able to submit again to try multiple usernames during username change. Now checks for `dirtySinceLastSubmit` to allow multiple submissions if they change the form inputs. Much more usable now, thank you!

kabeaty (approved):

PR #3984, opened 2022-06-16
[CORL-2529] Always show new comment count when `NEW_COMMENT_COUNT` is enabled

nick-funk:
## What does this PR do?

Some users want to be able to see whether they have visited a story and read the comments or not. By always showing the new comments since the user last visited the story (or never visited a story) we can indicated whether a story is new for them or if they have visited it in the past.

If they haven't ever seen a story and there are comments, the count script will show similar to: `3 Comments / 3 New`. This is because the story has never been visited and the 3 comments are both there and new.

If they have loaded a story and there are comments, the count script will show similar to: `22 Comments / 3 New`. This is because the story has been visited and since they last visited, 3 new comments were posted.

This helps commenters figure out whether they have visited a story or not and whether they should catch up on the commenting there.

No schema changes.

No. Any new clients that want this feature will have to enable `NEW_COMMENT_COUNT`. Existing clients using this feature already have `NEW_COMMENT_COUNT` enabled and do not require any updates.

No new indexes.

## How do I test this PR?

* Build and run Coral
* Spin up `multi-site-test` or a similar host client that has count scripts hooked up
* See that the counts are showing, but no new counts are shown
* Enable `NEW_COMMENT_COUNT` with the following mutation:

    ```
     mutation {
       enableFeatureFlag(input: {
         flag: "NEW_COMMENT_COUNT",
         clientMutationId: "1"
      }) {
         flags
       }
     }
    ```
* Reload the page and see that new counts now show
* Visit articles and check that the new comment count always shows and is correct
 
## How do we deploy this PR?

No special considerations for deploy. Existing clients that use this already have the `NEW_COMMENT_COUNT` feature flag set.


src/core/client/count/injectJSONPCallback.ts
```diff
@@ -57,35 +58,34 @@ function createCountElementEnhancer({
   // Update the innerHTML which contains the count and new value..
   element.innerHTML = html;
 
-  if (storyID) {
-    const previousCount = getPreviousCount(storyID);
-    if (previousCount !== null && previousCount < currentCount) {
```
kabeaty, line 62: Should we keep the `previousCount < currentCount` part of the check here? During my manual testing of this, I saw something funky like `-1 New` show up. Or maybe we check that `newCount` isn't somehow less than 0 and make it 0 if it is?

kabeaty (commented):
Looks good! I just have one question that I left.

src/core/client/count/injectJSONPCallback.ts
nick-funk: I haven't touched this in a while, but it looks like I re-added the `previousCount < currentCount` last time I was here.

nick-funk: I added a new check that only computes the diff if the `previousCount < currentCount`, otherwise the count is 0. This should allow us to always show the count as well as ensure it doesn't return a negative value similar to how it used to operate.

kabeaty (approved):

yogeshbeniwal:
@nick-funk Is there any additional settings required to use this feature? Using above suggested mutation is returning below error

`
{
  "error": {
    "errors": [
      {
        "message": "Expected type FEATURE_FLAG!, found \"NEW_COMMENT_COUNT\". Did you mean the enum value NEW_COMMENT_COUNT?",
        "locations": [
          {
            "line": 2,
            "column": 35
          }
        ],
        "extensions": {
          "code": "GRAPHQL_VALIDATION_FAILED"
        }
      }
    ]
  }
}
`

PR #3989, opened 2022-06-21
Make regenerating story trees run as a job with a job status reporter API

nick-funk:
## What does this PR do?

Moves the regenerate story trees functionality into a job processor queue.

Adds an api router at `<CORAL_URL>/api/jobStatus` that can be called with the `name`: `"regenerateStoryTrees` and return the progress of the story tree generation.

Adds optional `disableCommenting` and `disableCommentingMessage` fields to the `RegenerateStoryTreesInput` type.

No new environment variables.

No new indexes.

## How do I test this PR?

- Start up Coral
- Create some stories or delete the `tree` field off of some stories
- Call the regenerate story trees graphql query to `<CORAL_URL>/api/graphql` with:
     ```
     mutation {
       regenerateStoryTrees(input: {
         clientMutationId: "1",
         disableCommenting: true,
         disableCommentingMessage: "doing story tree regeneration, closing comments while I think really hard"
       }) {
         accepted
         jobID
       }
     }
     ```
- Remember the `jobID` this returns, you'll need it in a moment
- Wait a minute or so, the job takes a bit to get going and start posting progress since it does a count to check how many stories it's going to update so it can report accurate progress
- Call `<CORAL_URL>/api/jobStatus` authenticated with an admin user token with the payload until you see that the percentage is `100`
    - NOTE: this is rate limited, so you can only call it twice a minute
    ```
    {
       "name": "regenerateStoryTrees"
       "jobID": "<YOUR_JOB_ID>"
    }
    ```
- Check your stories, see that they have correct `tree` fields generated for all stories under the tenant.
 
## How do we deploy this PR?

Requires the same deployment process at the 7.1.0 story tree migration since this is a fix to remedy the story tree generation issues in 7.1.0.

### DIFF
--- a/src/core/server/app/handlers/api/jobStatus.ts
+++ b/src/core/server/app/handlers/api/jobStatus.ts
@@ -1,0 +1,96 @@
+import { AuthenticationError } from "apollo-server-express";
+import Joi from "joi";
+
+import { AppOptions } from "coral-server/app";
+import { validate } from "coral-server/app/request/body";
+import { RequestLimiter } from "coral-server/app/request/limiter";
+import { MongoContext } from "coral-server/data/context";
+import { RequestHandler, TenantCoralRequest } from "coral-server/types/express";
+
+import { GQLUSER_ROLE } from "coral-server/graph/schema/__generated__/types";
+
+const JobStatusQuerySchema = Joi.object().keys({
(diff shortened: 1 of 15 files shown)


marcushaddon (approved):
These changes look good to me, also tested locally and works as advertised!

PR #4000, opened 2022-06-30
Fix/corl 2593 spoiler links

marcushaddon:
## What does this PR do?
This PR fixes a bug where links hidden behind spoiler tags are immediately followed when people click to reveal them.

None.

No.

n/a

## How do I test this PR?
1. create a comment with a link in it
2. highlight the link and mark it as a spoiler
3. click on the spoiler in the created comment, note that the link is not followed when it is revealed, but can be followed with subsequent clicks (or touches if on mobile)
 
 
## How do we deploy this PR?


kabeaty (approved):
Looks great!

PR #4004, opened 2022-07-07
[CORL-2598]: only traverse after loading more comments if Z key clicked button

kabeaty:
## What does this PR do?

This addresses an issue where the keyboard shortcuts was overzealously traversing to the next unseen comment after a load more button is clicked (`Show more replies`, `Load more` or `View new comments`). It was traversing even when the button was just clicked, and we want it to only traverse if the button is clicked by the Z key.

none

no

## How do I test this PR?

Use Z key to interact with the stream. Set traversal focus to a comment. On another tab, create two new replies on different root comments via subscription below that comment with traversal focus. On the first tab, where the new `Show more replies` buttons showed up, click the second/bottom button. See that it reveals the new reply but does NOT affect traversal focus. However, if you then use the Z key, it should still click the topmost `Show more replies` button and reveal its new reply and set traversal focus to that reply.


nick-funk (approved):

kabeaty:
Closing, as these changes were included in https://github.com/coralproject/talk/pull/4010

PR #4007, opened 2022-07-14
Fix/corl 2261 children of rejected comments

marcushaddon:
## What does this PR do?
This PR makes disallows replies to comments that are themselves not rejected, but are descendants of rejected comments.

This adds the optional boolean field `canRespond` to the `Comment` type, which currently just checks for the presence of a `rejectedAncestor` field on the comment document, but could be utilized for other reasons to disallow replies.

No.

n/a

## How do I test this PR?
1. create root comment
2. create replies +2 levels deep from a couple of commenters
3. report a middle child as commenter
4. verify that middle child can be responded to
5. reject middle child as admin
6. verify that ancestor of rejected ancestor cannot be replied to in My Profile
9. report and reject root comment
10. verify that ancestors of root comment (so, all) cannot be replied to in My Profile tab as commenter
11. unreject root comment as admin
12. verify that ancestors of root comment (that are not also ancestors of rejected ancestor) can be replied to in Comment Stream AND My Profile tab as commenter
13. verify that ancestors of rejected ancestor (from step 3) reply still cannot be replied to
14. unreject rejected ancestor
15. verify that ancestors of now un-rejected ancestor can be replied to
 
## How do we deploy this PR?


src/core/client/stream/tabs/Profile/MyComments/HistoryCommentContainer.tsx
```diff
@@ -75,6 +75,7 @@ const enhanced = withFragmentContainer<Props>({
       id
       body
       createdAt
+      ...HistoryCommentFooterContainer_comment
```
kabeaty: It looks like this is already included here down at the bottom?

src/core/server/graph/resolvers/Comment.ts
```diff
@@ -64,6 +64,13 @@ export const Comment: GQLCommentTypeResolver<comment.Comment> = {
 
     return canModerate(ctx.user, { siteID: c.siteID });
   },
+  canRespond: (c, input, ctx) => {
```
kabeaty: It looks like we can remove `input` and `ctx` here since they're unused.

src/core/client/stream/test/comments/stream/banned.spec.tsx
```diff
@@ -78,6 +78,7 @@ afterAll(() => {
 });
 
 it("disables comment stream", async () => {
+  /* eslint-disable */
```
kabeaty: Why is this needed here?

src/core/server/stacks/rejectComment.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
kabeaty: Why is this needed here?

src/core/server/graph/schema/schema.graphql
```diff
@@ -3761,6 +3761,11 @@ type Comment @cacheControl(maxAge: 5) {
   """
   canModerate: Boolean!
 
+  """
+  canRespond returns true if the comment is able to be responded to.
+  """
+  canRespond: Boolean!
```
kabeaty: I think that naming this `canReply` or `canReplyTo` would be more consistent with the naming that we're using elsewhere in the code. What do you think?

kabeaty (commented):
Looks good overall and works as expected! I left a few small comments throughout. As discussed, it would also be great to have some frontend tests for this functionality.

src/core/client/stream/test/comments/permalink/permalinkViewRejectedComment.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+  commentWithRejectedReply,
+  replyableComment,
+  settings,
+  stories,
+  unreplyableComment,
+} from "../../fixtures";
+
+const story = stories[0];
+
+function customRenderWithContext(context: CoralContext) {
```
kabeaty: It shouldn't be necessary to redefine `customRenderAppWithContext` here. You should be able to import it from `coral-stream/test/customRenderAppWithContext`.

You'll just need to add something like this to `src/core/client/stream/test/comments/permalink/create.ts`. If you look at `src/core/client/stream/test/comments/create.ts`, you can use it as an example too:
```
export function createContext(params: CreateTestRendererParams) {
  return createTopLevelContext({
    ...params,
    initLocalState: (localRecord, source, environment) => {
      if (params.initLocalState) {
        params.initLocalState(localRecord, source, environment);
      }
    },
  });
}
```

src/core/client/stream/test/fixtures.ts
```diff
@@ -869,6 +870,49 @@ export const storyWithDeepestReplies = denormalizeStory(
   )
 );
 
+export const replyableComment: GQLComment = {
+  ...comments[0],
+  canReply: true,
```
kabeaty, line 888: Isn't `canReply` already set to `true` on the base comment? Can we just use `comments[0]` in the test instead of redefining it here?

src/core/client/stream/test/comments/permalink/permalinkViewRejectedComment.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+    unreplyableComment: disabledComment,
+  };
+};
+
+it("allows replies to comments with canReply = true", async () => {
+  const {
+    replyableComment: replyableCommentElement,
+  } = await createTestRenderer();
+
+  const replyButton = await within(replyableCommentElement).findByTestId(
```
kabeaty: Since this is a button, we can use the button role to find this element and assert against it. This is the top priority way to query for elements since it uses attributes that users interact with, querying by accessibility roles and names. Something like this: `screen.getByRole("button", { name: "Reply to comment by Markus" });`. Same with the next test too. Refactoring to query these elements this way also means that we shouldn't need to find the `replyableComment` and `unreplyableComment` to query within them anymore.

src/core/server/stacks/createComment.ts
```diff
@@ -247,6 +250,19 @@ export default async function create(
     );
   }
 
+  const ancestors = await retrieveManyComments(
```
kabeaty: I think we only need to check for a rejected ancestor if the comment has a `parent` and/or `ancestorIDs`. Should we add an `if` for that here so we aren't checking every time since we already have that info?

src/core/server/stacks/rejectComment.ts
```diff
@@ -82,4 +84,25 @@ const rejectComment = async (
...
+  const children = await mongo
+    .comments()
+    .find({
+      ancestorIDs: commentID,
+    })
+    .toArray();
+
+  const childIDs = children.map(({ id }) => id);
+
+  await mongo
```
kabeaty, line 104: Should we update to only do this if there are `childIDs`?

src/core/server/stacks/approveComment.ts
```diff
@@ -69,4 +71,33 @@ const approveComment = async (
...
+  mongo: MongoContext
+) => {
+  const children = await mongo
+    .comments()
+    .find({
+      parentID: commentID,
+    })
+    .toArray();
+
+  const nonRejectedChildIDs = children
```
kabeaty, line 89: Should we update to only do the rest of this if there are `children`?

kabeaty (changes requested):
Thanks for adding the tests! I left a few comments on the tests where I think things can be simplified, and then a few other comments throughout.

src/core/client/stream/test/fixtures.ts
marcushaddon: that field is set to true on the base comment, but not necessarily so. that array is just an array of comments, not "comments that can be replied to". I prefer to make things explicit rather than implicit to avoid tight coupling (I dont know why someone would change the values in that array, but if they made some valid change to one of them and a bunch of tests broke, I'd call that tight coupling)

src/core/server/stacks/createComment.ts
marcushaddon: good catch!

src/core/client/stream/test/comments/permalink/permalinkViewRejectedComment.spec.tsx
marcushaddon: Is there a way to find an element by its role (which I agree is better than its test id) but not depend on the arbitrary name of the commenter?

kabeaty: It's also possible to match with regex on substrings, like `name: /Reply to comment by/`, but that will match multiple comments. I think it makes sense to search by the full name of the button, including the commenter's name, because that's what we expect the user to see for the name of the button based on the comment fixtures that we've provided to the test.

marcushaddon: So is it desirable that the test will break if we change the wording of that button?

kabeaty: IMO it makes the tests more readable to have the actual text of the button there, and it's a way to test that translations are applied correctly and what we expect the user to see is what they see. It would break if either 1) the text of the button changes, but I think we'd both want to see that change and it would be fast to fix or 2) the username in the fixture was changed, which seems both unlikely and would also be fast to fix.

src/core/client/stream/test/fixtures.ts
```diff
@@ -869,6 +870,61 @@ export const storyWithDeepestReplies = denormalizeStory(
...
+export const rejectedComment: GQLComment = {
+  ...comments[1],
+  author: {
+    ...comments[1].author!,
+    username: "rejected comment author",
+  },
+  status: GQLCOMMENT_STATUS.REJECTED,
+};
+
+export const unreplyableComment: GQLComment = {
```
nick-funk: Typo: `unreplyable` -> `unrepliable`

I had to check this, but this is indeed the correct spelling, and to boot, it even means what it's supposed to mean!

image

It's also hilarious how it puts a `RARE` marker beside the word cause absolutely nobody is using this word in an every-day sentence haha

P.S. you don't actually have to fix this typo... I just thought it was funny

src/core/server/graph/resolvers/Comment.ts
```diff
@@ -64,6 +64,13 @@ export const Comment: GQLCommentTypeResolver<comment.Comment> = {
 
     return canModerate(ctx.user, { siteID: c.siteID });
   },
+  canReply: (c) => {
+    if (typeof c.rejectedAncestor === "undefined") {
+      return true;
+    }
+
+    return !c.rejectedAncestor;
```
nick-funk, line 72: This is so clever, I didn't even know there was a `rejectedAncestor` computed prop on comments.

src/core/server/models/comment/comment.ts
```diff
@@ -60,6 +60,12 @@ export interface Comment extends TenantResource {
    */
   ancestorIDs: string[];
 
+  /**
+   * rejectedAncestor indicates that a comment somewhere up this comment's
+   * chain of ancestors has been rejected.
+   */
+  rejectedAncestor?: boolean;
```
nick-funk, line 72: Oh wait, lol, I didn't know it existed cause you added it.

src/core/server/stacks/approveComment.ts
nick-funk: I am seeing this as resolved, but I don't see a guard clause (or similar check) to block if `!children || children.length === 0`. Am I missing something?

src/core/server/stacks/createComment.ts
nick-funk: Again, I see this as resolved, but I don't see a parent check in the code here. Un-marking as resolved to make sure the fix is pushed up.

nick-funk (changes requested):

src/core/server/stacks/rejectComment.ts
marcushaddon: ive added this check and pushed the change but I'm not seeing it show up here?

src/core/server/stacks/approveComment.ts
marcushaddon: I've pushed a change for this, but this conversation is not being marked as "Outdated", and the change isn't showing up here, but if you just look at the code on this branch on Github you should now see that check

src/core/server/stacks/createComment.ts
marcushaddon: same here, not seeing my change show up in this view, but you can view it at https://github.com/coralproject/talk/blob/fix/CORL-2261-children-of-rejected-comments/src/core/server/stacks/createComment.ts#L252

src/core/server/stacks/approveComment.ts
marcushaddon: https://github.com/coralproject/talk/blob/fix/CORL-2261-children-of-rejected-comments/src/core/server/stacks/approveComment.ts#L85

src/core/client/stream/test/fixtures.ts
marcushaddon: hahaha. im actually down to change this tbh

src/core/server/stacks/rejectComment.ts
marcushaddon: https://github.com/coralproject/talk/blob/fix/CORL-2261-children-of-rejected-comments/src/core/server/stacks/rejectComment.ts#L98 <- requested change

src/core/client/stream/test/comments/permalink/permalinkViewRejectedComment.spec.tsx
```diff
@@ -0,0 +1,55 @@
+import { screen } from "@testing-library/react";
+import customRenderAppWithContext from "coral-stream/test/customRenderAppWithContext";
```
kabeaty: Can you please update the imports in this file to be consistent with how we're importing elsewhere? (So, space before this import from "coral-stream/..." and then the imports from "./create" and "../../fixtures" below together with no space between)

marcushaddon: https://github.com/coralproject/talk/blob/fix/CORL-2261-children-of-rejected-comments/src/core/client/stream/test/comments/permalink/permalinkViewRejectedComment.spec.tsx#L5

marcushaddon:
@kabeaty Test has been added!

kabeaty (approved):
Looks good! Thanks for the updates.

nick-funk (approved):

PR #4010, opened 2022-07-22
[CORL-2608]: Virtualized rendering and next unseen updates

kabeaty:
## What does this PR do?

These changes implement virtualized rendering using Virtuoso in the stream. They update to load 20 comments initially, and then all of the rest of the stream comments in the background. For this, loading states were added to the `Load all comments` button that will be seen if anyone clicks on the button or tries to navigate to next unread before all comments are loaded in. Loading states were also added for new comments and new replies buttons.

They also update next unseen comment to work with virtualized rendering. Unmark all now unmarks all comments currently on a stream, not just those that are currently displayed/loaded.

In addition, there are updates to ignored comments and tombstones. Ignored comments are now always shown in the stream. Tombstone styles have been updated.

Media (twitter, youtube) in comments now have an `Expand` option to view them fully if they are over a certain height.

In alternate oldest view, if you click on the `Add a comment` button, it scrolls you down to the `Post a comment` box below all comments.

`allChildComments` is added to comments to be used by next unseen and unmark all
`markAllAsSeen` is added as an option to the `markCommentsAsSeen` mutation to support unmarking all comments

no

n/a

## How do I test this PR?

Test out the stream in general (adding comments, replying, ignoring users, rejecting comments, etc.).

To test out new loading states, wrap lines 178-180 and 203-210 in `AllCommentsTabContainer` in a setTimeout of something like 8000 or 10000, to mimic 8-10 seconds of loading time. You can test the new replies button loading state by wrapping 285-293 in `ReplyListContainer` in a similar setTimeout. Try to hit the `Load all comments` button before comments are loaded, and also try to navigate with Z key / mobile unread button before comments are loaded and see that you are shown a loading state.

Test out next unseen comment traversal, both via Z key and via the `Next unread` button in the mobile toolbar. Test in combination with ignored users, rejected comments, deleted comments. Test out looping around, new comment and reply subscriptions, unmarking all, when the next unseen comment is far down the stream, etc.

Ignore a user who has left comments. See the ignored comments changes (message/tombstone always shown in the stream). Check out the other tombstones displayed in the stream (moderate a comment and reject it in the stream, delete a comment).

Test out alternate oldest view by sorting by oldest and enabling the feature flag. See that clicking `Add a comment` takes you to the `Post a comment` box below all comments.

Test out expanding, showing, and hiding media (twitter, youtube) in comments in the stream.
 
## How do we deploy this PR?

n/a


src/core/client/stream/tabs/Comments/ReplyList/ReplyList.tsx
```diff
@@ -102,6 +100,7 @@ const ReplyList: FunctionComponent<ReplyListProps> = (props) => {
               fullWidth
               data-key-stop
               data-is-load-more
+              data-is-view-new
```
nick-funk, line 103: Was this intentional? I only ask as the only other changes here are for tombstone-ing. Completely fine if it is, but I'm checking this isn't a copy-paste flub since this is a `ReplyList` that I believe is a `data-is-load-more` versus a `data-is-view-new`.

src/core/client/stream/tabs/Comments/Stream/AddACommentButton.tsx
```diff
@@ -1,32 +1,80 @@
...
+                const stopElement = root.querySelector(
+                  `[data-index="${indexToScroll}"]`
+                );
+                if (stopElement !== undefined && stopElement !== null) {
+                  clearInterval(stopExists);
+                  postCommentForm.scrollIntoView();
+                }
+                if (count > 10) {
+                  clearInterval(stopExists);
+                }
```
nick-funk: I see what you mean by the non-ideal retry interval. That being said, this is a clever and nicely contained solution that doesn't add a lot of complexity or issues anywhere but here in this snippet of code. Well done!

src/core/client/stream/tabs/Comments/Stream/StreamContainer.tsx
```diff
@@ -265,7 +272,10 @@ export const StreamContainer: FunctionComponent<Props> = (props) => {
         {isRatingsAndReviews && <StoryRatingContainer story={props.story} />}
         {showCommentForm &&
           (alternateOldestViewEnabled ? (
-            <AddACommentButton isQA={isQA} />
+            <AddACommentButton
+              isQA={isQA}
+              currentScrollRef={currentScrollRef}
```
nick-funk: Thanks for keeping this `currentScrollRef` passed through cleanly. It's unfortunate it has to be plumbed through everywhere, but I appreciate that you kept the plumbing down to one variable instead of 3-10 and stored the rest in `Local`. 👍

src/core/server/graph/loaders/Comments.ts
```diff
@@ -384,6 +371,25 @@ export default (ctx: GraphContext) => ({
...
+    const story = await ctx.loaders.Stories.story.load(comment.storyID);
+    if (!story) {
+      throw new StoryNotFoundError(comment.storyID);
+    }
+    return retrieveCommentAllChildCommentsConnection(
+      ctx.mongo,
+      ctx.tenant.id,
+      comment,
+      {
+        first: 999,
```
nick-funk: We may want to make this something bigger, like 9999 or even more, there are known streams of more than 1000 comments.

src/core/server/graph/schema/schema.graphql
```diff
@@ -3661,6 +3668,20 @@ type Comment @cacheControl(maxAge: 5) {
...
   parentCount: Int!
 
+  """
+  allChildComments is a CommentsConnection that includes all comments that are
+  children of a comment (i.e. that have the comment included in their ancestorIDs)
+  """
+  allChildComments(
+    first: Int = 10
+    orderBy: COMMENT_SORT = CREATED_AT_DESC
+  ): CommentsConnection!
```
nick-funk, line 3671: Do we think it's clearer readability wise for this to be as it is named:`allChildComments` versus a flag/option called `all` (or similar) on the existing `comments` resolver?

src/core/server/models/comment/comment.ts
```diff
@@ -472,6 +472,33 @@ export const retrieveCommentRepliesConnection = (
...
+ * retrieveCommentAllChildCommentsConnection returns a Connection<Comment> for
+ * a given comment's child comments
+ *
+ * @param mongo database connection
+ * @param tenantID the tenant id
+ * @param storyID the id of the story the comment belongs to
+ * @param comment the comment to retrieve child comments of
+ * @param input connection configuration
+ */
+export async function retrieveCommentAllChildCommentsConnection(
```
nick-funk: Reading this makes me wonder if we might want to rename it to something like `retrieveChildrenForParent` or `retrieveAllChildrenForComment`? It doesn't quite match our existing naming, but I think we might be seeing that our current naming structure is confusing when it gets agglutinized longer and longer.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabVirtualizedComments.tsx
```diff
@@ -0,0 +1,307 @@
...
+  // It also takes into account the initial comments loaded since if we start
+  // with fewer than 20, we will never want to display load all as new comments
+  // come in.
+  const moreCommentsForLoadAll = useMemo(() => {
+    return (
+      comments.length > NUM_INITIAL_COMMENTS &&
+      ((initialComments && initialComments.length > NUM_INITIAL_COMMENTS) ||
+        (initialComments &&
+          initialComments.length === NUM_INITIAL_COMMENTS &&
+          initialComments.hasMore))
```
nick-funk: This is just a thought, not necessary for you to do. When I see complex logic like this, I tend to try and word it out via naming like so:

```
const haveMoreThanInitialComments = comments.length > NUM_INITIAL_COMMENTS;

const initialCommentsMoreThanExpected = initialComments && intialComments.length > NUM_INITIAL_COMMENTS;

const initialCommentsHasMore = initialComments &&
   initialComments.length === NUM_INITIAL_COMMENTS &&
   initialComments.hasMore;
   
return haveMoreThanInitialComments && (initialCommentsMoreThanExpected || initialCommentsHasMore);
```

Not saying this is a good example, but I know that the compiler/minifier can optimize this for me, so I tend to try and name/group logic together based on what my human brain was thinking when writing sections of it.

I suppose `moreCommentsForLoadAll` explains this, but each of the edge cases for why there are "more" is sometimes confusing without more explanation. On further thought, the comments may describe this thoroughly enough already.

Just curious what we think of naming logic bits for clarity.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabVirtualizedComments.tsx
```diff
@@ -0,0 +1,307 @@
...
+      }
+      if (!settings.loadAllComments) {
+        return !loadAllButtonHasBeenClicked;
+      } else {
+        return false;
+      }
+    } else {
+      return false;
+    }
+  }, [
```
nick-funk: This is perfect for handling this complexity. Very easy to jump in with console logging or break points. Thank you!

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -48,9 +45,34 @@ import { SetTraversalFocus } from "./SetTraversalFocus";
...
+    readonly id: string;
+    readonly allChildComments: {
+      readonly edges: ReadonlyArray<ChildComment>;
+    };
+    readonly deleted: boolean | null;
+    ignored?: boolean | null;
+    ignoredReplies?: string[];
+  };
+}
+
```
nick-funk, line 70: Thank you for using concrete interfaces and not trying to do a `Omit<Comment>(...)` on 1000 un-needed props on the comment. This is much clearer with the exact same type safety.

I dislike the `Omit<...>` as it does the exact opposite of an interface definition, it tells me only what I don't care aboue, and does not define what I DO care about!

nick-funk (commented):
Initial draft looks good, I will take another pass when you're done pushing up further changes!

src/core/client/stream/tabs/Comments/ReplyList/ReplyList.tsx
kabeaty: Yes, thanks for asking! This is the `Show more` button for viewing new replies, while the button for just showing more replies is above this one. I think I can actually remove this data attribute here and on the new comment subscription buttons now with the refactoring I've done though anyway.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabVirtualizedComments.tsx
kabeaty: I agree that especially in this case, better naming/grouping of logic here will improve the readability without relying on the comment to explain as much. I was also able to update to simplify the logic with the updates to the way that we are loading in comments now!

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabVirtualizedComments.tsx
kabeaty: Great! I was also able to simplify this with the updates to not have a configuration setting for `loadAllComments` anymore and with alternate oldest view just scrolling down to the last comment when clicking `Add a comment` now.

src/core/server/graph/loaders/Comments.ts
kabeaty: I was thinking this would be fine for replies, but I agree that it could be good to bump it up to be safe 👍

src/core/server/models/comment/comment.ts
kabeaty: Yes, I agree that rename would be clearer and that maybe this is a case to deviate from our existing naming structure. Will update to `retrieveChildrenForParentConnection` 👍

src/core/server/graph/schema/schema.graphql
nick-funk: Nevermind, I thought this was on the Query type and not the Comment type. This makes more sense.

src/core/client/stream/common/KeyboardShortcuts/KeyboardShortcuts.tsx
```diff
@@ -283,114 +290,615 @@ const KeyboardShortcuts: FunctionComponent<Props> = ({ loggedIn, storyID }) => {
...
+        await markSeen({
+          storyID,
+          commentIDs: [],
+          updateSeen: true,
+          markAllAsSeen: true,
+          orderBy: commentsOrderBy,
+        });
+      } catch (error) {
+        // eslint-disable-next-line no-console
+        console.error(error);
```
nick-funk, line 416: Did we want to leave this error log here?

nick-funk (approved):
Looks good, left a comment about an eslint ignore on a `console.error` and I found a file that was empty but floating around.

src/core/client/stream/tabs/Comments/Stream/AllCommentsTab/AllCommentsTabCommentVirtual.tsx:
image

Nothing that can hold up an approve though, great work!

nick-funk (approved):

PR #4018, opened 2022-08-09
[CORL-2614] Add development host and cdn testing environments to Coral

nick-funk:
## What does this PR do?

Adds a host and CDN test tooling to the Coral repo so we can duplicate production similar behaviour when testing locally with full STATIC_URI, hosted in a separate web app page, while using production builds.

None

No

No new indices.

## How do I test this PR?

Open a terminal:
```
npm run start:cdn
``` 

Create a `config.json` under `dev/host`:

```
{
  "sites": [
    {
      "port": 8000,
      "coralURL": "http://localhost:3000",
      "staticURI": "http://localhost:3001",
      "stories": [
        {
          "id": "67be7381-6233-443f-a568-0f8e696edbc3"
        }
      ]
    }
  ]
}
```

Open another terminal:
```
npm run start:host
```

Set the following environment variables in `.env` file:
```
SIGNING_SECRET=secret
STATIC_URI=http://localhost:3001
```

Open a third terminal:
```
npm run build
npm run start
```

Add `http://localhost:8000` to a site in Coral. (visit http://localhost:3000 to do this)

Visit `http://localhost:8000` and visit the story matching the one you created in the `config.json`.

Use the stream and test things out.

## How do we deploy this PR?

No special considerations.

### DIFF
--- a/dev/cdn/build/nativeNodeModules.js
+++ b/dev/cdn/build/nativeNodeModules.js
@@ -1,0 +1,37 @@
+const nativeNodeModulesPlugin = {
+  name: "native-node-modules",
+  setup(build) {
+    // If a ".node" file is imported within a module in the "file" namespace, resolve
+    // it to an absolute path and put it into the "node-file" virtual namespace.
+    build.onResolve({ filter: /\.node$/, namespace: "file" }, (args) => ({
+      path: require.resolve(args.path, { paths: [args.resolveDir] }),
+      namespace: "node-file",
+    }));
+
+    // Files in the "node-file" virtual namespace call "require()" on the
+    // path from esbuild of the ".node" file in the output directory.
+    build.onLoad({ filter: /.*/, namespace: "node-file" }, (args) => ({
+      contents: `
+        import path from ${JSON.stringify(args.path)}
+        try { module.exports = require(path) }
+        catch {}
+      `,
+    }));
+
+    // If a ".node" file is imported within a module in the "node-file" namespace, put
+    // it in the "file" namespace where esbuild's default loading behavior will handle
+    // it. It is already an absolute path since we resolved it to one above.
+    build.onResolve({ filter: /\.node$/, namespace: "node-file" }, (args) => ({
+      path: args.path,
+      namespace: "file",
+    }));
+
+    // Tell esbuild's default loading behavior to use the "file" loader for
+    // these ".node" files.
+    const opts = build.initialOptions;
+    opts.loader = opts.loader || {};
+    opts.loader[".node"] = "file";
+  },
+};
+
+module.exports = nativeNodeModulesPlugin;
--- a/dev/cdn/src/index.ts
+++ b/dev/cdn/src/index.ts
@@ -2,11 +2,11 @@ import Logger from "bunyan";
 import cors from "cors";
 import express from "express";
 
-const PORT = 3001;
+import { config } from "dotenv";
 
 const createLogger = () => {
   return Logger.createLogger({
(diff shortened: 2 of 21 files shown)


dev/cdn/src/index.ts
```diff
@@ -0,0 +1,37 @@
+import Logger from "bunyan";
+import cors from "cors";
+import express from "express";
+
+import { config } from "dotenv";
```
kabeaty, line 5: Can you please update to move this import up with the rest?

dev/host/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+This should have been done when you initialized the upper root Coral repo. If not, at the root of the repository, run:
+
+```
+npm install
+```
+
+## Config JSON for your sites
+
+Create a file named `config.json` at the root of your repo. Don't worry, it's already ignored in the `.gitignore`.
```
kabeaty, line 15: Isn't the `config.json` created under `dev/host/`, not at the root of the repo?

dev/host/README.md
```diff
@@ -0,0 +1,165 @@
...
+
+Enable `Login with Single Sign On` and also check `Allow Registration`.
+
+Copy the `Active` key's `Secret` field and use it for `<YOUR_SSO_SECRET_FROM_CORAL>` in your config.json.
+
+#### SSO User
+
+The user has an `id`, `username`, and `email`. These are all required fields.
+
+The user also has a `role` which can be `COMMENTER`, `MODERATOR`, or `ADMIN` as you choose. If you don't set it, Coral will defaul the user to be a `COMMENTER`.
```
kabeaty, line 107: typo, `defaul` --> `default`. Also, `MEMBER` or `STAFF` would be valid roles too, right? Could also just point to the role options where they are defined in the schema so this doesn't have to be kept updated if role options change.

dev/host/src/views/story.html
```diff
@@ -0,0 +1,70 @@
...
+    <link rel="stylesheet" href="/static/story.css">
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <script class="coral-script" src="{{ coralURL }}/assets/js/count.js" defer></script>
+    {%if section %}
+      <meta data-rh="true" property="article:section" content="{{ section }}" />
+    {% endif %}
+  </head>
+  <body>
+    <h2>Story: {{ storyID }}</h2>
+    <div>Story URL: <a href="{{ url }}">{{ url }}</a></div>
```
kabeaty, line 14: It might be nice to also add a link back to the main page with all story links for ease of switching between stories.

kabeaty (approved):
Left a few small notes, but looks and works great other than that! Very nice!

dev/cdn/build/nativeNodeModules.js
```diff
@@ -0,0 +1,37 @@
+const nativeNodeModulesPlugin = {
```
marcushaddon, line 1: So this is a custom esbuild plugin? Where did you get an overview on writing this?

nick-funk: Yes, this is a custom node modules resolver. Esbuild is still in a pretty hacky state at times, lots of folks just port over a webpack-esque thing to esbuild compatibility.

I took it from a GitHub issue from one of the esbuild maintainers when building pasteboard a few months back, I'm not sure where it's from at this point.

dev/host/README.md
nick-funk: Yeah, this is from me copy pasting from the other tool I built this on. I will fix this up!

PR #4019, opened 2022-08-10
set sentry sample rate to 0.2

tessalt:
## What does this PR do?

Sets the sample rate for sentry to 20% so we don't blow up the sentry quota. 

none

## How do we deploy this PR?
Once deployed, we need to re-enable error reporting on coral through the Sentry admin.

### DIFF
--- a/src/core/client/framework/lib/errors/reporter/sentry/sentry.tsx
+++ b/src/core/client/framework/lib/errors/reporter/sentry/sentry.tsx
@@ -67,7 +67,8 @@ export class SentryErrorReporter implements ErrorReporter {
       release: `coral@${process.env.TALK_VERSION}`,
       debug: Boolean(options.offlineDebug),
       transport: options.offlineDebug ? FakeDebugTransport : undefined,
-      whitelistUrls,
+      allowUrls: whitelistUrls,
+      sampleRate: 0.2,
       beforeSend: (event) => {
         // Drop all events if query string includes `fbclid` string
         // See: https://github.com/getsentry/sentry-javascript/issues/1811.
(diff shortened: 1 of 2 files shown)


src/core/client/framework/lib/errors/reporter/sentry/sentry.tsx
```diff
@@ -67,7 +67,8 @@ export class SentryErrorReporter implements ErrorReporter {
       release: `coral@${process.env.TALK_VERSION}`,
       debug: Boolean(options.offlineDebug),
       transport: options.offlineDebug ? FakeDebugTransport : undefined,
-      whitelistUrls,
+      allowUrls: whitelistUrls,
```
tessalt, line 70: sentry changed this field name, old name will be removed in v7

nick-funk (approved):

src/core/client/framework/lib/errors/reporter/sentry/sentry.tsx
```diff
@@ -67,7 +67,8 @@ export class SentryErrorReporter implements ErrorReporter {
...
+      sampleRate: 0.2,
```
kabeaty, line 71: This will set a sample rate of 20% for all events going to Sentry. My recommendation was that we set up a sampling rate only for events that have no domain tag value set since that's what we were seeing with these events we want to filter. https://docs.sentry.io/platforms/javascript/configuration/sampling/#setting-a-sampling-function

But we could also try a blanket sampling rate and see if we feel that's working well since it could limit any future spikes.

tessalt: yeah i'd like to start out quite conservatively just because we're sharing sentry with other teams. If it all seems manageable we can be more specific in the future.

kabeaty (approved):

PR #4023, opened 2022-08-12
Feat/corl 2474 allocate site mods

marcushaddon:
## What does this PR do?
This PR enables Organization Moderators to allocate Site Moderators to sites within their organization.

This updates the `UpdateUserRoleInput` type to accept an optional array of `siteIDs`.

No.

n/a

## How do I test this PR?
1. As an organization moderator navigate to the Community view. You should see that UserRoleChange buttons are available for all users with roles <= moderator (so not for org mods or admins).
2. Click on the Role name of a non-moderator user. You should see that only the role Site Moderator is enabled.
3. Click on this role, and you should be able to add sites to this user in the following modal. Completing the modal should change that user's role to Site Moderator. 
4. Now that the user is a site mod, the only roles available to you in the RoleChange component should be Site Moderator and Commenter (this should also be true for users that were already Site Moderators when you arrived).
5. Set a site moderator back to a Commenter.
6. Log in as any other role and see that managing user roles from the Community view works (or doesn't work) just as before.
 
 
## How do we deploy this PR?


src/core/client/admin/components/UserRole/UserRoleChange.tsx
```diff
@@ -6,6 +6,8 @@ import React, {
   useState,
 } from "react";
 
+import { validateRoleChange } from "coral-common/permissions";
```
kabeaty, line 9: Can you please remove the empty line below this to organize with the other coral imports?

src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
```diff
@@ -68,8 +68,14 @@ const UserRoleChangeContainer: FunctionComponent<Props> = ({
     [updateUserMembershipScopes, user.id]
   );
 
+  /**
+   * Q: MMODERATOR can changeRole requires there be further restrictions on
```
kabeaty: Did you get these questions answered and can remove this comment now?

src/core/client/admin/test/community/community-rtl.spec.tsx
```diff
@@ -253,21 +253,6 @@ it("change user role", async () => {
   expect(resolvers.Mutation!.updateUserRole!.called).toBe(true);
 });
 
-it("can't change role as a moderator", async () => {
```
kabeaty, line 261: Instead of just removing this test, what do you think about updating it to test that roles can now be changed as expected when the viewer is an org moderator?

src/core/common/permissions/types.ts
```diff
@@ -0,0 +1,18 @@
+// TODO: this annoys me
```
kabeaty: Is this TODO resolved?

src/core/common/permissions/types.ts
```diff
@@ -0,0 +1,18 @@
...
+  | "COMMENTER"
+  | "MEMBER"
+  | "STAFF"
+  | "MODERATOR"
+  | "ADMIN";
+
+export interface User {
+  id: string;
+  role: UserRole;
+  // tenantID: string;
```
kabeaty: Can this be removed?

src/core/common/permissions/validateRoleChange.ts
```diff
@@ -0,0 +1,73 @@
...
+
+  return false;
+};
+
+export const validateRoleChange = (
+  viewer: Readonly<User>,
+  user: Readonly<User>,
+  role: UserRole,
+  /**
+   * TODO: this optional argument that is only relelvant
```
kabeaty: Can you either remove this todo comment or update to remove typos for future reference on what we should do to change this?

src/core/server/models/user/helpers.ts
```diff
@@ -4,14 +4,10 @@ import { SSOUserProfile } from "coral-server/app/middleware/passport/strategies/
 
 import { GQLUSER_ROLE } from "coral-server/graph/schema/__generated__/types";
 
+import { isSiteModerationScoped } from "coral-common/permissions";
```
kabeaty, line 6: Can you please include this import up with that from `coral-server` in the same grouping?

src/core/server/services/users/user.updateRole.spec.ts
```diff
@@ -0,0 +1,209 @@
+import { MongoContext } from "coral-server/data/context";
+import { UserForbiddenError } from "coral-server/errors";
+import { GQLUSER_ROLE } from "coral-server/graph/schema/__generated__/types";
+import { Tenant } from "coral-server/models/tenant";
+import { User } from "coral-server/models/user";
+
+import { createMockMongoContex } from "coral-server/test/mocks";
```
kabeaty, line 6: Can you include this import with the above grouping from `coral-server`?

src/core/server/services/users/user.updateRole.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+    for (const user of nonAdmins) {
+      for (const role of ROLES) {
+        const res = await uut(mongo, mockTenant, admin, user, role);
+
+        expect(res.role).toEqual(role);
+      }
+    }
+  });
+
+  it("TODO: is this ok?! allows organization moderators to allocate site mods", async () => {
```
kabeaty: Did you want to update this todo? Looks like it's testing what we want it to, that org mods can add site mods?

src/core/server/services/users/user.updateRole.spec.ts
```diff
@@ -0,0 +1,209 @@
...
+        orgMod,
+        user,
+        GQLUSER_ROLE.MODERATOR
+      );
+
+      expect(res.role).toEqual(GQLUSER_ROLE.MODERATOR);
+    }
+  });
+
+  it("does not allow site mods, staff, members, or commenters to change anyones role", async () => {
```
kabeaty: Site mods are able to make other users site mods for their sites and members though, right?

src/core/server/services/users/users.ts
```diff
@@ -757,19 +761,50 @@ export async function updateUsernameByID(
...
+  if (user === null) {
+    throw new UserNotFoundError(userID);
+  }
+
+  const isScoped = !!siteIDs?.length;
+
+  const validUpdate = validateRoleChange(fullViewer, user, role, isScoped);
+
+  if (!validUpdate) {
+    throw new UserForbiddenError("TODO: reason", "user", "updateRole");
```
kabeaty: Can you please update this reason?

src/core/client/admin/components/UserRole/UserRoleChange.tsx
```diff
@@ -25,8 +28,8 @@ import UserRoleText from "./UserRoleText";
 import styles from "./UserRoleChange.css";
 
 interface Props {
-  username: string | null;
-  onChangeRole: (role: GQLUSER_ROLE_RL) => Promise<void>;
+  user: UserRoleChangeContainer_user;
```
kabeaty, line 30: Do we need to send through `user` separately here? It looks like we are still only using the info about the user that is already being passed through (`role` and `moderationScopes`) to validate the role changes, unless I'm missing something? 

If we do send `user` through like this, then I think we should update to not also send through `role`, `moderationScopes`, `membershipScopes`, `moderationScoped`, and `membershipScoped` separately.

kabeaty (dismissed):
Looks good! I left some comments throughout. I have a few overall questions on functionality, too. After I have an org mod promote someone to be a site mod, the option to then update them back to a `Commenter` is disabled. Shouldn't the org mod be able to make them a commenter again? 

Also, other role change dropdowns for admins and site mods don't include disabled roles like are displayed for org moderator (for example, Org Moderator, Admin, Staff, and Member are included in the org mod role change dropdown but grayed out as options). Is it intended that we want to include disabled options in this case? It also seems like an org moderator should be able to update a user to be a member?

I also see an empty spec file added (`UserRole.spec.tsx`). Can you please remove?

src/core/server/services/users/user.updateRole.spec.ts
marcushaddon: i don’t believe so, i think only organization moderators can do that. otherwise community mods for SBN could elevate their friends to moderators. i’ll confirm in standup tomorrow though

src/core/client/admin/components/UserRole/UserRoleChange.tsx
marcushaddon: this doesnt say outdated, but I'm not seeing an empty line below this import.

src/core/client/admin/test/community/community-rtl.spec.tsx
marcushaddon: A new test has been added! src/core/client/admin/test/community/community-rtl.spec.tsx

marcushaddon:
@kabeaty, I can hide rather than disable the invalid roles without much effort. I'd like to get input from Andrew on whether or not Org mods should be able to demote users back to commenters and / or allocate members. That empty test file should be deleted

src/core/server/services/users/user.updateRole.spec.ts
kabeaty: I wanted to follow up on this. If you go into the `Community` tab in the admin, site moderators currently can change users' roles to site moderator or member for their site(s), so I think this test should be updated.

src/core/common/permissions/rules.ts
```diff
@@ -0,0 +1,88 @@
...
+  isSiteModerator,
+  PermissionsActionRuleTest,
+} from "./types";
+// admins can do whatever they want to anyone except other admins
+const adminsArePowerful: PermissionsActionRuleTest = ({ viewer, user }) => ({
+  applies: viewer.role === "ADMIN" && user.role !== "ADMIN",
+  reason: "Admins may change any non admin's role or scopes",
+});
+
+// org mods can promote anyone < site mod to be an unscoped member
```
marcushaddon, line 13: This is a lot of comments, and was honestly to help me think, but I feel like this should be documented somewhere. Maybe I'll create a Slab doc and put a link to it here

src/core/common/permissions/types.ts
```diff
@@ -0,0 +1,120 @@
+import { isSiteModerationScoped } from "./isSiteModerationScoped";
```
marcushaddon, line 1: A lot of the types here are redundant but are needed (as far as I know) because we can't shouldn't import + depend on types from the backend in common. Please correct me if wrong!

src/core/common/permissions/validateRoleChange.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  return false;
+};
+
+export const validateRoleChange = (
+  viewer: Readonly<User>,
+  user: Readonly<User>,
+  role: UserRole,
+  /**
+   * TODO: this optional argument that is only relevant
```
marcushaddon, line 45: I intend to leave this TODO here until we break MODERATOR into SITE_MODERATOR + ORG_MODERATOR

src/core/server/graph/schema/schema.graphql
```diff
@@ -2779,6 +2779,12 @@ type UserModerationScopes {
   on. When null, it means that the user is allowed to moderate on all sites.
   """
   sites: [Site!]
+
+  """
+  sites (when not null) list the sites that the user is only allowed to moderate
+  on. When null, it means that the user is allowed to moderate on all sites.
+  """
+  siteIDs: [String!]
```
marcushaddon, line 2787: exposing these to the frontend because we have to compute these all over the place.

src/core/server/services/users/permissions/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    return (mongo: MongoContext, tenantID: string) =>
+      pullUserSiteModerationScopes(
+        mongo,
+        tenantID,
+        user.id,
+        user.moderationScopes.siteIDs
+      );
+  }
+
+  return async () => null; // believe it or not it is yelling at me unless i do this
```
marcushaddon: Currently looking into this. It is complaining that "not all code paths return a value" even though the return type is potentially undefined

src/core/server/services/users/user.updateRole.spec.ts
marcushaddon: ✅

src/core/client/admin/components/UserRole/SiteModeratorActionsModal.tsx
```diff
@@ -70,68 +72,53 @@ const SiteModeratorActionsModal: FunctionComponent<Props> = ({
...
-                  ) : (
-                    <Localized id="community-userNoLongerPermitted">
-                      <ModalBodyText>
-                        User will no longer be permitted to make moderation
-                        decisions or assign suspensions on:
-                      </ModalBodyText>
+                  <Localized id="community-siteModeratorsArePermitted">
+                    <ModalBodyText>
+                      Site moderators are permitted to make moderation
+                      suspensions on the sites they are assigned.
```
nick-funk: Has this been updated in `admin.ftl`?

src/core/common/permissions/rules.ts
```diff
@@ -0,0 +1,88 @@
...
+  };
+};
+
+/**
+ * validationRules represents all possible reasons that a pending
+ * action should be allowed. If no rule applies, the action
+ * should be disallowed.
+ */
+export const validationRules: PermissionsActionRuleTest[] = [
+  adminsArePowerful,
```
nick-funk, line 85: this variable name 😂

nick-funk (changes requested):
Found one small localization that may be missing from an ftl file. Otherwise, I really like this PR and appreciate all the refactoring you've done to make things cleaner here. It's really great to see all the permissions logic having some sense applied to it. The `siteIDs` be available from the API to avoid re-computation everywhere is also a great idea.

src/core/client/admin/components/UserRole/SiteModeratorActionsModal.tsx
marcushaddon: oh i actually didnt mean to update this! changed it back

nick-funk (approved):

PR #4031, opened 2022-08-22
Optimize tag count resolution via caching

nick-funk:
## What does this PR do?

Updates the logic for calculating the tag counts on a story to retrieve them from a cache of counts on the story document.

Also adds logic to generate the tag counts on the story document if they're not available (migrates existing stories).

No changes.

No.

No new indexes.

## How do I test this PR?

- Go to a story that you have used before
- Load the page
- Open your MongoDB compass, VS Code Mongo tools, etc and find the story document
- Confirm that a new field for `tags` has shown up under `commentCounts` on the story document
- Check that the counts are correct for how many comments are featured, written by an admin, moderator, staff, etc
- Use a Q&A and a Ratings & Review story and confirm the tags are accurate there as well while reviewing, asking questions, answering, experts, etc... 
 
## How do we deploy this PR?

No special considerations to deploy.

### DIFF
--- a/src/core/server/graph/mutators/Comments.ts
+++ b/src/core/server/graph/mutators/Comments.ts
@@ -36,6 +36,7 @@ import {
   GQLUnfeatureCommentInput,
 } from "coral-server/graph/schema/__generated__/types";
 
+import { hasTag } from "coral-server/models/comment";
 import { validateUserModerationScopes } from "./helpers";
 import { validateMaximumLength, WithoutMutationID } from "./util";
 
@@ -213,6 +214,9 @@ export const Comments = (ctx: GraphContext) => ({
       ctx.site!.id,
       ctx.mongo,
       ctx.redis,
+      // Create a diff where "before" tags does not have a
+      // featured tag and the after does since the previous
+      // `addTag` put the featured tag onto the comment
       comment.tags.filter((t) => t.type !== GQLTAG.FEATURED),
       comment.tags
     );
@@ -229,7 +233,33 @@ export const Comments = (ctx: GraphContext) => ({
     // Validate that this user is allowed to moderate this comment
     await validateUserModerationScopes(ctx, ctx.user!, { commentID });
 
-    return removeTag(ctx.mongo, ctx.tenant, commentID, GQLTAG.FEATURED);
+    const comment = await removeTag(
+      ctx.mongo,
+      ctx.tenant,
+      commentID,
+      GQLTAG.FEATURED
+    );
+
+    // If the tag is sucessfully removed (the tag is
+    // no longer present on the comment) then we can
+    // update the tag story counts.
+    const isFeatured = hasTag(comment, GQLTAG.FEATURED);
+    if (!isFeatured) {
+      await updateTagCommentCounts(
+        ctx.tenant.id,
+        comment.storyID,
+        ctx.site!.id,
+        ctx.mongo,
+        ctx.redis,
+        // Create a diff where "before" has the featured tag,
+        // and after does not since the result of the previous
+        // `removeTag` took the featured tag off of the comment
+        [...comment.tags, { type: GQLTAG.FEATURED, createdAt: new Date() }],
+        comment.tags
+      );
+    }
+
+    return comment;
   },
   markAsSeen: async ({
     commentIDs,
--- a/src/core/server/graph/mutators/Comments.ts
+++ b/src/core/server/graph/mutators/Comments.ts
@@ -2,6 +2,7 @@ import { ERROR_CODES } from "coral-common/errors";
 import { ADDITIONAL_DETAILS_MAX_LENGTH } from "coral-common/helpers/validate";
 import GraphContext from "coral-server/graph/context";
 import { mapFieldsetToErrorCodes } from "coral-server/graph/errors";
+import { hasTag } from "coral-server/models/comment";
 import { addTag, removeTag } from "coral-server/services/comments";
 import {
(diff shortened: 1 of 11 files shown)


src/core/server/models/comment/comment.ts
```diff
@@ -1043,6 +1047,52 @@ export async function retrieveStoryCommentTagCounts(
...
+    } else {
+      result.set(story.id, story.commentCounts.tags.tags);
+    }
+  }
+
+  return storyIDs.map((id) => {
+    const tags = result.get(id);
+    if (!tags) {
+      return {
+        [GQLTAG.ADMIN]: 0,
```
kabeaty: What are your thoughts on defining this in one common place instead of repeating since it would be easy to leave a tag out somewhere?

src/core/server/graph/mutators/Comments.ts
```diff
@@ -206,6 +207,16 @@ export const Comments = (ctx: GraphContext) => ({
       );
     }
 
+    await updateTagCommentCounts(
```
kabeaty, line 211: Do we also need to have this update tag comment counts when unfeaturing comments?

src/core/server/models/comment/counts/counts.ts
```diff
@@ -139,6 +163,27 @@ export function mergeCommentModerationQueueCount(
...
 
+export function mergeCommentTagCounts(
+  ...tags: CommentTagCounts[]
+): CommentTagCounts {
+  const merged = createEmptyCommentTagCounts();
+
+  for (const tagSet of tags) {
+    merged.total += tagSet.total;
+
+    merged.tags.EXPERT += tagSet.tags.EXPERT;
```
kabeaty, line 175: Do we not need `ADMIN` here?

kabeaty (changes requested):
Looks good overall! I left a few comments throughout.

src/core/server/graph/mutators/Comments.ts
nick-funk: Yes, we do! I will fix!

src/core/server/models/comment/counts/counts.ts
nick-funk: Yup yup! Fixed, thank you!

nick-funk: Yup yup! Fixed, thank you!

src/core/server/models/comment/comment.ts
nick-funk: I noticed we do a similar pattern in `empty.ts` for merging from zero-ed counts, have replaced all of these with the convenience function I used there.

src/core/server/graph/mutators/Comments.ts
nick-funk: Is fixed!

src/core/server/graph/mutators/Comments.ts
```diff
@@ -35,6 +36,7 @@ import {
   GQLUnfeatureCommentInput,
 } from "coral-server/graph/schema/__generated__/types";
 
+import { hasTag } from "coral-server/models/comment";
```
kabeaty: Can you please move this import up with the others from `coral-server`?

src/core/server/graph/mutators/Stories.ts
```diff
@@ -40,6 +41,7 @@ import {
   GQLUpdateStorySettingsInput,
 } from "coral-server/graph/schema/__generated__/types";
 
+import { initializeCommentTagCountsForStory } from "coral-server/models/comment";
```
kabeaty: Can you please move this import up with the others from `coral-server`?

kabeaty (changes requested):
Thanks for the updates! When testing in Q&A mode, I am seeing that the `QUESTION` tag doesn't seem to be incrementing when I ask a question, only the `UNANSWERED` tag. And when I answer a question, the `UNANSWERED` tag isn't decrementing. Can you look into this?

src/core/server/graph/mutators/Comments.ts
nick-funk: And here I thought TS sort imports was magic... I have now found its kryptonite. Yep, I'll fix this up!

nick-funk:
> Thanks for the updates! When testing in Q&A mode, I am seeing that the `QUESTION` tag doesn't seem to be incrementing when I ask a question, only the `UNANSWERED` tag. And when I answer a question, the `UNANSWERED` tag isn't decrementing. Can you look into this?

I believe I have a fix for the `UNANSWERED` increment and decrementing. Will be pushing up shortly.

As for the `QUESTION` not changing when you're in a Q&A stream, that would be because `QUESTION` is for Ratings & Reviews, and not for Q&A. I know, very confusing, but that mode also has a questions tab. I will re-test that as well to make sure it's working.

kabeaty (approved):
Looking great now, thanks for updating!

nick-funk:
> Looking great now, thanks for updating!

Awesome! Glad to hear it!

Also, I tested Ratings & Reviews, it looks to be working just fine since it doesn't add extra silly layers of add/remove tags in the code like Q&A does with its `UNANSWERED` tag.

This should be good to go now!

nick-funk:
Throwing a `HOLD MERGE` on this as it might be hotfixed into `main` after QA testing, or it may end up in `develop`. Just don't want it prematurely merged until @tessalt gives the okay.

tessalt (approved):

PR #4032, opened 2022-08-22
Feat/corl 2548 flatten replies depth

marcushaddon:
## What does this PR do?
Increases the depth at which replies are automatically flattened to 7.

None.

No.

n/a

## How do I test this PR?
1. Before checking Create a reply chain > 7 replies long (ideally 8 or 9), see that they are flattened only after the 7th comment.
 
 
## How do we deploy this PR?
No special considerations should be required


src/core/client/stream/tabs/Comments/Indent.css
```diff
@@ -43,6 +43,16 @@ $rootMargin: calc(var(--spacing-5) + 2px);
   border-left: 2px solid #e1e2e3;
 }
 
+.level7 {
+  margin-left: calc(6 * var(--spacing-3) + $rootMargin);
+  border-left: 2px solid #e1e2e3;
```
marcushaddon, line 48: i think we'll need Sam's input on what these color codes should be

nick-funk: It looks like we had a set hex value before. This should be fine for now as long as it's accurate. If you want to make it easier to refactor, maybe convert the hex into a css variable at the top of `Indent.css` so it's easy to find. Maybe called `@indentBarColor` or similar. Since it's a bar on the left that is indented in.

marcushaddon: we had a hard coded hex value for each level, this one's only repeated as a place holder. each one gets lighter but I'm not sure by how much the next level should get lighter

nick-funk (approved):

PR #4038, opened 2022-08-26
Check for and repair invalid comment tag counts

nick-funk:
## What does this PR do?

Instead of just checking if the tag counts field exists, check that all its sub fields and values are valid, otherwise we regenerate it.

Also checks on the GQL generated tag count types returned via GraphQL to ensure they're valid, or else it falls back to an empty tag counts to ensure the stream is always available even if a story is in a bad state. Also, if these are found, it logs warnings in the logs so we can catch these misbehaving stories.

Adding all the `GQLTag` types to the `tags` counts under `Story.commentCounts`. Now we have access to counts for `MODERATOR`, `ADMIN`, `STAFF`, etc... as well as the previous `FEATURED`, `QUESTION`, `REVIEW`, and `UNANSWERED`.

No

No new indices.

## How do I test this PR?

- Modify an existing stories tag counts to be broken.
   - If we have a story that looks like this in its `commentCounts.tags`:
     ```
      "tags": {
        "total": 12,
        "tags": {
          "ADMIN": 0,
          "EXPERT": 0,
          "FEATURED": 2,
          "MEMBER": 0,
          "MODERATOR": 10,
          "QUESTION": 0,
          "REVIEW": 0,
          "STAFF": 0,
          "UNANSWERED": 0
        }
      }
      ```
    - Ruin the structure by deleting some fields, say make it look like this:
      ```
      "tags": {
        "total": 12,
        "tags": {
          "MODERATOR": 3
        }
      }
      ```
- Visit the story
- Check the document in the DB again, make sure it has regenerated and is back to what it originally should have been
- You can also try deleting the `total`, deleting the `tags` entirely, etc, upgrading stories without `tags` in the the `commentCounts`, etc
 
## How do we deploy this PR?

No special considerations.

### DIFF
--- a/src/core/server/graph/schema/schema.graphql
+++ b/src/core/server/graph/schema/schema.graphql
@@ -3577,14 +3577,34 @@ CommentTagCounts provides count data for Tags made against Comment's.
 """
 type CommentTagCounts {
   """
+  ADMIN is the count of Comment's with the ADMIN tag.
+  """
+  ADMIN: Int!
+
+  """
+  EXPERT is the count of Comment's with the EXPERT tag.
+  """
+  EXPERT: Int!
+
+  """
   FEATURED is the count of Comment's with the FEATURED tag.
   """
   FEATURED: Int!
 
   """
-  UNANSWERED is the count of Comment's with the UNANSWERED tag.
+  MEMBER is the count of Comment's with the MEMBER tag.
   """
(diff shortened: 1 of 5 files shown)


kabeaty (approved):
Looks great and works as expected! Thank you for updating the tag count types/checks.

PR #4055, opened 2022-09-21
fix disappearing collapse button bug

marcushaddon:
## What does this PR do?
Fixes bug where collapse button disappears after level 5 of nested comments.

None.

No.

n/a

## How do I test this PR?
1. Create a reply chain longer than 8 comments long
2. Observe that collapse buttons are present on all comments until they start being flattened (in flatten replies mode)

### DIFF
--- a/src/core/client/stream/tabs/Comments/Comment/flattenReplies/index.ts
+++ b/src/core/client/stream/tabs/Comments/Comment/flattenReplies/index.ts
@@ -7,6 +7,6 @@ export const isReplyFlattened = (
   return (
     flattenedRepliesEnabled &&
     indentLevel &&
-    indentLevel >= MAX_REPLY_INDENT_DEPTH
+    indentLevel > MAX_REPLY_INDENT_DEPTH
   );
 };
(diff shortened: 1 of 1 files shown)


src/core/client/stream/tabs/Comments/Comment/flattenReplies/index.ts
```diff
@@ -7,6 +7,6 @@ export const isReplyFlattened = (
   return (
     flattenedRepliesEnabled &&
     indentLevel &&
-    indentLevel >= FlattenRepliesIndentLevel
+    indentLevel >= MAX_REPLY_INDENT_DEPTH
```
kabeaty: It looks like the collapse button is still disappearing one level too soon. I think this is because before the constant was 4 for 4 levels, and now it is 7 for 8 levels. If you change this to `indentLevel > MAX_REPLY_INDENT_DEPTH`, then I think it will show the collapse button for all expected levels.

nick-funk: Agreed, I think we have an off-by-one error here! 😆

nick-funk (changes requested):

src/core/client/stream/tabs/Comments/Comment/flattenReplies/index.ts
marcushaddon: ✅

kabeaty (approved):

PR #4056, opened 2022-09-21
[CORL-2451]: Update to React 18

kabeaty:
## What does this PR do?

These changes update Coral to use React 18. To achieve this, the following changes were required:

**package updates**
* updating Prettier, which is why there are some formatting/linting changes throughout
* updating @fluent/react, which included updates to `Localized` props. Vars are now all included in a `vars` prop and elements are now always included in an `elems` prop, instead of all separate props for each var/elem before.

A few other packages had to be updated, but these are the main ones that required significant changes along with their upgrades.

**React updates**
* `children` has to be explicitly included in a component's props now
* `createRoot` instead of `ReactDOM.render`

**test updates**
* some tests had to be converted to React Testing Library
* some tests needed additional acts/awaits

Note that unfortunately we cannot yet use `createRoot` for the stream. This is due to [this issue](https://github.com/petyosi/react-virtuoso/issues/714) in Virtuoso. So, the stream will use `ReactDOM.render` for now and be on React 17. Once this issue is resolved, we should be able to switch it over too.

none

no

n/a

## How do I test this PR?

You can test this by clicking around and seeing that everything still looks and works as expected in the stream/admin.


config/jest/client.config.js
```diff
@@ -1,5 +1,39 @@
...
+  "d3-scale",
+  "d3-scale-chromatic",
+  "d3-selection",
+  "d3-shape",
+  "d3-time",
+  "d3-time-format",
+  "d3-timer",
+  "d3-transition",
+  "d3-zoom",
+];
```
nick-funk, line 35: this is cool, didn't know we were using d3 in places! 👍

scripts/generateSupportedBrowsersRegExp.ts
```diff
@@ -24,6 +24,7 @@ function emitFileContent(outFile: string, verify = false) {
        * This file is generated by \`npm run generate:supportedBrowsersRegExp\`
        * Any manual changes will be overridden.
        */
+      // eslint-disable-next-line
       const supportedBrowsersRegExp = ${regexp};
```
nick-funk, line 28: Did it get mad about dumping a regexp into a formatted string?

src/core/client/account/routeConfig.tsx
```diff
@@ -8,10 +8,14 @@ import ConfirmRoute from "./routes/email/Confirm";
...
 
+interface CoralContainerProps {
+  children?: React.ReactNode;
+}
+
 /** Small wrapper that omits router props */
-const CoralContainer: FunctionComponent = ({ children }) => (
-  <CoralWindowContainer>{children}</CoralWindowContainer>
-);
+const CoralContainer: FunctionComponent<CoralContainerProps> = ({
```
nick-funk, line 16: ah, the "you have children is not assumed by default in the types" begins... (lol)

src/core/client/account/routes/password/Reset/ResetPasswordForm.tsx
```diff
@@ -81,7 +81,7 @@ const ResetPasswordForm: React.FunctionComponent<Props> = ({
                       </Localized>
                       <Localized
                         id="resetPassword-passwordDescription"
-                        $minLength={8}
+                        vars={{ minLength: 8 }}
```
nick-funk, line 84: well this is a pleasant improvement syntax wise, I didn't really like specifying by `$<varname>`... felt too much like PHP haha

src/core/client/admin/components/ModerateCard/ModerateCardContainer.tsx
```diff
@@ -99,20 +99,19 @@ const ModerateCardContainer: FunctionComponent<Props> = ({
...
-    fragment ModerateCardContainerLocal on Local {
-      moderationQueueSort
-    }
-  `);
+  const [{ moderationQueueSort }] =
+    useLocal<ModerateCardContainerLocal>(graphql`
+      fragment ModerateCardContainerLocal on Local {
+        moderationQueueSort
+      }
+    `);
```
nick-funk, line 107: looks like the new "recommended" typescript lint formatting has improved, I like the layout of this better. That or auto-formatting just wasn't run on this file in a while.

src/core/client/admin/index.tsx
```diff
@@ -27,7 +27,10 @@ async function main() {
   );
 
   // eslint-disable-next-line no-restricted-globals
-  ReactDOM.render(<Index />, document.getElementById("app"));
+  const container = document.getElementById("app");
+  const root = createRoot(container!);
```
nick-funk, line 31: Hah, this makes the 4th time they've changed the default "start doing React" part of React lol

nick-funk (approved):

scripts/generateSupportedBrowsersRegExp.ts
kabeaty: With the updates to prettier/linting, without this, it will get stuck in a loop when trying to commit of needing to regenerate this, but if it's generated, then the linting complains. So, I disabled the linting just for this line to break it out of this loop.

PR #4060, opened 2022-09-22
fix off by one error in MAX_REPLY_INDENT_DEPTH

marcushaddon:
## What does this PR do?
Fixes bug in which replies to deeply nested comments are incorrectly indented.

None.

No.

n/a

## How do I test this PR?
1. Enable Flatten Replies
2. Create a reply chain > 8 comments long
3. Reply to a reply past the 8th level (at which point all replies should be indented at the same level) and observe that the comment is in line with the comment preceding it (as opposed to the left of it).

### DIFF
--- a/src/core/client/stream/constants.ts
+++ b/src/core/client/stream/constants.ts
@@ -4,7 +4,7 @@ export const VIEWER_STATUS_CONTAINER_ID = "viewer-status";
 
 export const POST_COMMENT_FORM_ID = "post-comment-form";
 
-export const MAX_REPLY_INDENT_DEPTH = 7;
+export const MAX_REPLY_INDENT_DEPTH = 8;
 
 export const NUM_INITIAL_COMMENTS = 20;
 


src/core/client/stream/constants.ts
```diff
@@ -4,7 +4,7 @@ export const VIEWER_STATUS_CONTAINER_ID = "viewer-status";
 
 export const POST_COMMENT_FORM_ID = "post-comment-form";
 
-export const MAX_REPLY_INDENT_DEPTH = 7;
+export const MAX_REPLY_INDENT_DEPTH = 8;
```
nick-funk, line 7: I don't think making this 8 on its own will work. You have to make it 8 and subtract one elsewhere or leave this as 7 and subtract one in the cases where it needs to be one less to check when it should and shouldn't show things.

nick-funk (changes requested):

src/core/client/stream/constants.ts
kabeaty: Agree with Nick here. The test failures are due to this issue. To fix them, you need to update `const atLastLevelLocalReply =
    props.indentLevel === MAX_REPLY_INDENT_DEPTH && !flattenReplies;` in `ReplyListContainer.tsx` to `const atLastLevelLocalReply =
    props.indentLevel === MAX_REPLY_INDENT_DEPTH - 1 && !flattenReplies;` in `ReplyListContainer.tsx` (to subtract 1) so that the `Read more of this conversation` link is shown again at the deepest reply level when flatten replies isn't enabled.
    
   Also, isn't this update going to conflict with the disappearing collapse button fix? With updating this to 8, I think the check can stay `>=` instead of getting updated to `>`. It might actually make sense to include these two changes in one PR for this reason?

kabeaty (changes requested):

PR #4062, opened 2022-09-26
[CORL-2655] Fix no-text mode for host article count embed

nick-funk:
## What does this PR do?

Only show host article count text if the text class is present. If they set the `data-coral-notext="true"` you will only get a comment count number from the count embed instead of `12 Comments / 4 New` (or similar).

None

No

No new indexes.

## How do I test this PR?

- Spin up Coral
- In `multi-site-test` on `src/views/index.html` on line 28 in the `span` with `class="coral-count"`, set the attribute `data-coral-notext="true"`
- Spin up multi-site-test
- Go to your multi-site-test page
- See that the counts are just showing a number
- Remove the `data-coral-notext="true"` attribute
- See that it now shows count texts similar to `12 Comments / 4 New`
 
## How do we deploy this PR?

No special considerations.


src/core/client/count/injectJSONPCallback.ts
```diff
@@ -2,6 +2,8 @@ import { CountJSONPData } from "coral-common/types/count";
 import { COUNT_SELECTOR } from "coral-framework/constants";
 import getPreviousCountStorageKey from "coral-framework/helpers/getPreviousCountStorageKey";
 
+const TEXT_CLASS_NAME = "coral-count-text";
```
nick-funk, line 5: we can't put this in `src/core/common/constants.ts` because then the bundler will try and pull in too much of `src/core/common` into the embed and count script bundles. This is an issue with webpack/terser that I can't fix. So I'm just ignoring DRY principles for now and having this duplicated both here and in the `.../server/app/handlers/api/story/count.ts`

src/core/server/app/handlers/api/story/count.ts
```diff
@@ -107,6 +107,9 @@ export const countJSONPHandler = ({
 
     const count = story ? await calculateStoryCount(mongo, story) : 0;
 
+    // eslint-disable-next-line no-console
+    console.log(`NOTEXT`, notext);
```
kabeaty: Can you remove this?

nick-funk: yup! done!

kabeaty (approved):
Looks good, thanks for removing the debug!

marcushaddon (approved):

PR #4064, opened 2022-09-27
off by 1 errors

marcushaddon:
## What does this PR do?
Addresses https://vmproduct.atlassian.net/browse/CORL-2648 and https://vmproduct.atlassian.net/browse/CORL-2651

None.

No.

n/a

## How do I test this PR?
1. Enable Flatten Replies
3. Create a reply chain > 7 comments long
4. Reply to a reply past the 8th level (at which point all replies should be indented at the same level) and observe that the comment is in line with the comment preceding it (as opposed to the left of it).
5. Disable Flatten Replies
6. Create another reply chain 6 comments long (1 less than the maximum)
7. In another tab, reply to the 6th reply
8. In the original tab, note that the "Read more of this conversation" button appears


kabeaty (commented):
Looks good, I just have one question. It looks like these updates are making it so that the reply level is only up to a max depth of 7 now? Do we need to also change how we're updating the copy in the configuration for flatten replies to reflect this?

nick-funk (approved):
Am approving as this appears to be working well. I would second what Kathryn brought up about reply depth in the copy. Please update that. Otherwise this looks great!

kabeaty (approved):
Looks good, with the copy updates in the related PR for the flatten replies configuration!

PR #4068, opened 2022-10-04
[CORL-2261] Use story/ancestors state to determine whether we can reply to a comment

nick-funk:
## What does this PR do?

- Adds archived/is-archiving check to `canReply` resolver
- Uses retrieving ancestors by the `ancestorID`'s of a comment to determine if a parent was rejected when checking if a comment can reply.

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

No changes to schema

No

No new indexes

## How do I test this PR?

- Create a root comment with many replies
- Create replies to the replies using various commenter accounts
- Open single conversation view for one of the bottom most replies
- Reject a 1st (or near to root comment) reply
- Attempt to reply in the single conversation view
- See that it warns you an ancestor was rejected
- Refresh the page in single conversation view
- See that the reply buttons are greyed out below the rejected ancestor comment
- Approve the previously rejected reply/comment
- Refresh the single conversation view and see that the reply buttons are enabled again
- Make a reply and see that it is accepted
 
## How do we deploy this PR?

No special considerations

### DIFF
--- a/src/core/server/graph/resolvers/Comment.ts
+++ b/src/core/server/graph/resolvers/Comment.ts
@@ -17,9 +17,12 @@ import {
   hasPublishedStatus,
 } from "coral-server/models/comment/helpers";
 import { createConnection } from "coral-server/models/helpers";
-import { getURLWithCommentID, retrieveStory } from "coral-server/models/story";
+import { getURLWithCommentID } from "coral-server/models/story";
 import { canModerate } from "coral-server/models/user/helpers";
-import { getCommentEditableUntilDate } from "coral-server/services/comments";
+import {
+  getCommentEditableUntilDate,
+  hasRejectedAncestors,
+} from "coral-server/services/comments";
 
 import {
   GQLComment,
@@ -65,15 +68,16 @@ export const Comment: GQLCommentTypeResolver<comment.Comment> = {
     return canModerate(ctx.user, { siteID: c.siteID });
   },
   canReply: async (c, input, ctx) => {
-    const story = await retrieveStory(ctx.mongo, ctx.tenant.id, c.storyID);
+    const story = await ctx.loaders.Stories.find.load({ id: c.storyID });
     if (!story || story.isArchived || story.isArchiving) {
       return false;
     }
 
-    const rejectedAncestors = await comment.hasRejectedAncestors(
+    const rejectedAncestors = await hasRejectedAncestors(
       ctx.mongo,
       ctx.tenant.id,
-      c.id
+      c.id,
+      story.isArchived || story.isArchiving
     );
 
     return !rejectedAncestors;
--- a/src/core/server/graph/resolvers/Comment.ts
+++ b/src/core/server/graph/resolvers/Comment.ts
@@ -17,7 +17,7 @@ import {
   hasPublishedStatus,
 } from "coral-server/models/comment/helpers";
 import { createConnection } from "coral-server/models/helpers";
-import { getURLWithCommentID } from "coral-server/models/story";
+import { getURLWithCommentID, retrieveStory } from "coral-server/models/story";
 import { canModerate } from "coral-server/models/user/helpers";
 import { getCommentEditableUntilDate } from "coral-server/services/comments";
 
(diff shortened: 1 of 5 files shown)


src/core/server/graph/resolvers/Comment.ts
```diff
@@ -64,12 +67,20 @@ export const Comment: GQLCommentTypeResolver<comment.Comment> = {
...
 
-    return !c.rejectedAncestor;
+    const rejectedAncestors = await hasRejectedAncestors(
+      ctx.mongo,
+      ctx.tenant.id,
+      c.id,
+      story.isArchived || story.isArchiving
+    );
+
+    return !rejectedAncestors;
```
nick-funk, line 83: Apologies to the folks that wrote this originally, I don't mean to come in and rewrite all the functionality. There was a performance issue with the way this was working that needed to be fixed up. Otherwise, I wouldn't have touched any of this!

Explanation of my changes:

```
    const story = await ctx.loaders.Stories.find.load({ id: c.storyID });
```

I added a retrieval of the story so I can see if the story is archived or in the process of being archived. If it is, replying is disabled, so I added that here for consistency. The stories will already be locked from replying elsewhere, but it seemed right to check it here too so all our resolver feel logically consistent. Also, I used the data loader via `ctx.loaders` so that we can pull this from the request cache since this story has already been loaded once, saving us a trip to Mongo.

```
const rejectedAncestors = await hasRejectedAncestors(
      ctx.mongo,
      ctx.tenant.id,
      c.id,
      story.isArchived || story.isArchiving
    );
```

Second, I introduced `hasRejectedAncestors` which retrieves the ancestor comments by `id`, which is super fast cause Mongo loves retrieving data by id. This also means we don't have to write a `rejectedAncestor` flag on every comment beneath a rejected comment.

Together, these reduce a lot of read/write load on Mongo for rejected comments.

src/core/server/services/comments/comments.ts
```diff
@@ -466,3 +469,39 @@ export function retrieveChildrenForParentConnection(
...
+) {
+  const comment = await retrieveComment(
+    mongo,
+    tenantID,
+    commentID,
+    !isArchived
+  );
+  if (!comment) {
+    throw new CommentNotFoundError(commentID);
+  }
```
nick-funk, line 487: Technically, this would be faster if I had loaded it via a data loader `ctx.loaders.Comment...` and passed the comment in, but this function may get used in other places later and this makes it a better utility function that it can accept a `commentID` and be on its way.

src/core/server/services/comments/comments.ts
```diff
@@ -466,3 +469,39 @@ export function retrieveChildrenForParentConnection(
...
+
+  const { ancestorIDs } = comment;
+  if (!ancestorIDs || ancestorIDs.length === 0) {
+    return false;
+  }
+
+  const ancestors = await retrieveManyComments(mongo, tenantID, ancestorIDs);
```
nick-funk, line 494: I'm specifically using the `retrieveComment` and `retrieveManyComments` utility functions as they enforce the `tenantID` requirement to help indexes retrieve faster with multi-tenancy.

src/core/server/stacks/approveComment.ts
```diff
@@ -71,37 +69,4 @@ const approveComment = async (
...
 
-const enableRepliesToChildren = async (
-  commentID: string,
-  mongo: MongoContext
-) => {
-  const children = await mongo
-    .comments()
-    .find({
-      parentID: commentID,
-    })
```
nick-funk, line 82: This whole solution of tagging with `rejectedAncestor` would have worked pretty well if this filter was more specific. Namely, if it had `tenantID`, `storyID`, and `parentID` we likely wouldn't have had a performance concern.

I could have just added these filters, however, I opted to do the pure read `ancestorID` + rejected state lookup I did earlier because it avoids doing any writing of state to comments to get the same result.

src/core/server/stacks/approveComment.ts
```diff
@@ -71,37 +69,4 @@ const approveComment = async (
...
-  await mongo
-    .comments()
-    .updateMany(
-      { id: { $in: allChildIDs } },
-      { $set: { rejectedAncestor: false } }
-    );
-
-  await Promise.all(
-    nonRejectedChildIDs.map((id) => enableRepliesToChildren(id, mongo))
-  );
```
nick-funk, line 104: The recursive nature of doing many sequential Mongo calls is also why I opted for the read-only check for rejected ancestors.

Mongo isn't atomic, so we want to try and do all the updates in one `updateMany` call if possible. It helps avoids race conditions.

Also the way this is laid out, for a long and deep thread, we're going to see many recursive calls to write this `rejectedAncestor` flag out to all the children.

Which is what we see in Mongo when we look at the Profiler chart:

Screen Shot 2022-10-04 at 9 23 28 AM

All of that orange is the read/write updates to find by parentID and then update the flag as moderators go through and approve/reject comments. Unfortunately, writing a simple flag out adds a lot of traffic really quickly!

kabeaty (approved):
Looks good and works great! Thank you for including all of the detailed explanations of changes throughout.

PR #4071, opened 2022-10-06
[CORL-2674]: Fix React 18 upgrade bug and upgrade again

kabeaty:
## What does this PR do?

It fixes the bug related to the React 18 upgrade that was affecting the `MatchMedia` component. The bug is related to changes to effect cleanup timing: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#potential-issues. You can see the change in `onShadowRootWidthChange.ts`, line 73. We needed to capture `shadowRootFirstChild` so that it will not be set to `null` when the cleanup effect runs.

These changes also re-upgrade to React 18 since this bug is addressed.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [x] developers

none

no

n/a

## How do I test this PR?

To confirm that the bug is fixed, you can use Chrome dev tools to emulate a mobile phone, or you can go to `KeyboardShortcuts` and add `browserInfo.macOS` to line 1363 to get the keyboard shortcuts toolbar to show up on desktop, or you can run Coral locally on your mobile device. Whichever way you prefer to get the keyboard shortcuts toolbar (which uses `MatchMedia`) to show up. Make sure you are on the `All comments` tab on the stream. Click on `My Profile` and see it works as expected with no errors. Go back to the `All comments` tab. Change the sort order and see it works as expected with no errors. Go back to the `All comments` tab. Have at least one featured comment. See that clicking on the `Featured` tab works as expected with no errors.

You can also generally click around the stream/admin/install to test the React 18 upgrade more.


tessalt (approved):

src/core/client/ui/encapsulation/onShadowRootWidthChange.ts
```diff
@@ -81,7 +82,7 @@ export default function onShadowRootWidthChange(
     if (index > -1) {
       array.splice(index, 1);
       if (array.length === 0) {
-        value.resizeObserver.unobserve(shadowRoot.firstChild! as Element);
+        value.resizeObserver.unobserve(shadowRootFirstChild as Element);
```
nick-funk, line 85: Crazy how even today, in JS land, we're still basically dealing with a moving (ref)-pointer to an object. Since it changed out from underneath us, we can't clean up properly. Wild.

Good catch! Would be interested to hear how you sussed this out, as it looks like a very innocuous and hidden bug!

nick-funk (commented):
Wow, this was a sneaky one! Good hunting and well done!

src/core/client/ui/encapsulation/onShadowRootWidthChange.ts
marcushaddon: should we rewrite the frontend in Rust?

PR #4080, opened 2022-10-31
[CORL-2886] More generic title for "Log In" popout

marcushaddon:
## What does this PR do?
This PR changes the page title (as in the `<title></title>` element in the HTML head) from "Coral" to "Log In" to better suite users who want Coral to appear to be part of their own site. It also allows for custom variables to be passed into our server side templates.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [x] developers

None.

Yes. There is now an environment variable `SIGNIN_WINDOW_TITLE` that defaults to "Sign in", which should allow for overriding this value for localization/branding purposes.

n/a

## How do I test this PR?
As a commenter, click the "Sign in to join conversation" button, and note that the title of the new window that opens is "Log In" rather than "Coral"
 
## How do we deploy this PR?


src/core/server/app/router/client.ts
```diff
@@ -265,6 +272,7 @@ export async function mountClientRoutes(
       cacheDuration: false,
       disableFraming: true,
       entrypointLoader: manifestLoader.createEntrypointLoader("auth"),
+      templateVariables: { title: "Log In" },
```
nick-funk: How do we feel about making this `Sign in` to match our stream button language?

nick-funk (changes requested):

src/core/server/app/router/client.ts
marcushaddon: ✅ !

src/core/server/app/router/client.ts
```diff
@@ -265,6 +272,7 @@ export async function mountClientRoutes(
       cacheDuration: false,
       disableFraming: true,
       entrypointLoader: manifestLoader.createEntrypointLoader("auth"),
+      templateVariables: { title: "Sign in" },
```
nick-funk: Are we making this a configurable item via the env variables/config?

From what I see, this is only going to override the title with this "Sign in" default. Do we want to do a config via an env var? That fulfills the case notes requirements, but it might make sense to make configurable for non-english newsrooms.

If we don't do an env config, do we want to translate this? I don't know if this handler is initialized too early to use i18n and the tenant locale or not, but it would be nice if this title could be localized.

Thoughts?

nick-funk (commented):
This looks to cover the case requirements. Wanted to ask some refactor/robustness questions so I left this as comment. Let me know what you think on my ideas and we'll either push this through with an approve or refactor and re-review!

Thanks!

src/core/server/app/router/client.ts
marcushaddon: I think translation is the way to go, and these templates should be passed through translation. Looking into that.

marcushaddon: Scratch that, translation wasn't feasible, so I've added an env var

nick-funk (approved):

PR #4081, opened 2022-11-01
[CORL-2685] username aria labels

marcushaddon:
## What does this PR do?
This addresses an accessibility concern by giving usernames in the stream aria labels of the form "User <username>".

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None.

No.

n/a

## How do I test this PR?
Navigate to the stream, and inspect each instance of a username. Observe that usernames have aria labels of the form "User <username>".
 
 
## How do we deploy this PR?
No special considerations should necessary.


src/core/client/admin/components/Comment/InReplyTo.tsx
```diff
@@ -12,9 +12,15 @@ interface Props {
...
-    </BaseButton>
+    <Localized
+      id="common-username"
+      attrs={{ "aria-label": true }}
+      vars={{ username: children }}
+    >
+      <BaseButton onClick={onUsernameClick} className={styles.usernameButton}>
+        <span className={styles.username}>{children}</span>
+      </BaseButton>
+    </Localized>
```
nick-funk, line 23: I think we have to do similar in the stream side `InReplyTo.tsx`. Check `src/core/client/stream/tabs/Comments/Comment/InReplyTo.tsx`.

nick-funk (changes requested):
Looks good, just missing Localization on one of the InReplyTo components stream side. Rest of it looks to be using correct aria labels. Also, nice tests!

src/core/client/admin/components/Comment/InReplyTo.tsx
marcushaddon: I didn't update that one because it's not an interactive element, and from what I've read non interactive elements should not have aria labels, and should have descriptive visible text that the screen reader can read, and my screen reader reads this as "in reply to ...".

nick-funk (approved):

PR #4123, opened 2022-12-31
Multiple fixes for Akismet integration

### LINKED ISSUE #4122: Akismet blocking edited posts - not submitting recheck_reason?
It's end-of-year best-of list season over at my site, and many of our commenters are joining in with the writers and submitting their lists in the comments. You can see a number of examples [on this post](https://www.angrymetalguy.com/new-blood-same-bad-taste-the-amg-staff-pick-their-top-tenish-records-o-2022/) among others. This is the first year we've had Akismet turned on (as this is the first year we've started getting much spam posting to our Coral instance). While moderating, I've noticed the following problem:

If somebody submits a list post and then edits it, _most of the time_ Akismet will mark the edited post as spam, even though the original version was not.

Notably, the list posts are significantly wordier than the typical comments we get. My theory is that we're not currently correctly informing Akismet that this is an edit, and therefore it thinks it's a new post with identical content. When combined with being wordy and oddly formatted compared to normal comments, this seems to be enough to trip the spam filter reasonably reliably.

Looking at the Akismet API docs, I think the solution is fairly simple: if this is an edited comment, we should be sending `recheck_reason=edit`: https://akismet.com/development/api/#comment-check As you can see in the code, that's not currently happening: https://github.com/coralproject/talk/blob/36c7b84a48738ac14f0f3ea66cf8f8f20c6a0932/src/core/server/services/comments/pipeline/phases/spam.ts#L97

**Expected behavior:**

Non-spammy posts are not suddenly marked spam when they're edited.

**Actual behavior:**

Long but non-spammy posts are regularly marked spam when edited.

**Related Issues:**

None

**Versions:**

- NodeJS:
- NPM:
- MongoDB:
- Redis:
- Browser:
- OS:

Sentynel:
## What does this PR do?

This fixes a number of issues with the Akismet integration. This started as an attempt to fix #4122, but I ran into a number of other problems while I was working on it. Also, it doesn't seem to entirely fix that issue, but it does seem to have helped. Some of the other fixes here (successfully telling the Akismet API that something is not spam, for example) should also help. Hard to test definitively because spam detection is such a dark art.

Background: there are two separate parts of the flow in Coral that integrate with Akismet. When a comment is submitted, it's sent to Akismet for spam checking, and the results may affect what happens to the comment. Secondly, comments can be approved/unapproved, and the results of this for possible-spam comments will be sent to Akismet. Prior to this PR, these two sections are implemented entirely separately.

Here are the issues I've encountered, and fixes applied. Firstly, on the spam _checking_:
1. If the user-agent or referrer fields are not provided in the request or are empty, spam checking is abandoned. This means that you can bypass the spam filter by simply not supplying a user-agent; correct behaviour here is to tell Akismet that the user-agent was empty.
2. The recheck_reason field in the Akismet API request was not being populated for rechecked comments (for example, on edit), which seems to contribute to edited comments being marked as spam where their original versions were not. Annoyingly, the Akismet library used by this half of the integration does not correctly support the recheck_reason field, and it didn't look particularly actively maintained. I have instead merged this half with the direct Akismet API request made by the other half (it's not a particularly complicated API and most of the code is just argument marshalling we have to do anyway, so I didn't think the library was buying much).

There were more issues with the spam/not-spam submission, which I've ended up heavily rewriting:
1. Comments would be sent to Akismet as spam upon _user report_ of spam. This means that a random user can train the spam filter to block another target user, for example, which is clearly undesirable: submission should only happen on moderator confirmation. I've just disabled this hook.
2. Moderation approve/reject comment actions are intended to submit to Akismet if the comment was either detected as spam by Akismet, or reported as spam by a user. The checking of flag counts was incorrect and would never trigger. This is fixed.
3. The API requests sent to Akismet were invalid and would never be accepted, and the results were not checked. I have fixed the API requests, and added checks that the responses are as expected.
4. The approval path was confusing the request object of the _moderator_ making the request with the request object that created the comment, and would therefore submit the moderator's user-agent and referrer to Akismet as if they belonged to the spam comment. We don't store these for the comment, so don't know them in this situation. I have just removed sending them. (I don't know why this didn't happen on the near-identical reject path as well and resisted trying to work it out.)

There's quite a bit going on here, so I anticipate some review comments!

## These changes will impact:

- [x] commenters
- [x] moderators
- [ ] admins
- [ ] developers

None

No

None

## How do I test this PR?

Add some comments, edit those comments, report those comments as spam, and approve/reject them in the moderation panel. Verify that correct requests are generated to Akismet when this happens: added comments are spam checked; approved/rejected comments that were marked as possible spam are submitted; requests are not made in other situations. You may want to debug log the outgoing requests, and turn on the is_test API parameter to avoid accidentally banning yourself. You can set your username to "viagra-test-123" to force Akismet to declare you as spam.
 
## How do we deploy this PR?

No special requirements

### DIFF
--- a/src/core/server/services/comments/actions.ts
+++ b/src/core/server/services/comments/actions.ts
@@ -37,7 +37,6 @@ import {
   publishCommentFlagCreated,
   publishCommentReactionCreated,
 } from "../events";
-import { submitCommentAsSpam } from "../spam";
 
 export type CreateAction = CreateActionInput;
 
@@ -417,16 +416,6 @@ export async function createFlag(
     ).catch((err) => {
       logger.error({ err }, "could not publish comment flag created");
     });
-
-    const revision = getLatestRevision(comment);
-    if (
-      revision &&
-      tenant.integrations.akismet.enabled &&
-      action.actionType === ACTION_TYPE.FLAG &&
-      action.reason === GQLCOMMENT_FLAG_REPORTED_REASON.COMMENT_REPORTED_SPAM
-    ) {
-      await submitCommentAsSpam(mongo, tenant, comment, request);
-    }
   }
 
   return comment;
--- a/src/core/server/services/comments/pipeline/phases/spam.ts
+++ b/src/core/server/services/comments/pipeline/phases/spam.ts
@@ -1,4 +1,4 @@
-import { Client } from "akismet-api";
+import fetch from "node-fetch";
 
 import { SpamCommentError } from "coral-server/errors";
 import { ACTION_TYPE } from "coral-server/models/action/comment";
@@ -11,6 +11,19 @@ import {
   GQLCOMMENT_FLAG_REASON,
   GQLCOMMENT_STATUS,
 } from "coral-server/graph/schema/__generated__/types";
+
+interface AkismetParameters {
+  api_key: string;
+  blog: string;
+  user_ip: string;
+  user_agent: string;
+  referrer: string;
+  permalink: string;
+  comment_type: string;
+  comment_author: string;
+  comment_content: string;
+  recheck_reason?: string;
+}
 
 export const spam: IntermediateModerationPhase = async ({
   story,
@@ -54,12 +67,6 @@ export const spam: IntermediateModerationPhase = async ({
     return;
   }
 
-  // Create the Akismet client.
-  const client = new Client({
-    key: integration.key,
-    blog: integration.site,
-  });
-
   // Grab the properties we need.
   let userIP = "";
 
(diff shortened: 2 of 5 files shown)


tessalt:
Hi there! Thanks so much for your contribution, it looks very well researched and thorough. As the akismet integration is a fairly large piece of functionality used by many of our customers and open source users, we're going to have to take some time to make sure we've validated any changes we make to it. Our team is fairly small at the moment, so I'm not sure when we'll be able to get this merged, but we're adding it to our backlog to be prioritized shortly. Thanks again!

Sentynel:
Yes, of course. Happy to provide any details that might help. This is currently in production on my instance and I'm monitoring for issues.

nick-funk:
Thank you again @Sentynel! We have combed through this PR and integrated these fixes into a PR targeting our latest mono-repo changes (6.5.3+).

PR here: https://github.com/coralproject/talk/pull/4438

As well, we've updated the Akismet API so that it matches the latest expected types. We've also made sure that the `rechecked` state is being passed during edited/multi-revision comments so that this edge case is handled.

You should see it come out soon in a new release of Coral. Thank you so much for submitting a PR, bringing this to our attention and giving us so much to dig into along with your very thorough explanations!

PR #4127, opened 2023-01-05
[CORL-2376 ]quick domain moderation

marcushaddon:
## What does this PR do?
This PR implements a feature that allows moderators to create an email domain level ban at the same time they are modifying a user's ban status (in addition to the existing work flow available under Configure -> Moderation -> Email Domain

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

The `createEmailDomain` mutation has been updated to allow `MODERATOR` roles to execute it, and some additional checks have been put in place to prevent unauthorized access.

No.

n/a

## How do I test this PR?
1. As a moderator, site moderator, or admin, navigate to Admin -> Community
2. Choose a user who is not banned
3. Click the user's status and select "Manage Ban"
4. Observe that there is an option to "Ban all new accounts on {domain}" where domain is the domain of the user's email address
5. check this box, and ban the user (on all sites or just on some)
6. Observe that the user is banned
7. Navigate to Configure > Moderation > Email Domains and observe that the users email domain has been added to the list as "Ban all new commenter accounts"
8. This same workflow should be available from the Ban Modal when moderating reported comments (accessed via "b" hotkey
9. The options should not be available if the users email domain is already moderated
10. The option should not be available if there is no email stored for the user.
 
## How do we deploy this PR?


src/core/common/email.ts
```diff
@@ -0,0 +1,5 @@
+export const extractDomain = (email: string): string => {
+  const domain = email.split("@")[1];
```
marcushaddon: It could be argued that this isn't the exact definition of an email's domain, but this is the logic we use according to https://github.com/coralproject/talk/blob/develop/src/core/server/graph/schema/schema.graphql#L1799

nick-funk: yeah, there are proper ways to validate (we use Joi's built in `.email(...)` validator to do full IANA checking... but that's all server side and we don't want to include that package client side right now (it's huge).

I tried to be a turd and send myself an email with 2 @ signs on Outlook/Gmail and they both reject the email address, so I think we're safe to assume one @ per email.

I would guard clause the edge case that the split is less than length of 2 and return empty string just in case though. Never know when you're going to get a weird split with nothing after the @ (though I doubt it'll ever actually happen in practice).

src/core/server/services/tenant/tenant.ts
```diff
@@ -298,8 +305,42 @@ export async function createEmailDomain(
...
+
+  const tenantSites = await retrieveTenantSites(mongo, tenant.id);
+  const isAdmin = fullViewer.role === GQLUSER_ROLE.ADMIN;
+  const multiSiteEnabled = tenantSites.length > 1;
+  const modOnSingleSite = isModerator(fullViewer) && !multiSiteEnabled;
+  const orgModOnMultiSite = isOrgModerator(fullViewer) && multiSiteEnabled;
+
+  const allowed =
+    fullViewer.tenantID === tenant.id &&
+    (isAdmin || modOnSingleSite || orgModOnMultiSite);
```
nick-funk, line 334: Since this is allowing the moderator to be allowed whether they're single site or multi-site, does it matter if they're single or multi-site? Could we do away with retrieving whether the tenant has just one or more sites and simple check if the user's role is a moderator or higher?

Though, I could be wrong, please inform me if I'm missing some edge case here... as always, the single vs multi-site moderators is complicated.

nick-funk (changes requested):
Overall, this looks good. There are some small questions I wanted to ask before letting it through. One seemed like it could simplify the auth logic for the request and save you retrieving tenant sites, so please have a look at it. Let me know if you disagree or have questions, I could be wrong.

src/core/common/email.ts
marcushaddon: good catch!

src/core/server/services/tenant/tenant.ts
marcushaddon: If we did that it wouldn't catch the case in which their role is Moderator, but they are a site mod on a multi-site tenant, in which case we don't want them to be able to create email domain bans (according to the ticket). So we want to allow "Org mods on multi site tenants" or "any mod on single site tenants", but not "any mod on multi site tenants".

This is one of the many places where our logic is muddied by Org Mods and Site Mods not being distinct roles. If that were the case, all mods on single site tenants would have the Org Mod role, and we could just check for Org Mod role and ignore whether or not the tenant has multi site enabled.

nick-funk: Oh! I'm sorry, I saw `isModerator` and `isOrgModerator` and thought they were the same when I was reviewing. Looking at it fresh right now, I see they are different and this makes sense!

nick-funk (approved):

PR #4128, opened 2023-01-06
Shard Jest Unit Tests

wyattjoh:
This PR serves to use [CircleCI's matrix feature](https://circleci.com/docs/using-matrix-jobs/) to parallelize Jest tests to improve CI performance. Originally, testing took around 14 minutes to complete, now it only takes {{ waitingOnCI }}!

❤️


wyattjoh:
Ah, it turns out y'all are still on `jest` 26, `--shard` was introduced in 28. I attempted upgrading but it required upgrades to `ts-jest` and therefore `typescript`, which was too big a lift right now!

Closing for now until I can take a look!

PR #4131, opened 2023-01-11
Add stable classnames to more DOM elements in stream

tessalt:
## What does this PR do?

Adds more stable classnames to stream DOM elements to prevent usage of attribute selectors in custom stylesheets, making the DOM more future-proof and making customization simpler and more reliable. I tried to be conservative and base my additions on the custom stylesheets we maintain at Vox Media. The one exception is the addition of the `coral-icon` class to all icons. This somewhat contradicts vinh's initial direction of adding stable classnames to specific instances rather than components, which i assume was to encourage authors of custom stylesheets to use css variables where possible. I've made this exception for icons because they account for a large number of the attribute selectors i've seen at Vox Media and in the wild, as changing icons is a very common customization. In combination with the container classes already in place and that have been added here, this one classname should provide enough specificity to enable simple customization. 

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

## How do I test this PR?

View the stream and inspect impacted elements to see new classnames. Inspect the DOM in the "preferences" tab to see new classnames for all headings and sections.

### DIFF
--- a/src/core/client/admin/test/auth/__snapshots__/addEmailAddress.spec.tsx.snap
+++ b/src/core/client/admin/test/auth/__snapshots__/addEmailAddress.spec.tsx.snap
@@ -67,7 +67,7 @@ exports[`accepts valid email 1`] = `
       >
         <i
           aria-hidden="true"
-          className="Icon-root Icon-sm MessageIcon-root"
+          className="Icon-root Icon-sm coral coral-icon MessageIcon-root"
         >
           warning
         </i>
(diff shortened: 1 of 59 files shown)


nick-funk (approved):
These changes look good to me!

The fact that over half the PR is snapshot updates confirms to me just how useless and annoying snapshot tests are (lol)

marcushaddon:
Also looks good to me!

marcushaddon (approved):

PR #4135, opened 2023-01-17
[fix/CORL-2725] install and directly use bundlesize2

marcushaddon:
## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -129,7 +129,7 @@ jobs:
           no_output_timeout: 30m
       - run:
           name: Verify Bundle Size
-          command: npx bundlesize2 --enable-github-checks
+          command: ./node_modules/.bin/bundlesize --enable-github-checks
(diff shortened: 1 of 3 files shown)


nick-funk (approved):

.circleci/config.yml
```diff
@@ -129,7 +129,7 @@ jobs:
           no_output_timeout: 30m
       - run:
           name: Verify Bundle Size
-          command: npx bundlesize2 --enable-github-checks
+          command: ./node_modules/.bin/bundlesize --enable-github-checks
```
nick-funk, line 132: this threw me for a loop for a moment... did not realize that `bundlesize2` is using a binary called `bundlesize`... cause that's not confusing lol

this works great though, thanks for fixing this up!

PR #4141, opened 2023-01-20
[CORL-2727] Promoting site mods to org mods

marcushaddon:
## What does this PR do?
This PR fixes a bug in which a site moderator cannot be promoted to an org moderator. 

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

None.

No

n/a

## How do I test this PR?
1. As an admin user, navigate to admin/community
2. Find a site moderator user, and click their role, opening the role change dropdown
3. Click organization moderator
4. Observe that their role has been update to org mod
 
## How do we deploy this PR?

### DIFF
--- a/src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
+++ b/src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
@@ -38,11 +38,6 @@ const UserRoleChangeContainer: FunctionComponent<Props> = ({
   );
   const handleOnChangeRole = useCallback(
     async (role: GQLUSER_ROLE_RL, siteIDs?: string[]) => {
-      if (role === user.role) {
-        // No role change is needed! User already has the selected role.
-        return;
-      }
-
       await updateUserRole({ userID: user.id, role, siteIDs });
     },
     [user, updateUserRole]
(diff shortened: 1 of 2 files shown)


src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
```diff
@@ -38,11 +38,6 @@ const UserRoleChangeContainer: FunctionComponent<Props> = ({
   );
   const handleOnChangeRole = useCallback(
     async (role: GQLUSER_ROLE_RL, siteIDs?: string[]) => {
-      if (role === user.role) {
```
marcushaddon, line 41: I considered adding an check to see if in the case of matching roles, the roles were "MODERATOR" and all the siteIDs were the same as the user's current site ids but decided to just remove this check on the frontend because it is also handled on the backend and makes the code more understandable. Let me know if you disagree!

nick-funk: As long as we're still preventing them from clicking on "change role" via disabling/hiding options (which I believe we are) this should be fine.

nick-funk (approved):

PR #4158, opened 2023-02-17
[CORL-2729] Moderation Queue Focus 

marcushaddon:
## What does this PR do?
This PR fixes a bug in which upon navigating to `/admin/moderate`, focus immediately shifts to the first moderation card, which is an anti-pattern for accessibility. 

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

None.

No.

n/a

## How do I test this PR?
1. Ensure that there are multiple comments in the moderation queue
2. As an admin or moderator, navigate to `/admin/moderate`
3. Press the tab key
4. Observe that the focus begins at the first element on the page (the nav at the top) and cycles through the different nav options, rather than skipping them and jumping to the beginning of the mod queue
5. Observe that using the "j" and "k" keys still cycles through mod cards
 
 
## How do we deploy this PR?
No special considerations should be necessary


src/core/client/admin/routes/Moderate/Queue/Queue.tsx
```diff
@@ -99,6 +109,28 @@ const Queue: FunctionComponent<Props> = ({
...
   }, [window.document]);
 
+  useEffect(() => {
+    key(HOTKEYS.NEXT, QUEUE_HOTKEY_ID, selectNext);
+    key(HOTKEYS.PREV, QUEUE_HOTKEY_ID, selectPrev);
+
+    // key(HOTKEYS.BAN, QUEUE_HOTKEY_ID, onBan);
+
+    // key(HOTKEYS.APPROVE, QUEUE_HOTKEY_ID, onApprove);
+    // key(HOTKEYS.REJECT, QUEUE_HOTKEY_ID, onReject);
```
nick-funk: are these from past debugging or do they need to be uncommented?

src/core/client/admin/routes/Moderate/Queue/Queue.tsx
```diff
@@ -99,6 +109,28 @@ const Queue: FunctionComponent<Props> = ({
...
+
+    // The the scope such that only events attached to the ${id} scope will
+    // be honored.
+    key.setScope(QUEUE_HOTKEY_ID);
+
+    return () => {
+      // Remove all events that are set in the ${id} scope.
+      key.deleteScope(QUEUE_HOTKEY_ID);
+    };
+    // key.deleteScope(id);
```
nick-funk: old debug code?

nick-funk (changes requested):

src/core/client/admin/routes/Moderate/Queue/Queue.tsx
marcushaddon: ✅

nick-funk (approved):
Looks good! Nice work!

PR #4163, opened 2023-02-28
[CORL-2578] Fix Shift + A mark all comments seen keyboard shortcut

nick-funk:
## What does this PR do?

Fixes Shift + A mark all comments seen keyboard shortcut to mark all comments as seen and all nested replies.

It does this by marking all as seen using recursion to walk relay's data store.

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None

No

No new indices added.

## How do I test this PR?

- Enable `Z_KEY` and `COMMENT_SEEN`
- Login on stream
- See yellow not seen colouring (usually yellow)
- Press `Shift + A` 
- All comments should now be marked as seen (usually white)
 
## How do we deploy this PR?

No special considerations

### DIFF
--- a/src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
+++ b/src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
@@ -73,10 +73,12 @@ const markAllAsSeenRecursive = (
       continue;
     }
 
-    const childComments =
-      repliesConnection
-        .getLinkedRecords("edges")
-        ?.map((e) => e.getLinkedRecord("node")) || [];
+    const edges = repliesConnection.getLinkedRecords("edges");
+    if (!edges || edges.length === 0) {
+      continue;
+    }
+
+    const childComments = edges.map((e) => e.getLinkedRecord("node")) || [];
     if (!childComments || childComments.length === 0) {
       continue;
     }
@@ -103,9 +105,12 @@ const markAllAsSeen = (
     return;
   }
 
-  const comments = connection
-    .getLinkedRecords("edges")
-    ?.map((e) => e.getLinkedRecord("node"));
+  const edges = connection.getLinkedRecords("edges");
+  if (!edges || edges.length === 0) {
+    return;
+  }
+
+  const comments = edges.map((e) => e.getLinkedRecord("node"));
   if (!comments) {
     return;
   }
--- a/src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
+++ b/src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
@@ -52,18 +52,18 @@ const getReplies = (commentProxy: RecordProxy<{}>) => {
 
   const edges = repliesConnection.getLinkedRecords("edges");
   if (edges && edges.length > 0) {
-    allEdges.push(...edges);
+    edges.forEach((e) => allEdges.push(e));
   }
   const viewNewEdges = repliesConnection.getLinkedRecords("viewNewEdges");
   if (viewNewEdges && viewNewEdges.length > 0) {
-    allEdges.push(...viewNewEdges);
+    viewNewEdges.forEach((e) => allEdges.push(e));
   }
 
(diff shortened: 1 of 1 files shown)


src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
```diff
@@ -35,35 +37,87 @@ type Input = Omit<MarkCommentsAsSeenInput, "clientMutationId"> & {
...
+  }
+
+  const connection = ConnectionHandler.getConnection(story, "Stream_comments", {
+    orderBy: orderBy,
+  });
+  if (!connection) {
+    return;
+  }
+
+  const edges = connection.getLinkedRecords("edges");
```
kabeaty, line 161: We also need to get `viewNewEdges` and combine them as before, to maintain the behavior that shift + A marks all new top-level comments and replies that have come in since the stream was loaded but have not yet been revealed as read.

src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
```diff
@@ -35,35 +37,87 @@ type Input = Omit<MarkCommentsAsSeenInput, "clientMutationId"> & {
...
-        const allChildCommentsEdges =
-          proxy
-            .getLinkedRecord("allChildComments")
-            ?.getLinkedRecords("edges") || [];
-        for (const reply of allChildCommentsEdges) {
-          const replyID = reply.getLinkedRecord("node")?.getValue("id");
-          if (replyID) {
-            const replyProxy = store.get(replyID.toString());
-            if (replyProxy) {
-              replyProxy.setValue(!!input.updateSeen, "seen");
```
kabeaty, line 58: This was marking the `allChildComments` as seen, which is used in `KeyboardShortcuts` to detect whether the mobile toolbar should have its `Mark all as read` and `Next unread` buttons disabled or not. With these changes to remove this, if you use the toolbar to mark all comments as seen, these buttons are still enabled until the stream is refreshed.

kabeaty (changes requested):
I left two comments around functionality that's different with these changes. Just let me know if you have any questions about them!

src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
nick-funk: Ah, yes, I can pull in `viewNewEdges` for sure, I'll make that fix.

As for why I'm walking replies instead of `allChildComments`, I noticed that relay stops having links to `allChildComments` at the flatten replies barrier, leaving comments un-marked as seen. So I walked the replies connections instead as they are always guaranteed to exist in Relay. I know this might seem odd because we are querying `allChildComments` to fill Relay, but despite this, the stream is built on cached connections, not the queries we make, those connections are the root comment connection (stream) and the subsequent child streams (replies). To ensure this sets all comments as marked seen, I iterated on those connections since they are always available in Coral.

I forwarded the `input.updateSeen` value into my new function. I did my best to ensure that the `setValue` is pulling/setting the exact same value you had set/passed-in before.

nick-funk: Update: I've made it walk both the replies connection and the `allChildComments` in the relay cache. That should fix this issue with it not updating `KeyboardShortcuts`. I am also updating it to walk `viewNewEdges`.

nick-funk: Further update: made sure to pull in `edges` and `viewNewEdges` at all levels for the root comments connection, the replies, and also the `allChildComments`.

src/core/client/stream/tabs/Comments/Comment/MarkCommentsAsSeenMutation.tsx
nick-funk: Done! More details in my other replies on this PR.

nick-funk:
Closing and doing this only in redis-cache changes.

PR #4164, opened 2023-03-01
remove incorrect class name from reaction button icon

tessalt:
This code was adding the container class name to an icon, causing it to render incorrectly.


src/core/client/stream/tabs/Comments/Comment/ReactionButton/ReactionButtonContainer.tsx
```diff
@@ -126,7 +125,6 @@ const ReactionButtonContainer: FunctionComponent<Props> = ({
       readOnly={readOnly}
       isQA={isQA}
       author={comment.author?.username}
-      iconClassName={CLASSES.comment.actionBar.reactButton}
```
kabeaty, line 129: I think we can remove the `iconClassName` prop from `ReactionButton` now since this was the only place it was being used.

kabeaty (approved):
Looks good! Approving, but left a comment about removing the now-unused `iconClassName` prop too.

PR #4169, opened 2023-03-06
[CORL-2742] Make dashboard staff counts include admins, mods, site mods, and staff

nick-funk:
## What does this PR do?

Updates dashboard count metrics to include admins, mods, site mods, and staff comment counts in its calculations.

## These changes will impact:

- [ ] commenters
- [X] moderators
- [X] admins
- [ ] developers

No changes.

No new feature flags.

No new indexes.

## How do I test this PR?

- Go to `Admin > Dashboard`
    * Check the amount of `Staff + comments` you have
- Make some staff/admin/mod comments using various roles on a stream
- Go to `Admin > Dashboard`
- Check that comment counts for `Staff + comments` matches the counts you created
 
## How do we deploy this PR?

No special considerations.


src/locales/en-US/admin.ftl
```diff
@@ -1698,7 +1698,7 @@ dashboard-alltime-rejections-archiveEnabled = { $value } { $unit ->
     }
     *[other] unknown unit
   } average
-dashboard-today-staff-comments = Staff comments
+dashboard-today-staff-plus-comments = Staff + comments
```
kabeaty: Should this be `Staff+ comments` without a space instead of `Staff + comments`?

kabeaty (changes requested):
Looking good overall! I left one comment confirming spacing for the updated `Staff+` copy. 

Noting here for reference as well, but as mentioned, I also noticed while testing that when I leave a comment as an admin and refresh the dashboard as another user, the admin comment doesn't show up in the dashboard comment count until I leave a comment as another staff role (like moderator), and then all comments show correctly and it's incremented for admin as well. When I leave a comment as a moderator and refresh the dashboard as another user, the count increments as expected right away. Can you look into this?

src/locales/en-US/admin.ftl
nick-funk: Oh my... I am not going to lie, my brain 100% thought `Staff+` was a typo in the requirements doc and did not think on it one second further. `Staff+` makes so much more sense when you point it out, I will fix this!

nick-funk: Fixed!

nick-funk:
> Looking good overall! I left one comment confirming spacing for the updated `Staff+` copy.
> 
> Noting here for reference as well, but as mentioned, I also noticed while testing that when I leave a comment as an admin and refresh the dashboard as another user, the admin comment doesn't show up in the dashboard comment count until I leave a comment as another staff role (like moderator), and then all comments show correctly and it's incremented for admin as well. When I leave a comment as a moderator and refresh the dashboard as another user, the count increments as expected right away. Can you look into this?

Thank you for testing! I will definitely look into this. Hopefully, it's as simple as a tag not being set or broadcasted properly.

nick-funk:
> Looking good overall! I left one comment confirming spacing for the updated `Staff+` copy.
> 
> Noting here for reference as well, but as mentioned, I also noticed while testing that when I leave a comment as an admin and refresh the dashboard as another user, the admin comment doesn't show up in the dashboard comment count until I leave a comment as another staff role (like moderator), and then all comments show correctly and it's incremented for admin as well. When I leave a comment as a moderator and refresh the dashboard as another user, the count increments as expected right away. Can you look into this?

The admin comment count issue you described is now fixed, I needed to switch to the older cursor iteration format (different from the current mongo docs) as it was only sometimes iterating the cursor correctly.

It should now show correct counts!

kabeaty (approved):

PR #4172, opened 2023-03-07
[CORL-2660] Mitigate duplicate story ID errors by updating story url if a valid id is provided

nick-funk:
## What does this PR do?

During the retrieval of a comment stream via the createStreamEmbed, if the site provides a valid storyID that matches a real story but the url it is coming from differs from the one on record, we update the story url.

This prevents numerous "DUPLICATE_STORY_ID" errors when newsrooms modify their story URL multiple times in their CMS due to article title changes, slug changes, etc.

## These changes will impact:

- [X] commenters
- [X] moderators
- [X] admins
- [X] developers

None

No

No new indices.

## How do I test this PR?

- create a story in `multi-site-test`
- navigate to story and see that it creates the new story
- post a unique comment you'll remember
- using the id of the story you just created, navigate to `/story/alt/<STORY_ID>` instead
- see that it loads the same comment stream you just created despite the URL being different and has the comment you made previously
- check Mongo and query that story id, see that its URL is now the new `/alt/` URL
- go back to the old URL without the `/alt/` part
- see that is again still loads the stream with your comment on it
- check Mongo and query that story id, see that its URL is now back to the old URL without the `/alt/`
 
## How do we deploy this PR?

No special considerations


kabeaty (approved):
Looks great!

yogeshbeniwal:
Can this also take care of https://github.com/coralproject/talk-wp-plugin/issues/24

nick-funk:
> Can this also take care of [coralproject/talk-wp-plugin#24](https://github.com/coralproject/talk-wp-plugin/issues/24)

Sorry @yogeshbeniwal, unfortunately, this will not resolve the canonical URL support you're looking for. This merely keeps one story updated to one URL and does not handle multi-url to one-story mapping.

PR #4174, opened 2023-03-09
[CORL-2766] protect common email domains from domain moderation

marcushaddon:
## What does this PR do?
This PR creates a hardcoded list of email domains considered to common to be moderated domain-wide.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

None.

No.

n/a

## How do I test this PR?
1. As an org mod or admin attempt to ban any of the following domains, either through the checkbox presented while managing a user's ban status, or under Configure > Moderation > Email Domain
voxmedia.com
gmail.com
gmail.co.uk
aol.com
att.net
duck.com
verizon.net
yahoo.com
yahoo.co.nz
yahoo.com.au
yahoo.co.za
xtra.co.nz
btinternet.com
yahoo.se
comcast.net
outlook.com
icloud.com
rr.com
hotmail.com
hotmail.co.uk
hotmail.es
hotmail.fr
msn.com
me.com
live.co.uk
live.com
live.co.za
rogers.com
global.co.za
ncacoral.com.au
proton.me
westnet.com.au
laposte.net
free.fr
orange.fr
outlook.fr
yahoo.fr
online.no
live.no
yahoo.no
hotmail.no
2. Observe that the domain is not moderated (in the configure view, you should see a detailed error message. In the user ban status modal, you will not see any indication that the email domain ban failed, which is a preexisting behavior and is out of the scope of this ticket).
 
## How do we deploy this PR?
No special considerations should be necessary.


src/core/server/services/tenant/tenant.ts
```diff
@@ -41,6 +42,8 @@ import {
   GQLUSER_ROLE,
 } from "coral-server/graph/schema/__generated__/types";
 
+import { PROTECTED_EMAIL_DOMAINS } from "coral-common/constants";
```
kabeaty: Can you move these imports up?

nick-funk:
How do we feel about adding `fastmail.com` to this list? I only ask as I see we added `proton.me` to this list already.

nick-funk (approved):
Works as labelled! Ship it!

kabeaty (changes requested):
Looks good overall! Left a small comment. Also, I'm seeing an error thrown when testing out using the user ban status modal to apply a common email ban that I think we will need to handle? And are you able to add a quick test for these common email domains not being able to be banned, as well?

src/core/server/services/tenant/tenant.ts
marcushaddon: ✅

src/core/server/services/tenant/tenant.ts
```diff
@@ -41,6 +43,7 @@ import {
   GQLUSER_ROLE,
 } from "coral-server/graph/schema/__generated__/types";
 
+import { ERROR_CODES } from "coral-common/errors";
```
kabeaty: one more import to move up :)

kabeaty (approved):
Looks good, thanks for the updates! Just one more small comment, but approving!

src/core/server/services/tenant/tenant.ts
marcushaddon: ✅

PR #4175, opened 2023-03-10
Slovak translation

miroc:
## What does this PR do?

Adds partial Slovak (sk-SK) translations to the project.

## How do I test this PR?

Switch to Slovak language and manually check if translatable strings are translated.


nick-funk (approved):
Thank you for submitting translations!

PR #4181, opened 2023-03-16
[CORL-2775] match ban modal design

marcushaddon:
## What does this PR do?
This PR makes the Manage Ban Modal match the design spec.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

None.

No.

n/a

## How do I test this PR?
1. As a user with ban management permissions, open the ban modal either by clicking "Manage Ban" in a user row in the admin/community view, or by pressing "B" in the moderation queue
2. Observe that the ban modal more closely resembles the attached design picture
 
## How do we deploy this PR?
No special considerations necessary.


src/core/client/admin/components/BanModal.tsx
```diff
@@ -309,54 +313,131 @@ const BanModal: FunctionComponent<Props> = ({
...
-                        onChange={({ target }) => {
-                          setBanDomain(target.checked);
-                        }}
-                      >
-                        Ban all new accounts on <strong>{emailDomain}</strong>
-                      </CheckBox>
-                    </Localized>
+                    <Flex direction="column" className={styles.banDomainOption}>
+                      <HorizontalGutter spacing={2}>
+                        {/* doman ban header */}
```
kabeaty: typo --> `doman` --> `domain`

src/core/client/admin/components/BanModal.tsx
```diff
@@ -309,54 +313,131 @@ const BanModal: FunctionComponent<Props> = ({
...
+                {/* MAIN */}
+                <Flex
+                  direction="column"
+                  className={styles.form}
+                  justifyContent="flex-start"
+                >
+                  {/* BAN FROM/REJECT COMMENTS */}
+                  <Flex direction="column">
+                    {/* ban from header */}
+                    <Label className={styles.banFromHeader}>Ban From</Label>
```
kabeaty: Can you please localize this? Also, it looks like it should be `Ban from` (lowercase `from`) to match specs.

src/core/client/admin/components/BanModal.tsx
```diff
@@ -309,54 +313,131 @@ const BanModal: FunctionComponent<Props> = ({
...
-                          setBanDomain(target.checked);
-                        }}
-                      >
-                        Ban all new accounts on <strong>{emailDomain}</strong>
-                      </CheckBox>
-                    </Localized>
+                    <Flex direction="column" className={styles.banDomainOption}>
+                      <HorizontalGutter spacing={2}>
+                        {/* doman ban header */}
+                        <Label className={styles.domainBanHeader}>
```
kabeaty: Can you localize this label too?

src/core/client/admin/components/BanModal.tsx
```diff
@@ -309,54 +313,131 @@ const BanModal: FunctionComponent<Props> = ({
...
+                  justifyContent="flex-start"
+                >
+                  {/* BAN FROM/REJECT COMMENTS */}
+                  <Flex direction="column">
+                    {/* ban from header */}
+                    <Label className={styles.banFromHeader}>Ban From</Label>
+                    <Flex
+                      direction="row"
+                      className={styles.sitesOptions}
+                      justifyContent="flex-start"
```
kabeaty, line 332: I think this `Flex` component wrapping the sites options should still have the `spacing={5}` prop set. As is, the options look too close. It looks like the spacing is now being set with `gap`, but I think we should go with the `spacing` prop for consistency of how spacing styles are being applied when using `Flex`. 

Also, previously the class name used was `sitesToggle`. Can you clean up this old class name since it's unused now in favor of `sitesOptions`?

src/core/client/admin/components/BanModal.css
```diff
@@ -15,3 +15,34 @@ $ban-modal-text: var(--palette-text-500);
...
 }
+
+.domainBanHeader {
+  margin-bottom: var(--spacing-2);
+
+}
+
+
+.form {
+  gap: var(--spacing-3);
```
kabeaty: Can you set the `spacing` prop on this instead of using `gap` here to apply spacing, for consistency?

src/core/client/admin/components/BanModal.tsx
```diff
@@ -309,54 +313,131 @@ const BanModal: FunctionComponent<Props> = ({
...
+                        <Localized
+                          id="community-banModal-banEmailDomain"
+                          vars={{ domain: emailDomain }}
+                        >
+                          <CheckBox
+                            onChange={({ target }) => {
+                              setBanDomain(target.checked);
+                            }}
+                          >
+                            Ban all new accounts on{" "}
```
kabeaty: Should this be `Ban all new commenter accounts from {domain}`?

src/core/client/admin/components/BanModal.css
```diff
@@ -15,3 +15,34 @@ $ban-modal-text: var(--palette-text-500);
...
+  margin-bottom: var(--spacing-3);
+}
+
+.banFromHeader {
+  margin-bottom: var(--spacing-2);
+  margin-top: 0;
+  font-family: var(--font-family-primary);
```
kabeaty: Is it necessary to set `font-family` here? Looks like it's already set by the `Label` styles.

src/core/client/admin/components/BanModal.css
```diff
@@ -15,3 +15,34 @@ $ban-modal-text: var(--palette-text-500);
 .sitesToggle {
   margin-bottom: var(--spacing-3);
 }
+
+.domainBanHeader {
+  margin-bottom: var(--spacing-2);
+
```
kabeaty: Can you clean up this extra space here?

src/core/client/admin/components/BanModal.css
```diff
@@ -15,3 +15,34 @@ $ban-modal-text: var(--palette-text-500);
...
+
+.form {
+  gap: var(--spacing-3);
+  margin-bottom: var(--spacing-3);
+}
+
+.banFromHeader {
+  margin-bottom: var(--spacing-2);
+  margin-top: 0;
```
kabeaty: It doesn't look like setting `margin-top` to 0 is needed here...can you confirm and remove if so?

src/core/client/admin/components/BanModal.css
```diff
@@ -15,3 +15,34 @@ $ban-modal-text: var(--palette-text-500);
...
+  gap: var(--spacing-2);
+}
+
+.banDomainOption {
+  margin: var(--spacing-1) 0;
+}
+
+.customizeMessage {
+  align-self: flex-start;
+  color: -var(--palette-grey-800);
```
kabeaty: Looks like this has an extra dash in front of it making it invalid? It also looks like this darker grey is getting overridden by the button color as is...can you look into this?

kabeaty (changes requested):
Looks nice! I left a few comments/questions about copy and styles throughout.

src/core/client/admin/components/BanModal.tsx
marcushaddon: ✅

src/core/client/admin/components/BanModal.tsx
marcushaddon: ✅

src/core/client/admin/components/BanModal.tsx
marcushaddon: ✅

src/core/client/admin/components/BanModal.tsx
marcushaddon: ✅

src/core/client/admin/components/BanModal.css
marcushaddon: I'll need to refresh my memory as to why on monday, but without this, the header styles break when opening the Ban Modal from the moderation queue. There might be a better way to address that though.

src/core/client/admin/components/BanModal.css
marcushaddon: ✅

src/core/client/admin/components/BanModal.css
marcushaddon: ✅

src/core/client/admin/components/BanModal.css
marcushaddon: ah, you're right. The design shows it being darker. I think I need some input on what to do when the design doesn't fit our component library

src/core/client/admin/components/BanModal.tsx
marcushaddon: ✅

src/core/client/admin/components/BanModal.css
marcushaddon: ✅

src/core/client/admin/components/BanModal.css
marcushaddon: aaaaand after fixing the bug preventing me from reproducing this, I cannot reproduce. So, removed!
✅

src/core/client/admin/components/BanModal.css
marcushaddon: ✅

src/core/client/admin/components/BanModal.css
kabeaty: I'm still seeing an extra space here?

marcushaddon: ah, i only noticed the extra space within the brackets

marcushaddon: ✅

kabeaty (approved):
Looks great! Thanks for making the changes.

PR #4183, opened 2023-03-21
fix for undefined window on resize events

tessalt:
## What does this PR do?

One of our SaaS clients was seeing an issue when window was resized, either horizontally via window resizing, or vertically, via opening and closing the "reply" dialog, where they would get an error `cannot call removeEventListener on undefined`. We're not exactly sure what is leading to window being undefined in only this client's use case (inside a progressive web app) but the client has locally validated that adding this check does resolve the issue. 

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

### DIFF
--- a/src/core/client/ui/encapsulation/useWindowBreakpointClasses.ts
+++ b/src/core/client/ui/encapsulation/useWindowBreakpointClasses.ts
@@ -26,7 +26,9 @@ export default function useWindowBreakpointClasses() {
     };
     window.addEventListener("resize", callback);
     return () => {
-      window.removeEventListener("resize", callback);
+      if (window.removeEventListener) {
+        window.removeEventListener("resize", callback);
+      }
     };
   }, [className, window]);
   return className;
--- a/src/core/client/framework/hooks/useVisibilityState.ts
+++ b/src/core/client/framework/hooks/useVisibilityState.ts
@@ -35,7 +35,9 @@ function useVisibilityState() {
     window.document.addEventListener("visibilitychange", update);
 
     return () => {
-      window.document.removeEventListener("visibilitychange", update);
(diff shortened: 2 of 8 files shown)


kabeaty (commented):
Should we also add this check for where this is seen in `ClickOutside` and/or `KeyboardShortcuts`?

kabeaty (approved):

src/core/client/ui/encapsulation/useWindowBreakpointClasses.ts
```diff
@@ -26,7 +26,9 @@ export default function useWindowBreakpointClasses() {
     };
     window.addEventListener("resize", callback);
     return () => {
-      window.removeEventListener("resize", callback);
+      if (window.removeEventListener) {
```
nick-funk, line 29: If `window` or `removeEventListener` were not defined, that would set alarm bells in my head that something is weird for the context we're running in.

Is this due to Coral being run inside a web worker or service worker within an Electron or PWA app?

nick-funk (approved):

src/core/client/ui/encapsulation/useWindowBreakpointClasses.ts
rowanbeentje: Hmmm. This is definitely strange - agreed that this fix addresses the errors, but doesn't really explain why `window` is undefined and will likely be resulting in listener leaks. Weird one.

The PWA environment isn't executing Coral within a web worker or service worker (there is a service worker, but that's operating in prepopulated-cache-or-back-to-origin mode, and shouldn't touch these scripts). It definitely shouldn't mess with anything like `window` 😅

PR #4191, opened 2023-03-23
[CORL-2782] Separate tests into different sections to improve GHA performance

nick-funk:
## What does this PR do?

- Maintains existing `npm run test` functionality for devs
    - Devs don't have to do anything different to run tests
- Separates the GHA test runners into sections to improve multi-thread performance of each individual test section
- Prevents timeouts and other test failures when running tests using GitHub Actions

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

No new indices.

## How do I test this PR?

- run `npm run test`
- run `npm run test:server`
- run `npm run test:client:stream`
- run `npm run test:client:admin`
- run `npm run test:client:other`
- see that all of these pass

- also see that GHA ran these actions and passed

No

## How do we deploy this PR?

No special considerations


config/jest/client.other.config.js
```diff
@@ -0,0 +1,100 @@
...
+  const stats = fs.statSync(fullPath);
+  if (stats.isDirectory() && !excludedRoots.includes(fullPath)) {
+    roots.push(fullPath);
+  }
+}
+
+roots = roots.map((r) => `<rootDir>/${r}`);
+
+module.exports = {
+  displayName: "client",
```
kabeaty: Do we want to have `client:other` as the `displayName` here to match how the stream tests display name is configured?

config/jest/client.admin.config.js
```diff
@@ -0,0 +1,83 @@
...
+  "d3-shape",
+  "d3-time",
+  "d3-time-format",
+  "d3-timer",
+  "d3-transition",
+  "d3-zoom",
+];
+
+module.exports = {
+  displayName: "client",
```
kabeaty: Do we want to have `client:admin` as the `displayName` here to match how we're doing the stream?

config/jest/client.admin.config.js
```diff
@@ -0,0 +1,83 @@
+const path = require("path");
```
kabeaty: For all of these files that are sharing a lot of the same config, could we have something like a `baseConfig` and then just override what is specific about each of the admin/other/stream client configs? It would be nice to cut out some of the repetition and be able to see the differences better.

kabeaty (commented):
Very nice! I left a few suggestions in comments, but works well as is.

config/jest/client.admin.config.js
nick-funk: fixed

config/jest/client.admin.config.js
nick-funk: yup, good idea, refactored these to use a base config

config/jest/client.other.config.js
nick-funk: fixed

kabeaty (approved):
Thanks for the updates, looks great!

PR #4197, opened 2023-03-28
Fix edited comment subscriptions for Redis cache

nick-funk:
## What does this PR do?

Disables local redis comment caching for the subscription resolvers.

This ensures that the redis cache returns fresh data when it is fulfilling subscription requests like when a comment is edited.

Prior to this, it would carry an in-memory (similar to a data loader) version of the comments it had previously loaded (via a comment entered) and when that comment is edited, it would not go to Redis to retrieve the updated comment data when it was edited.

Now, like how caching is disabled for Mongo query request for subscriptions, it behaves the same within the Redis cache.

## These changes will impact:

- [X] commenters
- [X] moderators
- [ ] admins
- [ ] developers

None

No

No new indices.

## How do I test this PR?

- sign in as two separate users
- on one, post a comment
- see "x new comments" appeared for other user
    - click to load this comment in for the other user
- with the first user, edit the comment they just posted
- see that the comment updates for the other user
- test this for replies as well

No

## How do we deploy this PR?

No special considerations.

### DIFF
--- a/src/core/server/data/cache/commentCache.spec.ts
+++ b/src/core/server/data/cache/commentCache.spec.ts
@@ -46,6 +46,7 @@ const createFixtures = async (
     redis,
     null,
     logger,
+    false,
     options?.expirySeconds ? options.expirySeconds : 5 * 60
   );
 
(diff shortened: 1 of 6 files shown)


kabeaty (approved):
Looks good and works as expected!

tessalt (approved):

PR #4206, opened 2023-04-06
[CORL-2731] Add a message screen reader fix

marcushaddon:
## What does this PR do?
This PR fixes a bug in which a screen reader reads the word "Add" as "A D D" in the "Add message" buttons when configuring a stream.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

none.

No.

n/a

## How do I test this PR?
1. Enable a screen reader
2. As a moderator or admin, navigate to the "configure" tab of a stream, to the section where you can add a message at the top of the stream
3. Tab to the Add Message button, both with the editor open and closed (it will stay open after adding and then removing a message)
4. Observe that the screen reader correctly pronoucnes "ADD A MESSAGE"

No.

## How do we deploy this PR?
No special considerations needed.

### DIFF
--- a/src/core/client/stream/tabs/Configure/AddMessage/AddMessageClosed.tsx
+++ b/src/core/client/stream/tabs/Configure/AddMessage/AddMessageClosed.tsx
@@ -38,6 +38,7 @@ const AddMessageClosed: FunctionComponent<Props> = ({
           disabled={disableButton}
           upperCase
           data-testid="configure-addMessage"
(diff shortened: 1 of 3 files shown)


nick-funk (changes requested):
This looks good, update the snapshots (failing test reason) and we can get this approved and merged!

nick-funk (approved):

PR #4220, opened 2023-04-19
[CORL-2779] create a worker based word list processor

nick-funk:
## What does this PR do?

Replaces the existing vm based word list processor with a node worker based word list processor.

## These changes will impact:

- [ ] commenters
- [X] moderators
- [X] admins
- [X] developers

None

No

No new indices.

## How do I test this PR?

- start up Coral
- add some suspect and banned words
- post some comments with a non-staff user using those words
- check your rejected, reported, and unmoderated queues for the comments + highlighted words
- update the suspect/banned with new words
- use the new words in comments with a non-staff user
- check your rejected, reported, and unmoderated queues for the comments + highlighted words

None

## How do we deploy this PR?

This will cause pods to run multiple workers for word list processing, it might be advisable to consider slightly upping pod CPU resources so they can better multi-task the workers. This will effectively make the word list processing multi-threaded and allow Coral to better utilize multi-threading instead of linearly scaling on one processor as with most traditional node apps tend to be.


kabeaty (approved):
Looks good! Thanks for updating the worker script

PR #4233, opened 2023-04-27
[CORL-2820] make mark all as seen load newly marked comments

nick-funk:
## What does this PR do?

Returns the marked comments during mark seen mutations and then uses those id's to load the newly modified comments so that the seen flag can come through.

This appears to have been broken for some time now (prior to 8.0), I don't  know how mark all as seen worked prior to this as the code has never been returning and loading the marked comments all the way down the stack through the service calls. Maybe at one point we assumed all were marked and didn't use the reloaded comments client side?

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None

No

No new indices.

## How do I test this PR?

- enable `Z_KEY` and `COMMENT_SEEN` feature flags
- visit stream
- sign in as user
- ensure you have some yellow unseen comments
- press Shift + A to mark all as read
- see that stream changes all comments from yellow to white

No

## How do we deploy this PR?

No special considerations.

### DIFF
--- a/src/core/server/graph/mutators/Comments.ts
+++ b/src/core/server/graph/mutators/Comments.ts
@@ -291,7 +291,7 @@ export const Comments = (ctx: GraphContext) => ({
     markAllAsSeen,
   }: WithoutMutationID<GQLMarkCommentsAsSeenInput>) => {
     if (ctx.user) {
-      await markSeen(
+      const result = await markSeen(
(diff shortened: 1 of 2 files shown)


nick-funk:
Closing as @kabeaty has a better client side fix for this in another PR!

PR #4237, opened 2023-04-28
CPSRE-1124: Cache image builds on Coral GHA deploy workflow

existere:
## What does this PR do?
https://vmproduct.atlassian.net/browse/CPSRE-1124
Adds a GHA registry cache to help speed up the workflows with image builds.
There are no changes to the codebase. 

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

## How do I test this PR?
Testing will be done via merge and with the next deployment.

## How do we deploy this PR?
Merge and cut a versioned tag release; verify that the image build times were reduced.


iambrianchan (approved):
This is great. I verified that the docker build is using the image cache in the workflow: 
Screenshot 2023-05-09 at 9 05 27 AM

.github/workflows/build-and-test.yml
```diff
@@ -90,20 +101,24 @@ jobs:
...
-        uses: docker/build-push-action@v3
+        uses: docker/build-push-action@v4
         if: github.ref == 'refs/heads/develop'
         with:
           push: true
           tags: ${{ env.IMAGE_REPOSITORY }}:develop-latest
           build-args: |
               REVISION_HASH=${GITHUB_SHA}
+          cache-from: type=registry,ref=${{ env.IMAGE_CACHE_REPOSITORY }}:cache
+          cache-to: type=registry,ref=${{ env.IMAGE_CACHE_REPOSITORY }}:cache
```
nick-funk: I'm confused as to why these are cached? A build, tag, and push is frequently going to be a unique output. What is the purpose of caching this step?

When I am triggering an image build, I almost always want the build to be fresh and unique so as to ensure that its dependencies are installed for any package.json changes, the build is fresh, and the release is tagged appropriately.

Is this to re-use the build that was generated during the testing/continuous-integration phases prior to tagging and pushing out an image? If so, is there any possibility it might pull outdated files that mis-match the tagged release that is going out? Could another GHA build somewhere push newer files to the cache than what this is intended to be pulling from and cause it to pull the wrong files?

How does this "cache lock" deal with the race condition of multiple builds happening at the same time in GHA?

existere: We currently don't have a docker image build in the CI workflow, but i could add one strictly for caching purposes. The idea with the image cache in the build, tag, push step is that it uses docker image layers that remain unchanged instead of building the entire image every time.

existere: You bring up a good point, though. The image builds are only happening on `develop` and release tags, where caching is less important due to the fact that developers aren't waiting on those builds like they would be during active development while running the CI/testing workflow, and where waiting a little longer for a fresh build might actually be preferable. 

I suppose a better solution here would be to containerize the CI workflow and cache those builds to help speed up the development process.

nick-funk: I like your ideas around containerizing and caching the CI/testing workflow builds. I would be in favour of this as long as we could guarantee the dependencies (`npm i`) and resultant build would match the expected output for a develop/release branch.

As an alternative, it might be worth categorizing the cache-key based on branch (i.e. `develop`, `release-x.y.z`, etc) so that there aren't any chances for something to draw from a cached object other than its past self. This still has the same issues of needing to ensure dependencies and build are correct every time, but would at least avoid having to containerize the whole process right now.

existere: Great idea, I'll change the cache image tags to be branch-specific and discuss the containerization work with the rest of the premxp sre team.

tessalt (approved):

PR #4255, opened 2023-06-05
ES stream update

Lammland:
Please consider this two missing translations to visible elements on comment string.

- Add translation to "Refresh comments" button
- Add translation to footer navigation element


src/locales/es/stream.ftl
```diff
@@ -151,6 +151,14 @@ comments-rejectedTombstone =
 
 comments-featuredTag = Resaltado
 
+comments-refreshComments-closeButton =
+  .aria-label = Cerrar
+comments-refreshComments-refreshButton =
+  .aria-label = Actualizar comentarios
+comments-refreshQuestions-refreshButton =
+  .aria-label = Actualizar preguntas
+  comments-refreshReviews-refreshButton =
```
tessalt: hi there, can you fix the formatting here so that there's no space at the start of the line?

Lammland: Fixed! please check also the [en-US stream file line 320](https://github.com/coralproject/talk/blob/5c74a92cb1761603cc50cfcce1750788ec5fdcf4/src/locales/en-US/stream.ftl#L320) has the same mistake

PR #4275, opened 2023-06-26
[CORL-2384] upgrade node to v18.16+ and set compatibility flags

nick-funk:
## What does this PR do?

Update the target engine (node version) to `v18.16.0+` and set required compatibility flags so that the runtime doesn't complain about our older libraries living in a modern node environment.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

No new indices.

## How do I test this PR?

- install node 18 LTS
   - can do this via nvm as well via: `nvm install 18.16.0`
   - recommend setting an alias like: `nvm alias coral-new 18.16.0`
       - can then switch to that alias via `nvm use coral-new`
       - can check you're on the right version via `node --version`
- install npm 8.0.0 globally
   - `npm install -g npm@8.0.0`
   - if you do this when nvm is using the right version, will replace the `npm` version for that `nvm` mode for you, which is handy so you don't need to check with `npm` you have installed for `v18.16+`
- perform and `npm run build:development` build
- start up Coral via `npm run start:development`
- see that it runs and works correctly in stream + admin
- perform and `npm run build` build
- start up Coral via `npm run start`
- see that it runs and works correctly in stream + admin

No

## How do we deploy this PR?

Standard deploy procedure.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -43,10 +43,10 @@ jobs:
         name: Define SHORT_SHA with commit short sha
         run: echo "SHORT_SHA=`echo ${GITHUB_SHA} | cut -c1-8`" >> $GITHUB_ENV
       -
-        name: Setup Node14.x
+        name: Setup Node18.x
         uses: actions/setup-node@v3
         with:
-          node-version: '14.x'
+          node-version: '18.16.x'
       -
         name: Install npm 8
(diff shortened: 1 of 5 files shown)


nick-funk:
Closing as this is being implemented in another PR now that the client-server mono-repo is done!

PR #4283, opened 2023-06-30
Make `npm run watch` work on M1/2 macs

nick-funk:
## What does this PR do?

- allow `npm run watch` to work on M1/2 macs that also have the `arm64` build of `watchman` (https://facebook.github.io/watchman/) installed
- be sure to see testing notes for how to set it up and make it work

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

N/A

## How do I test this PR?

- have an M1/2 mac to test with
- install the `arm64` native version of `watchman` using `brew update && brew install watchman`
   - if you already had `watchman` installed, be sure to uninstall it via: `brew uninstall watchman` as it might be an x64 version running with Rosetta and you don't want that
- perform an `npm run build`
   - this is necessary because it generates manifests used by webpack to server the watch tooling
   - you only have to do this once, it should not be required to do this every time before you use `npm run watch`
- run `npm run watch`
- add `http://localhost:8080` to your permitted domain list
- test out the stream at `http://localhost:8080` and the admin at `http://localhost:8080/admin`

NOTE: nothing should change for intel (x64) macs, they just keep running as usual

No

## How do we deploy this PR?

No special considerations.

### DIFF
--- a/package.json
+++ b/package.json
@@ -266,6 +266,7 @@
     "@typescript-eslint/eslint-plugin": "^3.4.0",
     "@typescript-eslint/eslint-plugin-tslint": "^3.4.0",
     "@typescript-eslint/parser": "^3.4.0",
+    "address": "^1.2.2",
     "autoprefixer": "^9.8.6",
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^26.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -266,6 +266,7 @@
     "@typescript-eslint/eslint-plugin": "^3.4.0",
     "@typescript-eslint/eslint-plugin-tslint": "^3.4.0",
     "@typescript-eslint/parser": "^3.4.0",
+    "address": "^1.2.2",
```
nick-funk, line 269: for some weird reason, this core library is randomly missing if you install Webpack on an M1 mac. Oh well, threw it into the `devDependencies` and called it a day.

kabeaty (approved):

PR #4291, opened 2023-07-10
[CORL-2835] Reject all comments in queue

marcushaddon:
## What does this PR do?
This PR implements a feature in which selecting to reject all comments from a user while banning them will cause those comments to disappear from the moderation queue.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

The `User` type has been extended on the frontend to include a boolean flag for `allCommentsRejected`

No.

n/a

## How do I test this PR?
1. As a user with moderation privileges, navigate to the moderation queue.
2. Use the "b" key to open the ban modal for one of the comments' authors
3. Check the box for "reject all comments by this user"
4. Ban the user
5. Observe that any other comments by that user currently in the moderation queue should disappear.

No.

## How do we deploy this PR?


src/core/client/admin/test/moderate/regularQueue.spec.tsx
```diff
@@ -804,3 +812,69 @@ it("rejects comment in reported queue", async () => {
...
+    }),
+  });
+
+  const modCard = await screen.findByTestId(
+    `moderate-comment-card-${reportedComments[0].id}`
+  );
+
+  expect(modCard).toBeInTheDocument();
+
+  act(() => {
```
marcushaddon, line 868: I timeboxed my attempt to open the ban modal via keyboard shortcut, but if anyone has done this successfully, let me know!

src/core/client/admin/local/local.graphql
```diff
@@ -25,6 +25,10 @@ extend type CommentsConnection {
   viewNewEdges: [CommentEdge!]
 }
 
+extend type User {
+  allCommentsRejected: Boolean
```
kabeaty, line 32: Could you please add a comment explaining this field and what it's used for, for future reference?

src/core/client/admin/test/moderate/regularQueue.spec.tsx
```diff
@@ -18,6 +19,7 @@ import {
   replaceHistoryLocation,
 } from "coral-framework/testHelpers";
 
+import { commitLocalUpdate } from "relay-runtime";
```
kabeaty: Can you please move this import up into the correct section of imports?

buildlog
```diff
@@ -0,0 +1,606 @@
+
+> @coralproject/talk@8.2.3 build:development
```
kabeaty: Looks like this file accidentally got added?

kabeaty (changes requested):
This works well for the case where a user is banned across all sites via the ban modal. Should it also work when a user is banned across specific sites and has all comments rejected for the corresponding site(s)?

Also, I think it could be possible for the ban to send back an error but the comments to still be set to be filtered out in the mod queue. Would it be better to set the `allCommentsRejected` to true after the ban/rejection has successfully gone through?

src/core/client/admin/local/local.graphql
marcushaddon: ✅

src/core/client/admin/test/moderate/regularQueue.spec.tsx
marcushaddon: ✅

buildlog
marcushaddon: ✅

kabeaty (changes requested):
Thanks for the updates! It looks like it's still not working though when a user is banned across specific sites and has all comments rejected for the corresponding site(s). I think it's because `rejectExistingComments` isn't being passed through to the `updateUserBan` in `BanModal.tsx`. Looks like this was for some reason removed [here in this PR](https://github.com/coralproject/talk/pull/4127/files#diff-386f7d575ccfab29137cf084cb57b883484db9cc4b2fb9478addad6504299ffbR207). Can you take a look and add it back in?

kabeaty (approved):

PR #4296, opened 2023-07-13
[CORL-2838]: Spam ban via stream moderation

kabeaty:
## What does this PR do?

This PR makes it possible for site moderators and admin users to spam ban users that are obvious spam via the stream moderation actions.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

If a user is already banned, they can now be banned again by another user (it no longer throws an error and the query has been updated to find users with active bans too now). This is so that a user that is already banned can be spam banned (so, all of their comments can be rejected too).

no

none

## How do I test this PR?

You can test this PR by going to the comment stream and adding some comments as a commenter. Then log in as an admin. See that the moderation actions when you click the caret beside a comment now say "Spam ban" instead of "Ban user", and there is no longer any "Site ban" option. "BAN" button should be disabled until you type in "spam" (not case-sensitive). See a link to the single comment moderation view for more context. Click "BAN" and see that the ban is applied as expected and all comments are rejected (run using `start:development:withJobs` to see this take effect).

You can also test site moderators using the updated moderation actions. Site moderators should see the options of both "Site ban" and "Spam ban". "Site ban" still works to ban the user on the story's site. "Spam ban" does this PLUS rejects all comments by the user for that site.

After any spam ban, you should see a confirmation view. This will include a link to the `Community` tab in the admin to review the ban. You can close the confirmation view either by clicking the `Close` button or by clicking outside of the confirmation view modal.

A spam ban should always be able to be applied, regardless of whether the user is already banned (either site ban or all-site ban, whether it was done in the admin or via stream moderation actions). Anytime a user is re-banned, that new ban should appear in their ban history. However, if a user is re-banned, they should *not* receive another email notification about the ban.

A site moderator cannot be banned on their own site. If another site mod or an admin attempts to ban a site moderator on their own site, an error should be shown that says "Cannot ban accounts with moderator privileges".

Test on multi- and single-site (just has "Spam ban" option for single-site for mods and admins).

no


nick-funk (approved):
This looks great! Tested on dev2/locally + approved!

marcushaddon (approved):

PR #4297, opened 2023-07-18
[CORL-2872]: Add cors to single comment embed oembed endpoint

kabeaty:
## What does this PR do?

Adds cors to the oembed endpoint for single comment embeds. 

(Permissions are still determined via site permitted domains, this just ensures that `Access-Control-Allow-Origin` header for the permitted requesting origin is included if those checks are passed.)

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can uncomment the example of the Oembed API at the bottom of `commentEmbed.html` (you'll need to edit to add a comment id where noted). Inspect in the network tab of dev tools and see that the response still works as expected. 

Also copy over to a test site and embed there (update fetch url and comment ids as needed). See that the comment embed response now includes the `Access-Control-Allow-Origin` header set to the correct origin for the permitted request. Remove your test site from the permitted sites list. See that now you get a 401 Unauthorized error as expected.

### DIFF
--- a/src/core/server/app/middleware/commentEmbedWhitelisted.ts
+++ b/src/core/server/app/middleware/commentEmbedWhitelisted.ts
@@ -1,8 +1,11 @@
-import { AppOptions } from "coral-server/app";
+import { CorsOptionsDelegate } from "cors";
+
+import { MongoContext } from "coral-server/data/context";
 import { retrieveComment } from "coral-server/models/comment";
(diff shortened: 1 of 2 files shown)


nick-funk (approved):

nick-funk:
This looks good, approved! Thank you for limiting the `cors()` with an allow list handler, much appreciated!

PR #4300, opened 2023-07-24
[CORL-2875]: Improve consistency of Redis cache key for comment count

kabeaty:
## What does this PR do?

This makes the comment count jsonp Redis cache keys more consistently use just `storyUrl`. This also updates so that both the comment count html and the comment count text html are saved under the same key, and then the `notext` data attribute is used by the client at injection time to determine which html elements are needed.

These changes mean that only one key will be saved for each comment count story url.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can test this PR by using a page that has comment counts included (such as multi-site test index and story pages). Set your `JSONP_RESPONSE_CACHE` to `true` and the `JSONP_CACHE_MAX_AGE` to something like `3m`. Start it up and visit the page. Then go into Redis (`docker exec -it redis redis-cli  `) and look at the keys (`KEYS *`). See that the keys for the comment counts for the stories you loaded should look like `rmc:CoralCount&url=STORYURL`. The comment counts should still work as expected. Test comment counts both with and without the `data-coral-notext` attribute set to `"true"`) and confirm that with that, it shows the `Comments`/`New` text, and without it it shows just a comment count.

### DIFF
--- a/src/core/client/count/index.ts
+++ b/src/core/client/count/index.ts
@@ -10,12 +10,11 @@ import injectJSONPCallback from "./injectJSONPCallback";
 interface CountQueryArgs {
   id?: string;
   url?: string;
-  notext: boolean;
 }
 
 /** createCountQueryRef creates a unique reference from the query args */
 function createCountQueryRef(args: CountQueryArgs) {
-  return btoa(`${args.notext ? "true" : "false"};${args.id || args.url}`);
+  return btoa(`${args.url}`);
 }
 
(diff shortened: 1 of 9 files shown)


nick-funk (approved):
Looks good, thanks for doing additional testing as well! ✅

marcushaddon (approved):

PR #4308, opened 2023-08-01
[CORL-2875]: update Redis cache key name for comment embed jsonp

kabeaty:
## What does this PR do?

This fix ensures that when the comment embed jsonp is set in the Redis cache, each comment has its own consistent key specific to comment embeds. The comment count jsonp should still have a consistent Redis cache key specific to comment counts, as well.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can go to `/commentEmbed.html` and add a comment embed. Load the page. Go into Redis (`docker exec -it redis redis-cli`). Check its keys (`KEYS *`). See that there is a key that corresponds to the comment embed you loaded (something like `rmc:CoralCommentEmbed&commentID=COMMENT_ID`).

Also load comment counts somewhere (you can add `<span class="coral-count" data-coral-url="STORY_URL"></span>` to your embed index.html` or use multi-site test). Check the keys again and see that there's a key that corresponds to the count as well.

no

### DIFF
--- a/src/core/server/app/middleware/cache.ts
+++ b/src/core/server/app/middleware/cache.ts
@@ -37,13 +37,23 @@ async function get(
 }
 
 const cacheMiddleware =
-  (redis: Redis, ttl: number): RequestHandler =>
+  (
+    redis: Redis,
+    ttl: number,
+    jsonpType: "count" | "commentEmbed"
+  ): RequestHandler =>
   async (req, res, next) => {
     // Compute the cache key.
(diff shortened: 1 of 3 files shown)


marcushaddon:
I'm seeing the redis entry show up when loading the comment in the stream, but not when loading `commentEmbed.html` with a comment embedded

marcushaddon (approved):

PR #4309, opened 2023-08-02
[CORL-2877]: update to have comment count and embed jsonp use primarily Redis cache

kabeaty:
## What does this PR do?

When `JSONP_RESPONSE_CACHE` config is set to `true`, these updates make it so that the comment count and embed jsonp primarily use Redis cache. Whenever a comment is added, approved, or rejected, the comment count will be updated in the Redis cache. Whenever a comment is edited, the single comment embed (if present) will be cleared from the cache to ensure it will be set next time it's loaded with the correct updated content.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can test this PR by testing both comment counts and embeds being correctly added to the Redis cache, updated, and retrieved.

To test comment counts, you will probably want to use multi site test. Have two users signed in and adding comments to the same story. See that when you navigate to the index page, the comment counts have always been updated as expected. Check in the logs that they were retrieved from the cache, and see the correct updated comment counts in the network tab of dev tools for the relevant request. You can also look in the Redis cache (`docker exec -it redis redis-cli` and then `KEYS *`) for the relevant comment count key for the story and see it there. Also test that rejecting/approving comments updates the comment counts as expected as well.

To test single comment embeds, you can use `commentEmbed.html`. Copy an embed code from a comment and include it on the page. Navigate there to load the single comment embed. See that it's correctly saved in Redis cache after being loaded. Edit the comment. See that this clears the comment embed from the cache so that it can be correctly fetched with its edited content.

## How do we deploy this PR?

The `JSONP_CACHE_MAX_AGE` config should be increased now that Redis cache will be updated as comment counts update.


src/core/server/services/users/user.updateRole.spec.ts
```diff
@@ -198,8 +198,6 @@ describe("updateRole", () => {
         GQLUSER_ROLE.MODERATOR,
         true
       );
-      /* eslint-disable */
```
kabeaty, line 201: I just noticed this console.log was here, so removed it, but not otherwise related to these changes

marcushaddon (approved):

PR #4312, opened 2023-08-08
Update workers to use promise based callbacks instead of timeouts

wyattjoh:
Hey 👋🏻! I noticed that your worker implementation was using timeouts and looping to check for the worker results. This probably resulted in some spinning and wasted cycles in the event loop, so I updated it to use native promises instead which should be much more efficiant!


nick-funk (approved):
Thank you @wyattjoh for taking the time to contribute! I've done some extensive testing and this appears to have no memory leaks! Thank you for the refactor, much approved! ✅

PR #4321, opened 2023-08-22
[CORL-2859]: add defaults for flair badge config

kabeaty:
## What does this PR do?

Adds a default if flair badges aren't configured, since `badges` is non-nullable.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [x] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can test that when flair badges aren't configured on a tenant, the default of false for flairBadgesEnabled and empty badges array is there. There should be no error when loading the stream.

If flair badges are enabled but no badges configured, flairBadgesEnabled should be true and an empty badges array returned.

If flair badges are enabled and badges are configured, then both those values should be correctly returned.

no

### DIFF
--- a/src/core/server/graph/resolvers/Settings.ts
+++ b/src/core/server/graph/resolvers/Settings.ts
@@ -58,4 +58,15 @@ export const Settings: GQLSettingsTypeResolver<Tenant> = {
     args,
     ctx
   ) => embeddedComments,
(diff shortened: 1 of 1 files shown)


tessalt (approved):

nick-funk (approved):
This looks good! Thank you!

PR #4328, opened 2023-08-25
[CORL-2892]: allow additional origins for oembed api calls

kabeaty:
## What does this PR do?

These changes add in the ability to add additional tenant-wide domains that can make oEmbed calls for single comment embeds.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

This adds `oEmbedAllowedOrigins` to the `embeddedComments` config so that additional tenant-wide domains can be added that are permitted to call the oEmbed API.

no

n/a

## How do I test this PR?

You can test this by going into the admin and going to `Advanced` --> `Embedded comments`. Add/remove domains here and see that all functions/saves as expected. If an invalid URL is entered, you should see an error.

You can test that the domains are working by uncommenting and adding a comment id to the oEmbed example in `commentEmbed.html`. Remove the domain you are calling from from the site config, and see that the oEmbed throws an unauthorized error. Add the domain you are calling from to the new oEmbed api calls configuration and see that you no longer get an unauthorized error and that the domain is now permitted.


marcushaddon (approved):

src/core/server/app/middleware/commentEmbedWhitelisted.ts
```diff
@@ -37,6 +37,14 @@ export const commentEmbedWhitelisted =
...
           }
           if (origin) {
+            // if oEmbed API call, we also check oEmbed allowed origins on tenant
+            if (oembedAPI) {
+              if (
+                tenant.embeddedComments?.oEmbedAllowedOrigins.includes(origin)
+              ) {
+                return next();
+              }
+            }
```
nick-funk: since both of these nested `if`'s have no `else` how do you feel about doing the following instead?

```
// if oEmbed API call, we also check oEmbed allowed origins on tenant
if (oembedAPI && tenant.embeddedComments?.oEmbedAllowedOrigins.includes(origin)) {
    return next();
}
 ```
 
I only suggest this because I'm looking at the full file and seeing that we're at 6 levels of if-nesting happening. 

Not sure if we want to continue to add to it. Overall, it might be time for a refactor before this middleware gets out of hand.

Thoughts?

nick-funk (approved):
Approved! Good work, have one comment about the middleware you might find interesting. Not essential to merge this, but might help keep that code more read-able.

src/core/server/app/middleware/commentEmbedWhitelisted.ts
kabeaty: yep, updated--thank you for calling out!

PR #4334, opened 2023-08-31
[CORL-2891]: fix edit child reply to edited parent not updating

kabeaty:
## What does this PR do?

These changes make it so that we only check that the latest parent comment revision matches up with the parent revision id input when creating a comment reply. Previously, we were also checking whenever retrieving a parent comment (including when a comment is edited). 

This makes it so that child replies, to a parent comment that is edited after the reply has been created, can still be edited themselves.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

none

n/a

## How do I test this PR?

Create a comment. Create a reply to the comment. Edit the parent comment. Edit the reply comment. See that everything is edited/updated as expected and you get no errors.


src/core/server/stacks/createComment.ts
```diff
@@ -269,6 +274,17 @@ export default async function create(
...
+    // to.
+    if (input.parentRevisionID) {
+      const revision = getLatestRevision(parent);
+      if (revision.id !== input.parentRevisionID) {
+        throw new CommentRevisionNotFoundError(
+          parent.id,
+          input.parentRevisionID
+        );
+      }
+    }
```
nick-funk: It think the reason this wasn't being checked here is that it is possible for a comment to come in from a stale page (or conversation view) that has old revision ID's. If we do this, they will likely get errors when replying to a stale comment revision.

I recently had a person making a long reply to one of my comments on HN and I edited it while they were writing their reply, so I at leas know this can happen in the wild on other platforms.

In fast paced comment sections, I would imagine this happens a bit more frequently as well, where people are editing their comment simultaneously while others reply to them.

I'm not sure if we want to put this in here?

src/core/server/stacks/helpers/retrieveParent.ts
```diff
@@ -31,13 +30,6 @@ async function retrieveParent(
     throw new CommentNotFoundError(input.parentID);
   }
 
-  // Check to see that the most recent revision matches the one we just replied
-  // to.
-  const revision = getLatestRevision(parent);
-  if (revision.id !== input.parentRevisionID) {
-    throw new CommentRevisionNotFoundError(parent.id, input.parentRevisionID);
-  }
-
```
nick-funk: alright, now I'm tad bit confused, we were always checking against latest revision this whole time?

Does the check added to `createComment` matter then? Or was this only being called for `retrieveParent` which wasn't being used in `createComment`?

Hmmm... I'm going to have to test this to see what's going on here.

nick-funk (commented):
Going to need to test this, but have compiled my thoughts here.

src/core/server/stacks/helpers/retrieveParent.ts
kabeaty: Updating with what we discussed offline; removing this check altogether since we don't think it's necessary to check if replies are being made to stale comments or if replies to stale comment versions are being edited.

nick-funk (approved):
Glad we got that errant check out of the way, approved!

marcushaddon (approved):

PR #4341, opened 2023-09-06
update textHtml in redis cache for comment counts when count updates

kabeaty:
## What does this PR do?

These changes update the `textHtml` in the Redis cache for a story's comment count when the count updates. This ensures that the localized translation matches the count (for example, `1 Comment` and `2 Comments`, singular and plural).

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

## How do I test this PR?

You can use multi site test to test this. Create a new story with 0 comments. See that its count and text is added to the cache when you navigate to the index page for the story's site in multi site test. Go to the story page. Add a comment. Go back to the index page and see that it says `1 Comment` as expected, from the cache. Go back to the story page and add another comment. See that it now says `2 Comments` when you go back to the index page, from the cache.

Note that you'll need to make sure that you have `JSONP_RESPONSE_CACHE` and `JSONP_CACHE_MAX_AGE` set in your env.


nick-funk (approved):
Looks good!

marcushaddon (approved):
I think this is good for now but I think we should discuss adopting some sort of provider pattern to stop sprouting new arguments that get passed through 7+ layers of function calls without being accessed (which started long before this PR).

PR #4350, opened 2023-09-14
[CORL-2384] Upgrade to Node 18

nick-funk:
## What does this PR do?

Upgrades Coral to use Node 18 with npm 8.0.0.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

No new indices.

## How do I test this PR?

- install Node `v18.16.0` with npm `v8.0.0`
  - recommend using `nvm` for this
- run `sh scripts/clear-node_modules.sh`
  - this removes all existing node 14 module files that can cause conflicts
- run `sh npm-i.sh`
  - alternately, run `npm initialize.sh` if you've never run a build before or want to start fresh
- open a terminal in `client/` and run `npm run watch`
- open a terminal in `server/ and run `npm run watch`
- wait for the builds to finish
- navigate to http://localhost:8080 and test out the stream/admin
  - alternatively, do a server only build with `multi-site-test` if that's more to your taste
  
Extra testing:
- create an image using the `Dockerfile`
- spin up the image and see that it runs properly

No

## How do we deploy this PR?

No special considerations.


client/config/webpackDevServer.config.ts
```diff
@@ -69,7 +69,7 @@ export default function ({
     publicPath,
     // WebpackDevServer is noisy by default so we emit custom message instead
     // by listening to the compiler events with `compiler.plugin` calls above.
-    quiet: true,
+    quiet: false,
```
kabeaty: Just curious to know why we are setting `quiet` to `false` here now?

server/package.json
```diff
@@ -368,10 +368,10 @@
     "terser-webpack-plugin": "^4.2.3",
```
kabeaty, line 368: Can we remove this now? And also the `terser-webpack-plugin.d.ts` types file too?

kabeaty (approved):
Looking great and works as expected! Just had two questions/comments about the changes.

client/config/webpackDevServer.config.ts
nick-funk: just me accidentally leaving this on from debugging watch on node 18, I've set it back to `true`

server/package.json
nick-funk: removed!

kabeaty: yay thanks! looks like the types file was removed but not `terser-webpack-plugin` here too?

nick-funk: oh dang, I discarded at some point to try a cleaner approach to something or the other, must have lost removing this when I did that. I've fixed it up and pushed that change up now. Thanks for catching this!

tessalt (approved):

PR #4352, opened 2023-09-18
Update 01-01-installation.md

Shubhamsharma585:
Docs were outdated as in latest updated docker image has having different folder structure it becomes must to keep docs in sync

## These changes will impact:
- [ ] developers

### DIFF
--- a/docs/source/01-01-installation.md
+++ b/docs/source/01-01-installation.md
@@ -46,7 +46,7 @@ cat > docker-compose.yml <<EOF
 version: "2"
 services:
   talk:
-    image: coralproject/talk:6
+    image: coralproject/talk:8.5.0
     restart: always
     ports:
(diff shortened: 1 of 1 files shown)


nick-funk:
Thank you for taking the time to submit a PR!

Release/4 is specifically tagged to hold onto an older version of Coral prior to the v4 to v5 rewrite. The migration for v4 to specifically v5/6 is important. It is not advised to try and go straight from v4 to v8.5.0. As such, these docs are accurate for the steps we recommend to our external clients migrating from v4 to the newer Coral releases.

If you want to migrated from v4, the upgrade flow goes:
- v4 to v6
  - includes doing manual data migrations steps using an importer tool
- v6 validation and confirmation
  - inspect data to ensure the appropriate imports happened and the data is valid for v6
- v6 to v8+

PR #4366, opened 2023-10-06
Add DSAReport schema, mutation, model to DSA launch pad

kabeaty:
## What does this PR do?

It adds in the schema, initial createDSAReport mutation for creating DSAReports, creation in Mongo, etc.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

Adds `DSAReport` and related types.

no

none

## How do I test this PR?

You can test that everything works as expected still and using the mutation creates a `DSAReport` as expected.

no

### DIFF
--- a/server/src/core/server/models/dsaReport/report.ts
+++ b/server/src/core/server/models/dsaReport/report.ts
@@ -1,0 +1,100 @@
+import { v4 as uuid } from "uuid";
+
+import { Sub } from "coral-common/common/lib/types";
+import { MongoContext } from "coral-server/data/context";
+import { FilterQuery } from "coral-server/models/helpers";
+import { TenantResource } from "coral-server/models/tenant";
+
+import { GQLDSAReportStatus } from "coral-server/graph/schema/__generated__/types";
+
+export interface DSAReport extends TenantResource {
+  readonly id: string;
+
+  userID: string;
+
+  createdAt: Date;
+
+  lawBrokenDescription: string;
+
+  additionalInformation: string;
+
+  commentID: string;
+
+  submissionID?: string;
+
+  publicID: string;
+
+  status: GQLDSAReportStatus;
+}
+
+export type CreateDSAReportInput = Omit<
+  DSAReport,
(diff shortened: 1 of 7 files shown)


server/src/core/server/models/dsaReport/report.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  createdAt: Date;
+
+  lawBrokenDescription: string;
+
+  additionalInformation: string;
+
+  commentID: string;
+
+  submissionID?: string;
```
marcushaddon, line 23: Why is this optional if we are always supplying one?

kabeaty: the idea is that you can optionally supply the `submissionID` here in the cases when there are multiple comments reported together and you want comments to all be considered one submission. If it's not supplied, such as when it's just a single comment being reported, an id is generated down on line 54 of this file that is used, so every `DSAReport` will have a `submissionID`.

tessalt (approved):

server/src/core/server/models/dsaReport/report.ts
marcushaddon: That makes sense as to why it would be option in the *input* since it doesn't always have to be supplied, but the DSAReport itself will always end up with this field, and so it still seems odd for it to be optional. Not a big deal though

marcushaddon (approved):

server/src/core/server/models/dsaReport/report.ts
nick-funk: I agree with @marcushaddon here, if every report will have a submissionID, and since this is the domain/data model (not a GraphQL/API input type) I feel it should be required.

If folks want, I can fix this up and also implement the submissionID generation while I'm at it to shore this all up.

kabeaty: Thank you for reminding me...meant to follow up here, but I plan to add this as not optional in my other branch, I just merged this in before updating here. And the submissionID generation is already included!

PR #4376, opened 2023-10-19
DSA Notifications

nick-funk:
## What does this PR do?

Preliminarily adds notifications to the DSA epic branch.

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [X] developers

- Adds a new `Notification` type and corresponding Connection types

No.

```
db.notifications.createIndex({ tenantID: 1, id: 1 }, { unique: true });
```

This index creates the uniqueness constraint for the `tenantID` and `id` fields on the notifications collection

```
db.notifications.createIndex({ tenantID: 1, ownerID: 1, createdAt: 1 });
```

This index speeds up the retrieval of notifications by `tenantID`, `ownerID`, and `createdAt` which is the most common way of retrieving notifications for pagination in the notifications tab on the stream.

## How do I test this PR?

- Enable the DSA features under `Admin > Configure > General`
- Create some comments
- Feature, approve, or reject the comments
- Check `Notifications` for updates on these actions

No

## How do we deploy this PR?

- Merge into the dsa-launchpad branch
- Eventually merge the dsa-launchpad branch into develop
- Release
- Ensure appropriate indices (described above) are applied to any mongo instances using the notification features


server/src/core/server/graph/mutators/Comments.ts
```diff
@@ -254,6 +257,12 @@ export const Comments = (ctx: GraphContext) => ({
     // Publish that the comment was featured.
     await publishCommentFeatured(ctx.broker, comment);
 
+    await ctx.notifications.create(ctx.tenant.id, ctx.tenant.locale, {
+      targetUserID: comment.authorID!,
+      comment,
+      type: NotificationType.COMMENT_FEATURED,
+    });
+
```
nick-funk, line 285: @marcushaddon  and @kabeaty, this is what it looks like to publish a notification for a user. All you need is access to the `ctx.notifications` which is available for every query and mutation in Coral's API.

The content of the notification is also malleable and has many optional fields. It can also be extended to provide new interface types for various kinds of notifications. It's all very flexible to whatever needs to be inside the notification.

As well, there is a convenient `NotificationType` to categorize the notifications and is infinite in how many kinds it can be, but adds a type that makes it easy to create/sort/filter notifications in the future.

This is what I mean by a `one-liner` to publish a notification anywhere. This isn't relevant for either of your DSA work, but it shows how minimal it will be to send new notifications throughout both of your's code.

client/src/core/client/stream/App/TabBar.tsx
```diff
@@ -53,6 +60,10 @@ const AppTabBar: FunctionComponent<Props> = (props) => {
   );
   const myProfileText = getMessage("general-tabBar-myProfileTab", "My Profile");
   const configureText = getMessage("general-tabBar-configure", "Configure");
+  const notificationsText = getMessage(
+    "general-tabBar-notifications",
```
kabeaty: fyi, looks like this translation still needs to be added

kabeaty (approved):
Very nice initial framework for notifications to keep building on and extending!

client/src/core/client/stream/App/TabBar.css
```diff
@@ -5,6 +5,13 @@
   padding-right: var(--spacing-3);
 }
 
+.condensedTab {
+  padding-top: var(--spacing-2);
```
kabeaty: Can these styles be updated to just `padding: var(--spacing-2);`?

client/src/core/client/stream/tabs/Notifications/NotificationContainer.css
```diff
@@ -0,0 +1,86 @@
+.root {
+  font-family: var(--font-family-primary);
+  font-weight: var(--font-weight-primary-regular);
+  font-size: var(--font-size-2);
+
+  padding-left: var(--spacing-2);
```
kabeaty: Looks like this style is covered below with `padding`, so I think you can remove

client/src/core/client/ui/components/icons/ActiveNotificationBellIcon.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { FunctionComponent } from "react";
+
+const ActiveNotificationBellIcon: FunctionComponent = () => {
+  return (
+    <svg viewBox="-1.25 0 21 21" fill="none" xmlns="http://www.w3.org/2000/svg">
+      <path
+        d="M13.2291 4.85596C13.2291 5.519 13.4925 6.15488 13.9614 6.62372C14.4302 7.09257 15.0661 7.35596 15.7291 7.35596C16.3922 7.35596 17.0281 7.09257 17.4969 6.62372C17.9657 6.15488 18.2291 5.519 18.2291 4.85596C18.2291 4.19292 17.9657 3.55703 17.4969 3.08819C17.0281 2.61935 16.3922 2.35596 15.7291 2.35596C15.0661 2.35596 14.4302 2.61935 13.9614 3.08819C13.4925 3.55703 13.2291 4.19292 13.2291 4.85596Z"
+        fill="#EB5757"
```
kabeaty, line 8: Can you please remove? I don't think we want to set a fill color on the icon, but rather add it with css so that it's more reusable.

server/src/core/server/graph/schema/schema.graphql
```diff
@@ -3153,6 +3153,27 @@ type User {
...
   bio: String
+
+  """
+  hasNewNotifications returns true if the user has received new notifications
+  since they last viewed their notifications tab.
+  """
+  hasNewNotifications: Boolean!
+    @auth(
+      userIDField: "id"
+      roles: [ADMIN, MODERATOR]
```
kabeaty, line 3181: Just curious, why is this (and `lastSeenNotificationDate`) restricted to roles admin and moderator?

server/src/core/server/graph/schema/schema.graphql
```diff
@@ -4459,6 +4480,120 @@ type Queues {
...
+  """
+  id is the primary identifier of the DSA report.
+  """
+  id: ID!
+
+  """
+  publicID is the public identifier that is human readable
+  for the DSA report.
+  """
+  publicID: String!
```
kabeaty: fyi, I updated this to use `referenceID` instead of `publicID` here

server/src/core/server/models/notifications/notification.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (max.getTime() < date.getTime()) {
+      max = date;
+    }
+  }
+
+  const change: LastSeenNotificationChange = {
+    $set: { lastSeenNotificationDate: user.lastSeenNotificationDate },
+  };
+
+  if (
```
kabeaty: This works as is just fine, just a small (subjective) suggestion for this check for whether the `lastSeenNotificationDate` should be set to `max` and the one below as well. It may make it easier to read at a glance what's going on here if it was something more like:
```
const newNotificationSinceLastSeen = user.lastSeenNotificationDate && user.lastSeenNotificationDate.getTime() < max.getTime();
const notificationsNeverViewed = user.lastSeenNotificationDate === null || user.lastSeenNotificationDate === undefined;
if (newNotificationSinceLastSeen || notificationsNeverViewed) {
change.$set.lastSeenNotificationDate = max;
}
```

kabeaty (commented):
Echoing my previous review, fantastic framework for notifications set up here! Just left a few small comments/questions throughout.

client/src/core/client/stream/App/TabBar.tsx
nick-funk: yup, fixed!

server/src/core/server/graph/schema/schema.graphql
nick-funk: I believe this is saying that `hasNewNotifications` is restricted to the viewer's own notifications (by userID match) and can also be accessed by roles `ADMIN` and `MODERATOR`.

I have a commenter user who is not an admin or a moderator and because of the `userIDField` they are able to query their own notifications.

server/src/core/server/graph/schema/schema.graphql
nick-funk: Yup, I noticed this, updated and fixed!

server/src/core/server/models/notifications/notification.ts
nick-funk: I'm going to have to think about this, my brain is finding the above rather confusing, I do like your naming of the conditions being checked, that's cool. I'll have a think and do what I can to make this code more read-able!

client/src/core/client/ui/components/icons/ActiveNotificationBellIcon.tsx
nick-funk: this fill colour is specifically for the red dot on the bell, I'll convert it to a configurable value via CSS. Thanks for catching this!

server/src/core/server/graph/schema/schema.graphql
kabeaty: Ah okay, got it...thank you!

server/src/core/server/models/notifications/notification.ts
kabeaty: Fair enough, sounds good, ty...mostly that both the `if` and the `else if` led to the same outcome seemed like something unexpected to me when I was following along with the code

client/src/core/client/stream/App/TabBar.css
nick-funk: it can and I've updated as much, I honestly pulled it from the `smallTab` that does the same and already existed. Just following the prior styles. I personally don't mind explicit styles for each direction (did it a lot with XAML and WPF apps)

client/src/core/client/ui/components/icons/ActiveNotificationBellIcon.tsx
nick-funk: I've attempted to override and set this via CSS, not only does postcss make this extremely difficult, there is no way to cleanly override this on the icon itself due to the way `SvgIcon` instantiates its icon child. I think we need to discuss how we are going to allow sub-selected css overrides in our icon library, because as it is currently architected, I think it might not be possible.

client/src/core/client/stream/tabs/Notifications/NotificationContainer.css
nick-funk: yeah, fixed

server/src/core/server/models/notifications/notification.ts
nick-funk: refactored, I think it will make more sense now

PR #4377, opened 2023-10-20
[CORL-2948] emit moderation actions

marcushaddon:
## What does this PR do?
This PR updates our comment moderation pipeline to distinguish between `commentActions` and `moderationActions` during the pipeline. It causes certain phases to conditionally emit a `moderationAction` field that sets the status and terminates the pipeline.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none.

No.

n/a

## How do I test this PR?
1. On both dev, and then on this branch Create the following comments
	- [x] non rejected comments
	- [x] comment reported
	- [x] comment reported, then rejected
	- [x] comment reported, rejected, then approved
	- [x] comment reported then approved
	- [x] comment flagged for suspect word, approved
	- [x] comment flagged for suspect word, rejected
	- [x] comment flagged for suspect word
	- [x] comment flagged for banned word, rejected
	- [x] comment flagged for banned word, rejected, then approved
	- [x] comment rejected in stream by mod
	- [x]  comment rejected via reject all comments
	- [x]  comment rejected via reject all comments on this site
	- [x]  comment rejected via spam ban
	- [x] comment approved via external moderation
	- [x] comment flagged via external moderation
	- [x] comment rejected via external moderation
2. Confirm that all queues contain the same corresponding comments after creating these on each branch.

No.

## How do we deploy this PR?


server/src/core/server/services/comments/pipeline/helpers.ts
marcushaddon, line 1: Aggregating `commentActions` instead of `actions`. `moderationAction` is a single object rather than an array, and isn't aggregated because Moderation Actions set a status on the comment, and so short circuit the pipeline.

server/src/core/server/services/comments/pipeline/phases/detectLinks.ts
marcushaddon, line 1: Renaming field

server/src/core/server/services/comments/pipeline/phases/external.ts
```diff
@@ -102,9 +102,10 @@ export interface ExternalModerationRequest {
   tenantDomain: string;
 }
 
+// BOOKMARK: (marcushaddon) do we need a new type to maintain type safety with joi?
 export type ExternalModerationResponse = Partial<
-  Pick<PhaseResult, "actions" | "status" | "tags">
->;
+  Pick<PhaseResult, "status" | "tags">
+> & { actions: PhaseResult["commentActions"] };
```
marcushaddon, line 107: This is because our public API for external moderation phases still accepts the `actions` field, but their type is now internally called `commentActions`

server/src/core/server/services/comments/pipeline/phases/recentCommentHistory.ts
marcushaddon, line 1: Renaming fields

nick-funk (approved):
This make sense and thanks for adding in the new mod actions distinction! Solid refactor!

kabeaty (commented):
The refactor looks good overall!

I did notice that when banned words are rejected, their moderation cards don't include the `Banned word` marker next to `Details` at the bottom of the card. The `MarkersContainer` component is checking in actionCounts flag reasons for whether to show that. Is there a way to update so this is still included in the moderation card in this case?

kabeaty (approved):

PR #4379, opened 2023-10-24
[CORL-2934]: DSA - Moderating illegal content

kabeaty:
## What does this PR do?

These changes add the ability to moderate illegal content. They include the addition of a `Reports` tab to the admin that shows a table of all DSAReports made. When one of these reports is clicked on, a single report page is shown with that report's info, history, and the ability to add notes, download its info, change its status, and make an illegal or not illegal decision.

Still TODO:
- [x] add decision screen for moderating reports
- [x] update copy where needed - (will continue this in future PRs)
- [x] address TODOs in code
- [x] add tests - (will continue this in future PRs)
- [x] add link back to reports table from single report screen
- [x] update CSV file that's downloaded with design input

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

These changes add the ability to query for DSAReports and add to their history (add notes, change status, delete notes, download the report data, make a decision).

no

## How do I test this PR?

You can ensure that DSA is enabled. Have some DSAReports created. Go to the `Reports` tab in the admin. See the table of DSAReports. See that it can be sorted by newest/oldest. See that you can filter by closed reports. Click into a specific report. See its data. See its history, including when it was submitted. Add a note to its history. Delete a note from its history. Click `Share` button to download a csv file with the DSAReport's data. Change its status. Make a decision on a DSAReport. View its reported comment in the stream/moderation. See the user history drawer open if you click on the reporter username or the commenter's username. 

When you make a decision on a report, if it's a decision that it is illegal, you should see that the comment is rejected.

If you add a note to or download a CSV for a report that is status `Awaiting review` still, you will be prompted if you would like to move the report to `In review` with a modal. If you select `Completed` status for a report, you will be prompted by a modal to make a decision on that report.

To test how this works with GDPR deletion, you can have a user 1 comment, then user 2 report the comment for illegal content, and then user 1 request their account be deleted. Update their scheduled deletion date to something sooner. Run with jobs to have the deletion go through. See that the pending DSA report (must be in `In review` or `Awaiting review` status still) is set to `Void` status, and that you get a note about the reported comment being deleted when viewing the DSA report as well.


client/scripts/precommitLint.js
```diff
@@ -26,12 +26,12 @@ sgf((err, results) => {
...
   const eslintFiles = [];
   for (const item of results) {
-    const { filename } = item;
+    const { filename, status } = item;
 
     // only include valid, filtered extensions
     // this is primarily to keep eslint rampaging
     // over non-source files
-    if (!matchesExtension(extensions, filename)) {
+    if (!matchesExtension(extensions, filename) || status === "Deleted") {
```
nick-funk, line 34: Thanks for adding this, I vaguely remember adding something similar somewhere else, but maybe I missed this spot?

client/src/core/client/admin/App/Navigation/Navigation.tsx
```diff
@@ -31,6 +32,12 @@ const Navigation: FunctionComponent<Props> = (props) => (
         <NavigationLink to="/admin/dashboard">Dashboard</NavigationLink>
       </Localized>
     )}
+    {/* TODO: Any other permissions needed? */}
```
nick-funk, line 35: Ultimately, I think this is a question for design, but I would imagine this will be tied to an org mod level role? I don't know if, say, a site mod will be allowed to process DSA reports.

client/src/core/client/admin/App/Navigation/NavigationContainer.tsx
```diff
@@ -17,21 +18,22 @@ interface Props {
...
+      fragment NavigationContainerLocal on Local {
+        dsaFeaturesEnabled
+      }
+    `
+  );
+  return (
+    <Navigation
+      showDashboard={!!viewer && can(viewer, Ability.VIEW_STATISTICS)}
+      showConfigure={!!viewer && can(viewer, Ability.CHANGE_CONFIGURATION)}
+      showReports={!!dsaFeaturesEnabled}
```
nick-funk, line 33: following previous comment, I wonder if this needs to be given a similar limitation as the above `Ability` checks in `can` have for things like dashboard and config... probably a question for design. For now, the DSA flag is good enough to get this merged in!

client/src/core/client/admin/routes/Reports/ReportRowContainer.tsx
```diff
@@ -0,0 +1,99 @@
...
+  const statusIconMapping = useCallback((status: DSAReportStatus | null) => {
+    if (!status) {
+      return null;
+    }
+    switch (status) {
+      case "AWAITING_REVIEW":
+        return <SvgIcon Icon={SignBadgeCircleIcon} />;
+      case "UNDER_REVIEW":
+        return <SvgIcon Icon={NightMoonHalfIcon} />;
+      case "COMPLETED":
```
nick-funk: My only nitpick here would be that string constants are sometimes nicer when stuffed into a string type enum:

i.e.
```
export enum STATUS {
  AwaitingReview = "AWAITING_REVIEW",
  UnderReview = "UNDER_REVIEW",
  ...
}
```

client/src/core/client/admin/routes/Reports/ReportStatusMenu.css
```diff
@@ -0,0 +1,7 @@
+.statusLabel {
+  color: var(--palette-text-000);
```
nick-funk, line 2: lol, forgot we had a `...-text-000`, always seemed like a funny naming convention... wouldn't that make it transparent? 🤣

client/src/core/client/admin/routes/Reports/ReportedComment.tsx
```diff
@@ -0,0 +1,198 @@
...
+                    )}
+                  </div>
+                  {comment.rating && (
+                    <div>
+                      <StarRating rating={comment.rating} />
+                    </div>
+                  )}
+                  <div>
+                    <div>
+                      {/* TODO: Do we want to show message that comment has been rejected? */}
```
nick-funk: Just noting that we want to keep track of the various TODO's I'm seeing while reading. We should still merge this into the epic branch though.

client/src/core/client/framework/lib/rest.ts
```diff
@@ -45,6 +44,10 @@ const handleResp = async (res: Response) => {
     return res.text();
   }
 
+  if (type && type.includes("application/octet-stream")) {
+    return res.arrayBuffer();
+  }
+
```
nick-funk, line 50: what was the reason for this change? I'm sure it's necessary, but I'm at a loss as to what it's doing because this whole prior written handler is somewhat obscure.

server/src/core/server/app/router/api/dsaReport.ts
```diff
@@ -0,0 +1,15 @@
+import { AppOptions } from "coral-server/app";
+import { reportDownloadHandler } from "coral-server/app/handlers";
+import { userLimiterMiddleware } from "coral-server/app/middleware/userLimiter";
+
+import { createAPIRouter } from "./helpers";
+
+export function createDSAReportRouter(app: AppOptions) {
+  const router = createAPIRouter({ cacheDuration: "30s" });
+
+  router.use(userLimiterMiddleware(app));
```
nick-funk, line 10: I would recommend also putting an ip limiter on this

server/src/core/server/services/dsaReports/reports.ts
```diff
@@ -20,3 +46,178 @@ export async function createDSAReport(
...
+      cache,
+      config,
+      broker,
+      tenant,
+      commentID,
+      commentRevisionID,
+      userID,
+      now
+    );
+    // APPROVE IF NOT ILLEGAL
```
nick-funk: I'm unsure about approving a comment that is determined not illegal. Even if it's not illegal, it could still be toxic? Are we sure we don't want to just leave it with `NONE` and remain visible? Approving is like we're rubber stamping it from our mod team as good.

nick-funk (approved):
✅ This looks great! Left some comments for consideration, but nothing that would hold up a merge.

client/scripts/precommitLint.js
kabeaty: Yeah, I think it was in the Node 18 branch, so I pulled it in separately here when I ran into the issue since that's not in `develop` now. We could open up a small PR with only these updates into `develop` in the meantime, or else just get them in again with DSA.

client/src/core/client/admin/App/Navigation/Navigation.tsx
kabeaty: Yep, also what I think, just pending confirmation from design so I thought I'd leave the note in here so it's not forgotten.

client/src/core/client/admin/App/Navigation/NavigationContainer.tsx
kabeaty: Yeah, once we know how we want to limit, exactly!

client/src/core/client/admin/routes/Reports/ReportRowContainer.tsx
kabeaty: Updating to use `GQLDSAReportStatus` from our schema here, thanks for pointing out!

client/src/core/client/admin/routes/Reports/ReportedComment.tsx
kabeaty: Yep! The design input on this is probably not, so I can actually probably delete this one now.

client/src/core/client/framework/lib/rest.ts
kabeaty: This is to support the CSV download for DSA reports from the API. Otherwise, it won't work since it tries to respond with json and errors.

server/src/core/server/app/router/api/dsaReport.ts
kabeaty: Good point, updating to do that, thank you!

server/src/core/server/services/dsaReports/reports.ts
kabeaty: Good point, removing as agree that we don't want to do that! Thanks for catching.

PR #4387, opened 2023-11-07
[CORL-2950]: Update comment share and report interaction events

kabeaty:
## What does this PR do?

These changes fix the events that are emitted when a comment 1) has its share button clicked and 2) has its report button clicked.

Previously, the share event was firing multiple times upon the share button being clicked and share popover shown. Now, it only fires once.

Previously, the report event was not firing at all. Now, it fires once when the report button is clicked and report form is shown.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can test this PR by opening your dev tools and then going to a comment in the stream and clicking `Share`. You should see that one event is emitted, `showSharePopover` with the comment id. Click `Report`. You should see that one event is emitted, `showReportPopover` with the comment id.


nick-funk (approved):
Makes, nice clean fix! ✅

marcushaddon (approved):

PR #4388, opened 2023-11-07
[CORL-2920]: Community table fallbacks alignment updates

kabeaty:
## What does this PR do?

These changes update the styles for the community table in the admin so that all user data is displayed with the desired alignment.

If a username isn't available, the fallback message in this case should still be left aligned. All other columns should have their data vertically aligned in the middle, including the email data and its fallback message, member since, role, and status data.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

Go to the `Community` tab in the admin. If a username isn't available, the fallback message in this case should be left aligned. All other columns should have their data vertically aligned in the middle, including the email data and its fallback message, member since, role, and status data. (You can either update data so no username/email is available, or just comment out parts of the code that show the username/email so that the fallback shows)

### DIFF
--- a/client/src/core/client/admin/components/UserRole/UserRoleChange.css
+++ b/client/src/core/client/admin/components/UserRole/UserRoleChange.css
@@ -2,6 +2,7 @@
   width: 100%;
   justify-content: space-between;
   padding: 0;
+  line-height: 1.14;
 }
 
 .arrowsIcon {
--- a/client/src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
+++ b/client/src/core/client/admin/components/UserRole/UserRoleChangeContainer.tsx
@@ -124,8 +124,6 @@ const UserRoleChangeContainer: FunctionComponent<Props> = ({
 
   const handleOnChangeMembershipScopes = useCallback(
     async (siteIDs: string[]) => {
-      /* eslint-disable */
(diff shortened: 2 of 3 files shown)


nick-funk (approved):
Looks good! ✅

marcushaddon (commented):
I'm seeing the fallback message for username left aligned, but I also see the other columns being left aligned, rather than in the center, which is what I understood from the PR description as the requirement. Should they be in the middle?

kabeaty:
> I'm seeing the fallback message for username left aligned, but I also see the other columns being left aligned, rather than in the center, which is what I understood from the PR description as the requirement. Should they be in the middle?

The other columns should still be left aligned. The requirement is that all columns then also have their data **vertically** centered, in addition to their horizontal left alignment.

marcushaddon:
> > I'm seeing the fallback message for username left aligned, but I also see the other columns being left aligned, rather than in the center, which is what I understood from the PR description as the requirement. Should they be in the middle?
> 
> The other columns should still be left aligned. The requirement is that all columns then also have their data **vertically** centered, in addition to their horizontal left alignment.

Ah that makes sense.

marcushaddon (approved):

Issue #4375, opened 2023-10-19
Comments are not loading when scrolling 


Kristina-Dvorski:
**Expected behavior:** Comments should be loaded when I am scrolling through. 

**Actual behavior:** They are not loaded, instead I have white space. 

**Related Issues:**  Yes, but the problem is not solved. The bug is not in CSS on my side, because with normal Coral CSS the bug is also there. Please, read the following message (https://github.com/coralproject/talk/issues/4371#issuecomment-1761209897). I would appreciate that because, all users of my web-page have that problem and it is very unconformable. 
Thank you for your time.

**Versions:**

- NodeJS:  14
- NPM:  8.0.0
- MongoDB: 4.2
- Redis: 3.2
- Browser: Edge 118.0.2088.46 
- OS: Windows 11

losowsky:
Thanks Kristina for reporting this, and we're sorry this is happening. We run many variations of implementation across several CMSs and sites and aren't seeing this happen – we'll look into it more deeply soon and let you know.

losowsky:
@Kristina-Dvorski  Can you also confirm you're running the latest version of Coral?

Kristina-Dvorski:
Thank you, for taking your time to check my issue. Yes I am running the latest version of Coral (8.5.3). The Bug is happening when the container with comments is on `position: fixed`, when I change to `position: relative` the container is in the Page but the bug is not there.
I hope this helps

Kristina-Dvorski:
Here's a showcase of the issue with a clean installation of Coral and no custom CSS other than what is included in the `<style>` tag of this page: http://159.89.1.139/
Thank you for your time

losowsky:
Thanks Kristina. We have it in our queue to look into.

losowsky:
Hi Kristin, 

We've looked into it and I'm afraid we can't support putting Coral inside a fixed div due to limitations with Virtuoso. It is possible through some slightly hacky means (The Verge did something like that to make it work) but it's not something that we support out of the box. 

Sorry I don't have a better answer!

Kristina-Dvorski:
Hi @losowsky,

thank you for your time and effort. 

Have a nice day

oliver-dvorski:
For future users dealing with this bug - here's one way to patch the issue. Apparently the Virtuoso list will re-render when the browser window fires a `resize` event. Therefore, this makes the problem go away:

```js
document.querySelector('.coral-scroller').addEventListener(
    'scroll',
[...]

PR #4389, opened 2023-11-08
Update troubleshooting.md to showcase a workaround for issue #4375

oliver-dvorski:
## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

None

No

No


nick-funk:
Thanks for taking the time to write this up. 

As this isn't something we currently yet support, but would like to, we don't want to post a work around that we don't have time to test the full ramifications of how it might be used across our open source and SaaS services deployments.

While this work around is possibly viable for everyone, we don't have the time to vet it properly, and once something is written up and documented, whether it is labelled as a temporary work around or not, some clients or open source users may expect it to be fully supported.

Hopefully, we'll have time at some point to fix this issue so it doesn't require a work-around. We know that The Verge has Coral inside of a `position: fixed` div comment drawer without needing this work-around, so there may be another solution to this problem that we can include in our documentation.

PR #4403, opened 2023-11-17
DSA - Archived fixes

kabeaty:
## What does this PR do?

This makes it so that a DSAReport can be created for an archived comment. An archived comment's DSA Report can also have a decision made on it.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

`createAction` and `updateCommentActionCounts` now check for `isArchived` to determine which collection to use

no

n/a

## How do I test this PR?

Archive a story. Report an archived comment for illegal content. See that it submits without error. Look at the report in the Reports table in the admin. Click through to its single report page. Make an illegal decision on the report. See that the comment is rejected and the decision is successfully submitted. (Also can test making a legal decision on the report and seeing it works fine.)

### DIFF
--- a/server/src/core/server/graph/mutators/Comments.ts
+++ b/server/src/core/server/graph/mutators/Comments.ts
@@ -146,7 +146,7 @@ export const Comments = (ctx: GraphContext) => ({
         commentRevisionID,
       }
     ),
-  createIllegalContent: ({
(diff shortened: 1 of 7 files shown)


nick-funk (approved):
Nice work! ✅

PR #4405, opened 2023-11-17
[CORL-296]7 view conversation rejection reasons

marcushaddon:
## What does this PR do?
This PR fixes a bug in which the moderation reasons modal doesn't close after seelcting a reason when in the user drawer. It also updates the user deletion flow to include comment rejection reasons when rejecting the deleted users comments.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

None.

No.

n/a

## How do I test this PR?
Navigate to a user drawer for any usera as an user with mod privileges, with DSA enabled. Reject a comment from the drawer. Observe that a reason is requried, and when one is supplied, the Moderation Reason modal closes.

Find a user with comments with no replies. As that user, request that your account be deleted. In the database, set the scheduled deletion time to something soon. Observe that the account is successfully deteleted, and the comments with no replies are rejected, and the resulting CommentModerationActions specify a reason.

No

### DIFF
--- a/server/src/core/server/services/users/delete.ts
+++ b/server/src/core/server/services/users/delete.ts
@@ -12,10 +12,10 @@ import {
   GQLCOMMENT_STATUS,
   GQLDSAReportStatus,
   GQLREJECTION_REASON_CODE,
+  GQLRejectionReason,
 } from "coral-server/graph/schema/__generated__/types";
 
 import { moderate } from "../comments/moderation";
-import { Moderate } from "../comments/moderation/moderate";
 import { I18n } from "../i18n";
 import { AugmentedRedis } from "../redis";
 
@@ -134,7 +134,7 @@ async function moderateComments(
   targetStatus: GQLCOMMENT_STATUS,
   now: Date,
   isArchived = false,
-  rejectionReason?: Moderate["rejectionReason"]
+  rejectionReason?: GQLRejectionReason
 ) {
   const tenant = await retrieveTenant(mongo, tenantID);
   if (!tenant) {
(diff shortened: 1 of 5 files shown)


server/src/core/server/services/users/delete.ts
```diff
@@ -131,7 +133,8 @@ async function moderateComments(
   filter: FilterQuery<Comment>,
   targetStatus: GQLCOMMENT_STATUS,
   now: Date,
-  isArchived = false
+  isArchived = false,
+  rejectionReason?: Moderate["rejectionReason"]
```
kabeaty: Should the type here be `GQLRejectionReason`?

marcushaddon: ✅

kabeaty (approved):
Looks good!

server/src/core/server/services/users/delete.ts
```diff
@@ -309,7 +313,11 @@ async function deleteUserComments(
     },
     GQLCOMMENT_STATUS.REJECTED,
     now,
-    isArchived
+    isArchived,
+    {
+      code: GQLREJECTION_REASON_CODE.OTHER,
+      detailedExplanation: "User account deleted",
```
nick-funk, line 319: We can translate this using the following:

```
import { I18n, translate } from "coral-server/services/i18n";
...

const text = "User account deleted";
const key = "common-userDeletion-userAccountDeleted";
const args = {};

const bundle = this.i18n.getBundle(tenant.locale);
const result = translate(bundle, text, key, args);
```

Then the translations for this can be specified in `common.ftl` under `server/src/core/server/lcoales/en-US`.

PR #4412, opened 2023-11-20
DSA notifications: stop double notifying rejected comment during illegal report decisions

nick-funk:
## What does this PR do?

Prevents the reject comment flow when determining a comment is illegal from posting a second notification. We already post an illegal comment rejection, so we don't need the `rejectComment` function doing it again for us.

## These changes will impact:

- [X] commenters
- [X] moderators
- [ ] admins
- [ ] developers

None

No

None

## How do I test this PR?

- post a comment
- report it as illegal
- use an admin/mod to process the report as illegal
- check the original offending user's notifications
- see they only have one notification, about their comment being rejected for being illegal

No

## How do we deploy this PR?

- Merge into DSA epic branch

### DIFF
--- a/server/src/core/server/services/dsaReports/reports.ts
+++ b/server/src/core/server/services/dsaReports/reports.ts
@@ -233,7 +233,7 @@ export async function makeDSAReportDecision(
         detailedExplanation,
       },
       req,
-      true,
+      false, // set to false because we're about to notify below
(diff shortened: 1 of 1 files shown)


kabeaty (approved):
Looks good! Thanks for adding a comment for future reference too.

PR #4413, opened 2023-11-20
Fix/localize rejection reason

marcushaddon:
## What does this PR do?
This PR localizes the string passes as the explanation of the rejection reason when rejecting a users comments as part of the account deletion flow.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

None.

No.

n/a

## How do I test this PR?
Change your tenants locale to something other than english.
In the FTL file for that locale, add a translation for `common-accountDeleted`.
Delete a user's account, observe that the moderation actions for the resulting rejected comments have a `detailedExplanation` field matching your FTL entry.

No.

## How do we deploy this PR?
No special considerations should be needed

### DIFF
--- a/server/src/core/server/services/users/delete.ts
+++ b/server/src/core/server/services/users/delete.ts
@@ -6,7 +6,7 @@ import { ACTION_TYPE } from "coral-server/models/action/comment";
 import { Comment, getLatestRevision } from "coral-server/models/comment";
 import { DSAReport } from "coral-server/models/dsaReport";
 import { Story } from "coral-server/models/story";
-import { retrieveTenant, Tenant } from "coral-server/models/tenant";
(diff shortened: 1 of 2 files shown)


server/src/core/server/services/users/delete.ts
```diff
@@ -290,14 +289,21 @@ async function deleteUserComments(
     isArchived
   );
 
+  const bundle = i18n.getBundle(tenant.locale);
+  const translatedExplanation = translate(
+    bundle,
+    "User account deleted",
```
kabeaty, line 295: Do we want this fallback text to match the translation text? (so, `User account was deleted.`)?

kabeaty (approved):
Looks good, just left one small comment!

PR #4414, opened 2023-11-21
Fix: DSA - download void report csv

kabeaty:
## What does this PR do?

These changes make it so that downloading a `VOID` status CSV no longer causes an error. If status is void, we don't try to add the deleted comment info. Also, it updates to add the void status change to the report's history. This shows up in the report timeline on the single report page now.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can create a User 1, have them comment. Then have another User 2 report their comment for illegal content. Then have the User 1 request their account be deleted. Go in and update their scheduled deletion date to be now. Run Coral with jobs so that they will be deleted (might want to update to make account deletion run more often in accountDeletion.ts so you don't have to wait). See that the user is successfully deleted with no errors. The illegal content report against their comment should now be set to status Void. Try to download the CSV for the report and see that there are no errors. Also see that the status update is shown in the report timeline (that the user deleted their account and the report is void).

### DIFF
--- a/server/src/core/server/services/dsaReports/download.ts
+++ b/server/src/core/server/services/dsaReports/download.ts
@@ -95,7 +95,7 @@ export async function sendReportDownload(
     translate(bundle, "Action", "dsaReportCSV-action"),
     translate(bundle, "Details", "dsaReportCSV-details"),
   ]);
-  if (reportedComment) {
+  if (reportedComment && report.status !== GQLDSAReportStatus.VOID) {
     csv.write([
       formatter.format(report.createdAt),
       reporter?.username,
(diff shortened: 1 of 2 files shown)


server/src/core/server/services/dsaReports/download.ts
```diff
@@ -95,7 +95,7 @@ export async function sendReportDownload(
     translate(bundle, "Action", "dsaReportCSV-action"),
     translate(bundle, "Details", "dsaReportCSV-details"),
   ]);
-  if (reportedComment) {
+  if (reportedComment && report.status !== GQLDSAReportStatus.VOID) {
```
nick-funk, line 98: ah, yeah, that'll fix it up real good

nick-funk (approved):
Looks good!

PR #4415, opened 2023-11-21
Fix: DSA add additional comments validation update

kabeaty:
## What does this PR do?

Don't check for errors when disabling add additional comments button. You will still get errors if you try to submit an invalid comment URL, and it will be disabled if there is no input.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

Add additional comments to a DSA report. Add some that error. See that you can still add new URLs and the `Add` button isn't disabled after you get an error. You should still see correct errors upon clicking `Add`.


nick-funk (approved):
Makes sense! Approved!

PR #4416, opened 2023-11-21
Fix: DSA copy and style update for report form

kabeaty:
## What does this PR do?

Updates the copy for the report illegal content link in report forms for both archived and non-archived comments. Also updates styles so that on mobile the link text doesn't overflow.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

Click `Report` for a comment in the stream. See the updated `This comment contains illegal content` text and icon for the report link. Check on mobile. Text should wrap instead of overflowing out of the report form (note that with the copy update, you may need to make text longer while testing to make it wrap).


nick-funk (approved):
Looks good!

PR #4419, opened 2023-11-21
DSA - update Reject and moderation reasons modal

kabeaty:
## What does this PR do?

This updates the Reject and moderation reasons modal to match updated designs. Note that it leaves adding the required custom reason for `Other` to a future PR since it needs to be discussed further how to handle.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can try to reject a comment in the stream or in the admin, and you should see the updated rejection reasons modal. Compare with designs. Try out selecting different rejection reasons, adding explanations, rejecting, going back to change reasons, etc. Make sure to test the `Other` option since it requires a detailed explanation to be able to reject.

### DIFF
--- a/client/src/core/client/admin/components/ModerationReason/DetailedExplanation.css
+++ b/client/src/core/client/admin/components/ModerationReason/DetailedExplanation.css
@@ -4,12 +4,18 @@
 
 .explanationLabel {
   margin-bottom: var(--spacing-2);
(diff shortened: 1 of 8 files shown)


marcushaddon (approved):
Looks great! ty for picking this up!

PR #4428, opened 2023-11-25
New tenant utilities script for installing new tenants

nick-funk:
## What does this PR do?

Adds a script under `utilities/tenant` that includes an installation script for new tenants on existing clusters that already have tenants.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

No

## How do I test this PR?

- perform a development build of coral using `sh scripts/build-development.sh`
- `cd` into `utilities/tenant`
- run `npm i`
- perform a tenant install via ` npm run start -- --mongoURI mongodb://localhost:27017 --mongoDBName coral createTenant`
- follow the prompts for a new tenant
- check your Mongo db to see the new tenant was created
- change the domain of the new tenant to localhost
- modify the existing coral tenant to not be localhost
- start up Coral
- see that the new tenant is now being used instead of the old one

None

## How do we deploy this PR?

No special considerations.


server/src/core/server/models/tenant/helpers.ts
```diff
@@ -10,29 +10,37 @@ import { AuthIntegrations } from "../settings";
...
+  bundle?: FluentBundle
+): Tenant["badges"] => {
+  return {
+    label: bundle ? translate(bundle, "Staff", "staff-label") : "Staff",
+    adminLabel: bundle ? translate(bundle, "Staff", "staff-label") : "Staff",
+    staffLabel: bundle ? translate(bundle, "Staff", "staff-label") : "Staff",
+    moderatorLabel: bundle
+      ? translate(bundle, "Staff", "staff-label")
+      : "Staff",
+    memberLabel: bundle ? translate(bundle, "Member", "member-label") : "Staff",
```
kabeaty: Should this default to "Member" instead of "Staff" here in the case that `bundle` is false?

kabeaty (approved):
Works as expected and looks good! Just left one small comment.

server/src/core/server/models/tenant/helpers.ts
nick-funk: This is copied from the prior functionality. I haven't changed any of the logic. You may have discovered a bug? Or maybe this was done deliberately?

nick-funk: Ah, looking closer, the default for all labels is just `Staff` if it's not available. This looks to be deliberately done in this way?

nick-funk: You know what, this makes no sense to keep it as `Staff`, I'm changing it to `Member` as default.

kabeaty: Thanks for looking into this and updating!

PR #4436, opened 2023-11-28
[CORL-2979]: Add custom reason field to Other rejection reason

kabeaty:
## What does this PR do?

This adds a required `Custom reason` textbox to the rejection reason modal in the case that the `Other` reason is selected. This supports rejecting comments for custom reasons.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

This adds the `customReason` field to `RejectionReason` to support this.

no

n/a

## How do I test this PR?

You can go to a comment in the stream and reject it. Select the `Other` reason. See that there's now a required `Custom reason` field when `Other` is selected. Type a custom reason and click reject. See that the custom reason is saved. An additional explanation can also still be added. Other rejection reasons should still behave as before.

Also see that the same changes are reflected in the admin for rejection reason modal.

### DIFF
--- a/client/src/core/client/admin/components/ModerationReason/DetailedExplanation.css
+++ b/client/src/core/client/admin/components/ModerationReason/DetailedExplanation.css
@@ -1,5 +1,5 @@
 .detailedExplanation {
-  margin-bottom: var(--spacing-3);
+  margin-bottom: var(--spacing-2);
 }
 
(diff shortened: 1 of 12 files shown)


marcushaddon (approved):

nick-funk (approved):
Looks good! ✅

PR #4439, opened 2023-11-29
[CORL-2978]: Add rejection reasons to comment card and user drawer

kabeaty:
## What does this PR do?

This adds rejection reasons, if the comment is rejected and there are rejection reasons, to comment moderation cards and in the user drawer.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

Adds in the `customReason` introduced in another PR so that it can be displayed here if available.

no

n/a

## How do I test this PR?

You can reject some comments for different reasons (including `Other` and `Illegal content`), with and without detailed explanations. See that if a comment is rejected, the rejection decision info will be shown when the `DETAILS` is clicked in the comment card in mod queues and user drawer. That the comment is rejected and the rejection reason code should always be shown, in addition to the timestamp of when the comment was rejected. The legal grounds, detailed explanation, and custom reason should only show if they exist. If there is a decision, it should be the first tab shown. 

If a comment is not rejected and there is no decision to show, then `INFO` should still be the tab shown.

### DIFF
--- a/client/src/core/client/admin/components/ModerateCard/DecisionDetailsContainer.css
+++ b/client/src/core/client/admin/components/ModerateCard/DecisionDetailsContainer.css
@@ -13,3 +13,17 @@
   text-transform: uppercase;
   color: var(--palette-grey-500);
 }
+
+.rejected {
+  color: var(--palette-text-000);
+  background-color: var(--palette-error-500);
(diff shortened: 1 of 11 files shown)


nick-funk (approved):
Looks good!

PR #4444, opened 2023-12-05
Fix translation for comment rejection reason when user is banned

kabeaty:
## What does this PR do?

The translation string used in the case of comments being rejected when a user is banned is now sent through in the correct order and found in the correct translation file, as well.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can ban a user. Then check the moderate card for the comment(s) rejected when they were banned in the admin and see that it says "User was banned." under detailed explanation and not "common-userBanned".


kabeaty:
Closing this PR to handle here https://github.com/coralproject/talk/pull/4455

PR #4455, opened 2023-12-12
[CORL-2984]: Add ability to select rejection reason for Reject all comments when DSA enabled

kabeaty:
## What does this PR do?

These changes add the ability to select a rejection reason when banning a user and rejecting all of their comments when DSA is enabled.

They also ensure that reason of `SPAM` is what is sent through when using spam ban in-stream moderation.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

`rejectionReason` info is added to the input for banning a user and updating a user ban so that it can be used for the reason for rejected existing comments

no

n/a

## How do I test this PR?

Go into the admin and enable DSA. Go to the `Community` tab and find a user to ban (you'll want them to have a few comments that you can reject too). Ban them on 
1) all sites with `Reject all comments` selected. See that once you select `Reject all comments`, the `Save` button is disabled until you choose a reason for rejection. It should be enabled when you choose a reason. You should be able to add a detailed explanation as well. If you choose the `Other` reason, you will need to add a `custom reason` as well to be able to save and submit the ban. Submit and then see that all comments for the user are rejected with the given reason/explanation/custom reason combo.
2) specific sites with `Reject all comments` selected. Use same testing instructions as for all sites, as should see same options.

Now disable DSA. See that the flow is as before; you can select `Reject all comments` but there is no required rejection reason, etc.

With DSA enabled, also go to the stream as an admin and spam ban a user with multiple comments. See that when you check those comments' rejection reason, it says `SPAM`. If DSA is disabled, you shouldn't see this `SPAM` reason added.


nick-funk (approved):
Looks good!

PR #4457, opened 2023-12-13
fix: add german suspend email template

tmair:
The suspend template will be sent to a suspended user, if it is not customized. This is a user facing template and should be translated in order to prevent emails with english and german text content.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers


kabeaty (approved):
Looks good, thank you for adding!

PR #4465, opened 2023-12-15
[CORL-2993]: Top commenter backend work

kabeaty:
## What does this PR do?

These changes implement the backend to support the `Top commenter badge` feature.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

These changes add `lastFeaturedDate` and `featuredCommenter` to users.

no

n/a

## How do I test this PR?

You can feature a comment, and then go check and see that the commenter for that comment has a `lastFeaturedDate` set to the time when that comment was featured. Feature another comment by that same commenter.  See that there is a new `lastFeaturedDate` for that user. Unfeature one of the comments. The `lastFeaturedDate` should now be the date the other comment was featured. If you unfeature all comments for a commenter, `lastFeaturedDate` should be null.

Check that this works with and without the data cache.

### DIFF
--- a/server/src/core/server/data/cache/userCache.ts
+++ b/server/src/core/server/data/cache/userCache.ts
@@ -160,7 +160,9 @@ export class UserCache implements IDataCache {
     return {
       ...parsed,
       createdAt: new Date(parsed.createdAt),
-      lastFeaturedDate: new Date(parsed.lastFeaturedDate),
+      lastFeaturedDate: parsed.lastFeaturedDate
+        ? new Date(parsed.lastFeaturedDate)
+        : parsed.lastFeaturedDate,
       status: {
         ...parsed.status,
         suspension: {
--- a/server/src/core/server/data/cache/userCache.ts
+++ b/server/src/core/server/data/cache/userCache.ts
@@ -160,6 +160,7 @@ export class UserCache implements IDataCache {
     return {
       ...parsed,
       createdAt: new Date(parsed.createdAt),
+      lastFeaturedDate: new Date(parsed.lastFeaturedDate),
       status: {
         ...parsed.status,
         suspension: {
--- a/server/src/core/server/graph/resolvers/User.ts
+++ b/server/src/core/server/graph/resolvers/User.ts
@@ -102,4 +102,8 @@ export const User: GQLUserTypeResolver<user.User> = {
   lastSeenNotificationDate: ({ lastSeenNotificationDate }) => {
     return lastSeenNotificationDate ?? new Date(0);
   },
+  featuredCommenter: ({ lastFeaturedDate }) => {
+    const tenDaysAgo = new Date(new Date().setDate(new Date().getDate() - 10));
+    return lastFeaturedDate && lastFeaturedDate >= tenDaysAgo;
+  },
 };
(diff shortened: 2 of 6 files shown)


server/src/core/server/data/cache/userCache.ts
```diff
@@ -160,6 +160,7 @@ export class UserCache implements IDataCache {
     return {
       ...parsed,
       createdAt: new Date(parsed.createdAt),
+      lastFeaturedDate: new Date(parsed.lastFeaturedDate),
```
nick-funk: Is `lastFeaturedDate` sometimes null/undefined? If so, do we want to check that and set the value accordingly?

server/src/core/server/graph/resolvers/User.ts
```diff
@@ -102,4 +102,8 @@ export const User: GQLUserTypeResolver<user.User> = {
   lastSeenNotificationDate: ({ lastSeenNotificationDate }) => {
     return lastSeenNotificationDate ?? new Date(0);
   },
+  featuredCommenter: ({ lastFeaturedDate }) => {
+    const tenDaysAgo = new Date(new Date().setDate(new Date().getDate() - 10));
```
nick-funk: Would this be more readable?

```
const tenDaysAgo = new Date(Date.now() - 10 * 24 * 60 * 60 * 1000)
```

server/src/core/server/models/comment/comment.ts
```diff
@@ -1701,3 +1701,30 @@ export async function retrieveFeaturedComments(
...
+  tenantID: string,
+  userID: string,
+  excludedCommentID: string
+) {
+  const $match: FilterQuery<Comment> = {
+    tenantID,
+    authorID: userID,
+    commentID: { $ne: excludedCommentID },
+    "tags.type": GQLTAG.FEATURED,
+    status: { $in: PUBLISHED_STATUSES },
```
nick-funk: I think, since all featured comments are always published, you can remove this status filter to improve the performance of this query.

If it is ever rejected, the featured tag will also be removed.

server/src/core/server/models/comment/comment.ts
```diff
@@ -1701,3 +1701,30 @@ export async function retrieveFeaturedComments(
...
+export async function retrieveLatestFeaturedCommentForAuthor(
+  mongo: MongoContext,
+  tenantID: string,
+  userID: string,
+  excludedCommentID: string
+) {
+  const $match: FilterQuery<Comment> = {
+    tenantID,
+    authorID: userID,
+    commentID: { $ne: excludedCommentID },
```
nick-funk: What is the purpose of an `excludedCommentID`?

nick-funk (commented):
Had a few questions and suggestions before I approve. Overall, looks very good!

server/src/core/server/data/cache/userCache.ts
kabeaty: good point, updated, thanks!

server/src/core/server/graph/resolvers/User.ts
kabeaty: sure, updated!

server/src/core/server/models/comment/comment.ts
kabeaty: Okay, thanks...I can remove for `retrieveFeaturedComments` too probably then I guess (it's basically the same query, just I added in `authorID` for this one)

server/src/core/server/models/comment/comment.ts
kabeaty: I was thinking that way it couldn't ever grab the newly unfeatured comment here...but looking at this again, this shouldn't happen, so probably best to just simplify and remove.

nick-funk (approved):
Looks great, approved!

PR #4466, opened 2023-12-18
[CORL-2965]: Add DSA input fields max length validations

kabeaty:
## What does this PR do?

These changes add max length validations to all DSA input fields. The DSA "what law do you believe has been broken" field has a max of 500 characters, additional info a max of 1000 characters, and additional comment URLs a max of 100 characters. The first two of those now show a "remaining characters" countdown. If too many characters are entered in the fields, the `Submit report` button will remain disabled. If an additional comment URL is attempted to be added that is too long, the relevant error will be shown.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can test this PR by enabling DSA and then going to a comment to report it and clicking the link to report illegal content.

Once on the DSA report form, you should now see info about how many characters are remaining for the first two input fields. Type in those fields and see that the remaining characters update. See that if you enter input in all required fields (law believed to be broken, additional info, and checkbox) below the max limits, the `Submit report` button is enabled. If you enter too many characters in one of the fields, it becomes disabled again.

If you attempt to add an additional comment URL that is beneath the max and valid, it is added as expected. If you attempt to add an additional comment URL over the max limit of 100 characters, you should see a relevant error.

### DIFF
--- a/client/src/core/client/stream/tabs/Comments/IllegalContentReportView/AddAdditionalComments.tsx
+++ b/client/src/core/client/stream/tabs/Comments/IllegalContentReportView/AddAdditionalComments.tsx
@@ -7,9 +7,13 @@ import React, {
 } from "react";
 import { Field, useForm } from "react-final-form";
 
+import { MAX_DSA_ADDITIONAL_COMMENT_URL_LENGTH } from "coral-common/common/lib/constants";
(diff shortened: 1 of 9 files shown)


nick-funk (approved):
Looks good!

PR #4474, opened 2024-01-02
[CORL-3005]: Refactor and optimize tests including CommentContainer and BanModal specs

kabeaty:
## What does this PR do?

These changes remove the flaky `CommentContainer` tests. They are updated to move their elements into tests where these things were already being tested. For example, tests for disabled commenting and closed stories already existed elsewhere and have been included there, replies are tested elsewhere already, same with rendering of comments in the stream. This ensures these elements are still tested while not having these flaky tests rendering the whole app to test each element out of its context of how a user would interact with it. Fixtures used only by these tests have also been removed.

The `BanModal` spec was one of our slowest tests. They were taking awhile because they were looping through and rendering the app and running tests for every email domain in the protected domains list; these updates now test the functionality against just one of the protected email domains.

These changes also update to use the `--maxWorkers=50%` option for running Jest tests. This has made running tests locally much more efficient.

In addition, these changes provide an empty string fallback in all cases where there is no username available for a comment's commenter. Previously, this was being provided in some places but not others, leading to errors in localization vars that were showing up when tests were running.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can test this PR by running tests locally and seeing that they efficiently run. Look at the latest test run in GitHub Actions for this PR and see that the `admin` tests especially are showing around 30s faster than other branches without these changes.

no


nick-funk (approved):
Looks good!

PR #4476, opened 2024-01-04
[CORL-3001]: clean up dependencies in client and server post-monorepo update

kabeaty:
## What does this PR do?

These changes audit the dependencies for both `client` and `server` post-monorepo updates. We had a lot of dependencies in both when they only needed to be in either `client` or `server`.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

Run `sh scripts/clear-node_modules.sh` and `sh npm-i.sh`. Then run Coral and click around, seeing that there are no errors and all still works as expected.


nick-funk (approved):
Tested and this works, approved!

PR #4481, opened 2024-01-11
[CORL-2805]: Delete user button in admin

kabeaty:
## What does this PR do?

These changes add a button to the user account history drawer to be able to delete a user's account if you are an admin. You can also cancel the deletion request there if one has already been made. The history of account deletion requests and cancellations is also now displayed in the user account history drawer.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [x] admins
- [ ] developers

These changes add user deletion history to a user's status history to keep track of requests and cancellations for deletion. The changes also add mutations for scheduling deletions and cancelling deletions to be used by admin users other than the target user.

no

n/a

## How do I test this PR?

You can test this PR by logging in as an admin user and clicking on another user's username in the Community tab of the admin to open that user's account history drawer. Then click `Account History`. You should see a button at the top that says `Delete Account`. When you click it, you will see that it now says the account has been scheduled for deletion along with when and the option to cancel it. You should also see this deletion request at the top of the account history with the username of the user who you used to request it. Click `Cancel User Deletion` button. See that the request is canceled, with the deletion cancellation request shown at the top of the account history and the `Delete Account` button shown again. 

If there is an error when scheduling the account deletion, its message will be displayed in the account deletion popover. If there is an error canceling the account deletion, it will be displayed in account history near the cancel deletion callout. You can also test these if you like.

You can also confirm that:
* only admins can see the button when logged into the admin
* if a user requests their own deletion via `My profile` -> `Account`, the deletion history is updated and shown in the user account history drawer in the admin accordingly
* once the deletion time has passed after an admin-requested deletion is scheduled, as with a user-scheduled deletion, the user is deleted


server/src/core/server/graph/resolvers/UserDeletionStatus.ts
```diff
@@ -0,0 +1,14 @@
+import * as user from "coral-server/models/user";
+
+import { GQLUserDeletionStatusTypeResolver } from "coral-server/graph/schema/__generated__/types";
+
+export type UserDeletionStatusInput = user.ConsolidatedUserDeletionStatus & {
+  userID: string;
+};
+
+export const UserDeletionStatus: Required<
+  GQLUserDeletionStatusTypeResolver<UserDeletionStatusInput>
```
nick-funk, line 10: Are we using an input as a resolver result type here? Maybe time for a new interface or maybe there is a generated schema type that would make more sense?

server/src/core/server/graph/resolvers/UserStatus.ts
```diff
@@ -57,6 +58,10 @@ export const UserStatus: Required<GQLUserStatusTypeResolver<UserStatusInput>> =
       ...user.consolidateUsernameStatus(username),
       userID,
     }),
+    deletion: ({ userID, deletion }): UserDeletionStatusInput => ({
```
nick-funk, line 61: Same here, seems like the resolver might have generated a result type instead of an Input.

nick-funk (approved):
Looks good, made a couple comments about some resolver return types that might be worth looking into.

PR #4486, opened 2024-01-17
DSA fix: only check for duplicate report if not anonymous user

kabeaty:
## What does this PR do?

Update so that we only check for duplicate DSA reports submitted by a user for a particular comment if the user is signed in and it is **not** an anonymous report.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

Make sure you're signed out. Try to DSA report the same comment twice. See that there are no errors and both show in the `DSA Reports` tab.


nick-funk (approved):
Makes sense!

PR #4489, opened 2024-01-18
[CORL-3018]: update client handler for local dev to add port

kabeaty:
## What does this PR do?

These changes update the client handler for local dev to add a port to the root URL. This enables improved experience for local dev for the wordpress plugin.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

## How do I test this PR?

Test out AMP support with the WP plugin and see that it works as expected. (Without this, root URL will not be correct and AMP local dev won't work.)


nick-funk (approved):
Makes sense!

PR #4499, opened 2024-01-23
add file-loader as a dev-dependency in client

### LINKED ISSUE #4105: Cannot find module 'file-loader'
When `npm install --legacy-peer-deps` and `npm run build`, the following error occurred.

```
> @coralproject/talk@7.4.3 build:server
> gulp server


> @coralproject/talk@7.4.3 build:client
> ts-node --transpile-only ./scripts/build.ts

[04:45:13] Using gulpfile ~/talk/gulpfile.js
[04:45:13] Starting 'server'...
[04:45:13] Starting 'server:schema'...
[04:45:14] Finished 'server:schema' after 710 ms
[04:45:14] Starting 'server:scripts'...
[04:45:14] Starting 'server:static'...
[04:45:14] Starting 'locales'...
Creating an optimized production build...
{"name":"coral","hostname":"ls.leafia.net","pid":1620389,"level":20,"config":"{\n  \"env\": \"production\",\n  \"port\": 3003,\n  \"dev_port\": 8080,\n  \"defaultLocale\": \"en-US\",\n  \"fallbackLocale\": \"en-US\",\n  \"generateReport\": false,\n  \"disableSourcemaps\": false,\n  \"disableMinimize\": false,\n  \"disableChunkSplitting\": false,\n  \"enableTreeShake\": false,\n  \"enableReactProfiler\": false,\n  \"maxCores\": 4\n}","msg":"loaded configuration","time":"2022-12-02T19:45:14.662Z","v":0}
Failed to compile.

Cannot find module 'file-loader'
Require stack:
- /srv/coral/talk/src/core/build/createWebpackConfig.ts
- /srv/coral/talk/scripts/build.ts


ERROR: "build:client" exited with 1.
```
This is fixed by doing `npm install file-loader --leacy-peer-deps`. Please add it `package.json`.

coral/talk is db8f4c0c3013b80fdc103a4d7e094469bbd6fbb0, `v7.4.3`.

**Expected behavior:**
`npm install` finished successfully.

**Actual behavior:**
Error, Cannot find module 'file-loader'.

**Related Issues:** 

**Versions:**

- NodeJS: `v14.21.1`
- NPM: `8.19.3`
- MongoDB: `6.0.3`
- Redis: `6.0.16`
- OS: Debian `11.5`

oliver-dvorski:
## What does this PR do?

Fixes https://github.com/coralproject/talk/issues/4105

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

None

No

## How do I test this PR?

- Make a fresh clone of the Coral repo
- `cd` into `client`
- Run `npm i --legacy-peer-deps` (without the flag for legacy peer deps, the install will fail, but that has nothing to do with this PR)
- Run `npm run watch`
- The command will fail, reporting that `file-loader` is nowhere to be found (unless this package is globally installed on your machine)
- Apply this patch
- Run `npm i --legacy-peer-deps`
- Run `npm run watch`
- Now it works

No

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -149,6 +149,7 @@
         "extract-loader": "^5.1.0",
         "fake-indexeddb": "^3.1.2",
         "farce": "^0.4.5",
+        "file-loader": "^6.2.0",
         "final-form": "^4.20.1",
         "final-form-arrays": "^3.0.2",
         "flat": "^5.0.0",
@@ -9682,6 +9683,7 @@
     },
     "node_modules/argparse": {
       "version": "1.0.10",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "sprintf-js": "~1.0.2"
@@ -12102,16 +12104,6 @@
       "dev": true,
       "license": "MIT"
     },
-    "node_modules/bufferutil": {
-      "version": "4.0.2",
-      "hasInstallScript": true,
-      "license": "MIT",
-      "optional": true,
-      "peer": true,
-      "dependencies": {
-        "node-gyp-build": "^4.2.0"
-      }
-    },
     "node_modules/builtin-modules": {
       "version": "1.1.1",
       "dev": true,
@@ -13826,6 +13818,7 @@
     },
     "node_modules/cosmiconfig": {
       "version": "3.1.0",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "is-directory": "^0.3.1",
@@ -18580,8 +18573,6 @@
       "resolved": "https://registry.npmjs.org/file-loader/-/file-loader-6.2.0.tgz",
       "integrity": "sha512-qo3glqyTa61Ytg4u73GultjHGjdRyig3tG6lPtyX/jOEJvHif9uB0/OCI2Kif6ctF3caQTW2G5gym21oAsI4pw==",
       "dev": true,
-      "optional": true,
-      "peer": true,
       "dependencies": {
         "loader-utils": "^2.0.0",
         "schema-utils": "^3.0.0"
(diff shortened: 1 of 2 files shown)


nick-funk:
Thank you for taking the time to submit a PR!

A few of the items in your testing instructions caught my eye and I was hoping to point you in the right direction:

1. You've indicated to use `npm i --legacy-peer-deps`. This is likely due to the fact you are still using the legacy node 14. If you check the `package.json` you'll notice that it states the following:

    ```
    "engines": {
        "node": "^18.16.0",
        "npm": "8.0.0"
     },
     ```
     
     This means you need to use Node `18.16.0` and npm `8.0.0`. I'd recommend using `nvm` to install this (https://github.com/nvm-sh/nvm).
     
     With that installed, you should no longer need to use the `--legacy-peer-deps` flag.
     
2. I see you're wanting us to add in the `file-loader` package. We don't use the `file-loader` package because it runs into a memory issue on macOS/linux and instead we use a more memory safe tooling from Meta called `watchman` (https://facebook.github.io/watchman/). If you install this, you won't need the `file-loader` package.

    This can be brought in via a simple `brew install watchman` if you have `homebrew` (https://brew.sh/) installed.

I hope that helps resolve your development issues!

oliver-dvorski:
Thank you for taking the time to help me, @nick-funk. I appreciate that.

You are completely correct. I managed to eliminate both issues. But not by upgrading node or installing Watchman, I already have Node 18 and I've got Watchman installed. The issue went away when I downgraded npm from 8.19.4 to 8.0.0.

However, this doesn't remove the issue which is being reported in newer versions of npm. The client requires React 18 **and** `@giphy/react-components@5.4.0` which requires React 16.10.2 - 17. But that's an issue for another pull request.

The topic of this PR is the webpack `file-loader`. I understand that you're using Watchman to get notified about file changes, but my understanding is that the webpack `file-loader` has nothing to do with tracking file changes. It's a file processing utility. If you take a look at this line:

https://github.com/coralproject/talk/blob/6d621e4e755a84f077e1976cd0b36abe6314a99b/client/src/core/build/createWebpackConfig.ts#L528

you'll notice that Coral does have a direct dependency on the package. It is being installed by npm 8.0.0 because it comes with `url-loader@4.1.1`, which is listed in the dev-dependencies. So it's there by chance and Coral does depend on it. If we want to be able to update npm in the future, this package has to be listed as a dependency, or we need to break the dependency on it.

I appreciate your time and work very much and I do hope we can find a solution for this issue.

ihardyslide:
I can confirm I have the same issue since Coral `v7.4.4`.
I am currently upgrading step by step from 7.3.0 to latest version.
Node: `18.17.0`
NPM: `8.19.4`

Downgrading NPM to `8.0.0` did not help in my case.

PR #4502, opened 2024-01-23
[CORL-3042]: Fix feature/unfeature counts bug

kabeaty:
## What does this PR do?

These changes address a bug where if a comment was featured or unfeatured multiple times by different mods, then the count would be updated. This makes it so that if a comment is already featured and someone tries to feature it again, it doesn't update counts or otherwise re-feature. If a comment is already unfeatured and someone tries to unfeature it again, it doesn't update counts or otherwise re-unfeature.

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

## How do I test this PR?

See that if you have two windows open with different logged-in mods, and each mod features the same comment, the count will increment by 1 in both cases and it will say "featured". On refresh, the count will still be the same and the comment will still be featured.

See that if you have two windows open with different logged-in mods, and each mod unfeatures the same comment, the count will decrement by 1 in both cases and it will say "unfeatured". On refresh, the count will still be the same and the comment will be unfeatured.

### DIFF
--- a/server/src/core/server/graph/mutators/Comments.ts
+++ b/server/src/core/server/graph/mutators/Comments.ts
@@ -248,7 +248,7 @@ export const Comments = (ctx: GraphContext) => ({
     // Validate that this user is allowed to moderate this comment
     await validateUserModerationScopes(ctx, ctx.user!, { commentID });
 
-    const comment = await addTag(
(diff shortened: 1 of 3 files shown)


nick-funk (approved):
Looks good!

PR #4503, opened 2024-01-23
Create a live icon for showing notification counts

nick-funk:
## What does this PR do?

Adds a live icon that shows notification counts.

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None, piggybacks on existing `notificationCount(...)` query.

Adds the `notificationsPollRate` env var.

N/A

## How do I test this PR?

- post a comment
- have another account reply to it
- see the little bell icon increment a count

None

## How do we deploy this PR?

Merge into epic branch and release that.

### DIFF
--- a/client/src/core/client/stream/tabs/Notifications/LiveBellIcon.tsx
+++ b/client/src/core/client/stream/tabs/Notifications/LiveBellIcon.tsx
@@ -1,8 +1,12 @@
+import cn from "classnames";
 import { useLocal } from "coral-framework/lib/relay";
-import React, { FunctionComponent } from "react";
+import React, { FunctionComponent, useMemo } from "react";
 import { graphql } from "relay-runtime";
 
 import { LiveBellIconLocal } from "coral-stream/__generated__/LiveBellIconLocal.graphql";
+import CLASSES from "coral-stream/classes";
+import { NotificationBellIcon, SvgIcon } from "coral-ui/components/icons";
+import styles from "./LiveBellIcon.css";
 
 interface Props {
   userID?: string;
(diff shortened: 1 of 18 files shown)


client/src/core/client/stream/tabs/Notifications/LiveBellIcon.tsx
```diff
@@ -0,0 +1,64 @@
...
+export const LiveBellIcon: FunctionComponent<Props> = ({ size = "md" }) => {
+  const [{ notificationCount }] = useLocal<LiveBellIconLocal>(graphql`
+    fragment LiveBellIconLocal on Local {
+      notificationCount
+    }
+  `);
+
+  const count = useMemo(() => {
+    if (!notificationCount) {
+      return "";
```
kabeaty: What if we returned `null` when there's no `notificationCount` or it's <= 0? This would simplify the check below, could just do `count && ()` on line 52 then.

config/lib/config.ts
```diff
@@ -77,6 +77,12 @@ export interface StaticConfig {
    * moderation and illegal reporting are enabled on the tenant.
    */
   dsaFeaturesEnabled?: boolean;
+
+  /**
+   * notificationsPollRate determins the polling rate at which the client checks
```
kabeaty: typo, `determins` --> `determines`

kabeaty (commented):
Looks good overall! Left a few small comments. Also, I did notice that the live bell is only working for admins/moderators and it's not working for regular commenters. This seems to be due to the `@auth` placed on the `notificationCount`.

client/src/core/client/stream/tabs/Notifications/LiveBellIcon.tsx
nick-funk: updated!

config/lib/config.ts
nick-funk: Fixed!

nick-funk:
> Looks good overall! Left a few small comments. Also, I did notice that the live bell is only working for admins/moderators and it's not working for regular commenters. This seems to be due to the `@auth` placed on the `notificationCount`.

I looked into it and have fixed this issue. The check is now handled gracefully in the resolver and no longer needs an `@auth(...)` directive since the conditions are simple. Should work properly now for all users!

kabeaty (approved):
looks good! thanks for the update

PR #4506, opened 2024-01-25
Add a `secretBadges` array to the user model

nick-funk:
## What does this PR do?

Adds a `secretBadges` array to the user model which is only configurable via Mongo and used for special cases by the organization.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

N/A

## How do I test this PR?

- use multi-site test with SSO users
- find an SSO user in Mongo, add a `secretBadges` array field to their user document and add an entry (they are string values in an array)
- visit the site and post a comment as that user via SSO
- see that the user has the "secret badge" you gave them in Mongo
- see that any other SSO badges also exist

No

## How do we deploy this PR?

If we want to formally release this, rebase this off of `develop` (instead of v8.7.1) and merge it as usual into `develop`.


server/src/core/server/graph/resolvers/User.ts
```diff
@@ -106,4 +106,19 @@ export const User: GQLUserTypeResolver<user.User> = {
     const tenDaysAgo = new Date(Date.now() - 10 * 24 * 60 * 60 * 1000);
     return lastFeaturedDate && lastFeaturedDate >= tenDaysAgo;
   },
+  badges: ({ badges, secretBadges }) => {
+    if (!badges && !secretBadges) {
+      return undefined;
```
nick-funk, line 111: I did this to retain the original behaviour of the `badges` possibly being `undefined` and thus making sure we aren't messing things up by returning an empty array when something might expect `undefined` instead. I doubt there is any issue with returning an empty array, but why mess with an edge case that might exist?

server/src/core/server/graph/resolvers/User.ts
```diff
@@ -106,4 +106,19 @@ export const User: GQLUserTypeResolver<user.User> = {
...
+    }
```
nick-funk, line 112: I did this to retain the original behaviour of the `badges` possibly being `undefined` (type is `string[] | undefined` on the `User` model) and thus making sure we aren't messing things up by returning an empty array when something might expect `undefined` instead.

I doubt there is any issue with returning an empty array, but why mess with an edge case that might exist when we can easily return what we know works well?

kabeaty (approved):

PR #4512, opened 2024-01-30
[CORL-3020]: Update to use pnpm instead of npm

kabeaty:
## What does this PR do?

These changes update to use `pnpm` instead of `npm` as Coral's package manager.

To achieve this, it was necessary to:
- Update scripts to use `pnpm`
- Install some dependencies that weren't directly installed before
- Update `transformIgnorePatterns` in jest config to include `.pnpm` symlink path (See https://jestjs.io/docs/configuration#transformignorepatterns-arraystring and its `Tip` section)

In this PR, we also override `protobufjs` in our server `package.json` to address a security update needed there.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

First, you'll need to get set up with `pnpm`. You can run `npm install -g pnpm@8.14.3`
Then you'll want to clear your node modules. Run `sh scripts/clear-node_modules.sh`
Then you can run `sh scripts/pnpm-ci.sh` to install from the new `pnpm-lock.yaml`. (This is equivalent to `package-lock.json`)
Then you can do all the usual things. 
`pnpm run watch` inside client and server to run Coral
(Might need to run `sh scripts/build.sh` first if you see linting errors)
`pnpm run lint` inside client and server to check linting (or `sh scripts/lint.sh` to run linting everywhere)
`pnpm run test` inside client and server to check tests


nick-funk (approved):
This looks good!

Feel free to remove the now outdated `package-lock.json` files and also update the GHA CI yml's to no longer cache/load the `node_modules/` folders since we no longer need that with `pnpm`'s install speediness!

Thank you!

Issue #4522, opened 2024-02-02
Finnish translation: Notification email footer issues


juhog:
I have a PR ready to be submitted to fix this bug. My colleague has submitted similar Finnish translation fixes before in a few PRs (https://github.com/coralproject/talk/pull/2811, https://github.com/coralproject/talk/pull/3192, https://github.com/coralproject/talk/pull/3515).

**Expected behavior:**

Finnish language notification emails should show the organization name and an unsubscribe link in the email footer.

**Actual behavior:**

Finnish language notification emails are missing both the organization name and the unsubscribe link:
- Organization name is shown as "$organizationName"
- Unsubscribe link is just text, not a link that could be clicked

![email_footer](https://github.com/coralproject/talk/assets/1115979/9580fce2-a84a-4413-ba3c-f8d4d05b34be)

kabeaty:
Hello, thanks for finding this. We welcome the PR that you mention you have ready to be submitted to fix this bug. Please do open it when you get a chance, and we'll get the changes into an upcoming release.

losowsky:
Thanks - we'll take a look.

losowsky:
Now merged so am closing the issue! Thanks for submitting a PR.

PR #4524, opened 2024-02-05
Finnish translations: Fix notification email footer issues

juhog:
## What does this PR do?

Finnish language notification emails are missing both the organization name and the unsubscribe link. As described in issue https://github.com/coralproject/talk/issues/4522.

This PR fixes both of those issues. Also, this PR removes unnecessary trailing white-space from some lines in the translations file.

| Before | After |
|--------|--------|
| email_footer | email fix |

### DIFF
--- a/server/src/core/server/locales/fi-FI/email.ftl
+++ b/server/src/core/server/locales/fi-FI/email.ftl
@@ -13,7 +13,7 @@ email-template-accountNotificationForgotPassword =
 email-subject-accountNotificationBan = Sinut on asetettu kirjoituskieltoon
 email-template-accountNotificationBan =
   { $customMessage }<br /><br />
-  Jos kielto on mielestäsi aiheeton, ota yhteyttä ylläpitoon 
+  Jos kielto on mielestäsi aiheeton, ota yhteyttä ylläpitoon
   <a data-l10n-name="organizationContactEmail" >{ $organizationContactEmail }</a>.
 
 email-subject-accountNotificationPasswordChange = Salasanasi on päivitetty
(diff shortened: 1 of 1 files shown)


server/src/core/server/locales/fi-FI/email.ftl
```diff
@@ -84,8 +84,8 @@ email-template-accountNotificationDeleteRequestCompleted =
 # Notification
 
 email-footer-notification =
-  <br /><br />Ystävällisin terveisin<br />$organizationName<br /><br />
-  <i>Etkö enää halua tällaisia viestejä? <a data-l10n-name="unsubscribeLink">Paina tästä</a></i>
+  <br /><br />Ystävällisin terveisin<br />{ $organizationName }<br /><br />Etkö enää halua tällaisia viestejä? <a data-l10n-name="unsubscribeLink">Paina tästä</a>
```
kabeaty: I see that the `en-US` translation for this also includes an organization link like so: `<a data-l10n-name="organizationLink">{ $organizationName }</a>`. Could you add it in here as well while you're updating?

kabeaty (approved):
Thank you for these updates! I left one small comment about how it would be great to get the organization link added in around the updated `{ $organizationName }` variable with these changes as well.

juhog:
@kabeaty Thanks for the review. Is there something I can do to get the PR merged? I'm not quite familiar with the process yet.

kabeaty:
> @kabeaty Thanks for the review. Is there something I can do to get the PR merged? I'm not quite familiar with the process yet.

@juhog I'm working on the next release now and will be merging it in soon. So nothing else required from you now. Thanks for asking!

PR #4526, opened 2024-02-05
[CORL-3070]: filter out rejected and deleted comments from notifications

kabeaty:
## What does this PR do?

Filters out any comment reply notifications with comment replies that are rejected or deleted so that these will not show in the notification view.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can reply to comments and then reject the replies. Go to the original comment author and see that the reply doesn't show in their notification tab. Approve the reply comment. Now a notification should show in their notifications tab for it.

Reply to a comment then delete the author who replied (which will also delete their comment). See that this reply notification doesn't appear in the original commenter's notification tab.


nick-funk (approved):
Looks good!

PR #4531, opened 2024-02-09
[CORL-3076]: Only check protected email domains if banning, not for always premod

kabeaty:
## What does this PR do?

These changes make it so that protected email domains can be set to always premod, but they still cannot be set for ban new commenter accounts.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

Attempt to set a protected email domain to always premod in Configure > Moderation > Users. See that it works. Attempt to set it to ban new users. See that it still throws an error.


nick-funk (approved):
Looks good!

PR #4538, opened 2024-02-14
[CORL-3091/92] load/replace default notifications preferences from Mongo

nick-funk:
## What does this PR do?

- create a default state for notification preferences
- use the default if they are undefined on a user in the resolver
- use the default if they are undefined on a user in the internal notifications context

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None

No

N/A

## How do I test this PR?

- remove a users `inPageNotifications` in Mongo
- post a comment as that user
- with another user, reply to the comment
- on that first user, see they have notification counts (live bell count goes up) even though they don't have any preferences set
- modify the preferences for first user under My Profile to no longer notify for replies
- using another user, again reply to comment
- see count does not increment in notifications (live bell count does not go up/appear)

No

## How do we deploy this PR?

Merge into epic branch.

### DIFF
--- a/server/src/core/server/graph/resolvers/User.ts
+++ b/server/src/core/server/graph/resolvers/User.ts
@@ -88,15 +88,11 @@ export const User: GQLUserTypeResolver<user.User> = {
     // Get the Stories!
     return ctx.loaders.Stories.story.loadMany(results.map(({ _id }) => _id));
   },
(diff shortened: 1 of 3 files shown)


kabeaty (approved):
Looks good!

PR #4542, opened 2024-02-15
[CORL-3095]: disable e-mail notifications when in-page notifications is enabled

kabeaty:
## What does this PR do?

These changes make it so that when in-page notifications is enabled, e-mail notifications are disabled. There is also a note about this added to the admin configuration.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [x] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can enable in-page notifications. See that e-mail notifications and their settings are no longer shown. Also see that if you reply to a user (who has e-mail settings for replies enabled) while in-page notifications is enabled, that e-mail notification won't get sent out. 

If you disable in-page notifications again, they can then change their e-mail settings and e-mail notifications are sent out again on reply, etc. In-page notification settings are no longer shown in this case, and neither are in-page notifications.


nick-funk (approved):
Looks good!

PR #4548, opened 2024-02-20
[CORL-3020]: update to pnpm

kabeaty:
## What does this PR do?

**These changes add fixes to address an issue with scripts not running everything as expected to remove `npm-run-all` and just use `pnpm`, plus updates how some args are passed through.**

These changes update to use `pnpm` instead of `npm` as Coral's package manager.

To achieve this, it was necessary to:
- Update scripts to use `pnpm`
- Install some dependencies that weren't directly installed before
- Update `transformIgnorePatterns` in jest config to include `.pnpm` symlink path (See https://jestjs.io/docs/configuration#transformignorepatterns-arraystring and its `Tip` section)

In this PR, we also override `protobufjs` in our server `package.json` to address a security update needed there.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

First, you'll need to get set up with `pnpm`. You can run `npm install -g pnpm@8.14.3`
Then you'll want to clear your node modules. Run `sh scripts/clear-node_modules.sh`
Then you can run `sh scripts/pnpm-ci.sh` to install from the new `pnpm-lock.yaml`. (This is equivalent to `package-lock.json`)
Then you can do all the usual things. 
`pnpm run watch` inside client and server to run Coral
(Might need to run `sh scripts/build.sh` first if you see linting errors)
`pnpm run lint` inside client and server to check linting (or `sh scripts/lint.sh` to run linting everywhere)
`pnpm run test` inside client and server to check tests


server/package.json
```diff
@@ -47,7 +48,7 @@
     "test:cache": "node --trace-warnings scripts/test.cache.js --env=jsdom",
     "tscheck:scripts": "tsc --project ./tsconfig.json --noEmit",
     "tscheck:server": "tsc --project ./src/tsconfig.json --noEmit",
-    "tscheck": "npm-run-all --parallel tscheck:*",
+    "tscheck": "pnpm run --parallel '/^tscheck:.*/'",
```
nick-funk, line 51: this has me wondering if this was even working properly before now? 🤣

client/package.json
```diff
@@ -22,38 +22,39 @@
...
+    "generate:relay-persist": "pnpm run generate:relay:account --persist && pnpm run generate:relay:admin --persist && pnpm run generate:relay:auth --persist && pnpm run generate:relay:framework --persist && pnpm run generate:relay:install --persist && pnpm run generate:relay:stream --persist",
     "generate:relay:account": "ts-node --transpile-only ./scripts/compileRelay --bundle account",
     "generate:relay:admin": "ts-node --transpile-only ./scripts/compileRelay --bundle admin",
     "generate:relay:auth": "ts-node --transpile-only ./scripts/compileRelay --bundle auth",
     "generate:relay:framework": "ts-node --transpile-only ./scripts/compileRelay --bundle framework",
     "generate:relay:install": "ts-node --transpile-only ./scripts/compileRelay --bundle install",
     "generate:relay:stream": "ts-node --transpile-only ./scripts/compileRelay --bundle stream",
-    "generate:relay": "npm-run-all --parallel generate:relay:*",
+    "generate:relay": "pnpm run generate:relay:account && pnpm run generate:relay:admin && pnpm run generate:relay:auth && pnpm run generate:relay:framework && pnpm run generate:relay:install && pnpm run generate:relay:stream",
```
nick-funk: since we have `pnpm run --parallel` can these be broken out into parallel ops again?

client/config/webpackDevServer.config.ts
```diff
@@ -107,7 +108,7 @@ export default function ({
           ].some((p) => p === lc || lc.startsWith(`${p}/`));
         },
         target: `127.0.0.1:${serverPort}`,
-        onError: (err, req, res) => {
+        onError: (err: Error, req: IncomingMessage, res: ServerResponse) => {
```
nick-funk, line 111: nice clean up of types here, thank you!

nick-funk (approved):
Looks good! Made a few comments, see if they make sense to you.

server/package.json
kabeaty: It did--`pnpm` just uses regex to run multiple scripts

client/package.json
kabeaty: Updated to make this run parallel with a multiple scripts regex, thanks!

PR #4567, opened 2024-03-06
[CORL-3125]: increment and decrement notificationCount on comment rejection/approval

kabeaty:
## What does this PR do?

When a comment is rejected, these changes check for any reply notifications for that comment and decrement the notification count for the relevant user if any are found (only decrement if user has no last seen notification date or one that is before the reply notification for the deleted comment). If a comment is rejected upon initial notification creation, we never decrement the user notification count. If a previously rejected comment is approved, then we want to increment the user count (only increment if user has no last seen notification date or one that is before the reply notification for the newly approved comment). 

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can test these changes by:
* replying to a user's comments, with a window open on the stream for the user to whom you are replying. See their notification count increase
* Go in and reject one of the replies. See that user's notification count decrease
* If you submit a comment that is immediately rejected based on including a banned word, the notification count for the user never increments
* approve a reply that you rejected. See that the notification count increments

Also check that these decrements/increments only happen when the user either has no last seen notification date **or** a last seen notification date that is before these reply notifications.

### DIFF
--- a/server/src/core/server/models/notifications/notification.ts
+++ b/server/src/core/server/models/notifications/notification.ts
@@ -59,6 +59,17 @@ export const retrieveNotificationsConnection = async (
   return resolveConnection(query, input, (n) => n.createdAt);
 };
 
+export const retrieveNotificationByCommentReply = async (
+  mongo: MongoContext,
+  tenantID: string,
+  replyCommentID: string
+) => {
+  const notification = await mongo
+    .notifications()
+    .findOne({ replyID: replyCommentID, tenantID });
+  return notification;
+};
+
 export const createNotification = async (
   mongo: MongoContext,
   notification: Notification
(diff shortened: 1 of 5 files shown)


server/src/core/server/models/notifications/notification.ts
```diff
@@ -59,6 +59,17 @@ export const retrieveNotificationsConnection = async (
...
 };
 
+export const retrieveNotificationByCommentReply = async (
+  mongo: MongoContext,
+  tenantID: string,
+  replyCommentID: string
+) => {
+  const notification = await mongo
+    .notifications()
+    .findOne({ replyID: replyCommentID, tenantID });
```
nick-funk, line 69: Do we want to create an index for this? There could be millions of notifications to search through by `replyID`.

nick-funk (approved):
Approving, as this makes sense, but we may want to add an index for the retrieval as we'll have a lot of notifications to look through by `replyID` at some point.

server/src/core/server/models/notifications/notification.ts
kabeaty: Yep, good point, I have added, ty!

PR #4569, opened 2024-03-07
Notifications: show different rejection message for rejected pending comments

nick-funk:
## What does this PR do?

When a comment is in the pending state (not visible stream side due to pre-moderation) and is then rejected, show a slightly different text that indicates it was never visible stream side (does not say it was removed from stream).

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

Adds a `previousStatus` field to notifications.

No

None

## How do I test this PR?

- post a comment as a commenter
- using an admin/mod, reject the comment
- check the commenter's notifications and see they have a rejected notification saying their comment violated guidelines and was removed
- set a commenter to `always premoderate` in the community tab
- post a comment as that commenter
- using an admin/mod, reject this comment
- check the commenter's notifications and see that they have a different message that says it violated guidelines, but does not say it was removed from the stream

No

## How do we deploy this PR?

Merge into epic branch.


kabeaty (approved):
Looks good!

PR #4570, opened 2024-03-07
notifications: add and emit viewnotificationsfeed event

kabeaty:
## What does this PR do?

Adds in an event for when the notifications feed is viewed and emits this event upon view.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [x] admins
- [x] developers

## How do I test this PR?

View the notifications feed and see that this event is emitted.


nick-funk (approved):
Looks good, approved!

PR #4573, opened 2024-03-12
Notifications: hide pre-mod reply notifications until they are approved

nick-funk:
## What does this PR do?

- Hides pre-mod notifications in the notification feed until they are approved

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None

No

N/A

## How do I test this PR?

Check that pre-mod replies are hidden:
- post a comment as a user (call them user A for simplicity)
- with another user (user B), who is pre-moderated, reply to A
- check A's notifications that it does not show a count increment, or a new notification in the feed
- approve the reply made by user B
- check the notifications for user A and see the reply shows up

Check that we increment notification count if the user hasn't seen the notifications, but a reply is approved out of pre-mod before they visit notifications:
- have user B reply again to user A
- approve the reply written by B before user A looks at their notifications
- see that the notification count increments for A and if they open the notifications feed, the reply from B is there

No

## How do we deploy this PR?

Merge into develop and release with next release.

### DIFF
--- a/server/src/core/server/services/dsaReports/reports.ts
+++ b/server/src/core/server/services/dsaReports/reports.ts
@@ -254,6 +254,7 @@ export async function makeDSAReportDecision(
         targetUserID: rejectedComment.authorID,
         type: GQLNOTIFICATION_TYPE.ILLEGAL_REJECTED,
         comment: rejectedComment,
+        previousStatus: comment.status,
         legal: {
           legality: input.legality,
           grounds: legalGrounds,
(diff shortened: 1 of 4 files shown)


server/src/core/server/services/dsaReports/reports.ts
```diff
@@ -254,6 +254,7 @@ export async function makeDSAReportDecision(
         targetUserID: rejectedComment.authorID,
         type: GQLNOTIFICATION_TYPE.ILLEGAL_REJECTED,
         comment: rejectedComment,
+        previousStatus: comment.status,
```
nick-funk, line 257: these are just myself doing clean up to add in the previousStatus where possible so we can trust it if it's made available to us inside the context for future notifications work.

kabeaty (approved):
Looks good!

PR #4577, opened 2024-03-15
update git hooks to use pnpm

nick-funk:
## What does this PR do?

- Update the git hooks to use `pnpm`

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

N/A

## How do I test this PR?

- run `sh initialize.sh`
- make a test branch
   ```
   git branch precommit-check
   git checkout precommit-check
   ```
- modify the readme or other trivial file
- save the file
- try and commit, see that it works
   ```
   git commit -m "don't care, gonna throw away"
   ```
- let the pre-commit operations happen
- see that it commits properly since the git hooks now have access to `pnpm`
- switch back to `develop` and delete our test branch
   ```
   git checkout develop
   git branch -D precommit-check
   ```

No

## How do we deploy this PR?

Merge into develop.

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -8,8 +8,8 @@ FAST_LINT=true
 
 cd server
 
-npm run precommit
+npx pnpm run precommit
 
 cd ../client
 
-npm run precommit
+npx pnpm run precommit
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -6,10 +6,11 @@
 
 FAST_LINT=true
 
-cd server
+export PATH=$PATH:~/.nvm/versions/node/v18.16.0/bin
 
-npx pnpm run precommit
+cd server
+pnpm run precommit
 
 cd ../client
 
-npx pnpm run precommit
+pnpm run precommit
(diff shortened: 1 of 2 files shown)


scripts/initialize.sh
```diff
@@ -21,6 +21,10 @@ else
   brew install watchman
 fi
 
+echo "installing correct pnpm version globally (needed for git hooks)"
+nvm i v18.16.0
+npm install -g pnpm@8.14.3
```
nick-funk, line 26: I'm creating a v18.16.0 node instance inside `nvm` here and then installing `pnpm` globally for our target version so that it can be mounted and accessed by the githooks.

It has to be specifically `v18.16.0` so that I can mount that path inside the git hook for duration of the git hook while `pnpm` runs.

scripts/hooks/pre-commit
```diff
@@ -6,10 +6,11 @@
 
 FAST_LINT=true
 
-cd server
+export PATH=$PATH:~/.nvm/versions/node/v18.16.0/bin
```
nick-funk, line 9: this is how we tell the git hook where `pnpm` is so it stops whining about it not existing.

tessalt (approved):

PR #4578, opened 2024-03-18
update docs with correct node version

tessalt:
## What does this PR do?
Update the node version referenced in the docs. What should we be putting in terms of exact versions vs. greater than? 

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers


nick-funk (approved):
Looks good!

PR #4589, opened 2024-03-29
[CORL-3120]: Add initialStatus to comment and use for pre-mod labels in moderate cards

kabeaty:
## What does this PR do?

These changes add an `initialStatus` field to `Comment`. This helps have a clearer history from the initial creation of the comment into what its status history is. A `status` field is added to comment revisions, so we also have status per revision now. Then we use the first revision's status to populate `initialStatus`.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

This adds an `initialStatus` field to the `Comment`. It also adds `status` to `CommentRevision`.

no

n/a

## How do I test this PR?

You can set a user to pre-mod. Create comments as that user. See that they have the pre-mod label in moderate cards, in moderation queues and in user history drawer. Approve/reject these comments. See that they still have the pre-mod label based on their initial pre-mod status.

### DIFF
--- a/client/src/core/client/admin/components/ModerateCard/MarkersContainer.tsx
+++ b/client/src/core/client/admin/components/ModerateCard/MarkersContainer.tsx
@@ -3,6 +3,7 @@ import React, { useMemo } from "react";
 import { graphql } from "react-relay";
 
 import { withFragmentContainer } from "coral-framework/lib/relay";
+import { GQLCOMMENT_STATUS } from "coral-framework/schema";
 import {
   Flex,
   HorizontalGutter,
@@ -27,14 +28,14 @@ const markers: Array<
   (c: MarkersContainer_comment) => React.ReactElement<any> | null
 > = [
   (c) =>
-    (c.status === "PREMOD" && (
+    (c.status === GQLCOMMENT_STATUS.PREMOD && (
       <Localized id="moderate-marker-preMod" key={keyCounter++}>
         <Marker color="pending">Pre-Mod</Marker>
       </Localized>
     )) ||
     null,
   (c) =>
-    (c.status === "PREMOD" &&
+    (c.status === GQLCOMMENT_STATUS.PREMOD &&
       c.author &&
       c.author.premoderatedBecauseOfEmailAt && (
         <Localized id="moderate-marker-preMod-userEmail" key={keyCounter++}>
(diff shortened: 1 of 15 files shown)


client/src/core/client/admin/components/ModerateCard/MarkersContainer.tsx
```diff
@@ -42,6 +42,13 @@ const markers: Array<
         </Localized>
       )) ||
     null,
+  (c) =>
+    (c.status !== "PREMOD" && c.initialStatus === "PREMOD" && (
```
nick-funk: Do we have string type enums that could be imported here instead of using raw string comparisons?

server/src/core/server/models/comment/comment.ts
```diff
@@ -110,6 +110,11 @@ export interface Comment extends TenantResource {
    */
   status: GQLCOMMENT_STATUS;
 
+  /**
+   * initialStatus is the initial Comment Status.
+   */
+  initialStatus: GQLCOMMENT_STATUS;
```
nick-funk, line 116: Thoughts on storing/exposing this in the comment revisions? We follow a history based structure everywhere else in the server, might be cool to have the status per-revision?

nick-funk (commented):
Curious what your thoughts are with respect to storing this in the `CommentRevision` instead of an `initialStatus` field on the comment.

nick-funk (approved):
I'm approving because technically `status` is also a persisted computed state that is calculated from the mod actions... so this `intialStatus` makes sense in that regard too...

Coral is so weird... I'll leave it up to you which style you prefer cause Coral seems to do both...

server/src/core/server/models/comment/comment.ts
kabeaty: Yeah, I like this idea so that we have access to it for each revision this way. Updated to then use the first revision status to populate `initialStatus` for comments then.

client/src/core/client/admin/components/ModerateCard/MarkersContainer.tsx
kabeaty: Yep, updated to use that in all places status is referenced in this component...thanks!

nick-funk: awesome, thank you!

server/src/core/server/models/comment/comment.ts
nick-funk: amazing, thank you!

nick-funk (approved):
Updates look great, thank you once again! Approved!

PR #4604, opened 2024-04-26
[CORL-3136] Fix Akismet spam/not-spam comment submissions

nick-funk:
## What does this PR do?

Update approve/reject spam submission to use `FLAG__COMMENT_...` for retrieving action count status instead of `COMMENT_...`.

When we access the raw actionCounts record, we have to use the serialized values or else they will come back as undefined.

## These changes will impact:

- [X] commenters
- [X] moderators
- [ ] admins
- [X] developers

None

No

N/A

## How do I test this PR?

- run Coral in a dev mode so you can attach with a debugger
    - add a break point to `server/src/core/server/stacks/approveComment.ts` on line 70
        - on the `submitCommentAsNotSpam` call
    - add a break point to `server/src/core/server/stacks/rejectComment.ts` on line 132 
        - on the `submitCommentAsSpam` call
- enable Akismet spam detection (both radio buttons on)
    - use an akismet key from one of the dev instances 
- make a commenter user
- with that user, post a long comment
    - may I suggest 3-5 paragraphs of Star Trek ipsum? https://vlad-saling.github.io/star-trek-ipsum/
- edit the comment and paste in a link at the bottom of the comment
    - I used a random music video from YouTube as a URL 
- see that the comment is submitted for moderation
- in the admin, see that the comment has been detected as spam
    - in the details for the marker 
- approve/reject and see that the debug breakpoints you placed are hit
    - this confirms we are submitting corrections to Akismet
    - if you want to check the Akismet submission results, you can hop into those functions and debug out the results from the `AkismetClient` submission
        - spoiler: it's a boring and simple boolean true/false value 

No

## How do we deploy this PR?

Merge into `develop` and then release `develop` into `main` as usual.

### DIFF
--- a/server/src/core/server/stacks/approveComment.ts
+++ b/server/src/core/server/stacks/approveComment.ts
@@ -64,8 +64,8 @@ const approveComment = async (
   if (
     revision &&
     tenant.integrations.akismet.enabled &&
-    (revision.actionCounts.COMMENT_REPORTED_SPAM > 0 ||
-      revision.actionCounts.COMMENT_DETECTED_SPAM > 0)
+    (revision.actionCounts.FLAG__COMMENT_REPORTED_SPAM > 0 ||
+      revision.actionCounts.FLAG__COMMENT_DETECTED_SPAM > 0)
   ) {
     await submitCommentAsNotSpam(mongo, tenant, result.before, request);
   }
--- a/server/src/core/server/stacks/rejectComment.ts
+++ b/server/src/core/server/stacks/rejectComment.ts
@@ -126,8 +126,8 @@ const rejectComment = async (
   if (
     revision &&
     tenant.integrations.akismet.enabled &&
-    (revision.actionCounts.COMMENT_REPORTED_SPAM > 0 ||
-      revision.actionCounts.COMMENT_DETECTED_SPAM > 0)
+    (revision.actionCounts.FLAG__COMMENT_REPORTED_SPAM > 0 ||
+      revision.actionCounts.FLAG__COMMENT_DETECTED_SPAM > 0)
   ) {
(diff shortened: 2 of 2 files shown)


server/src/core/server/stacks/approveComment.ts
```diff
@@ -64,8 +64,8 @@ const approveComment = async (
   if (
     revision &&
     tenant.integrations.akismet.enabled &&
-    (revision.actionCounts.COMMENT_REPORTED_SPAM > 0 ||
-      revision.actionCounts.COMMENT_DETECTED_SPAM > 0)
+    (revision.actionCounts.FLAG__COMMENT_REPORTED_SPAM > 0 ||
+      revision.actionCounts.FLAG__COMMENT_DETECTED_SPAM > 0)
```
nick-funk, line 68: the fact that we can accidentally not use the serialized value for this here is rather annoying... everywhere else in the code-base, this is `COMMENT_REPORTED_SPAM` and `COMMENT_DETECTED_SPAM`. You just have to "know" that it's going to do a weird `FLAG__` when serializing and persisting this enum into Mongo.

This speaks to a limitation of using GraphQL enum types inside your data persistence instead of creating concrete maps and persistence in a data layer. It's important to know you're accurately saving, retrieving, and comparing data types.

Doubly annoying is, this `actionCounts` object is not typed in any way. It is a `Record<string, number>` and has no indication of what any of the counts are indexed as. It's all just hidden "magic" key values and there is no way to know this by reading the code. You have to debug this and realize "oh wow, the enum serialization does not match the names of the types/values we use everywhere else".

Frustrating.

kabeaty (approved):
Oof. Good catch and fix here!

tessalt (approved):

PR #4605, opened 2024-04-30
[CORL-3165]: only notify on approval when comment previously pending

kabeaty:
## What does this PR do?

This checks that the previous status of the comment was a `PENDING_STATUS` (premod or system withheld) and only creates a notification when a comment is approved in this case.

## These changes will impact:

- [x] commenters
- [x] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

Create pending comments as a user (put the user on premod, for example). As another user, approve these comments. See that you get the expected notification of approval.

Report a user's comment. Then approve that comment as another user. See that there is no notification.

Create a comment with a suspect word. Then approve that comment as another user. See that there is no notification.

### DIFF
--- a/server/src/core/server/stacks/approveComment.ts
+++ b/server/src/core/server/stacks/approveComment.ts
@@ -21,6 +21,12 @@ import {
 
 import { publishChanges } from "./helpers";
 
+const notVisibleStatuses = [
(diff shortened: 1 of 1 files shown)


nick-funk (approved):
Looks good!

PR #4625, opened 2024-05-29
Fix `node-gyp`

nick-funk:


nick-funk:
Closing, this is not the fix, was simply testing a node-gyp workflow in GHA.

PR #4631, opened 2024-06-05
Change order of moderation pipeline to ensure external phases can run (Fixes #4630)

### LINKED ISSUE #4630: Moderation Pipeline will never run external phases when PREMOD is set
**Expected behavior:**
External moderation phases are not skipped when pre-moderation phase is enabled.

**Actual behavior:**
When pre-moderation is enabled in admin settings, external moderation phases will never run as the `statusPreModerate.ts` will set a `status` and therefore the pipeline will abort early.

-----

I wonder if it really makes sense to run the `statusPreModerate` phase before external phases. When premod is disabled and an external phase is unavailable (e.g. due to network issues or errors), this could potentially result in every comment being published immediately.

However, if external phases would run before the premod phase, they would have the opportunity to set a status and if not, the premod phase would still ensure that moderation is required.

The only necessary change would be to change the order of the very last phases. Instead of running external phases last, they should run before any phase that returns a status.

bastiankistner:
## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

### DIFF
--- a/server/src/core/server/services/comments/pipeline/phases/index.ts
+++ b/server/src/core/server/services/comments/pipeline/phases/index.ts
@@ -62,11 +62,17 @@ export const moderationPhases: IntermediateModerationPhase[] = [
   spam,
   detectLinks,
 
-  // Apply any pre-existing conditions to these comments.
+  // If user-specific moderation is enabled, it should run before
+  // external phases as those won't have the required information
+  // about the user to decide on their own.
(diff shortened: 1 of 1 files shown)


losowsky:
Closed as this was handled in a separate PR https://github.com/coralproject/talk/pull/4641

PR #4649, opened 2024-07-17
add max width to tenor image

tessalt:
## What does this PR do?
Set a max-width on gifs so they don't overflow

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers


nick-funk (approved):
makes sense to me

Issue #4664, opened 2024-09-12
Assistive technologies read landmark label twice


michalrusina:
**Expected behavior:**

When using assistive technologies (screenreaders), all tabbed controls read the word "Tabs" twice. Once because of `aria-role="tabs"` and second time because there is an `span` element  containing text "Tabs".

**Actual behavior:** 

"Tabs" should be read only once, only use `aria-role="tabs"` where appropriate.

**Related Issues:**

This antipattern occurs on all tabbed controls and possibly also on other components. 

**Versions:**

- NodeJS: -
- NPM: -
- MongoDB: -
- Redis: -
- Browser: Chrome 128.0.6613.85
- OS: MacOS

losowsky:
Thanks - we'll add this to our backlog

losowsky:
Closing as this was included in the latest release, 9.4.0

PR #4667, opened 2024-09-27
[CORL-3190]: remove extra Tab definition for screen readers

kabeaty:
## What does this PR do?

As mentioned in issue https://github.com/coralproject/talk/issues/4664, we shouldn't need the aria info span that says `Tab:` since we have `role="tab"` on the tabs.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can use a screen reader. For example, you can test this by pressing Command-F5 to enable `VoiceOver`. Then you can use caps lock + right arrow to navigate through the page and hear that tab is only said once (without this change, `tab` is said twice).

no


nick-funk (approved):
Makes sense!

PR #4669, opened 2024-09-30
[CORL-3180]: update to not autoplay tenor gifs in mod queue

kabeaty:
## What does this PR do?

These changes update so that Tenor gifs are not autoplaying in the mod queue. For Tenor gifs, we are now retrieving the `mp4` and `gifpreview` formats from the Tenor api and attaching those to the media embed on the comment, so that they can be used to play/pause the gifs in the mod queue, similar to how Giphy gifs display if enabled.

To support any Tenor gifs that have been added since Tenor has been an optional and prior to these changes, this adds a fallback of show/hide Gif, similar to what is seen for Gifs in the comment stream.

I also moved Tenor types out into `common` where Giphy types currently live for consistency and so that we can use them in different places.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [x] admins
- [ ] developers

These changes add `still`, `video`, `width`, and `height` to `TenorMedia` schema, to make it consistent with what Giphy has to enable it to be play/paused in the mod queue.

no

n/a

## How do I test this PR?

You can have Tenor enabled for gifs, and add some comments with gifs to the comment stream. Go to the admin and find the comments you added in the moderation queue. See that you can play/pause the gifs and that they are not autoplaying.

Add comments with Tenor gifs on `develop` or another branch without these changes, and see that these gifs (when back on this branch) include a show/hide Gif button that can be toggled. This provides a fallback for any comments with Tenor gifs that don't have a still/video to use for play/pause.


client/src/core/client/admin/components/MediaContainer/GifMedia.tsx
```diff
@@ -12,19 +12,56 @@ interface Props {
   width: number | null;
   height: number | null;
   video: string | null;
+  url: string | null;
 }
 
-const GiphyMedia: FunctionComponent<Props> = ({
+const GifMedia: FunctionComponent<Props> = ({
```
kabeaty, line 18: I updated the name of this component since it is now used for displaying both Tenor and Giphy gifs.

server/src/core/server/app/handlers/api/tenor/index.ts
```diff
@@ -133,6 +93,7 @@ export const tenorSearchHandler =
     url.searchParams.set("key", apiKey);
     url.searchParams.set("limit", `${SEARCH_LIMIT}`);
     url.searchParams.set("contentfilter", contentFilter);
+    url.searchParams.set("media_filter", "gif,nanogif");
```
kabeaty, line 96: I added this `media_filter` to only return the response object formats that we need when we are calling the Tenor api here.

tessalt (approved):

client/src/core/client/admin/components/ModerateCard/CommentRevisionContainer.tsx
```diff
@@ -100,6 +107,10 @@ const enhanced = withFragmentContainer<Props>({
           ... on TenorMedia {
             url
             title
+            width
+            height
+            tenorStill: still
+            tenorVideo: video
```
nick-funk, line 113: If we're aliasing these to `still` and `video` when we consume them from the API client side, do we want to call them `TenorMedia > tenorStill` and `TenorMedia > tenorVideo`? Seems like a duplication of the naming?

Maybe `TenorMedia > still` and `TenorMedia > video`?

server/src/core/server/services/tenor/tenor.ts
```diff
@@ -0,0 +1,59 @@
...
+        status: res.status,
+      });
+    }
+
+    const data = await res.json();
+    return validateSchema(TenorRetrieveResponseSchema, data);
+  } catch (err) {
+    // Ensure that the API key doesn't get leaked to the logs by accident.
+    if (err.message) {
+      err.message = err.message.replace(tenant.media.gifs.key, "[Sensitive]");
```
nick-funk, line 53: this works well to hide the api keys, also, we can use bunyan's `pvt` and `pub` fields to hide these details if you want

nick-funk (approved):
Looks good to me, nice refactors around the presentation logic too!

client/src/core/client/admin/components/ModerateCard/CommentRevisionContainer.tsx
kabeaty: Hmm, I was trying to keep this consistent with the naming for `GiphyMedia`, except that we might not have a `still` or `video` for some Tenor gifs, so types are different.

PR #4671, opened 2024-09-30
[CORL-3179]: Create notification on previously rejected comment being approved if DSA is enabled

kabeaty:
## What does this PR do?

If DSA is enabled, then with these changes, we now create a notification if a comment was previously rejected (whether automatically for including a banned word or by a moderator) and then is approved.

If DSA is not enabled, then there are no changes; no notification is created when a comment is rejected, and there is also no subsequent notification created if it is later approved after rejection.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

This adds a new type to the `NOTIFICATION_TYPE` enum of `PREVIOUSLY_REJECTED_COMMENT_APPROVED`.

no

n/a

## How do I test this PR?

You can test out these changes by:
- Enabling DSA
- As User A, creating a comment with a banned word in it (or creating a comment and then rejecting it); see that it is rejected and gets a notification about this
- As a User B (with role moderator/admin), approving the rejected comment
- See that User A has a notification that the comment was previously rejected but has been restored to the page

Then, also:
- Disabling DSA
- Repeating the other steps above, but observing that no notifications are created for rejection/approval

### DIFF
--- a/client/src/core/client/stream/tabs/Notifications/NotificationContainer.tsx
+++ b/client/src/core/client/stream/tabs/Notifications/NotificationContainer.tsx
@@ -48,6 +48,9 @@ const getIcon = (type: NOTIFICATION_TYPE | null): ComponentType => {
   if (type === GQLNOTIFICATION_TYPE.COMMENT_APPROVED) {
     return CheckCircleIcon;
   }
+  if (type === GQLNOTIFICATION_TYPE.PREVIOUSLY_REJECTED_COMMENT_APPROVED) {
+    return CheckCircleIcon;
+  }
   if (type === GQLNOTIFICATION_TYPE.COMMENT_FEATURED) {
     return RatingStarRibbonIcon;
   }
@@ -81,6 +84,13 @@ const getTitle = (
       bundles,
       "notifications-yourCommentHasBeenApproved",
       "Your comment has been published"
+    );
+  }
+  if (type === GQLNOTIFICATION_TYPE.PREVIOUSLY_REJECTED_COMMENT_APPROVED) {
+    return getMessage(
+      bundles,
+      "notifications-yourPreviouslyRejectedCommentHasBeenApproved",
+      "Your comment was previously rejected. It has now been restored to the page."
     );
   }
   if (type === GQLNOTIFICATION_TYPE.COMMENT_FEATURED) {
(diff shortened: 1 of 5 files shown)


tessalt (approved):

client/src/core/client/stream/tabs/Notifications/NotificationContainer.tsx
```diff
@@ -83,6 +86,13 @@ const getTitle = (
       "Your comment has been published"
     );
   }
+  if (type === GQLNOTIFICATION_TYPE.PREVIOUSLY_REJECTED_COMMENT_APPROVED) {
+    return getMessage(
+      bundles,
+      "notifications-yourPreviouslyRejectedCommentHasBeenApproved",
+      "Your comment was previously rejected. It has now been restored to the page."
```
nick-funk: "restored to the page"

Have we checked this copy with Andrew?

server/src/core/server/stacks/approveComment.ts
```diff
@@ -106,6 +106,18 @@ const approveComment = async (
     });
   }
 
+  // create notification if dsa enabled upon approval of previously rejected comment
+  if (tenant.dsa?.enabled) {
+    if (previousComment?.status === GQLCOMMENT_STATUS.REJECTED) {
```
nick-funk: am I missing something on the nested if's here? Can it be simplified to the following?

```
if (tenant.dsa?.enabled && previousComment?.status === GQLCOMMENT_STATUS.REJECTED) {
```

nick-funk (approved):
Looks good to me, only thought is to confirm the copy with Andrew. Could just be me but I don't think we use "page" in our terminology when referring to the comment stream so it stuck out to me.

If I'm wrong and that was the expected copy, am cool with it, feel free to merge!

client/src/core/client/stream/tabs/Notifications/NotificationContainer.tsx
kabeaty: This is what was included in the issue. @losowsky can you confirm that this is the copy we want for this?

server/src/core/server/stacks/approveComment.ts
kabeaty: yep, it can be...updating! thanks

client/src/core/client/stream/tabs/Notifications/NotificationContainer.tsx
losowsky: I think Nick is right, let's change it to "Your comment was previously rejected. It has now been approved."

kabeaty: Updated...thank you both!

PR #4674, opened 2024-10-02
Add a safety check for tenant.dsa in external phase

rigperro:
## What does this PR do?

Adds a simple check that tenant.dsa is defined for the external moderation phases file. In our local deployment, merging from 8.5.3 to 9.3.1 introduced this line that broke our moderation (external phase). I don't know if it's due to the existing tenant being older than the setting or the external moderation phase, but nonetheless this fixes the issue. 

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

In our case, all of them but it is a single character change after all.

None

No

No new indexes.

## How do I test this PR?

I described our issue, but this probably needs no testing.

No

## How do we deploy this PR?

As per usual.


kabeaty (approved):
Thank you!

nick-funk (approved):
makes sense, good catch, approved!

Issue #4662, opened 2024-09-11
Can't fetch Comment count for URL in Kannada Language


nawazsuventure:
I have added whatever is required to fetch the comment count. But it is  **not** fetching the comment count properly for URLs in **Kannada** language.

Example of URL in kannada language:

https://prajavani-training-web.quintype.io/op-ed/olanota/ಕೋವಿಡ್-ಹೆಚ್ಚಳ-ನಾವು-ಜಾಗರೂಕರಾಗಿರಬೇಕು-ಭಯಪಡುವ-ಅಗತ್ಯವಿಲ್ಲ-ಮಾಂಡವೀಯ-2-1275643
![Screenshot 2024-09-11 at 12 15 16 PM](https://github.com/user-attachments/assets/cd11b623-612e-4bfd-a608-31cfb303360a)

Example of URL  English language (working fine):

https://prajavani-training-web.quintype.io/todays-news/test-photo-m14n2-231-1391213
![Screenshot 2024-09-11 at 12 15 37 PM](https://github.com/user-attachments/assets/cc717ff3-cc87-4a7c-bb44-695109d0cac7)
 
 
 Also, I have tried Encoding and Decoding the URL before passing.

losowsky:
Thanks - we've added this to our backlog.

losowsky:
Fixed in our latest version 9.4.0

PR #4675, opened 2024-10-02
[CORL-3189]: Add support for non-latin character range for urls for comment counts

kabeaty:
## What does this PR do?

These changes add support for non-Latin characters in story urls for comment counts. Previously, we were getting `Uncaught InvalidCharacterError: Failed to execute 'btoa' on 'Window': The string to be encoded contains characters outside of the Latin1 range.` when attempting to create a count query ref with characters outside of the range. This updates as recommended [here](https://developer.mozilla.org/en-US/docs/Glossary/Base64#the_unicode_problem) to address this issue.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

You can test that comment counts still work as expected. You can also run these changes--the new `bytesToBase64` function and `bytesToBase64(new TextEncoder().encode(URL))`--in the console with a story url such as is mentioned in [this issue](https://github.com/coralproject/talk/issues/4662) and confirm that you get a ref that can be used back and not an error.

### DIFF
--- a/client/src/core/client/count/index.ts
+++ b/client/src/core/client/count/index.ts
@@ -1,6 +1,9 @@
 import { COUNT_SELECTOR } from "coral-framework/constants";
-import detectCountScript from "coral-framework/helpers/detectCountScript";
-import resolveStoryURL from "coral-framework/helpers/resolveStoryURL";
+import {
+  bytesToBase64,
+  detectCountScript,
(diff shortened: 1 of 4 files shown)


tessalt (approved):
great investigation

PR #4680, opened 2024-10-07
Update Mongo driver to `v6.9.0`

nick-funk:
## What does this PR do?

- Upgrades Mongo to the latest driver (v6.9.0)

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

N/A

## How do I test this PR?

This touches all of Coral's data retrieval and persistence, all features must be re-tested in a regression test.

No

## How do we deploy this PR?

- Deploy to staging/testing environments
- Perform a full regression test
    - Fix any critical bugs that are found
- Issue out a release


server/src/core/server/models/comment/metrics.ts
```diff
@@ -116,10 +118,12 @@ export async function retrieveTodayCommentMetrics(
       continue;
     }
 
+    const mod = moderator;
```
kabeaty, line 121: just curious why we're assigning this here now? Looks like we're still using `moderator` on line 124 though? (Same question about similar change below in the file too)

server/src/core/server/models/helpers/query.ts
```diff
@@ -65,15 +70,16 @@ export default class Query<T> {
    * orderBy will apply sorting to the query filter when executed.
    * @param sort the sorting option for the documents
    */
-  public orderBy(sort: object): Query<T> {
+  public orderBy(sort: { [key: string]: SortDirection }): Query<T> {
+    // todo: merge sort's together
```
kabeaty: Is this still a TODO?

server/src/core/server/services/mongodb/index.ts
```diff
@@ -6,7 +6,10 @@ import logger from "coral-server/logger";
 async function createMongoClient(mongoURI: string): Promise<MongoClient> {
   try {
     return await MongoClient.connect(mongoURI, {
-      useNewUrlParser: true,
+      // believe we don't need this since the new driver only uses
+      // the new URL parser. But am leaving this here in case we have
+      // issues with URL's and want to investigate into it.
+      // useNewUrlParser: true,
```
kabeaty, line 12: 👍

server/src/core/server/models/site/index.ts
```diff
@@ -1,5 +1,5 @@
 import { identity, isNumber } from "lodash";
-import { MongoError } from "mongodb";
+import { Collection, MongoError } from "mongodb";
 import { v4 as uuid } from "uuid";
```
kabeaty, line 4: I'm seeing a warning in the console: "(node:85676) [MONGODB DRIVER] Warning: cursor.count is deprecated and will be removed in the next major version, please use `collection.estimatedDocumentCount` or `collection.countDocuments` instead". 

Looks like we are using deprecated `count()` on line 135 in this file. Should we update to one of the suggested replacements now?

kabeaty (approved):
Everything seems to be working great with the update! I left a few questions throughout.

server/src/core/server/models/comment/metrics.ts
nick-funk: it's a bug in typescript, it sometimes fails to process types after a `continue` op. If you re-assign, it works. So I did that for the few lines that were "erroring" (but compile and run fine).

server/src/core/server/models/helpers/query.ts
nick-funk: nope! I "sorted" this earlier! I will update!

nick-funk: fixed!

server/src/core/server/models/site/index.ts
nick-funk: looking into this, will see if I can update the query

nick-funk: yup, was an easy fix, done!

rigperro:
Very nice! What mongo version are you looking to use with this?

PR #4683, opened 2024-10-09
[CORL-3195]: more error handling and type check fixes

kabeaty:
## What does this PR do?

These changes:
- Make sure that we're checking that a tenant has `media` and `gifs` defined before checking that it is enabled/supported. This was throwing errors when `gifs` wasn't defined.
- Because non-error exception volume increased after Tenor option was introduced, attempt to check type of new error caught and thrown within these changes, and update what's thrown if it is not an instance of `Error`.

Also, if after these changes we are still seeing issues with non-error exceptions being thrown, we can add a check in `services/errors/sentry` for instance of `Error` and handle accordingly. I kept out of these changes to try one step at a time.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

none

no

n/a

## How do I test this PR?

Check that comments with Tenor gifs still work as expected.

### DIFF
--- a/server/src/core/server/app/handlers/api/tenor/index.ts
+++ b/server/src/core/server/app/handlers/api/tenor/index.ts
@@ -69,7 +69,7 @@ export const tenorSearchHandler =
       return;
     }
 
-    const gifsEnabled = tenant.media?.gifs.enabled ?? false;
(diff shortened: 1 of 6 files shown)


nick-funk (approved):
This looks good to me!

PR #4691, opened 2024-11-07
add helper function for checking if user is ignored

tessalt:
## What does this PR do?

We were getting errors `Cannot read properties of null (reading 'some')` from [here](https://github.com/coralproject/talk/blob/develop/server/src/core/server/services/notifications/internal/context.ts#L69). Obviously we need to add a check for `user.ignoredUsers` before checking `.some` on that array, but when I looked for this pattern elsewhere, I noticed multiple `// TODO: extract function` notes, so I extracted it into a function. 

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

NA

NA

## How do I test this PR?
- with notifications enabled, as user 1, reply to user 2
- as user 2, ensure you have received a notification
- as user 1, ignore user 2
- as user 2, reply to user 1
- as user 1, ensure that you DO NOT receive a notification
- as user 1, un-ignore user 2
- as user 2, reply to user 1
- as user 1, ensure you DO receive a notification

### DIFF
--- a/server/src/core/server/models/user/helpers.ts
+++ b/server/src/core/server/models/user/helpers.ts
@@ -6,7 +6,7 @@ import { GQLUSER_ROLE } from "coral-server/graph/schema/__generated__/types";
 import { isSiteModerationScoped } from "coral-common/common/lib/permissions";
 
 import { MODERATOR_ROLES, STAFF_ROLES } from "./constants";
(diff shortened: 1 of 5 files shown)


kabeaty (approved):
looks good to me, and all manual testing functions as expected!

nick-funk (approved):
good fix, thank you!

PR #4715, opened 2024-12-12
fix: refresh comments button shouldn't show on initial load

kabeaty:
## What does this PR do?

These changes make it so that the refresh comments button only appears when comments are navigated away from and back to.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

See that upon initial load of the page, the refresh comments button isn't showing. When going to another tab then back to comments, then it shows.

You can test against develop, where you'll see it is showing on initial load.


nick-funk (approved):
that makes sense

PR #4723, opened 2025-01-17
[CORL-3212]: throw no username included error for sso with url for setting username

kabeaty:
## What does this PR do?

These changes make it so that if an sso token does not include a username for a user, but does include a url for account management to set a username, a helpful error will be thrown letting the user know how to set their username to be able to comment.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can use multi site test to send through an sso user who does not have a username set but does have a url for account management. You should see that

### DIFF
--- a/client/src/core/client/framework/lib/errors/relayNetworkRequestError.ts
+++ b/client/src/core/client/framework/lib/errors/relayNetworkRequestError.ts
@@ -11,6 +11,14 @@ interface DupErrorObj {
   };
 }
 
+interface UsernameNotProvidedErrorObj {
+  error?: {
+    traceID: string;
+    code: string;
+    message: string;
+  };
+}
+
 const parseDuplicateEmailError = (
   error: RRNLRequestError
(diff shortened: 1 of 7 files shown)


client/src/core/client/framework/lib/errors/relayNetworkRequestError.ts
```diff
@@ -30,6 +38,25 @@ const parseDuplicateEmailError = (
...
+    if (!json || !json.error || json.error.code !== "USERNAME_NOT_PROVIDED") {
+      return null;
+    }
+
+    return json;
+  } catch {
+    return null;
+  }
+};
+
```
nick-funk, line 59: heh, looks like we're going to have to be generalizing this at some point... so many error object types popping up to be parsed through the response text 😅

nick-funk (approved):
Looks good to me!

PR #4726, opened 2025-01-28
optimize the account deletion process

nick-funk:
## What does this PR do?

- do not update comment actions (recs) as it is unnecessary
- use an appropriate index with conditional moderation to dramatically speed up the moderation process and cursor retrieval for large comment collections

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [X] admins
- [X] developers

None

No

N/A

## How do I test this PR?

- ensure that `DISABLE_JOB_PROCESSORS` is set to `false` in your `.env`
  - user deletion is processed as a job
  - if using `pnpm run start:development` use `pnpm run start:development:withJobs` instead
- enable user deletion in `Admin > Config`
- create a new user to delete
- post some comments
  - recommended that you also have another user reply to some of their comments
  - different things happen if a deleted comment has children or not
- schedule the user for deletion
  - use account options for deletion in stream tabs
- modify the scheduled deletion date for the user in Mongo
- let the job run
  - may want to modify the job schedule to run more frequently via `server/src/core/server/cron/accountDeletion.ts`
    ```
    const job = new ScheduledJob(options, {
      name: `Twice Hourly ${NAME}`,
      cronTime: "0,30 * * * *",
      command: deleteScheduledAccounts,
    });
    ```
- see that user is deleted

No

## How do we deploy this PR?

- merge into `develop`
- issue out a release

### DIFF
--- a/server/src/core/server/services/users/delete.ts
+++ b/server/src/core/server/services/users/delete.ts
@@ -1,22 +1,19 @@
-import { Collection, Document, Filter } from "mongodb";
+import { Collection, Document, WithId } from "mongodb";
 import { v4 as uuid } from "uuid";
 
 import { Config } from "coral-server/config";
 import { MongoContext } from "coral-server/data/context";
-import { ACTION_TYPE } from "coral-server/models/action/comment";
-import { Comment, getLatestRevision } from "coral-server/models/comment";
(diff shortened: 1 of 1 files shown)


kabeaty (approved):
Looks good, and works as expected!

PR #4736, opened 2025-02-10
[CORL-4733] encodes unicode chars for WebKit engines

### LINKED ISSUE #4733: "Invalid Regular Expression" Error in Safari / iOS / iPad Browsers
We're encountering a sporadic but persistent issue in Safari that is proving difficult to isolate. We're reaching out for help in identifying the root cause and potential solutions.

In Safari, the following error occurs sporadically:
`SyntaxError: Invalid regular expression: range out of order in character class`
This prevents the script `stream.26a79a60c687283b1c933f8108b48401.js` from loading properly. The issue does not occur in other browsers.

The problem does not appear on every page load but sometimes only after 20 or more refreshes. We tested a fresh, unconfigured Coral Talk installation, including rolling back to a version from four months ago, and the issue still persists. This suggests that the problem is not related to custom configurations but rather an underlying issue in the codebase.

The error is reproducible on **iPhones and iPads in any browser**, since all browsers use WebKit on iOS. On macOS, it only occurs in Safari, as Chrome and other browsers use their own rendering engines.

**Reproduction Steps**
We managed to reproduce the issue in the simplest possible setup:
- A freshly set up Docker Compose environment running the latest Coral Talk 9.x version
- Only basic configuration, no customizations or external modifications
- The script block was copied directly from the Coral Talk admin panel and placed on a completely blank HTML page

However, we could not reproduce the issue when running Coral Talk locally in development mode (pnpm watch) on our machines. This strongly suggests that the problem is tied to the build process and does not occur in the unminified development version.

**Caching Considerations**
Given the sporadic nature of the issue, we initially suspected caching as a possible cause. To rule this out, we:
- Added a Traefik proxy to enforce Pragma: no-cache and related headers
- Enabled "Disable Cache" in Safari Developer Tools
- Tested in Safari’s Private Browsing mode

Despite these measures, the issue persisted, suggesting that it is not directly related to caching.

**Suspicious Code in common/lib/helpers/validate.ts**
During our analysis, we identified a potentially problematic regular expression in the file `common/lib/helpers/validate.ts`:

`export const USERNAME_REGEX = new RegExp(/^[a-zA-ZÀ-ÖØ-öø-ÿ0-9_.]+$/);`

This regex was flagged as invalid in multiple syntax validation tools. Specifically, the character range À-ÖØ-öø-ÿ may be problematic, as Safari strictly enforces correct character class ranges in regular expressions, while other browsers may be more lenient.

**Furthermore, a similar issue has been reported in another project, [gpu.js Issue #743](https://github.com/gpujs/gpu.js/issues/743). In that case, UglifyJS minification altered a regular expression in a way that led to the exact same error in Safari. Since we were unable to reproduce the issue in development mode but consistently encountered it in the built production version, it is likely that a minification step (Webpack/UglifyJS/Terser) is transforming the regex in a way that Safari interprets as invalid.**
We think this is a very important clue in solving this issue.

Since the issue is consistent across different versions of Coral Talk and occurs even in a minimal setup, it seems to be either a Safari-specific regex parsing issue or a minification-related bug.

**We Need Help!**
At this point, we are struggling to pinpoint the exact cause of this issue and would greatly appreciate any insights from the community.

- Has anyone else encountered similar issues with regular expressions breaking in Safari after minification?
- Are there any known workarounds or ways to prevent UglifyJS/Terser from modifying specific regex patterns?
- Could there be another factor we're overlooking?

**Any guidance or suggestions would be immensely helpful!**

Proof of Human / Disclaimer for transparency: I wrote / translated this issue with the help of ChatGPT bc. I'm no native english speaker. The whole conversation can be seen here: https://chatgpt.com/share/67a5f908-cb00-8003-9f9e-af1bc2a2b244

czepan:
## What does this PR do?

This PR fixes a Webkit-specific issue where the regular expression used for username validation (`/^[a-zA-ZÀ-ÖØ-öø-ÿ0-9_.]+$/`) caused a `SyntaxError: Invalid regular expression: range out of order in character class`. This issue was sporadic and only occurred in WebKit due to its handling of character classes with certain Unicode ranges.

The fix replaces the problematic regex with a WebKit-compatible version:
`/^[a-zA-Z0-9_.\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u00FF]+$/`
which maintains the same character range but avoids WebKit's parsing issue.

Fixes: https://github.com/coralproject/talk/issues/4733

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

N/A

## How do I test this PR?

- Open Coral in a WebKit Browser like Safari (both standard and private browsing modes).
- Attempt to refresh the page multiple times (~20+ times) to check for the `SyntaxError: Invalid regular expression`
- Verify that user input validation still works as expected for usernames.

no

## How do we deploy this PR?

No special deployment steps are required


czepan:
Here is the "Explain" section from Regexr as proof that the character sequences remain unchanged.

image

losowsky:
we'll take a look, thanks!

kabeaty (approved):
This looks good and works well! Thank you for looking into this issue and implementing this fix.

Issue #3976, opened 2022-06-07
"INTERNAL_ERROR: DUPLICATE_EMAIL" after moving to SSO


jacotec:
To overcome the limitations with the internal registration, I've changed the authentication to OpenID-Connect (Keycloak). It's all pretty much working, but not my own user which was present with email authentication before.

Whenever I want to log in via SSO, I'm getting this error:

![image](https://user-images.githubusercontent.com/3321655/172401838-da7c7325-a77f-4c75-b1b0-be63ab601a6f.png)

This is pretty annoying as I need to get my account back.

I've still found the old entry in the database and have deleted this old user with the same email address, afterwards I have restarted Coral. But I'm still getting this error and need a solution where this messages comes from (as there is **no** user with the same email in the database!).

**Expected behavior:**

Coral should detect collisions of the same email address and merge the info / handle the login properly.

**Actual behavior:**

Although I've removed the old user from the database, I'm still getting this error.

**Related Issues:** 

**Versions:**

- NodeJS: 14.18.0
- NPM: 8.0.0
- MongoDB: 4.4
- Redis:
- Browser:
- OS:

tessalt:
check to make sure you have no users with the same email in both the `user.email` field and the `user.profiles[x].email` field. For this error to occur there must still be a user in the database with the same email address or ID.

rigperro:
Im also experiencing this. We have OpenID and SSO set up, but for some reason when deleting an user created with OpenID, the email **sometimes** remains in the database. The Mongo call in here https://github.com/coralproject/talk/blob/develop/src/core/server/services/users/delete.ts#L279 
succeeds for the most part, but fails to unset the email. I can't find anything in the logs regarding this. The profiles field gets unset and the deletedAt attribute is added. Other times, the email gets unset as well.

Issue #3947, opened 2022-05-12
Account linking doesn't work with OIDC


grob:
Linking existing accounts with oidc doesn't work, imho because the method [linkUsersAvailable](https://github.com/coralproject/talk/blob/develop/src/core/server/models/tenant/helpers.ts#L79) only checks for `local`, `google` or `facebook`.

Is there a specific reason why `oidc` is excluded from these checks?

tessalt:
Hi there, thanks for your report. We haven't had the resources to look into the OIDC flow in a while but it is in our backlog to look into in the future.

PR #4740, opened 2025-02-11
Feature: allow linking sso and oidc profiles

bastiankistner:
## What does this PR do?

Add capability to link OIDC and SSO profiles.

Resolves 
- https://github.com/coralproject/talk/issues/3947
- https://github.com/coralproject/talk/issues/3976

## These changes will impact:

- [x] commenters
- [x] moderators
- [x] admins
- [ ] developers

None

No

## How do I test this PR?

**For OIDC**
1. setup any OIDC provider and enable it in coral admin UI
2. register a new user via email / password
3. logout
4. login with the OIDC provider
5. link your accounts

**For SSO**
1. create a user (not with SSO)
2. create an SSO token according to https://docs.coralproject.net/sso with the email of the previously created user
3. use the SSO token for requests against the API
4. check the user's profiles (will now have an SSO entry)

## Open Questions to discuss:

- [ ] would this change introduce any problems 
  The change is quite simple and I wonder why this hasn't been implemented yet. What were the reasons to restrict linking to `google` and `facebook` so far ?
- [ ] do you consider it problematic to not request the user to link their existing account with an SSO profile? Since the user does already exist and there is not really an authentication flow tied to Coral (since we're manually issuing the SSO token), I don't see where this should take place.

### DIFF
--- a/server/src/core/server/models/tenant/helpers.ts
+++ b/server/src/core/server/models/tenant/helpers.ts
@@ -95,7 +95,9 @@ export function linkUsersAvailable(
   return (
     hasEnabledAuthIntegration(config, tenant, "local") &&
     (hasEnabledAuthIntegration(config, tenant, "facebook") ||
-      hasEnabledAuthIntegration(config, tenant, "google"))
+      hasEnabledAuthIntegration(config, tenant, "google") ||
+      hasEnabledAuthIntegration(config, tenant, "sso") ||
+      hasEnabledAuthIntegration(config, tenant, "oidc"))
   );
 }
 
--- a/server/src/core/server/models/user/user.ts
+++ b/server/src/core/server/models/user/user.ts
@@ -811,6 +811,35 @@ export async function findOrCreateUserInput(
   return user;
 }
 
+export async function linkSsoUser(
+  mongo: MongoContext,
+  tenantID: string,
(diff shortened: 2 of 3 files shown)


tessalt:
hi there, thanks for your contribution. We are unlikely to merge this PR right now, as the user `findOrCreate` method is a critical code path that has been stable and unchanged for a very long time, and requires thorough testing and planning to modify. This is a change we've been considering implementing, but we haven't come to a conclusion on the future of auth profiles just yet. We will discuss internally but unfortunately I can't give you a timeline for a decision right now.

bastiankistner:
Hi @tessalt ! I very much appreciate your feedback and I understand that it's a critical code path. You're saying that you haven't come to a conclusion regarding the future of auth profiles yet. Wouldn't it then be a better idea to open up this topic for discussion instead of closing this issue and postponing it further? I saw that some tickets related to this issue have remained untouched for almost 3 years. 

Being able to link SSO and OIDC profiles is (probably not only for our own goals) a very important capability. We have several hundreds of thousands of users who have all signed up with email / password but we have to migrate to a different flow that is based on OIDC with another system like keycloak as the leading component and will also need to leverage SSO tokens to communicate between a facade and Coral Talk. 

Would be great if you could provide a shared platform to push this forward together.

PR #4747, opened 2025-02-19
[CORL-3221] Allow multiple profile linking for Coral auth (SSO/Local) + integration testing environment

nick-funk:
## What does this PR do?

- Introduces in-memory Redis and Mongo support for the testing environments
- Creates a bunch of mocks and helpers to support testing full request handlers
- Adds tests for the signup and auth flows (passport)
- Allow linking an SSO profile to an existing local auth user
  - Includes integration tests around this flow
- Allow linking an SSO profile to an existing SSO user if they share the same email
  - Includes integration tests around this flow

## These changes will impact:

- [X] commenters
- [ ] moderators
- [ ] admins
- [X] developers

None

No

No

## How do I test this PR?

- Run the local cache tests via `cd server && pnpm run test:cache`
- Use `multi-site-test` and create two sso users with different id's but same email
  - see that you can login as both and check in database that profiles were added to the user with identifiers for old and new id
  - example sso users for `multi-site-test` below:
  ```
  {
    "id": "926576b4-9e7e-425d-aaea-c2cc1f605ec5",
    "username": "existingsso",
    "email": "existingsso@test.com",
    "role": "COMMENTER"
  },
  {
    "id": "ebfc7170-798f-4b49-9064-ea236e0392b2",
    "username": "existingsso",
    "email": "existingsso@test.com",
    "role": "COMMENTER"
  },
  ```

No

## How do we deploy this PR?

- Merge into `develop`


server/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
```diff
@@ -131,11 +147,70 @@ export async function findOrCreateSSOUser(
...
+  // attempt to find an sso profile user
+  if (!user) {
+    user = await retrieveUserWithProfile(mongo, tenant.id, {
+      type: "sso",
+      id,
+    });
+  }
+
+  // attempt to find a local user that matches by id
+  if (!user && id) {
```
kabeaty, line 161: wondering if this section through line 188 could be refactored and simplified a bit to something like:
```
if (!user) {
  const idToMatch = id ?? email;
  user = await retrieveUserWithProfile(mongo, tenant.id, {
    type: "local",
    id: idToMatch,
  });
  
  if (user) {
    user = await ensureHasSSOProfile(mongo, user, {
      type: "sso",
      id,
    });
  }
}
```

server/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
```diff
@@ -131,11 +147,70 @@ export async function findOrCreateSSOUser(
...
+
+    if (user) {
+      user = await ensureHasSSOProfile(mongo, user, {
+        type: "sso",
+        id,
+      });
+    }
+  }
+
+  // attempt to find an existing sso user by email
```
kabeaty, line 190: ty for adding this comment!

server/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
```diff
@@ -131,11 +147,70 @@ export async function findOrCreateSSOUser(
...
+    const matchedUser = await findUserByEmail(mongo, tenant.id, email);
+    if (matchedUser && hasSSOProfile(matchedUser)) {
+      user = await addProfileToUser(mongo, matchedUser, {
+        type: "sso",
+        id,
+      });
+    }
+  }
+
+  // // attempt to find an existing sso user by email
```
kabeaty: It looks like these commented-out lines can just be removed?

server/src/core/server/services/users/integrationTests/strategies.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+
+import {
+  createVerifiers,
+  verifyAndRetrieveUser,
+} from "coral-server/app/middleware/passport/strategies/jwt";
+import config from "coral-server/config";
+import { GQLUSER_ROLE } from "coral-server/graph/schema/__generated__/types";
+import { createTestEnv, createTestLocalUser } from "coral-server/test/helpers";
+
+it("sso user, can verify and retrieve user", async () => {
```
kabeaty, line 12: Yay, love all these tests! Very nice.

server/src/core/server/services/users/users.ts
```diff
@@ -2630,3 +2631,59 @@ export async function link(
...
+    {
+      $set: {
+        profiles: uniqueProfiles,
+      },
+    },
+    { returnDocument: "after" }
+  );
+};
+
+export const findUserByEmail = async (
```
kabeaty, line 2645: Should this also be added in `models/user/user.ts` instead?

server/src/core/server/services/users/users.ts
```diff
@@ -2630,3 +2631,59 @@ export async function link(
...
+    for (const p of existing.profiles) {
+      profiles.set(`${p.type}:${p.id}`, p);
+    }
+  }
+
+  profiles.set(`${profile.type}:${profile.id}`, profile);
+
+  const uniqueProfiles = Array.from(profiles.values());
+
+  return await mongo.users().findOneAndUpdate(
```
kabeaty: Should this be added in `models/user/user.ts` instead and imported here?

kabeaty (commented):
Looking good! Thank you for adding all the tests, very nice to see. Manual testing all worked well. I left a few comments/questions throughout.

server/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
nick-funk: Unfortunately, not 😞 

This is because there is a possibility of having both email and id and we want to attempt to match both of them. It is possible for both to be defined, email turns up nothing, but id does, so we have to check and do both. An `id` takes precedence over an `email` match in our system of rules though, so we do that check first.

server/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
nick-funk: fixed

server/src/core/server/services/users/users.ts
nick-funk: fixed!

server/src/core/server/services/users/users.ts
nick-funk: fixed!

server/src/core/server/app/middleware/passport/strategies/verifiers/sso.ts
kabeaty: ah, yeah, makes sense...thanks!

kabeaty (approved):
Looks great, and thanks for the updates!

tessalt (approved):
this looks great. shockingly less code than I expected given all the tests!

PR #4758, opened 2025-02-28
[CORL-3227]: Add domain alias pre-mod filter

kabeaty:
## What does this PR do?

These changes add a domain alias pre-mod filter. If a new user signs up with a domain alias, they will be set to pre-mod when this is enabled.

## These changes will impact:

- [ ] commenters
- [x] moderators
- [ ] admins
- [ ] developers

adds `domainAliases` to the pre-moderate email address config

no

n/a

## How do I test this PR?

You can test by having the feature disabled (admin --> configure --> users --> email address --> pre-moderate domain alias) and seeing that regular domains and domain aliases can sign up just fine.

Enable the feature and see that users with domain aliases are pre-moderated. Users with regular domains can sign up just fine.

Domain aliases of banned domains should still be banned (no change in functionality from before here).

### DIFF
--- a/server/src/core/server/services/users/emailPremodFilter.ts
+++ b/server/src/core/server/services/users/emailPremodFilter.ts
@@ -104,6 +104,26 @@ const emailIsAnAliasOfExistingUser = async (
   return similarUsers.length > 0;
 };
 
+const emailIsDomainAlias = (email: string | undefined) => {
+  if (!email) {
+    return false;
+  }
+  const emailSplit = email.split("@");
+  if (emailSplit.length < 2) {
+    return false;
+  }
+
+  const domain = emailSplit[1].trim().toLowerCase();
+
+  let periodCount = 0;
+  for (const char of domain) {
+    if (char === ".") {
+      periodCount++;
+    }
+  }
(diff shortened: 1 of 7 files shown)


nick-funk (approved):
Makes perfect sense, nice feature and fast turnaround!

server/src/core/server/services/users/emailPremodFilter.ts
```diff
@@ -104,6 +104,26 @@ const emailIsAnAliasOfExistingUser = async (
...
+  const domain = emailSplit[1].trim().toLowerCase();
+
+  let periodCount = 0;
+  for (const char of domain) {
+    if (char === ".") {
+      periodCount++;
+    }
+  }
+  return periodCount > 1;
+};
```
tessalt, line 125: I believe this will catch too many legitimate email addresses, including institutional emails (@mail.school.edu, @army.mil.us) and second-level domains like @gmail.co.uk. Can we do the same thing we do for email aliases, where we strip out the subdomain, then check if that domain is banned, then premod?

kabeaty:
Closing for now over concern that this will catch too many false positives.

PR #4806, opened 2025-08-26
fix: improve url parsing and handling

kabeaty:
## What does this PR do?

These changes improve our url parsing and handling.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

### DIFF
--- a/client/src/core/client/admin/test/dsaReports/dsaReportsSingle.spec.tsx
+++ b/client/src/core/client/admin/test/dsaReports/dsaReportsSingle.spec.tsx
@@ -175,7 +175,7 @@ it("includes expected information about dsa report", async () => {
     within(container).getByRole("link", { name: "View comment in stream" })
   ).toHaveProperty(
     "href",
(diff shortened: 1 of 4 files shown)


client/src/core/client/admin/test/dsaReports/dsaReportsSingle.spec.tsx
```diff
@@ -175,7 +175,7 @@ it("includes expected information about dsa report", async () => {
     within(container).getByRole("link", { name: "View comment in stream" })
   ).toHaveProperty(
     "href",
-    "http://localhost/admin/reports/report/dsa-report-1?commentID=comment-regular-0"
+    "http://localhost:8080/?commentID=comment-regular-0"
```
kabeaty, line 178: Before, because the story url for this comment was an empty string, it was falling back to the `window.location`, set at the top of this test, which is not the correct story/comment link. Now, this updated expected link is using the new story url added to this comment's story for usage in other tests too.

tessalt (approved):

PR #4810, opened 2025-09-08
CSS Loading Timeout

colettak:
## What does this PR do?
The goal of this code is to prevent coral from hiding forever due to css never loading. This is most likely due to weird Safari font loading interactions, but certain css files are never returning a response at all, causing the coral css that hides coral on the page to never be removed. This effect essentially forces it to show after 3 seconds, whether or not all the extra css has loaded or not.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None

No
## How do I test this PR?
Most easily reproducible on Safari with fonts being imported in the css, though even then it's intermittent.


nick-funk (approved):
this makes sense and is a good practice to ensure we're not allowing external resources (CSS) to block our app from working

kabeaty (approved):

PR #4813, opened 2025-09-12
Emit comment count data

colettak:
## What does this PR do?

There is an internal event already, commentCount, that keeps track of the changes to commentCount. We want to also emit this event externally, so pages can render comment counts however they see fit.

So, we change the event from an internal event (no prefix) to a new event, updating the name to `count.comments` and then ensuring it flows properly to the `EventEmitter`. We also update the internal tracking to use the new name.

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

None

## How do I test this PR?
If you make an embed and listen to the events, you should see `counts.comment` firing whenever the comment count is updated.


client/src/core/client/embed/test/basic.spec.ts
```diff
@@ -95,8 +95,10 @@ describe("Basic integration test", () => {
             "_removeListener": false,
             "delimiter": ".",
             "listenerTree": Object {
-              "commentCount": Object {
-                "_listeners": [Function],
+              "count": Object {
```
nick-funk, line 98: since we already had a `commentCount` event, why are we replacing it? Why not re-use?

also, if we replace this with `count`, it might break other prior uses of `commentCount`?

are we sure this isn't a breaking change?

i.e. are we sure that `commentCount` was a 100% internal event?

nick-funk (commented):
I have one question that I'd like to have answered

if you're sure this won't break other prior `commentCount` listeners, we can go ahead and approve this to merge for next release

client/src/core/client/embed/test/basic.spec.ts
colettak: We totally could just reuse it, it's just that the withEventEmitter code checks for prefixes. I could totally just allow that specific event to flow through, it just felt more hacky than giving it a namespace. 100% fine either way, so if you think it is safer to just use the OG, I can just update the withEventEmitter code to pass that specific event forward.

colettak: It does turn it into a 1 line change, which is nice.

nick-funk: eh, lets test on staging like you suggested, see where this goes, we can always refactor later

nick-funk (approved):
Let's see how this behaves out on staging. I think the rename is probably fine. Approved!

kabeaty (approved):
Looks good, let's get it tested out!

PR #4821, opened 2025-10-08
Setting to collapse top-level replies by default

tessalt:
## Collapse First-Level Replies by Default

Adds a new tenant setting to collapse first-level replies in the comment stream by default, giving moderators control over the initial display density of threaded conversations.

### Changes

**Server**
- Add `collapseReplies` boolean setting to tenant configuration
- Add GraphQL schema field for the new setting

**Admin**
- Add "Collapse Replies" configuration toggle in General settings
- When enabled, top-level replies (depth 1) default to collapsed state

**Client**
- Refactor `CollapsableComment` to use local GraphQL store for persistence
- Track user toggles in `collapsedCommentSettings` (stores comment IDs toggled opposite to default)
- Comments that are manually expanded/collapsed maintain their state within the session
- Newly created comments always render expanded regardless of tenant setting

### Behavior

- **Setting OFF (default)**: First-level replies display expanded (existing behavior)
- **Setting ON**: First-level replies display collapsed, users can expand individually
- toggle states persist on scroll (this is a change from previous behaviour)
-
## These changes will impact:

- [x] commenters
- [ ] moderators
- [x] admins
- [ ] developers

## How do I test this PR?

- Turn on the setting in General > Collapse Replies
- Go to a stream with threaded comments, observe that the first level of replies is collapsed. 
- open up a thread, and see that further nested comments are not collapsed
- scroll down and back, see that the state you recently set is maintained 
- reply to a top-level comment, see that the new comment is visible


client/src/core/client/admin/routes/Configure/sections/General/CollapseRepliesConfig.tsx
```diff
@@ -0,0 +1,48 @@
+import { Localized } from "@fluent/react/compat";
+import {
+  FieldSet,
+  FormField,
+  FormFieldDescription,
+  Label,
+} from "coral-ui/components/v2";
+import React, { FunctionComponent } from "react";
+import { graphql } from "react-relay";
```
kabeaty: Not a big deal, but would be nice to sort the imports up here to maintain consistency with other files. (So, first external imports, space, internal imports, space, local imports).

client/src/core/client/stream/tabs/Comments/Comment/CollapsableComment.tsx
```diff
@@ -7,13 +12,66 @@ interface InjectedCollapsableCommentProps {
...
+        }
+      }
+    `);
+
+  // If comment was just created, always start un-collapsed
+  const isNewlyCreated = comment?.lastViewerAction === "CREATE";
+
+  const collapsed = useMemo(() => {
+    // If comment was just created, always show un-collapsed
+    if (isNewlyCreated) {
```
kabeaty, line 41: This check means that until a user reloads the stream, they cannot collapse their new reply. I'm not sure how often someone would want to do that, but the new comment does show an expand/collapse option still, so it could be a bit confusing that it won't collapse.

client/src/core/client/stream/tabs/Comments/ReplyList/ReplyListCommentContainer.tsx
```diff
@@ -57,7 +57,11 @@ const ReplyListCommentContainer: FunctionComponent<Props> = ({
         allowTombstoneReveal={allowIgnoredTombstoneReveal}
       >
         <HorizontalGutter spacing={commentSeenEnabled ? 0 : undefined}>
-          <CollapsableComment>
+          <CollapsableComment
+            defaultCollapsed={indentLevel === 1 && settings.collapseReplies}
+            commentID={comment.id}
```
kabeaty: do we need to pass through both `comment` and `commentID` here? I think we can just pass in `comment`

kabeaty (commented):
Looks good overall! The maintaining collapsed/expanded state on scroll is definitely a nice improvement. 

I left a few small comments throughout. My main question is if we want to update so that it's possible to expand/collapse new replies after they have initially loaded in if the user who added the replies clicks on the expand/collapse.

client/src/core/client/stream/tabs/Comments/Comment/CollapsableComment.tsx
tessalt: ah interesting. I'm not suuure I care but I'll play around with it for a few minutes. good catch.

kabeaty (approved):

client/src/core/client/framework/helpers/supportedBrowsersRegExp.ts
```diff
@@ -3,5 +3,5 @@
  * Any manual changes will be overridden.
```
nick-funk, line 3: thank you for updating this

nick-funk (approved):
looks good to me

PR #4822, opened 2025-10-14
create a new external notifications service

nick-funk:
## What does this PR do?

allows events to be forwarded to an external notifications service for external consumption

## These changes will impact:

- [ ] commenters
- [ ] moderators
- [ ] admins
- [x] developers

None

- `EXTERNAL_NOTIFICATIONS_API_URL`
- `EXTERNAL_NOTIFICATIONS_API_KEY`

N/A

## How do I test this PR?

- set the `EXTERNAL_NOTIFICATIONS_API_URL` to the api url you want to forward notifications to
- set the `EXTERNAL_NOTIFICATIONS_API_KEY` to the value of the api key you will use with your external notifications service
- reply and rec content
- see that notifications calls are made to your external service

No

## How do we deploy this PR?

- merge into `develop`
- release with other changes via `develop` into `main`

### DIFF
--- a/server/src/core/server/config.ts
+++ b/server/src/core/server/config.ts
@@ -505,6 +505,18 @@ const config = convict({
     default: ms("3000s"),
     env: "NOTIFICATIONS_POLL_RATE",
   },
+  external_notifications_api_url: {
+    doc: "URL to forward notifications information to an external url.",
+    format: "url",
+    default: null,
+    env: "EXTERNAL_NOTIFICATIONS_API_URL",
+  },
+  external_notifications_api_key: {
+    doc: "API key to use when forwarding notifications to an external url",
+    format: String,
+    default: null,
+    env: "EXTERNAL_NOTIFICATIONS_API_KEY",
+  },
 });
 
 export type Config = typeof config;
--- a/server/src/core/server/config.ts
+++ b/server/src/core/server/config.ts
@@ -508,13 +508,13 @@ const config = convict({
   external_notifications_api_url: {
     doc: "URL to forward notifications information to an external url.",
     format: "url",
-    default: null,
+    default: "",
     env: "EXTERNAL_NOTIFICATIONS_API_URL",
   },
   external_notifications_api_key: {
     doc: "API key to use when forwarding notifications to an external url",
     format: String,
-    default: null,
+    default: "",
     env: "EXTERNAL_NOTIFICATIONS_API_KEY",
   },
 });
--- a/server/src/core/server/config.ts
+++ b/server/src/core/server/config.ts
@@ -508,7 +508,7 @@ const config = convict({
   external_notifications_api_url: {
     doc: "URL to forward notifications information to an external url.",
     format: "url",
-    default: "http://localhost:7003",
+    default: "http://localhost:7003/api",
     env: "EXTERNAL_NOTIFICATIONS_API_URL",
   },
   external_notifications_api_key: {
--- a/server/src/core/server/config.ts
+++ b/server/src/core/server/config.ts
@@ -508,7 +508,7 @@ const config = convict({
   external_notifications_api_url: {
     doc: "URL to forward notifications information to an external url.",
     format: "url",
-    default: "",
+    default: "http://localhost:7003",
     env: "EXTERNAL_NOTIFICATIONS_API_URL",
   },
   external_notifications_api_key: {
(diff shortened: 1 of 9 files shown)


server/src/core/server/services/comments/actions.ts
```diff
@@ -353,6 +355,20 @@ export async function createReaction(
     ).catch((err) => {
       logger.error({ err }, "could not publish comment flag created");
     });
+
+    if (externalNotifications.active()) {
+      const parentAuthor = comment.authorID
```
kabeaty: Using `parentAuthor` and `author` here for recs doesn't seem as clear to me as it does for reply notifications. What about something more like `commentAuthor` and `recAuthor` or `recUser` or `userWhoRecced`?

server/src/core/server/config.ts
```diff
@@ -505,6 +505,18 @@ const config = convict({
...
+  external_notifications_api_url: {
+    doc: "URL to forward notifications information to an external url.",
+    format: "url",
+    default: null,
+    env: "EXTERNAL_NOTIFICATIONS_API_URL",
+  },
+  external_notifications_api_key: {
+    doc: "API key to use when forwarding notifications to an external url",
+    format: String,
+    default: null,
```
kabeaty: I think this needs to have a default of empty string instead of null, as otherwise there is an error when starting up the server if nothing is set for the external notifications api key. Same note for the api key; it also seems to need a default url.

I guess then the check in the notifications service to see if it's active needs to be updated too to check for not empty string, etc.

server/src/core/server/services/notifications/externalService.ts
```diff
@@ -0,0 +1,188 @@
...
+    const data = {
+      query: CreateNotificationsMutation,
+      variables: {
+        input: {
+          items: [notification],
+        },
+      },
+    };
+
+    const response = await fetch(this.url!, {
```
kabeaty, line 187: I'm getting 
```
"message": "INTERNAL_ERROR: fetch is not defined",
```

I think this needs to be updated to use `import { createFetch } from "coral-server/services/fetch";` to create fetch here to use?

kabeaty (commented):
This is looking good overall! I left a few comments throughout where I am seeing errors while testing this out.

server/src/core/server/services/notifications/externalService.ts
nick-funk: oh geeze, nah, it's cause we need to use `node-fetch` in node v18 still, I was running on v20 for the vscode config, repairing now!

nick-funk: fixed

server/src/core/server/config.ts
nick-funk: it turns out that empty strings are falsy, so that works for us

updated and fixed!

server/src/core/server/services/comments/actions.ts
nick-funk: great idea, updated!

server/src/core/server/config.ts
kabeaty: The `external_notifications_api_key` needs to have a valid default url though, not just empty string. When starting up without anything set, it gives this error now:
```
 "message": "external_notifications_api_url: must be a URL",
```

server/src/core/server/services/comments/actions.ts
kabeaty: nice, thanks for updating!

server/src/core/server/stacks/createComment.ts
```diff
@@ -443,6 +445,24 @@ export default async function create(
         reply: comment,
         type,
       });
+
+      // if we have an external notifications service hooked up
```
tessalt, line 449: should we skip doing default notification if the external service is enabled?

kabeaty (changes requested):
Thank you for the updates! This looks and works great overall now. Once the comment is addressed about providing a default url for `external_notifications_api_key` so that there is no error on server startup if it's not set, should be good to go.

server/src/core/server/stacks/createComment.ts
nick-funk: good call, this is now hooked up to disable internal notifications when the external service is active

server/src/core/server/config.ts
nick-funk: I've set this to the default url of `http://localhost:7003/api`, that should stop the errors when this isn't set in the env vars

this also works for the active state because you also have to set a valid API key for the service to become active 🙂

kabeaty (approved):
looking good!

PR #4837, opened 2025-11-24
send approve, reject, and feature actions to external notifications

nick-funk:
## What does this PR do?

- send approve, reject, and feature actions to external notifications

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [x] developers

## What changes to the GraphQL/Database Schema does this PR introduce?

None

## Does this PR introduce any new environment variables or feature flags?

No

## If any indexes were added, were they added to `INDEXES.md`?

N/A

## How do I test this PR?

- configure env to send external notifications
- approve, reject, feature comments
- notifications should be sent through

## Were any tests migrated to React Testing Library?

No

## How do we deploy this PR?

- merge into `develop`
- cut release from `develop` into `main`

PR #4838, opened 2025-11-24
Update when send reply notifications to external service when comments are premod, system withheld, or rejected

kabeaty:
## What does this PR do?

These changes make it so that reply notifications are not sent to the external service when comments are created with a status of premod, system withheld, or rejected.

When a comment is approved, if its previous status was one of these, a reply notification is created.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

none

no

n/a

## How do I test this PR?

You can test this PR by seeing that comments that are pre-mod (use a pre-mod user to create a comment) do not send out a reply notification to the external service when it is active.

Comments that are rejected (for example, include a banned word) do not send out a reply notification to the external service.

Comments that are system withheld (for example, include a suspect word) do not send out a reply notification to the external service.


kabeaty:
closing since this is handled in here now https://github.com/coralproject/talk/pull/4837

PR #4839, opened 2025-12-01
Hide internal notification preferences when external notifications are enabled

kabeaty:
## What does this PR do?

When external notifications are enabled, then the internal notification preferences shown in `My Profile` --> `Preferences` shouldn't be shown because the external notifications are responsible for external notification preferences.

## These changes will impact:

- [x] commenters
- [ ] moderators
- [ ] admins
- [ ] developers

Added external notifications to settings so we could know whether it's active or not to determine whether to show internal notifications of any sort (in-page or email).

## How do I test this PR?

You can test this by enabling external notifications and seeing that no notification preferences are shown under `My Profile`. 

Disable external notifications and the notification preferences should be shown again. They should correctly show either in-page notifications settings (if enabled) or email notification settings (if in-page are not enabled).


nick-funk (approved):
Looks good, nice!

PR #4851, opened 2026-01-08
bulk send approve notifications out as one request

nick-funk:
## What does this PR do?

- process external notifications on a job queue
    - no longer holds up the approve/reject flow with in-line awaits
- bulk send notifications out as one request inside the job
    - this makes it much faster rather than awaiting two net requests
- adds bulk send to the external notifications service as well

## These changes will impact:

- [ ] commenters
- [x] moderators
- [ ] admins
- [ ] developers

None

No

N/A

## How do I test this PR?

- spin up the service with jobs `pnpm run start:development:withJobs`
- create a premodded comment that is a reply to another published comment
- go to admin and approve the premodded comment
- compare the speed of this on `develop` versus in this branch
    - should be significantly faster to approve/reject

No

## How do we deploy this PR?

- merge into `develop`
- cut out release from `develop` into `main`


server/src/core/server/queue/index.ts
```diff
@@ -91,7 +96,14 @@ export function createQueue(options: QueueOptions): TaskQueue {
...
   });
   const webhook = createWebhookTask(queueOptions, options);
-  const rejector = createRejectorTask(queueOptions, options);
+  const externalNotifications = createExternalNotificationsTask(
+    queueOptions,
+    options
+  );
+  const rejector = createRejectorTask(queueOptions, {
+    ...options,
+    externalNotificationsQueue: externalNotifications,
```
kabeaty, line 105: Is this argument supposed to go to the rejector task? I think these got mixed up and it should go to the external notifications task?

kabeaty (commented):
looking good overall! I left one comment with a question.

server/src/core/server/queue/index.ts
nick-funk: nope, the rejector will also send out notifications when it performs its job, so it needs to be able to schedule out other "tasks" (jobs) the same way the notifier and the mailer are chained task queues 😃

kabeaty (approved):

PR #4871, opened 2026-05-06
hopefully fix netlify docs deploy

tessalt:
- netlify was using npm, not pnpm, so was not getting a lockfile, and ended up with bad webpack versions. this will hopefully fix it

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -62,5 +62,10 @@
   },
   "engines": {
     "node": ">=18.0"
+  },
+  "pnpm": {
+    "overrides": {
+      "webpack": "5.95.0"
+    }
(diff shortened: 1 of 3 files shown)


kabeaty (approved):

docs/package.json
```diff
@@ -62,5 +62,10 @@
   },
   "engines": {
     "node": ">=18.0"
+  },
+  "pnpm": {
```
helderdb, line 66: This feels hacky.
I've taken a quick look at fully upgrading pnpm & node to a more recent version. Wouldn't it be better to align this better as a general update and make sure all versions are more up to date?

